PR #18, opened 2019-09-21
Corrections to Tutorial 3

jmorgan29:
I found an issue with the code while running the "Tutorial_3_Dynamic_Flowsheets" example. When plotting the component concentrations, the concentration of "NaOH" is plotted for the curve labeled "Ethanol". Since it was a simple issue, I created this PR in which the .py and .ipynb files are modified so that the "Ethanol" concentration is plotted as suggested by the label.

### DIFF
--- a/idaes/examples/tutorials/Tutorial_3_Dynamic_Flowsheets.ipynb
+++ b/idaes/examples/tutorials/Tutorial_3_Dynamic_Flowsheets.ipynb
@@ -809,7 +809,7 @@
     "         list(m.fs.Tank1.outlet.conc_mol_comp[:, \"SodiumAcetate\"].value),\n",
     "         label='SodiumAcetate')\n",
     "plt.plot(m.fs.time,\n",
-    "         list(m.fs.Tank1.outlet.conc_mol_comp[:, \"NaOH\"].value),\n",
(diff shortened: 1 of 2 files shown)


jghouse88:
@andrewlee94 Do we want to review this or we revamping tutorials? Also, @ksbeattie what happens to your update process of pushing dev to pse if and when we have a public PR like this?

jghouse88:
@jmorgan29 Will take a look. Seems simple enough but need to figure out the logistics before the merge is approved.

ksbeattie:
> @andrewlee94 ... Also, @ksbeattie what happens to your update process of pushing dev to pse if and when we have a public PR like this?

Unfortunately, it makes a mess for us.  @jmorgan29, would you mind re-creating this PR in idaes-dev?  It will make it a ton simpler for me.

blnicho:
@ksbeattie I don't think @jmorgan29 has access to idaes-dev. This is the first instance of user contributed code.

ksbeattie:
ok, thank you @blnicho, I'm going to re-play this PR in idaes-dev.

@jmorgan29, thank you for the contribution.  Very sorry we have this limitation on accepting PRs directly in idaes-pse.  Hopefully I'll figure out a way to manage this split between the private idaes-dev and the public idaes-pse, but for now this is the fastest, easiest way to get you work into idaes-pse believe it or not.

PR #54, opened 2020-11-11
Add CI workflows using GitHub Actions

lbianchi-lbl:
## Motivation

This PR implements part of the CI functionality that we had with CircleCI using GitHub actions.

Advantages of GitHub actions include (but are not limited to):

- Run tests on macOS and Windows
- Much more granular control on events that trigger workflow runs
- Better integration with the GitHub interface and PR workflow
  - e.g. built-in support for running CI workflows in forks without the need for additional setup

### Functionality

- I've tested this on my fork (<https://github.com/lbianchi-lbl/idaes-pse/actions>) and things seem to be working just fine
- A good fraction of the CircleCI functionality should be covered:
  - Build Sphinx docs
  - Install and run non-integration tests
  - Run static analysis with pylint
  - Integration tests including IDAES/examples-pse
- All of these are triggered by pull requests and push events, except the integration tests that have to be triggered manually (until I get the "approved tag" workflow to run reliably)

### Caveats/limitations
- For simplicity, relies on the built-in action `actions/setup-python` to install Python in the global environment of the CI runner (as opposed to setting up a Conda environment as we were doing in CircleCI)
  - This shouldn't have any effect until we start relying on `conda` more heavily for installing dependencies 
  - There's nothing in particular preventing us from using Conda environments, it just requires a bit more tinkering

## Summary

I believe that there are large margins of improvement for the current functionality; that said, perhaps it's better to merge this in and restore the PR workflow using this version and then implement further changes as separate future PRs.

- Add GitHub Actions workflow definition files 

### Legal Acknowledgement

----

## Reviewer Checklist
* Reviewer 1
* Reviewer 2


ksbeattie (approved):
Love this!

.github/workflows/pull-request-main.yml
```diff
@@ -0,0 +1,127 @@
+name: Pull request (main) CI tests
+
+on:
+  push:
+    branches:
+      - main
+      # TODO "gh-actions" branch included to facilitate development
+      # we could also consider simply going myfork:main -> upstream:main for CI PRs
+      - gh-actions
```
jsiirola, line 9: Should this be removed before merging?

.github/workflows/pull-request-main.yml
```diff
@@ -0,0 +1,127 @@
...
+          python -m pip install --progress-bar off -r requirements.txt
+      - name: Display IDAES version
+        run: |
+          idaes --version
+      - name: Install IDAES extensions
+        run: |
+          idaes get-extensions --verbose
+          # TODO: these commands are Unix-shell specific 
+          # and will need to be changed if we want to test on CMD.exe/Powershell
+          find $(idaes data-directory) -ls
```
jsiirola, line 54: Is this comment out of date, as GHA makes the bash shell available on Windows?

jsiirola (dismissed):

.github/workflows/pull-request-main.yml
lbianchi-lbl: Good point, I think so. It looks like the `on: push` trigger works just fine, and in case there are more reviews requesting changes that I need to test, I should be able to trigger the workflows manually.

.github/workflows/pull-request-main.yml
lbianchi-lbl: Here I was making a note that having `find` requires a Unix shell, i.e., they will work on Windows as long as `shell: bash` is set or, conversely, they will fail on Windows with the other available shells, including the case where `shell` is not specified, since (as I understand) it's the only available OS where the default shell is not a Unix shell.

ksbeattie (approved):
LGTM

jghouse88 (approved):
Customary LGTM.

PR #55, opened 2020-11-11
Control volume scaling

TimBartholomew:
**NOTE:** When this PR is approved, you *must* do the following so it can be merged:
* Add the "approved" tag to the pull-request
* Trigger tests by pushing a commit or manually re-running in CircleCI

See [this wiki page](https://github.com/IDAES/idaes-dev/wiki/Github-Pull-Request-Workflow-with-Approval-Tag) for details.

----
## Summary/Motivation:
The control volume `calculate_scaling_factors` method should set the scaling factors for all variables and transform all constraints that the control volume creates.

- Adds scaling methods to reaction_base.py, so reaction packages can support default scaling factors and setting scaling factors. These methods were already included in property_base.py
- Updates ControlVolume0D and ControlVolume1D to automatically scale the variables and constraints associated with the true configuration arguments: 
   - `dynamics`
   - `has_rate_reactions`
   - `has_equilibrium_reactions`
   - `has_mass_transfer`
   - `has_heat of reaction`
   - `has_heat_transfer`
   - `has_work_transfer`
   - `has_enthalpy_transfer`
- Organizes calculate_scaling_factors to first set scaling factors for variables and then transform constraints
- Establishes that all control volume scaling does not overwrite any user provided scaling factors
- Fixes typos in the edited files

### Legal Acknowledgement

----

## Reviewer Checklist
* Reviewer 1
* Reviewer 2


TimBartholomew:
This is a transferred PR from ideas-dev (IDAES/idaes-dev#1204)

TimBartholomew:
@andrewlee94 and @eslickj this is ready to be reviewed with the following notes:
- I did my best to estimate the scaling factors, but they probably need to be reviewed closely. For example, I now include the scaling factor for length in the calculations for the 1D control volume variables that are per unit length, but I'm not 100% sure that is correct.
- I assumed mw_comp used in the element flow basis is a parameter, however, mw_comp is on the StateBlock rather than the PhysicalParameterBlock. I think mw_comp should be moved, but I left it as it was originally implemented.
- For variables that are indexed just by time and location, I set the scaling factor at the indexed variable (parent) rather than the var data. I think that will make it easier for users to adjust scaling factors for those variables, if they are dependent on time and location then the user can iterate through them and set them.

andrewlee94:
@TimBartholomew I will start reviewing this today, but as a quick response to you comment about `mw_comp`, I don't think it should be moved. Molecular weight is an unusual case, in that it is a property of the material but just one that never changes - thus to me it makes sense that you could find it in the state blocks with all the other properties (not that it is generally just a reference to the parameter block).

Secondly, unit models (almost) never look to the property parameter block for information (and those cases where they do are slowly being phased out). So, from a consistency point of view, it probably makes more sense to access `mw_comp` through the state block.

andrewlee94:
OK, next observation - all the tests are failing. I am not completely certain, but I think that you have created a circular import loop somewhere.

idaes/core/control_volume0d.py
```diff
@@ -1531,112 +1531,243 @@ def _get_stream_table_contents(self, time_point=0):
...
+                if iscale.get_scaling_factor(v) is None:
+                    sf = iscale.get_scaling_factor(
+                        self.properties_out[t].get_material_flow_terms(p, j),
+                        default=1, warning=True)
+                    iscale.set_scaling_factor(v, sf)
+
+        if hasattr(self, "rate_reaction_extent"):
+            if iscale.get_scaling_factor(self.rate_reaction_extent) is None:
+                # extent of reaction typically between 0.01 and 1
+                iscale.set_scaling_factor(self.rate_reaction_extent, 10)
```
andrewlee94, line 1607: Reaction extent is equal to the reaction rate multiplied by the stoichiometric coefficient. Given the coefficients are generally only order 1, this should be equal to the scaling factor of the reaction rate from the Reaction Block.

idaes/core/control_volume0d.py
```diff
@@ -1531,112 +1531,243 @@ def _get_stream_table_contents(self, time_point=0):
...
+
+        if hasattr(self, "equilibrium_reaction_extent"):
+            if iscale.get_scaling_factor(self.equilibrium_reaction_extent) is None:
+                # extent of reaction typically between 0.01 and 1
+                iscale.set_scaling_factor(self.equilibrium_reaction_extent, 10)
```
andrewlee94, line 1612: As mentioned earlier, it is hard to know what this should be - basically the scaling here should be the same as for the equilibrium reaction generation for now.

idaes/core/control_volume0d.py
```diff
@@ -1531,112 +1531,243 @@ def _get_stream_table_contents(self, time_point=0):
...
+                    sf *= iscale.get_scaling_factor(
+                        self.phase_fraction[t, p],
+                        default=10, warning=True)
+                    sf *= iscale.get_scaling_factor(
+                        self.properties_out[t].get_material_density_terms(p, j),
+                        default=1, warning=True)
+                    # element comp scaling factor is the inverse of its value
+                    sf *= (self.properties_out[t].params.element_comp[j][e])**-1
+                    sf_list.append(sf)
+                sf = min(sf_list)
```
andrewlee94, line 1635: 1. We should set separate scaling factors for each component where possible, as components might have very different scaling.
2. Elemental holdup scaling should use the same approach as the normal material scaling terms, just with the addition of the stoichiometric conversion (and conversion from mass basis if appropriate).

idaes/core/control_volume0d.py
```diff
@@ -1531,112 +1531,243 @@ def _get_stream_table_contents(self, time_point=0):
...
                 iscale.set_scaling_factor(v, sf)
 
+        if hasattr(self, "element_accumulation"):
+            for (t, e), v in self.element_accumulation.items():
+                if iscale.get_scaling_factor(v) is None:
+                    sf = 100 * iscale.get_scaling_factor(
+                        self.element_holdup[t, e])
```
andrewlee94, line 1642: Same applies to elemental accumulation as for normal accumulation.

idaes/core/control_volume0d.py
```diff
@@ -1531,112 +1531,243 @@ def _get_stream_table_contents(self, time_point=0):
...
+                    iscale.set_scaling_factor(v, sf)
+
         if hasattr(self, "energy_accumulation"):
-            for i, v in self.energy_accumulation.items():
-                sf = 100*iscale.get_scaling_factor(
-                    self.energy_holdup[i], default=1, warning=True)
-                iscale.set_scaling_factor(v, sf)
+            for (t, p), v in self.energy_accumulation.items():
+                if iscale.get_scaling_factor(v) is None:
+                    sf = 100 * iscale.get_scaling_factor(self.energy_holdup[t, p])
```
andrewlee94, line 1679: As for material accumulation.

idaes/core/control_volume0d.py
```diff
@@ -1531,112 +1531,243 @@ def _get_stream_table_contents(self, time_point=0):
...
-        # Material Balance Constraints
         if hasattr(self, "material_balances"):
             mb_type = self._constructed_material_balance_type
             if mb_type == MaterialBalanceType.componentPhase:
                 for (t, p, j), c in self.material_balances.items():
                     sf = iscale.get_scaling_factor(
                         self.properties_in[t].get_material_flow_terms(p, j),
-                        default=1,
-                        warning=True)
+                        default=1, warning=True)
```
andrewlee94, line 1732: Minor comment on PEP8 standards (as I understood them when I read it ages ago): if arguments are broken over multiple lines, then you should generally have one argument per line.

andrewlee94 (changes requested):
I haven't looked at the 1D version yet, but I assume a lot of the comments I have made will apply there as well. I am not completely familiar with the scaling tools, but I think what is there is a good start. However, there are a few places where  think we can get better initial scaling factors for a few terms (notably accumulation and reaction terms).

I did not pay as much attention to the constraint scaling section as it needs a little more understanding of what happens behind the scenes - I am hoping @eslickj can take a deeper look there (we can split the work).

idaes/core/reaction_base.py
```diff
@@ -92,6 +94,66 @@ def build(self):
...
         self._validate_property_parameter_properties()
 
+    # This is a dict to store default property scaling factors. They are
+        # defined in the parameter block to provide a universal default for
+        # quantities in a particular kind of state block.  For example, you can
+        # set flow scaling once instead of for every state block. Some of these
+        # may be left for the user to set and some may be defined in a property
+        # module where reasonable defaults can be defined a priori. See
+        # set_default_scaling, get_default_scaling, and unset_default_scaling
+        self.default_scaling_factor = {}
```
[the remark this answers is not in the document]
TimBartholomew: I agree. I'll move it there and remove it from reaction_base and property_base.

idaes/core/control_volume0d.py
```diff
@@ -1531,112 +1531,243 @@ def _get_stream_table_contents(self, time_point=0):
...
-
-        # Set defaults where scale factors are missing
-        _fill_miss_with_default("volume", heat_sf_default)
-        _fill_miss_with_default("heat", heat_sf_default)
-        _fill_miss_with_default("work", work_sf_default)
-        _fill_miss_with_default("phase_fraction", phase_frac_sf_default)
-
-        if hasattr(self, "energy_holdup"):
-            for (t, p), v in self.energy_holdup.items():
+        # Set scaling for variables in order of appearance -----------------------
```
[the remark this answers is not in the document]
eslickj: I think it's okay here.  This is called after the model is completely built (normally) so any order is fine.

idaes/core/reaction_base.py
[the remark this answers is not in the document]
eslickj: I think part of the reason it is where it is, is that it is designed for things with a parameter block.  The state block looks to the parameter block for a default scale factor.  In general process blocks aren't configured that way, and the way we use this it doesn't necessarily make sense in general process blocks.

We could think more about setting default scaling factors in general.

idaes/core/util/testing.py
```diff
@@ -309,6 +310,9 @@ def build(self):
         # Attribute to switch flow basis for testing
         self.basis_switch = 1
 
+        self.set_default_scaling("reaction_rate", 101, 'r1')
```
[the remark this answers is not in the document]
eslickj: The reason for this order is that the index is optional.  But I agree.  Style-wise, it may be better to use named arguments and flip it. ```self.set_default_scaling("raction_rate", index="r1", value=101)```

eslickj:
What's the status on this?  Looks like @andrewlee94 requested changes a while back, but it has been a while since anything changed.  Let me know if you need my help or are ready for re-review.

TimBartholomew:
@eslickj there's been no change since @andrewlee94's comments. I've been pushing it off to the next week for a while now.

There are a couple of test failures (a circular import loop that fails for python 3.6 and then something happening with the HX1D model with constraint_scaling_transform).

idaes/core/control_volume0d.py
TimBartholomew: Oh I didn't know what reaction extent was, I assumed it was the standard term "extent of reaction" which goes from 0 to 1.

lbianchi-lbl:
Re: the import errors on Python 3.6, from a cursory look it seems that it could be a bug in 3.6 (then fixed in 3.7), rather than a documented API/functionality change; see e.g. https://bugs.python.org/issue30024.

I'll look more carefully into this, and try to come up with a fix.

idaes/core/reaction_base.py
```diff
@@ -31,7 +31,14 @@
...
 from idaes.core.util.misc import add_object_reference
-import idaes.core.util.scaling as iscale
+try:
+    # WHY on Python 3.6, this fails with "AttributeError: module 'idaes' has no attribute 'core'"
+    # this is likely due to a bug/limitation in how the Python import mechanism resolves circular imports
+    # for more information, see https://stackoverflow.com/questions/24807434
+    # and the official Python bug report: http://bugs.python.org/issue30024
+    import idaes.core.util.scaling as iscale
+except (AttributeError, ImportError):
+    from idaes.core.util import scaling as iscale
```
andrewlee94, line 41: As a more general solution, would it be better to just import scaling this way and be done with it?

I.e. Is there a benefit to trying the other way first with the `try/except`?

andrewlee94 (approved):
Looks good to me.

idaes/core/reaction_base.py
lbianchi-lbl: That would work, too. My goal was to emphasize the fact that the original syntax caused an error, but the same could be achieved with a comment.

andrewlee94 (approved):
Subject to the tests finally passing, I think this is ready to merge.

lbianchi-lbl (approved):
Rubber-stamped!

PR #58, opened 2020-11-11
Address all pylint errors in the current codebase

lbianchi-lbl:
## Summary/Motivation

Get rid of all pylint errors to obtain a "clean slate" so that pylint can be activated as a required check for future PRs.

## Strategy

This PR implement changes aimed to address everything that pylint reports as an error in different ways, depending if the reported error is a *false positive*, i.e. a statement that pylint considers a cause of runtime error, but that doesn't actually cause errors at runtime, or a *true positive*, i.e. a statement that would (or does) result in a runtime error if that portion of the code is run:

- For **false positives**, the fixes only affect pylint and **do not change the runtime behavior of the code**
  - For *one-off* errors, add a [pylint directive](http://pylint.pycqa.org/en/latest/user_guide/message-control.html) instructing pylint to ignore a particular error for that statement or scope
  - For *systematic* errors, i.e. a construct or implementation that's used in multiple places in the codebase, develop general solutions that address all related error at the same time. Typically this is done by either:
    - Adding a *global directive* affecting the entire codebase as a pylint configuration option
    - Creating a [pylint transform plugin](http://pylint.pycqa.org/en/latest/how_tos/transform_plugins.html) that "explains" the offending construct to pylint so that it matches the runtime behavior
- For **true positives**, more care is needed as the fixes **will change the runtime behavior of the code**
  - If the cause of the error is clear, the intended behavior of the code is unambiguous, and the fixed implementation has no or very little chance of affecting the rest of the code, then I'll try to fix the issue myself, asking reviewers to confirm that the fix is valid
  - If these conditions do not apply, I'll refrain from making assumptions and ask the original author(s) to review the affected statement and comment or suggest a fix
  - For intermediate cases, I'll suggest a fix in a code comment, but refrain from actually implementing that fix until it has been reviewed
  - For all cases, I've marked the comment with `PYLINT-TODO` so that it's easy to find these fixes within the codebase

## Changes proposed in this PR

- Create `.pylint` top-level directory to store all pylint-related code and configuration
- Move pylint configuration to `pylintrc` file
- Add pylint transform plugins to address systematic false-positive errors specific to IDAES and/or Pyomo
- Add custom pylint reporter to display a real-time log of errors in the CI interface and simultaneously save a machine-readable (JSON) dump of the error messages for further
- Multiple unrelated changes throughout the codebase addressing true-positive errors detected by pylint

## Reviewing

- Changes that are already implemented in the code (i.e. anything that can be reviewed from the usual GitHub "[Files changed](https://github.com/IDAES/idaes-pse/pull/58/files)" interface): please feel free to review as normal
- Changes that are not implemented in the code: **UPDATE 2021-05-21** All changes, including adding directives for errors still needing to be addressed, have been made in the code; see https://github.com/IDAES/idaes-pse/pull/58#issuecomment-846317973 for details
  - ~~Remaining errors are listed in the checkist below. There is a link to each affected file in the header, and links pointing to the specific line where the error was detected (using the GitHub `/blame/` URL endpoint to display the edit history of the affected line)~~
  - ~~I'll be pinging the author(s) of the affected code (as indicated by `git blame` and/or previous conversations) both by requesting a PR review and by @-mentioning them pointing to specific lines; of course, everyone is more than welcome to chime in and clarify~~
  - ~~The same applies to errors with an unchecked checkbox where I haven't yet mentioned anyone (either for lack of a clear author, because I forgot, etc)~~
  - ~~In general, I expect that a number of these errors to be in parts of the code that are rarely or not used anymore (which is why they haven't been fixed until now); if this is the case, let me know if it makes  sense for me to try to address the pylint errors  at all or they can be skipped (because e.g. the files will be removed or deprecated soon)~~
- pylint-specific functionality: as mentioned above, this only affects pylint and not the IDAES code itself, but feel free to ask questions about the intended behavior, implementation, impact on existing CI workflows, etc

---

## Errors left to fix (list compiled on 2021-05-07, redundant as of 2021-05-21)

- The set of errors still needing to be address is now encoded directly in the code through `# PYLINT-TODO` sentinel comments; see https://github.com/IDAES/idaes-pse/pull/58#issuecomment-846317973 for details

<details>
<summary>Click to display the original checklist</summary>
 
#### [`idaes.core.reaction_base`](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/core/reaction_base.py)
- [x] At line 536, in `ReactionBlockDataBase.__getattr__`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/core/reaction_base.py#L536)) (`E1305`) `Too many arguments for format string` (`too-many-format-args`)
- ~~@andrewlee94: the error comes from the format string having two "slots" but three arguments being used: `.format(self.name, attr, attr)`. Would it be OK to change that to `.format(self.name, attr)`?~~

#### [`idaes.core.util.dyn_utils`](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/core/util/dyn_utils.py)
- [x] At line 614, in `copy_non_time_indexed_values`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/core/util/dyn_utils.py#L614)) (`E0602`) `Undefined variable 'varname'` (`undefined-variable`)
- ~~@Robbybp what should be used here instead of the non-existing `varname`?~~

#### [`idaes.core.util.unit_costing`](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/core/util/unit_costing.py)
- [ ] At line 411, in `pressure_changer_costing`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/core/util/unit_costing.py#L411)) (`E1305`) `Too many arguments for format string` (`too-many-format-args`)
- [ ] At line 438, in `pressure_changer_costing.base_pump_rule`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/core/util/unit_costing.py#L438)) (`E1305`) `Too many arguments for format string` (`too-many-format-args`)
- [ ] At line 506, in `pressure_changer_costing`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/core/util/unit_costing.py#L506)) (`E1305`) `Too many arguments for format string` (`too-many-format-args`)
- [ ] At line 574, in `pressure_changer_costing`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/core/util/unit_costing.py#L574)) (`E1305`) `Too many arguments for format string` (`too-many-format-args`)
- [ ] At line 661, in `pressure_changer_costing`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/core/util/unit_costing.py#L661)) (`E1305`) `Too many arguments for format string` (`too-many-format-args`)
- [ ] At line 858, in `platforms_ladders.CPL_rule`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/core/util/unit_costing.py#L858)) (`E1305`) `Too many arguments for format string` (`too-many-format-args`)
- [ ] At line 865, in `platforms_ladders`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/core/util/unit_costing.py#L865)) (`E1305`) `Too many arguments for format string` (`too-many-format-args`)
- [ ] At line 1005, in `fired_heater_costing.CB_rule`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/core/util/unit_costing.py#L1005)) (`E1305`) `Too many arguments for format string` (`too-many-format-args`)
- @MAZamarripa , all of these look similar (and similarly easy to fix). Could you take a look and let me know what the correct format args are supposed to be?

#### [`idaes.core.util.convergence.convergence_base`](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/core/util/convergence/convergence_base.py)
- [ ] At line 652, in `Stats.to_dmf`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/core/util/convergence/convergence_base.py#L652)) (`E0602`) `Undefined variable 'stats'` (`undefined-variable`)
- @eslickj the variable `stats` doesn't exist, but this is a method of the `Stats` class. Should `stats` be `self` instead?

#### [`idaes.core.util.convergence.mpi_utils`](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/core/util/convergence/mpi_utils.py)
- [ ] At line 137, in `ParallelTaskManager.allgather_global_data`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/core/util/convergence/mpi_utils.py#L137)) (`E0602`) `Undefined variable 'global_data_list_of_lists'` (`undefined-variable`)

#### [`idaes.commands.base`](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/commands/base.py)
- [ ] At line 108: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/commands/base.py#L108)) (`E1120`) `No value for argument 'verbose' in function call` (`no-value-for-parameter`)
- [ ] At line 108: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/commands/base.py#L108)) (`E1120`) `No value for argument 'quiet' in function call` (`no-value-for-parameter`)
- @eslickj is this function still used here? if so, what arguments should be provided?

#### [`idaes.generic_models.unit_models.pressure_changer`](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/generic_models/unit_models/pressure_changer.py)
- [x] At line 620, in `PressureChangerData.model_check`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/generic_models/unit_models/pressure_changer.py#L620)) (`E1305`) `Too many arguments for format string` (`too-many-format-args`)

#### [`idaes.power_generation.flowsheets.supercritical_steam_cycle.supercritical_steam_cycle`](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/power_generation/flowsheets/supercritical_steam_cycle/supercritical_steam_cycle.py)
- [ ] At line 230, in `create_model.mixer_pressure_constraint`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/power_generation/flowsheets/supercritical_steam_cycle/supercritical_steam_cycle.py#L230)) (`E0102`) `function already defined line 182` (`function-redefined`)
- [ ] At line 275, in `create_model.mixer_pressure_constraint`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/power_generation/flowsheets/supercritical_steam_cycle/supercritical_steam_cycle.py#L275)) (`E0102`) `function already defined line 182` (`function-redefined`)
- [ ] At line 326, in `create_model.mixer_pressure_constraint`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/power_generation/flowsheets/supercritical_steam_cycle/supercritical_steam_cycle.py#L326)) (`E0102`) `function already defined line 182` (`function-redefined`)
- @eslickj @andrewlee94: referring to our previous discussion about this in this PR, I'd do one of the following:
  - **A** Add a pylint directive so that this (false-positive) error is ignored (my choice all else being equal)
  - **B** Change the names of the functions so that they're unique (I'd need your input on what new names to use)

#### [`idaes.surrogate.main`](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/main.py)
- [ ] At line 432, in `Pysmo_rbf.handle_results`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/main.py#L432)) (`E1101`) `Instance of 'RadialBasisFunctions' has no 'rbf_generate_expression' member` (`no-member`)
- [ ] At line 484, in `Pysmo_kriging.handle_results`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/main.py#L484)) (`E1101`) `Instance of 'KrigingModel' has no 'kriging_generate_expression' member` (`no-member`)
- @mengleCMU  both of these errors are due non-existing methods, but the classes have methods with similar names that do exist: can you check if my guesses/assumptions in the comments above the affected line makes sense?

#### [`idaes.surrogate.ripe`](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/ripe/__init__.py)
- [ ] At line 32: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/ripe/__init__.py#L32)) (`E0603`) `Undefined variable name 'massact' in __all__` (`undefined-all-variable`)

#### [`idaes.surrogate.ripe.write`](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/ripe/write.py)
- [ ] At line 19, in `writeripe`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/ripe/write.py#L19)) (`E0401`) `Unable to import 'ripe'` (`import-error`)
- [ ] At line 73, in `writemodel`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/ripe/write.py#L73)) (`E0401`) `Unable to import 'ripe'` (`import-error`)
- [ ] At line 74, in `writemodel`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/ripe/write.py#L74)) (`E0401`) `Unable to import 'alamopy'` (`import-error`)
- [ ] At line 294, in `writeminlp`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/ripe/write.py#L294)) (`E0401`) `Unable to import 'ripe'` (`import-error`)
- @mengleCMU these errors are fixable, but the file starts with a comment saying that it is no longer used. Can you clarify what is the status of the file, and whether it would make sense for me to try and address the pylint errors or not?

#### [`idaes.surrogate.helmet.Helmet`](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/helmet/Helmet.py)
- [ ] At line 243, in `viewMultResults`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/helmet/Helmet.py#L243)) (`E1101`) `Module 'idaes.surrogate.helmet.Plotting' has no 'multSSECombo' member` (`no-member`)

#### [`idaes.surrogate.helmet.GAMSDataWrite`](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/helmet/GAMSDataWrite.py)
- [ ] At line 214, in `SNDdt`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/helmet/GAMSDataWrite.py#L214)) (`E1306`) `Not enough arguments for format string` (`too-few-format-args`)
- [ ] At line 215, in `SNDdt`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/helmet/GAMSDataWrite.py#L215)) (`E1306`) `Not enough arguments for format string` (`too-few-format-args`)
- [ ] At line 216, in `SNDdt`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/helmet/GAMSDataWrite.py#L216)) (`E1306`) `Not enough arguments for format string` (`too-few-format-args`)

#### [`idaes.surrogate.alamopy.doalamo`](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/alamopy/doalamo.py)
- [ ] At line 462, in `getTrainingData`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/alamopy/doalamo.py#L462)) (`E1102`) `almerror is not callable` (`not-callable`)
- [ ] At line 464, in `getTrainingData`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/alamopy/doalamo.py#L464)) (`E1102`) `almerror is not callable` (`not-callable`)
- [ ] At line 635, in `addBasisConstraints`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/alamopy/doalamo.py#L635)) (`E0602`) `Undefined variable 'group_constraints'` (`undefined-variable`)
- [ ] At line 637, in `addBasisConstraints`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/alamopy/doalamo.py#L637)) (`E0602`) `Undefined variable 'basis_constraints'` (`undefined-variable`)
- [ ] At line 801, in `parseKwargs`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/alamopy/doalamo.py#L801)) (`E1102`) `almerror is not callable` (`not-callable`)
- [ ] At line 996, in `readTraceFile`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/alamopy/doalamo.py#L996)) (`E0602`) `Undefined variable 'writethis'` (`undefined-variable`)
- [ ] At line 1213, in `_construct_mock`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/alamopy/doalamo.py#L1213)) (`E0602`) `Undefined variable 'debug'` (`undefined-variable`)

#### [`idaes.surrogate.alamopy.allcard`](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/alamopy/allcard.py)
- [ ] At line 22, in `allcard`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/surrogate/alamopy/allcard.py#L22)) (`E0401`) `Unable to import 'alamopy.writethis'` (`import-error`)

</details>

---

## Original Summary/Motivation (superseded and only partially relevant as of 2021-05-07)

<details>
Transferred here from IDAES/idaes-dev#1199

This PR implements changes aimed at getting rid of false-positives errors, i.e. portions of the code that do not cause runtime errors, but are erroneously flagged as such by pylint because of limitations in pylint's static analysis.

In particular, it introduces a pylint transform plugin that makes pylint aware of `ProcessBlock` subclasses that are dynamically generated at runtime. This eliminates the pylint errors described in IDAES/idaes-dev#1159 without requiring any changes to the code.

This is part of the efforts towards IDAES/idaes-dev#1149. Specifically, this PR contains only changes that affect pylint exclusively, and thus do not alter the runtime behavior of the code in any way, i.e.:

- pylint configuration
- pylint commands in the CI configuration
- pylint directives expressed as comments throughout the codebase

The idea is to keep those sets of changes that do affect the code itself (and that therefore could alter the runtime behavior) in dedicated PRs, to streamline both the fixes and the review process. In other words, reviews for this PR can concentrate entirely on the pylint analysis, without having to take into account compatibility, introducing regressions, or other code-specific considerations.

In addition, these changes aggregate most of the configuration into the pylint RC file located at .pylint/pylintrc, allowing to share the same settings among different use cases in addition to CI checks, e.g. direct invocation, and integration with IDEs/editors supporting static analysis with pylint.

## Changes proposed in this PR

- Add pylint transform plugin addressing `ProcessBlock` subclasses false positives
- Restructure pylint setup so that the same configuration can be reused in multiple contexts such as CI and IDE/editor code check

</details>

---

### Legal Acknowledgement

### DIFF
--- a/.pylint/idaes_transform.py
+++ b/.pylint/idaes_transform.py
@@ -8,7 +8,7 @@ import functools
 import logging
 
 import astroid
-from astroid.builder import extract_node
+from astroid.builder import extract_node, parse
 
 
 _logger = logging.getLogger('pylint.ideas_plugin')
@@ -78,7 +78,10 @@ def create_declared_class_node(decorated_cls_node: astroid.ClassDef):
 
 
 def is_idaes_module(mod_node: astroid.Module):
-    return 'idaes' in mod_node.name
+    mod_name = mod_node.name
+    if mod_name:
+        _display(f'analyzing module: {mod_name}')
+    return 'idaes' in mod_name
 
 
 def register_process_block_class(decorated_cls_node: astroid.ClassDef):
@@ -101,6 +104,35 @@ def register_process_block_classes(mod_node: astroid.Module):
         _display(f'{decorated_cls_node.name} -> {decl_cls_node.name}')
 
 
+def is_base_pyomo_var_class(node):
+    try:
+        _display(f'node.qname()={node.qname()}')
+        return 'pyomo.core.base.var.Var' in node.qname()
+    except AttributeError:
+        pass
+    return False
+
+
+def get_concrete_pyomo_var_class(node: astroid.ClassDef, context=None) -> astroid.ClassDef:
+    # node_from_import = extract_node('from pyomo.core.base.var import SimpleVar; SimpleVar')
+    # simple_var_cls_node = astroid.helpers.safe_infer(node_from_import)
+    clsdef_code = """
+    from pyomo.base.var import SimpleVar, IndexedVar
+    class ConcreteVar(SimpleVar, IndexedVar):
+        pass
+"""
+    simple_var_cls_node = extract_node(clsdef_code)
+    print(simple_var_cls_node)
+    return simple_var_cls_node
+
+
+def infer_concrete_var_instance(node: astroid.ClassDef, context=None):
+    _display(f'abstract var class: {node}')
+    concrete_cls_node = get_concrete_pyomo_var_class(node, context=context)
+    _display(f'concrete var class: {concrete_cls_node}')
+    return iter([concrete_cls_node.instantiate_class()])
+
+
 def register(linter):
     "This function needs to be defined for the plugin to be picked up by Pylint"
 
@@ -110,4 +142,8 @@ astroid.MANAGER.register_transform(
     # but it doesn't seem to be the case at this point
     # astroid.ClassDef, register_process_block_class, has_declare_block_class_decorator
     astroid.Module, register_process_block_classes, is_idaes_module,
+)
+
+astroid.MANAGER.register_transform(
+    astroid.ClassDef, astroid.inference_tip(infer_concrete_var_instance), is_base_pyomo_var_class
 )
--- a/.pylint/idaes_transform.py
+++ b/.pylint/idaes_transform.py
@@ -6,6 +6,7 @@ See #1159 for more information.
 """
 import functools
 import logging
+import typing
 
 import astroid
 from astroid.builder import extract_node, parse
@@ -104,59 +105,6 @@ def register_process_block_classes(mod_node: astroid.Module):
         _display(f'{decorated_cls_node.name} -> {decl_cls_node.name}')
 
 
-def is_base_pyomo_var_class(node):
-    try:
-        _display(f'node.qname()={node.qname()}')
-        return 'pyomo.core.base.var.Var' in node.qname()
-    except AttributeError:
-        pass
-    return False
-
-
-def get_concrete_pyomo_var_class(node: astroid.ClassDef, context=None) -> astroid.ClassDef:
-    # node_from_import = extract_node('from pyomo.core.base.var import SimpleVar; SimpleVar')
-    # simple_var_cls_node = astroid.helpers.safe_infer(node_from_import)
-    clsdef_code = """
-    from pyomo.base.var import SimpleVar, IndexedVar
-    class ConcreteVar(SimpleVar, IndexedVar):
-        pass
-"""
-    simple_var_cls_node = extract_node(clsdef_code)
-    return simple_var_cls_node
-
-
-def is_base_pyomo_rangeset_class(node):
-    try:
-        return 'pyomo.core.base.set.RangeSet' in node.qname()
-    except AttributeError:
-        pass
-    return False
-
-
-def get_concrete_pyomo_rangeset_class(node: astroid.ClassDef, context=None) -> astroid.ClassDef:
-    clsdef_code = """
-    from pyomo.base.set import RangeSet, _FiniteRangeSetData
-    class ConcreteRangeSet(RangeSet, _FiniteRangeSetData):
-        pass
-"""
-    concrete_cls_node = extract_node(clsdef_code)
-    return concrete_cls_node
-
-
-def infer_concrete_var_instance(node: astroid.ClassDef, context=None):
-    _display(f'abstract var class: {node}')
-    concrete_cls_node = get_concrete_pyomo_var_class(node, context=context)
-    _display(f'concrete var class: {concrete_cls_node}')
-    return iter([concrete_cls_node.instantiate_class()])
-
-
-def infer_concrete_rangeset_instance(node: astroid.ClassDef, context=None):
-    _display(f'abstract var class: {node}')
-    concrete_cls_node = get_concrete_pyomo_rangeset_class(node, context=context)
-    _display(f'concrete var class: {concrete_cls_node}')
-    return iter([concrete_cls_node.instantiate_class()])
-
-
 def is_config_block_class(node: astroid.ClassDef):
     # NOTE might be necessary to use node.qname() instead, but that's more prone to breaking
     # if the internal package structure is changed
@@ -179,6 +127,30 @@ def disable_attr_checks_on_slots(node: astroid.ClassDef):
     del node.locals['__slots__']
 
 
+def has_conditional_instantiation(node: astroid.ClassDef, context=None):
+    if 'pyomo' not in node.qname():
+        return
+    try:
+        # check if the class defines a __new__()
+        dunder_new_node: astroid.FunctionDef = node.local_attr('__new__')[0]
+    except astroid.AttributeInferenceError:
+        return False
+    else:
+        # _display(node)
+        # find all return statements; if there's more than one, assume that instances are created conditionally,
+        # and therefore the type of the instantiated object cannot be known with static analysis
+        # to be more accurate, we should check for If nodes as well as maybe the presence of other __new__() calls
+        return_statements = list(dunder_new_node.nodes_of_class(astroid.node_classes.Return))
+        return len(return_statements) > 1
+
+
+def make_node_create_uninferable_instance(node: astroid.ClassDef, context=None):
+    def _instantiate_uninferable(*args, **kwargs):
+        return astroid.Uninferable()
+    node.instantiate_class = _instantiate_uninferable
+    return node
+
+
 def register(linter):
     "This function needs to be defined for the plugin to be picked up by Pylint"
 
@@ -191,13 +163,12 @@ astroid.MANAGER.register_transform(
 )
 
 astroid.MANAGER.register_transform(
-    astroid.ClassDef, astroid.inference_tip(infer_concrete_var_instance), is_base_pyomo_var_class
-)
-
-astroid.MANAGER.register_transform(
-    astroid.ClassDef, astroid.inference_tip(infer_concrete_rangeset_instance), is_base_pyomo_rangeset_class
-)
-
-astroid.MANAGER.register_transform(
     astroid.ClassDef, disable_attr_checks_on_slots, is_config_block_class
+)
+
+
+astroid.MANAGER.register_transform(
+    astroid.ClassDef,
+    make_node_create_uninferable_instance,
+    predicate=has_conditional_instantiation
 )
--- a/.pylint/idaes_transform.py
+++ b/.pylint/idaes_transform.py
@@ -6,6 +6,7 @@ See #1159 for more information.
 """
 import functools
 import logging
+import time
 import typing
 
 import astroid
@@ -80,8 +81,6 @@ def create_declared_class_node(decorated_cls_node: astroid.ClassDef):
 
 def is_idaes_module(mod_node: astroid.Module):
     mod_name = mod_node.name
-    if mod_name:
-        _display(f'analyzing module: {mod_name}')
     return 'idaes' in mod_name
 
 
@@ -151,10 +150,6 @@ def make_node_create_uninferable_instance(node: astroid.ClassDef, context=None):
     return node
 
 
-def register(linter):
-    "This function needs to be defined for the plugin to be picked up by Pylint"
-
-
 astroid.MANAGER.register_transform(
     # NOTE both these options were tried to see if there was a difference in performance,
     # but it doesn't seem to be the case at this point
@@ -172,3 +167,49 @@ astroid.MANAGER.register_transform(
     make_node_create_uninferable_instance,
     predicate=has_conditional_instantiation
 )
+
+
+from pylint.reporters import BaseReporter, text, JSONReporter
+
+
+class MultiReporter(BaseReporter):
+    name = 'multi'
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.console = text.ColorizedTextReporter()
+        self.json = JSONReporter(output=open('pylint.json', 'w'))
+        self._start = None
+
+    def __iter__(self):
+        yield self.console
+        yield self.json
+
+    def handle_message(self, msg):
+        for rep in self:
+            rep.handle_message(msg)
+
+    def writeln(self, **kwargs):
+        self.console.writeln(**kwargs)
+
+    @property
+    def elapsed(self):
+        f = time.clock
+        if self._start is None:
+            self._start = f()
+            return 0.
+        return f() - self._start
+
+    def on_set_current_module(self, module, filepath):
+        self.writeln(string=f'{self.elapsed:07.3f} {filepath}')
+        self.console._template = self.console.line_format
+
+    def display_messages(self, layout):
+        self.json.display_messages(layout)
+
+    def _display(self, layout):
+        ...
+
+
+def register(linter):
+    "This function needs to be defined for the plugin to be picked up by pylint"
+    linter.register_reporter(MultiReporter)
--- a/.pylint/idaes_transform.py
+++ b/.pylint/idaes_transform.py
@@ -1,0 +1,113 @@
+"""
+Pylint transform plugin to make Pylint aware of custom `ProcessBlock` classes
+dynamically created through the `declare_process_block_class()` decorator.
+
+See #1159 for more information.
+"""
+import functools
+import logging
+
+import astroid
+from astroid.builder import extract_node
+
+
+_logger = logging.getLogger('pylint.ideas_plugin')
+
+
+# TODO figure out a better way to integrate this with pylint logging and/or verbosity settings
+_display = _notify = print
+
+
+def has_declare_block_class_decorator(cls_node, decorator_name="declare_process_block_class"):
+    if 'idaes' not in cls_node.root().name:
+        return False
+    decorators = cls_node.decorators
+    if not decorators:
+        return False
+    for dec_subnode in decorators.nodes:
+        if hasattr(dec_subnode, 'func'):
+            # this is true for decorators with arguments
+            return dec_subnode.func.as_string() == decorator_name
+    return False
+
+
+# this will be called N times if running with N processes
+# returning the cached result on subsequent calls
+@functools.lru_cache(maxsize=1)
+def get_base_class_node():
+    _notify('Getting base class node')
+    import_node = extract_node('from idaes.core.process_block import ProcessBlock; ProcessBlock')
+    cls_node = next(import_node.infer())
+    return cls_node
+
+
+def add_attribute_nodes(node: astroid.ClassDef, attr_names):
+    for attr_name in attr_names:
+        rhs_node = astroid.Unknown(
+            lineno=node.lineno,
+            parent=node,
+        )
+        node.locals[attr_name] = [rhs_node]
+        node.instance_attrs[attr_name] = [rhs_node]
+
+
+def create_declared_class_node(decorated_cls_node: astroid.ClassDef):
+    decorators = decorated_cls_node.decorators
+    call = decorators.nodes[0]
+    name_arg_node = call.args[0]
+    decl_class_name = name_arg_node.value
+
+    base_class_node = get_base_class_node()
+    decl_class_node = astroid.ClassDef(
+        decl_class_name,
+        # TODO the real doc should be available as the "doc" kwarg of the decorator
+        # but it's not clear if we're going to need it anyway
+        doc=f"Declared from {decorated_cls_node.name}",
+        parent=decorated_cls_node.parent,
+        lineno=decorated_cls_node.lineno,
+    )
+    decl_class_node.bases.extend(
+        [
+            base_class_node,
+            decorated_cls_node,
+        ]
+    )
+    # doesn't seem to be needed at the moment
+    # add_attribute_nodes(decl_class_node, ['_orig_name', '_orig_module'])
+    return decl_class_node
+
+
+def is_idaes_module(mod_node: astroid.Module):
+    return 'idaes' in mod_node.name
+
+
+def register_process_block_class(decorated_cls_node: astroid.ClassDef):
+    module_node = decorated_cls_node.parent
+    _display(module_node.name)
+    decl_class_node = create_declared_class_node(decorated_cls_node)
+
+
+def iter_process_block_data_classes(mod_node: astroid.Module):
+    for node in mod_node.body:
+        if isinstance(node, astroid.ClassDef):
+            if has_declare_block_class_decorator(node):
+                yield node
+
+
+def register_process_block_classes(mod_node: astroid.Module):
+    _display(mod_node.name)
+    for decorated_cls_node in iter_process_block_data_classes(mod_node):
+        decl_cls_node = create_declared_class_node(decorated_cls_node)
+        _display(f'{decorated_cls_node.name} -> {decl_cls_node.name}')
+
+
+def register(linter):
+    "This function needs to be defined for the plugin to be picked up by Pylint"
+
+
+astroid.MANAGER.register_transform(
+    # NOTE both these options were tried to see if there was a difference in performance,
+    # but it doesn't seem to be the case at this point
+    # astroid.ClassDef, register_process_block_class, has_declare_block_class_decorator
+    astroid.Module, register_process_block_classes, is_idaes_module,
+)
--- a/.pylint/idaes_transform.py
+++ b/.pylint/idaes_transform.py
@@ -15,7 +15,7 @@ _logger = logging.getLogger('pylint.ideas_plugin')
 
 
 # TODO figure out a better way to integrate this with pylint logging and/or verbosity settings
-_display = _notify = print
+_display = _notify = lambda *a, **kw: None
 
 
 def has_declare_block_class_decorator(cls_node, decorator_name="declare_process_block_class"):
@@ -122,13 +122,37 @@ def get_concrete_pyomo_var_class(node: astroid.ClassDef, context=None) -> astroi
         pass
 """
     simple_var_cls_node = extract_node(clsdef_code)
-    print(simple_var_cls_node)
     return simple_var_cls_node
+
+
+def is_base_pyomo_rangeset_class(node):
+    try:
+        return 'pyomo.core.base.set.RangeSet' in node.qname()
+    except AttributeError:
+        pass
+    return False
+
+
+def get_concrete_pyomo_rangeset_class(node: astroid.ClassDef, context=None) -> astroid.ClassDef:
+    clsdef_code = """
+    from pyomo.base.set import RangeSet, _FiniteRangeSetData
+    class ConcreteRangeSet(RangeSet, _FiniteRangeSetData):
+        pass
+"""
+    concrete_cls_node = extract_node(clsdef_code)
+    return concrete_cls_node
 
 
 def infer_concrete_var_instance(node: astroid.ClassDef, context=None):
     _display(f'abstract var class: {node}')
     concrete_cls_node = get_concrete_pyomo_var_class(node, context=context)
+    _display(f'concrete var class: {concrete_cls_node}')
+    return iter([concrete_cls_node.instantiate_class()])
+
+
+def infer_concrete_rangeset_instance(node: astroid.ClassDef, context=None):
+    _display(f'abstract var class: {node}')
+    concrete_cls_node = get_concrete_pyomo_rangeset_class(node, context=context)
     _display(f'concrete var class: {concrete_cls_node}')
     return iter([concrete_cls_node.instantiate_class()])
 
@@ -168,6 +192,10 @@ astroid.MANAGER.register_transform(
 
 astroid.MANAGER.register_transform(
     astroid.ClassDef, astroid.inference_tip(infer_concrete_var_instance), is_base_pyomo_var_class
+)
+
+astroid.MANAGER.register_transform(
+    astroid.ClassDef, astroid.inference_tip(infer_concrete_rangeset_instance), is_base_pyomo_rangeset_class
 )
 
 astroid.MANAGER.register_transform(
--- a/.pylint/idaes_transform.py
+++ b/.pylint/idaes_transform.py
@@ -133,6 +133,28 @@ def infer_concrete_var_instance(node: astroid.ClassDef, context=None):
     return iter([concrete_cls_node.instantiate_class()])
 
 
+def is_config_block_class(node: astroid.ClassDef):
+    # NOTE might be necessary to use node.qname() instead, but that's more prone to breaking
+    # if the internal package structure is changed
+    return 'ConfigDict' in node.name
+
+
+def disable_attr_checks_on_slots(node: astroid.ClassDef):
+    # ConfigDict/ConfigBlock defines __slots__, which trigger a pylint error
+    # whenever a value is set on a non-slot attribute
+    # in reality, ConfigDict/ConfigBlock objects do support setting attributes at runtime
+    # via their __setattr__() method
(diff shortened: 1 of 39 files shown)


idaes/core/process_base.py
```diff
@@ -268,6 +268,9 @@ def report(self, time_point=0, dof=False, ostream=None, prefix=""):
                           f"Activated Blocks: {nb}")
 
         if performance is not None:
+            # PYLINT-WHY: pylint has no way of knowing that performance is supposed to be dict-like
+            # pylint: disable=unsubscriptable-object
+            # PYLINT-TODO: alternatively, have the function return an empty dict and test with `if performance:`
```
andrewlee94, line 273: I do not see a problem with this suggestion (although I cannot guarantee it won't break something somewhere else). I don't think there was a particular reason for having `get_performance_contents` return `None` by default.

idaes/core/property_base.py
```diff
@@ -198,10 +198,12 @@ def build_state_block(self, *args, **kwargs):
             for i in initialize.keys():
                 initialize[i]["parameters"] = self
 
-        return self.state_block_class(*args,
-                                      **kwargs,
-                                      default=default,
-                                      initialize=initialize)
+        return self.state_block_class(  # pylint: disable=not-callable
```
andrewlee94, line 201: For this case, I don't see any reason why we could not make `state_block_class` a method rather than a `property` - that way pylint should be able to tell that it is callable.

idaes/core/property_meta.py
```diff
@@ -345,6 +345,7 @@ def _create_derived_units(self):
             raise PropertyPackageError(
                 "{} cannot determine derived units, as property package has "
                 "not defined a set of base units.".format(self.name))
+                # PYLINT-TODO: maybe str(self) or self.__class__.__name__?
```
andrewlee94, line 348: What is the specific pylint error/warning associated with this? I am guessing that either fix suggested here would be fine - this might be a case of me thinking that `self` was a Pyomo component (where `name` is defined) when it is in fact a metadata object.

idaes/core/reaction_base.py
```diff
@@ -196,10 +197,12 @@ def build_reaction_block(self, *args, **kwargs):
             for i in initialize.keys():
                 initialize[i]["parameters"] = self
 
-        return self.reaction_block_class(*args,
-                                         **kwargs,
-                                         default=default,
-                                         initialize=initialize)
+        return self.reaction_block_class(  # pylint: disable=not-callable
```
andrewlee94, line 200: We could use the same approach here as for `state_block_class` above.

idaes/core/reaction_base.py
```diff
@@ -44,9 +44,10 @@
 # Some more information about this module
 __author__ = "Andrew Lee, John Eslick"
 
-__all__ = ['ReactionBlockData',
-           'ReactionBlock',
-           'ReactionParameterBlock']
+__all__ = [
+    'ReactionBlock',  # pylint: disable=undefined-all-variable
+    'ReactionParameterBlock'  # pylint: disable=undefined-all-variable
+]
```
andrewlee94, line 50: @eslickj Do we even actually use this `__all__` attribute?

idaes/core/tests/test_control_volume_base.py
```diff
@@ -48,7 +48,7 @@ def test_energy_balance_type():
 
     # Test that error is raised when given non-member
     with pytest.raises(AttributeError):
-        EnergyBalanceType.foo
+        EnergyBalanceType.foo  # pylint: disable=no-member
```
andrewlee94, line 51: I think the better response here would be to find a better way to test the `Enum`, rather than turning off pylint functionality. One could argue there isn't really need to test that an `Enum` returns an `AttributeError` if given a non-member. - we should just check that it has the expected members and be done.

idaes/generic_models/properties/helmholtz/helmholtz.py
```diff
@@ -475,6 +475,7 @@ def p(self, **kwargs):
 
     def v(self, **kwargs):
         blk, delta_liq, delta_vap, tau, x = self.basic_calculations(**kwargs)
+        # PYLINT-TODO check if self.mw (flagged as no-member by pylint) isn't actually self.param.mw
```
andrewlee94, line 478: `self.mw` should be a reference (of some form) to `self.param.mw`. However, they might show up as different objects.

idaes/power_generation/flowsheets/supercritical_steam_cycle/supercritical_steam_cycle.py
```diff
@@ -224,6 +224,10 @@ def mixer_pressure_constraint(b, t):
     @m.fs.hotwell.Constraint(m.fs.time)
     def mixer_pressure_constraint(b, t):
         return b.condensate_state[t].pressure == b.mixed_state[t].pressure
+    # PYLINT-TODO check if the "function-redefined" pylint error can be addressed
```
andrewlee94, line 227: This is a tricky one. Pylint is correctly identifying that the methods are being redefined, although that does not affect code functionality as they are being used in different namespaces (which pylint cannot easily see). This could easily be fixed by changing the name of the method (e.g. appending a prefix or suffix for the unit in question), but we would need to make sure that other parts of the code are not looking for these constraints by name.

Alternatively, using call syntax would work fine, and avoid any issues with changing the name of the constraint. The rules are sufficiently different to preclude writing a common method for all, but each constraint could define a new method with a unique name and not affect the code.

idaes/power_generation/unit_models/helm/condenser_ntu.py
```diff
@@ -293,6 +294,8 @@ def saturation_eqn(b, t):
 
     def set_initial_condition(self):
         if self.config.dynamic is True:
+            # PYLINT-TODO: pylints reports undefined-variable for hot_side and cold_side
```
andrewlee94, line 297: Yes, this should be `self.hot_side` and `self.cold_side`.

idaes/power_generation/unit_models/helm/valve_steam.py
```diff
@@ -56,11 +56,13 @@ def valve_function(b, t):
 def _quick_open_callback(blk):
     @blk.Expression(blk.flowsheet().config.time)
     def valve_function(b, t):
-        return sqrt(b.valve_opening[t])
+        # PYLINT-TODO: check if fix OK for "undefined-variable 'sqrt'"
+        return pyo.sqrt(b.valve_opening[t])
```
andrewlee94, line 60: This looks correct.

idaes/power_generation/unit_models/helm/valve_steam.py
```diff
@@ -56,11 +56,13 @@ def valve_function(b, t):
...
 
 
 def _equal_percentage_callback(blk):
-    blk.alpha = Var(initialize=1, doc="Valve function parameter")
+    # PYLINT-TODO: check if fix OK for "undefined-variable 'Var'"
+    blk.alpha = pyo.Var(initialize=1, doc="Valve function parameter")
```
andrewlee94, line 65: This also looks correct.

idaes/power_generation/unit_models/helm/valve_steam.py
```diff
@@ -198,7 +200,8 @@ def build(self):
         if vfselect == ValveFunctionType.linear:
             _linear_callback(self)
         elif vfselect == ValveFunctionType.quick_opening:
-            _quick_callback(self)
+            # PYLINT-TODO: check if fix OK for "undefined-variable '_quick_callback'"
+            _quick_open_callback(self)
```
andrewlee94, line 204: I think this is also correct.

andrewlee94 (commented):
A few comments on things that I am responsible for or understand. For the most part, these look good, but I would tend towards fixing our issues rather than turning off pylint features.

idaes/power_generation/flowsheets/supercritical_steam_cycle/supercritical_steam_cycle.py
andrewlee94: Short answer - I suggest refactoring the code, which should not be hard.

lbianchi-lbl:
Thanks a lot @andrewlee94, this is exactly the type of review feedback I was looking for! I'll address your comments and reply or confirm the proposed fix as appropriate.

idaes/power_generation/flowsheets/supercritical_steam_cycle/supercritical_steam_cycle.py
lbianchi-lbl: Thanks for the explanation, it was useful for me to understand the context a bit better. On second thought, my personal preference would be to add the pylint directive (leaving the code as it is) rather than making changes to the code:

- I initially thought that the functionality (i.e. the inner function's body) was identical in all cases, but as you point out, this is not the case
- This particular portion of the code doesn't seem to be covered by unit tests (might be by component/integration tests)
- I peeked at the source code for `Constraint` in `pyomo.core.base.constraint` and, unless the implementation used in this part of the IDAES code is significantly simpler, I think I've seen enough ~dark magic~ metaprogramming to make me hesitant to jump in and make the changes without more research or ~amulets~ tests
- In general, I would argue that the type of error that pylint is flagging here does not apply in this particular situation, as the decorator is applied directly to the function definition; therefore, I'd lean more into considering this a false positive

idaes/core/property_base.py
lbianchi-lbl: One consequence that this change would have is that clients of `state_block_class` would have to use it as a callable (`state_block_class()`), i.e. breaking the current API. This would not only require changing all call sites (about 20 from a quick search), but also having to use the (IMO slightly awkward) double-parens syntax to call the class itself (e.g. `state_block = my_block.state_block_class()(**state_block_params)`).

Taken these into account, and considering that I don't think there's anything wrong with the current solution of using a callable through a property (a reasonable strategy to effectively "build" an instance-specific method at runtime), I can see the argument for leaving the pylint directive and avoid changing the code.

idaes/core/property_meta.py
lbianchi-lbl: The error is the missing `name` attribute: `E1101: Instance of 'PropertyClassMetadata' has no 'name' member (no-member)`. If `self.name` not existing is actually expected, I'll go ahead and set it to `str(self)`.

idaes/core/process_base.py
lbianchi-lbl: A quick check showed at least two unit tests in `test_process_base()` that would need to be updated. Apart from the (however minor) chance of breaking the existing API, I'm not sure I agree with pylint that this is an error: personally, for this and similar cases, I don't think there's anything particularly wrong with returning `None` as a sentinel value.

Seeing that all errors of the same type are related to the use of the same `_get_performance_contents()` in derived classes, I would propose to add this specific error (`E1128: Assigning result of a function call, where the function returns None (assignment-from-none)`) to the list of globally ignored checks, and leave the similar but semantically quite different `E1111: Assigning result of a function call, where the function has no return (assignment-from-no-return)` active, as IMO it covers a case that is much more likely to be accidental ("forgetting" to return a value where the client expects some, versus explicitly using `return None`).

idaes/generic_models/properties/helmholtz/helmholtz.py
lbianchi-lbl: Just to be sure I understand for the purpose of addressing the pylint error: is it OK to change `self.mw` to `self.param.mw` here (which would make pylint happy without needing directives), or not?

idaes/core/tests/test_control_volume_base.py
lbianchi-lbl: I see your point, but now I realize that this fix can probably be disregarded as we're (purposefully) not running pylint on test files. I tested it at some point, and it turned out to be lots of headaches (at least 5x runtime and lots of errors, most likely because of pylint) for very little gain (at least at the level of catching runtime errors not covered by tests, as tests are by definition run during... well, test runs).

idaes/__init__.py
```diff
@@ -6,6 +6,12 @@
 import os
 
 from . import config as cfg
```
eslickj, line 8: I don't know why I wrote it like this, but I probably had a reason at some point then changed this around.  As it is, it doesn't make much sense.

We can probably just just change it to ```from . import config``` then change the few ```cfg```'s to config.

idaes/power_generation/flowsheets/supercritical_steam_cycle/supercritical_steam_cycle.py
andrewlee94: I agree that it is a false positive, but I could also see it confusing someone reading the code as well. For me, I am somewhat hesitant to add pylint directive unless absolutely necessary, in case this ends up masking real problems in the future.

That said, this is not something I would ask you to fix - if we were going to change the model code I would ask the powerplant team to do it.

idaes/core/property_meta.py
andrewlee94: This is probably just me think that `name` was defined when it was not - `str(self)` should be fine.

idaes/generic_models/properties/helmholtz/helmholtz.py
andrewlee94: I note that a couple of later methods use `self.params.mw` instead of `self.mw`, so we are not even consistent internally with how we are doing this. I did check the code, and `self.mw` is declared (as part of the `build` method), so it will exist, but it seems pylint doesn't see this (probably because the current method is a peer of build).

Given some other things that could potentially crop up, I think making this `self.params.mw` is probably OK.

idaes/__init__.py
lbianchi-lbl: Sounds good, I've made this change and things seem to work just fine.

Robbybp:
> * At line 614, in `copy_non_time_indexed_values`: ([link](https://github.com/IDAES/idaes-pse/blame/60bfa4f38cafb9bbc9b61c8e71dd051eea6bc62c/idaes/core/util/dyn_utils.py#L614)) (`E0602`) `Undefined variable 'varname'` (`undefined-variable`)
> 
> * @Robbybp what should be used here instead of the non-existing `varname`?

`var_src.name` is the correct replacement. Should I open a PR into this branch or is it easier for you to change?

lbianchi-lbl:
Thanks @Robbybp and @andrewlee94. 3 done, 34 to go!

.pylint/idaes_transform.py
```diff
@@ -0,0 +1,173 @@
+"""
```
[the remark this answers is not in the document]
lbianchi-lbl: Good point, I'm adding it to the dev call agenda.

lbianchi-lbl:
After a first iteration of trying to tackle the remaining pylint errors within this PR, and having in mind our plans for the near future to tackle areas of the codebase with no coverage (which is where all *true positive* pylint errors reside), @ksbeattie and I decided to switch to a slightly different strategy:

- Instead of having to wait for all errors to be fixed before merging in this PR (which is holding us back from enabling the non-bypassed pylint check to be run and enforced on new PRs), I've added a local directive to silence each individual error, together with a short description of what the error refers to and possible ways to address it
- In addition to the pylint directive (a comment of the form `# pylint: disable=...`), I've also added a separate sentinel comment of the form `#  PYLINT-TODO`
- This facilitates searching the codebase for the remaining pylint-detected errors, which can be fixed in smaller and more targeted batches in dedicated PRs
- Since the eventual goal is to address all remaining pylint errors and therefore getting rid of all `PYLINT-TODO`s, I've added a simple check to the pylint action that searches for that pattern using `grep` on the entire codebase to display all  remaining `PYLINT-TODO` items (in addition to any pylint message originating from newly added or modified code)
- Unlike the "unleashed" pylint check, this "`PYLINT-TODO` scan" is only informative, i.e. it will not cause the check to fail if any items are detected, but it can nonetheless serve as an "IOU" to help settle this small "pylint debt"

### TL;DR/Summary

- We'd like to wrap this PR up and use pylint as part of PR checks sooner rather than waiting for all remaining errors to be fixed here
- As of now, all pylint errors are addressed, and this PR is ready to be reviewed and can in principle be merged at any point
- Sentinel comments have been added to the pylint directives bypassing the remaining errors so that it's easy to find and fix them in smaller dedicated future PRs

ksbeattie (dismissed):
Let's get this in, and keep dealing with pylint suggestions as we go.

ksbeattie (approved):
LGTM

Issue #96, opened 2020-11-22
Improvements to distillation unit model


jghouse88:
Pending items:

- Initialization improvement to include a linear temperature profile before activating energy balance
- Add compatibility to scaling framework
- Add compatibility to generic properties framework
- Separate out `make_phase_split` and `propagate_state` methods as utility methods and use these in condenser and reboiler unit models

ksbeattie:
@jghouse88 is this still possible for Aug?  Is there someone else who could take it over?

adowling2:
@agarciadiego Are you using this branch?

lbianchi-lbl:
@jghouse88 @adowling2 we're looking at open PRs in preparation for the August release. Is this work expected to be completed in the next few days? Otherwise, we can keep tracking this PR as part of the November release.

ksbeattie:
@jghouse88, seems like this should get a new assignee.  Can you pick one?

ksbeattie:
@jghouse88 is this still relevant?

andrewlee94:
@ksbeattie It is still very relevant, but we lack the resources to address it right now.

adowling2:
@agarciadiego has been working on extensions about on this. We have ideas on improving initialization.

andrewlee94:
I was thinking @agarciadiego would be a good person to take this over, however he probably won;t have time in the next few months as I believe he is committed to other projects.

Issue #73, opened 2020-08-11
Update tests for distillation units


jghouse88:
From review of IDAES/idaes-dev#1082:

   - Any test that involves a solver (or depends on the result of a solver) should be tagged as a component test. This would be the initialization, solver, solution and conservation tests.
  -  To allow component tests to run independently of unit tests, setting the degrees of freedom should be done in the fixture, not in the dof test itself (otherwise if the component tests are run in isolation, the dof are not set).
  -  Could you separate the different model structures (e.g. state variables) into separate test classes. This will help with debugging in the future, as each test case can be debugged in parallel. As it is, the first failure will mask any subsequent failures.
 - Test other combinations of state variables other than FTPz and FcTP

andrewlee94:
Most of this should be fixed as part of https://github.com/IDAES/idaes-pse/pull/1050 if it hasn't already.

Issue #65, opened 2020-11-14
Add last unit models to library


andrewlee94:
Copied from https://github.com/IDAES/idaes/issues/129

The following needs to be completed before next release (carried over from IDAES/idaes-dev#40):

- [x] Equilibrium reactor

- [x] Tray Model

- [x] Feed Block

- [x] Feed Block with Flash

- [x] Product Block

- [x] Translator Block

- [x] Ideal property package

- [x] CEOS property package

- [x] NRTL property package

- [ ] distillation column model

Documentation will also be required.

PR #97, opened 2020-11-22
Adding tray unit model to support HDA case study

jghouse88:
**NOTE:** When this PR is approved, you *must* do the following so it can be merged:
* Add the "approved" tag to the pull-request
* Trigger tests by pushing a commit or manually re-running in CircleCI

See [this wiki page](https://github.com/IDAES/idaes-dev/wiki/Github-Pull-Request-Workflow-with-Approval-Tag) for details.

----
## Fixes
Partial addresses to #65; #73 

## Summary/Motivation:
- Adding the tray column model and changes to initialization methods of the individual `tray`, `condenser`, and `reboiler` unit models. 
- These commits have been tested with the HDA case study. Once approved, I will open a PR to add the HDA example with distillation on examples-pse. 
- Pending items being tracked in #96

- Tray column unit model with a working initialization routine
- Updates to other unit models to support the tray column model and initialization

### Legal Acknowledgement

----

## Reviewer Checklist
* Reviewer 1
* Reviewer 2


idaes/generic_models/unit_models/distillation/reboiler.py
```diff
@@ -545,40 +560,107 @@ def rule_bottoms_enth(self, t):
...
         else:
-            init_log.warning(
-                "Solver not provided during initialization, proceeding"
-                " with deafult solver in idaes.")
-            solver = get_default_solver()
+            # In this case, outlet temperature has already been fixed by user
+            # and boil up ratio is assumed not to be fixed, otherwise the
+            # system is over specified.
+
+            self.eq_boilup_ratio.deactivate()
```
andrewlee94, line 631: The logic here seems a little off. You deactivate the boilup ratio equation and expect the problem to have 0 degrees of freedom, solve it and then reactivate the boilup ratio equation. This means that either the boilup ratio equation is only used to calculate the boilup ratio (I assume this is the case) in which case why don't you just leave it active, or that the problem with the equation is not square.

idaes/generic_models/unit_models/distillation/reboiler.py
```diff
@@ -545,40 +560,107 @@ def rule_bottoms_enth(self, t):
...
+                    format(idaeslog.condition(res))
+                )
+            else:
+                raise Exception("State vars fixed but degrees of freedom "
+                                "for reboiler is not zero during "
+                                "initialization.")
+
+            self.eq_boilup_ratio.activate()
+
+        if degrees_of_freedom(self) == 0:
```
andrewlee94, line 647: Do you need to check degrees of freedom again? Once we've got this far, shouldn't we know already? This might mean storing the DoF (or a boolean) earlier to use here.

idaes/generic_models/unit_models/distillation/tests/test_tray_column.py
```diff
@@ -0,0 +1,215 @@
...
+from idaes.core.util.testing import get_default_solver, \
+    PhysicalParameterTestBlock, initialization_tester
+
+
+# -----------------------------------------------------------------------------
+# Get default solver for testing
+solver = get_default_solver()
+
+
+@pytest.mark.build
```
andrewlee94, line 41: Need to have a `pytest.unit` mark here.

idaes/generic_models/unit_models/distillation/tests/test_tray_column.py
```diff
@@ -0,0 +1,215 @@
...
+    assert m.fs.unit.tray[5].config.is_feed_tray
+
+    assert hasattr(m.fs.unit, "condenser")
+    assert hasattr(m.fs.unit, "reboiler")
+
+    assert hasattr(m.fs.unit, "liq_stream")
+    assert hasattr(m.fs.unit, "vap_stream")
+
+
+@pytest.mark.build
```
andrewlee94, line 69: Pytest mark again.

idaes/generic_models/unit_models/distillation/tests/test_tray_column.py
```diff
@@ -0,0 +1,215 @@
...
+        m.fs.unit.condenser.reflux_ratio.fix(1.4)
+        m.fs.unit.condenser.condenser_pressure.fix(101325)
+
+        m.fs.unit.reboiler.boilup_ratio.fix(1.3)
+
+        assert degrees_of_freedom(m) == 0
+
+        return m
+
+    @pytest.fixture(scope="class")
```
andrewlee94, line 107: For future, I would suggest doing the two cases as separate test classes, just to keep them separate and help future readers follow what is happening.

idaes/generic_models/unit_models/distillation/tests/test_tray_column.py
```diff
@@ -0,0 +1,215 @@
...
+        m.fs.unit.condenser.condenser_pressure.fix(101325)
+
+        m.fs.unit.reboiler.boilup_ratio.fix(1.3)
+
+        assert degrees_of_freedom(m) == 0
+
+        return m
+
+    @pytest.mark.unit
+    def test_build(self, btx_ftpz, btx_fctp):
```
andrewlee94, line 144: I would suggest breaking this into two separate tests, one for each case.

idaes/generic_models/unit_models/distillation/tests/test_tray_column.py
```diff
@@ -0,0 +1,215 @@
...
+
+        assert hasattr(btx_fctp.fs.unit, "condenser")
+        assert hasattr(btx_fctp.fs.unit, "reboiler")
+
+        assert hasattr(btx_fctp.fs.unit, "liq_stream")
+        assert hasattr(btx_fctp.fs.unit, "vap_stream")
+
+    @pytest.mark.solver
+    @pytest.mark.component
+    def test_initialize(self, btx_ftpz):
```
andrewlee94, line 167: Where are the initialization, solver and solution tests for the other case?

idaes/generic_models/unit_models/distillation/tray.py
```diff
@@ -547,7 +547,46 @@ def rule_flow(self, t, i):
...
+
+                    expr = Expression(self.flowsheet().time,
+                                      phase_set,
+                                      component_set,
+                                      rule=rule_flow)
+                    self.add_component("e_" + local_name + port.local_name,
+                                       expr)
+                    port.add(expr, k)
+                else:
+                    raise PropertyNotSupportedError(
```
andrewlee94, line 585: I would suggest a PropertyPackageError rather than ProeprtyNotSupported - the "not supported" error is intended to indicate that a unit model asked for a property that the property package did not support. This looks like the opposite however - the property package told a unit model to use something the unit model does not know how to handle.

idaes/generic_models/unit_models/distillation/tray_column.py
```diff
@@ -0,0 +1,311 @@
...
+        """Build the model.
+
+        Args:
+            None
+        Returns:
+            None
+        """
+        # Call UnitModel.build to setup dynamics
+        super(TrayColumnData, self).build()
+
```
andrewlee94, line 164: Given that we allow for some critical config argument (e.g. number of trays) to have default values of `None`, I think we need to have some validation here. I agree that `None` is the correct default, but we should tell the user that they need to provide the value if they do not.

idaes/generic_models/unit_models/distillation/tray_column.py
```diff
@@ -0,0 +1,311 @@
...
+                                      self.config.has_heat_transfer,
+                                      "has_pressure_change":
+                                      self.config.has_pressure_change,
+                                      "property_package":
+                                      self.config.property_package,
+                                      "property_package_args":
+                                      self.config.property_package_args}})
+
+        # Add condenser
+        self.condenser = Condenser(
```
andrewlee94, line 202: I assume conditional options for these are future work (i.e. a `has_condenser` option).

idaes/generic_models/unit_models/distillation/tray_column.py
```diff
@@ -0,0 +1,311 @@
...
+                     "property_package": self.config.property_package,
+                     "property_package_args":
+                     self.config.property_package_args})
+
+        # Construct arcs between trays, condenser, and reboiler
+        self._make_arcs()
+        TransformationFactory("network.expand_arcs").apply_to(self)
+
+    def _make_arcs(self):
+        # make arcs
```
andrewlee94, line 223: A comment for future work: One option for avoiding issues with the SD tool would be to write our won constraints for linking trays rather than using Arcs.

A second point - we should use test the Pyomo model reduction tools against this to see if we can eliminate these additional variables and constraints via post-processing. We have a lot of redundant components here, so this would be a really good test case for those tools, and if they work here they will hopefully work for a flowsheet as well.

idaes/generic_models/unit_models/distillation/tray_column.py
```diff
@@ -0,0 +1,311 @@
...
+                return {"source": self.reboiler.vapor_reboil,
+                        "destination": self.tray[i - 1].vap_in}
+            else:
+                return {"source": self.tray[i].vap_out,
+                        "destination": self.tray[i - 1].vap_in}
+
+        self.liq_stream = Arc(self.liq_stream_index, rule=rule_liq_stream)
+        self.vap_stream = Arc(self.vap_stream_index, rule=rule_vap_stream)
+
+    def propagate_stream_state(self, source=None,
```
andrewlee94, line 252: I think we have a utility function that already does this  in `idaes.core.util.initialization`.

andrewlee94 (changes requested):
A number of comments, but I think most of them are just little things.

idaes/generic_models/unit_models/distillation/condenser.py
```diff
@@ -196,10 +174,6 @@ def build(self):
             balance_type=self.config.energy_balance_type,
             has_heat_transfer=True)
 
-        self.control_volume.add_momentum_balances(
```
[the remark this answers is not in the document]
jghouse88: Done.

idaes/generic_models/unit_models/distillation/reboiler.py
```diff
@@ -545,40 +560,107 @@ def rule_bottoms_enth(self, t):
...
 
         init_log = idaeslog.getInitLogger(self.name, outlvl, tag="unit")
         solve_log = idaeslog.getSolveLogger(self.name, outlvl, tag="unit")
 
+        if solver is None:
+            init_log.warning("Solver not provided. Default solver(ipopt) "
+                             " being used for initialization.")
+            solver = get_default_solver()
+
+        if state_args is None:
```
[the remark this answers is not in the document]
jghouse88: Good point.

idaes/generic_models/unit_models/distillation/reboiler.py
jghouse88: I deactivate the boil up ratio constraint which is (`V/B == boilup_ratio`) and fix the outlet reboiler temperature to a temperature within the two phase envelope. This is an easier spec to initialize than by having the constraint active and figuring out the temperature that satisfies the system of equations. Let me know if you have any other suggestions for this step. I based the temperature guess on the following:

```
alpha = 0.5 * (
                self.control_volume.properties_in[0].temperature_dew.value -
                self.control_volume.properties_in[0].
                temperature_bubble.value) + \
                self.control_volume.properties_in[0].temperature_bubble.value

            self.eq_boilup_ratio.deactivate()
            self.control_volume.properties_out[0].temperature.fix(alpha)
```

andrewlee94: That is true for the first part of the if statement I believe, but this is for the case when T was already fixed by the user.

In this case, I think you should just be able to solve for boilup from the beginning (i.e. not need to deactivate it).

jghouse88: Okay, good question but see comment on lines 613-615. If the outlet temperature was fixed, the boil up ratio var cannot be fixed and hence, I have a check to ensure that the DOF=0.  I can do better by raising appropriate error messages when DOF is not 0. Not sure if I answered what you asked for.

idaes/generic_models/unit_models/distillation/reboiler.py
jghouse88: Agreed. Removing this but updated the error message to better explain what is expected to be fixed.

idaes/generic_models/unit_models/distillation/tray_column.py
jghouse88: Added a check.

idaes/generic_models/unit_models/distillation/tray_column.py
jghouse88: Yes. For now, the tray column is a simple conventional column with a condenser and reboiler.

idaes/generic_models/unit_models/distillation/tray_column.py
jghouse88: I still would prefer using Pyomo's `Arcs` for this and instead fix the SD tool that looks at the top level. I still need to file the issue with Pyomo. As for the second point, I agree. Something to check.

idaes/generic_models/unit_models/distillation/tray_column.py
jghouse88: I had to write my own because the utility function did not support indexed arcs.

jghouse88: My idea is to spin off these and the `make_phase_split` methods as distillation only utility methods.

jghouse88:
@andrewlee94 Have addressed your comments. Can you take another look?

idaes/generic_models/unit_models/distillation/tray_column.py
andrewlee94: As a thought, maybe we should update the utility method to optionally handle indexed Arcs then?

jghouse88: Considered it but thought will merge later..

andrewlee94 (approved):
Looks OK to me.

jghouse88:
@eslickj can you please take a look at this when you get time?

jghouse88:
@eslickj Can you take a look at this?

eslickj (approved):

PR #115, opened 2020-12-08
Removing old turbine and valve models that use any prop pack

eslickj:
## Summary/Motivation:
Removes the old power plant turbine and valve models that were not Helmholtz EoS specific.  These models were causing confusion and getting out-of-date.  There will be a couple follow up PRs to update the docs and expand the tests for the remaining steam cycle models.

This is part of the December power plant model release milestone.

### Legal Acknowledgement


eslickj:
Had to bring back the generic steam valve model.  I guess after this, I'll write a generic valve model to replace it.  The valve is use in the controller test, and I missed that.

MAZamarripa (approved):
lgtm

andrewlee94 (approved):
Looks Ok as best I can tell.

PR #117, opened 2020-12-08
New Property Package - CO2-H2O System

anujad95:
## Fixes
NA

## Summary/Motivation:
Addition of a new property package that represents the CO2-H2O system, with ideal vapor and liquid phase.

- New property package - CO2_H2O_Ideal_VLE.py, built on the current IDAES framework
- A property package test - test_ CO2_H2O_Ideal_VLE.py

### Legal Acknowledgement


jghouse88 (dismissed):
LGTM.

idaes/generic_models/properties/core/examples/tests/test_CO2_H2O_Ideal_VLE.py
```diff
@@ -0,0 +1,390 @@
...
+            pytest.approx(1, abs=1e-4)
+        assert model.props[1].mole_frac_phase_comp["Vap", "H2O"].value == \
+            pytest.approx(0.11501, abs=1e-4)
+        assert model.props[1].mole_frac_phase_comp["Vap", "CO2"].value == \
+            pytest.approx(0.88499, abs=1e-4)
+        assert model.props[1].phase_frac["Vap"].value == \
+            pytest.approx(0.56498, abs=1e-4)
+            
+    @pytest.mark.integration
+    def test_temp_swing(self,model):
```
andrewlee94, line 320: You can take the `model` argument out here, as you are not actually using the fixture.

idaes/generic_models/properties/core/examples/tests/test_CO2_H2O_Ideal_VLE.py
```diff
@@ -0,0 +1,390 @@
...
+                              -268740.26687,-265479.46928,-261958.54546,
+                              -258130.54709,-253936.4179,-249300.81058,
+                              -244126.04073,-238283.13381,-231598.19109,
+                              -223830.94744,-214639.75318,-203521.76902,
+                              -189705.16711,-171941.46564,-148070.35054,
+                              -114001.22402,-60937.07508,-3219.88715,
+                              -2631.66029,-2043.09203,-1454.17752,-864.91582,
+                              -275.30623]
+        
+        outvals = zip(expected_vapor_frac,expected_heat_duty)
```
andrewlee94, line 372: Just as an observation (don't go a and change it), but I would have just created the `dict` of expected values directly. The reason being it would be a little more readable and maintainable for future, as you could easily cross reference  temperature with expected results. That said, the current form works so it is fine.

andrewlee94 (changes requested):
Only one trivial change required. And to emphasise - you don't need to change anything for the second comment, it was just something to think about for the future.

idaes/generic_models/properties/core/examples/tests/test_CO2_H2O_Ideal_VLE.py
anujad95: Yes, i will make this change. Thanks.

andrewlee94 (approved):
Looks really good.

jghouse88 (approved):
LGTM

PR #118, opened 2020-12-09
Adding checks on state_bounds argument to find unexpected keys

andrewlee94:
## Summary/Motivation:
A couple of people have noted that the generic properties framework does not inform the user if they mis-type an entry in the `state_bounds` configuration dict. This PR adds checks for unexpected keys and raises a `ConfigurationError` if one is found (or  logs a warning in the case of mole fractions if they are a state variable).

-
-

### Legal Acknowledgement

### DIFF
--- a/idaes/generic_models/properties/core/state_definitions/FPhx.py
+++ b/idaes/generic_models/properties/core/state_definitions/FPhx.py
@@ -23,6 +23,11 @@ from idaes.generic_models.properties.core.generic.utility import \
     get_bounds_from_config
 from idaes.generic_models.properties.core.state_definitions.FTPx import (
     state_initialization)
+from idaes.core.util.exceptions import ConfigurationError
+import idaes.logger as idaeslog
+
+# Set up logger
+_log = idaeslog.getLogger(__name__)
 
 
 # TODO : Need a way to get a guess for T during initialization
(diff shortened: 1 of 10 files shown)


eslickj (approved):
Looks good to me.

idaes/generic_models/properties/core/state_definitions/FPhx.py
```diff
@@ -39,6 +44,23 @@ def define_state(b):
     # TODO: should have some checking to make sure developers implement this properly
     b.always_flash = True
 
+    # Check that only necessary state_bounds are defined
+    expected_keys = ["flow_mol", "enth_mol", "temperature", "pressure"]
```
jghouse88, line 48: Question: Why is `temperature` an expected key when the state definition is `FPhx`?

idaes/generic_models/properties/core/state_definitions/FcPh.py
```diff
@@ -40,6 +45,18 @@ def define_state(b):
     # TODO: should have some checking to make sure developers implement this properly
     b.always_flash = True
 
+    # Check that only necessary state_bounds are defined
+    expected_keys = ["flow_mol_comp", "enth_mol", "temperature", "pressure"]
```
jghouse88, line 49: Same as previous: Why `temperature`?

jghouse88 (approved):
Minor comment. LGTM.

idaes/generic_models/properties/core/state_definitions/FPhx.py
andrewlee94: Temperature is there to allow users to provide bounds for T as they are generally better known than bounds on enthalpy. These are also very helpful (if not necessary) for initializing the flash calculations.

Issue #104, opened 2020-10-19
Improve costing implementation


TimBartholomew:
After adding costing to some of my NAWI work, I think there are a few places where we can change our costing implementation.

Here are two things that I think are easy changes:

1. We should use `**kwargs` in the Unit_Model_costing function call in the get_costing method for each unit model. This way the financial modules don't need to include all the key word arguments.
2.  We should make it standard for financial modules to create a block with all the costing parameters on the flowsheet. This would act similarly as the properties parameter block. This would help with sensitivity analyses and I think is just a better way than an example costing module I saw on IDAES that created a new parameter on the unit model costing block.

Other bigger picture things we should think about:
- Should a "costing package" be more implemented like a "properties package"?
    - where we define it on the flowsheet with `m.fs.costing = financials.CostingParameterBlock"
    - we feed it into unit models with `"costing_package" = m.fs.costing`
    - if a costing package is provided to the unit model, then the unit costing block is created along with the costing variables and constraints. Currently, we must call the get_costing method on each unit model in the flowsheet script to do this.
 - We may need to set up naming conventions for costing. We already do that for what the functions in financial modules must be called, but we don't have standard variable names. Generally the naming conventions won't be as big of a deal as property packages, but it would be important if we envision someone using 2 financial modules (costing packages) in the same flowsheet.

andrewlee94:
Closing this in light of the ongoing development of a new costing framework.

PR #134, opened 2020-12-18
Generic costing: improvements to break the cost into multiple units, initialize costing, and scale model

MAZamarripa:
## Fixes
Improves generic costing methodology
fixes part of issue (Improve costing implementation #104)

## Summary/Motivation:
Improve costing methodology to include number of units to be costed to take advantage of the economics of scale
Improve costing methodology for better initialization and scaling.

- base cost is now an expression
- added base_cost_per_unit, cost number of units
- added number_of_units (Variable that can be continuous or integer)
- initialization method
- scaling method

### Legal Acknowledgement


idaes/core/unit_model.py
```diff
@@ -652,16 +652,26 @@ def initialize(blk, state_args=None, outlvl=idaeslog.NOTSET,
...
         with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
             results = opt.solve(blk, tee=slc.tee)
 
         init_log.info_high(
             "Initialization Step 2 {}.".format(idaeslog.condition(results))
         )
 
+        # if costing block exists, activate and initialize
+        if hasattr(blk, "costing"):
+            import idaes.core.util.unit_costing as costing
```
andrewlee94, line 669: Whilst this might work, I am not sure that I like the solution.

Firstly, it is generally considered good practice to put all imports at the beginning of the file.
Secondly, this isn't really addressing the problem. It just hides it by putting delaying the import until a point where costing has probably been imported by something else already (the thing causing the circular import).

I would much rather we work out what the actual issue is and fix that.

idaes/core/util/unit_costing.py
```diff
@@ -905,61 +965,201 @@ def p_factor_rule(self):
...
+def pfr_costing(self, alignment='horizontal', Mat_factor='carbon_steel',
+                    weight_limit='option1', L_D_range='option1', PL=True):
     vessel_costing(self, alignment=alignment,
                    Mat_factor=Mat_factor,
                    weight_limit=weight_limit,
                    L_D_range=L_D_range,
                    PL=PL)
+
+
+def initialize(self):
```
eslickj, line 1041: This PR generally looks good to me.  I don't know too much about the costing framework and why it is implemented the way it is, but looking at this initialize method makes me want to raise this question again.  Miguel and I talked about this a little, so I'm not going to insist on a different implementation.  I'm just curious about really.

There are a lot of if hasattrs in this.  I would think the normal why to do this would be to have the costing block class be derived from ProcessBase, then build up the initialize methods though inheritance.   Like there is a common costing block, then from that derive one for pumps, from that specific pumps.  Naively that's how I would have done this.

eslickj (approved):
Looks good as far as I can tell.  Had a comment on the general design of the cost models, but I don't expect you to rework it.  Maybe something for the future.

idaes/core/unit_model.py
MAZamarripa: "This approach doesn't contradict Python syntax, as the Python documentation says: "It is customary but not required to place all import statements at the beginning of a module (or script, for that matter)"."

idaes/core/util/unit_costing.py
eslickj: This also addresses (indirectly) to @andrewlee94's import comment.

andrewlee94: I did ask the same question, but we left it this way for now due to time constraints. However, I agree that going forward we should change this to inherit from ProcessBase and have different classes for each type of unit operation, each with their own initialize method.

docs/technical_specs/core/util/unit_costing.rst
```diff
@@ -28,17 +28,24 @@ In addition, the first time `get_costing` is called for a unit operation within
 on the flowsheet object (i.e. `flowsheet.unit.costing`) in order to hold any global parameters relating to costing. The most 
 common of these paramters is the cost normalization parameter based on the year selected by the user.
 
+An initialize method for costing components has been developed. This method is called by the unit.initialize() or it can be imported by the user to initialize costing blocks.
```
andrewlee94, line 31: This is not quite correct, at least from the developer's stand point. I would reword this to be something like:

The unit costing module also contains an `initialize` method which can be used to estimate good initial guesses for costing variables based on the current stat of the associated unit model. This method can be called directly from the `unit_costing` module to initialize a specific costing block, or can be incorporated into a unit model initialization procedure. This method has been incorporated into the `initialize` method of all the models in the core unit model library,

docs/technical_specs/core/util/unit_costing.rst
```diff
@@ -310,7 +323,11 @@ A centrifugal pump is usually driven by an electric motor, the `self.costing.mot
 
 .. math:: self.motor_purchase_cost = FT * self.costing.motor\_base\_cost * (CE_{index}/500)  (Eq. 22.20)
 
-.. math:: self.costing.motor\_base\_cost = \exp{(5.8259 + 0.13141\log{PC} + 0.053255(\log{PC})^{2} + 0.028628(\log{PC})^{3} - 0.0035549(\log{PC})^{4})}  (Eq. 22.19)
+.. math:: self.costing.motor\_base\_cost = self.costing.motor\_base\_cost\_per\unit * self.costing.number\_of\_units
```
andrewlee94, line 326: Given the size of this, I would suggest we break this section up into sub-pages in future, i.e. a separate page for the equations for each type of model. This might be part of the reworking of costing into using ProcessBlocks.

idaes/core/flowsheet_model.py
```diff
@@ -203,10 +203,11 @@ def visualize(self, model_name, **kwargs):
         """
         visualize(self, model_name, **kwargs)
 
-    def get_costing(self, module=costing, year=None):
+    def get_costing(self, module=costing, year=None, integer_n_units=False):
```
andrewlee94, line 206: We probably should add a method doc string here to explain what this method does and what the arguments are. Seeing as this is a public facing method, I can see someone trying to use help on this, so it would be good to provide some information. This would also end up in the autodocs for the flowsheet block class, meaning that people would see the costing methods there and maybe come looking for more details.

idaes/generic_models/unit_models/heat_exchanger.py
```diff
@@ -29,7 +29,7 @@
     Block,
     units as pyunits
 )
-
+from pyomo.util.calc_var_value import calculate_variable_from_constraint
```
andrewlee94, line 32: Why was this added in this PR? I don't see it being used in the new code below?

idaes/generic_models/unit_models/heat_exchanger.py
```diff
@@ -540,10 +538,9 @@ def initialize(
 
         init_log.info("Initialization Completed, {}".format(idaeslog.condition(res)))
         # if costing block exists, activate
-        try:
+        if hasattr(self, "costing"):
             self.costing.activate()
-        except AttributeError:
-            pass
+            costing.initialize(self.costing)
```
andrewlee94, line 543: Did you import the costing module above? I don't see it, so I am wondering if this is being missed by the tests as well.

idaes/generic_models/unit_models/tests/test_flash.py
```diff
@@ -37,7 +37,7 @@
 from idaes.core.util.testing import (get_default_solver,
                                      PhysicalParameterTestBlock,
                                      initialization_tester)
-
+import idaes.core.util.unit_costing as costing
```
andrewlee94, line 40: Is this still needed?

idaes/core/util/unit_costing.py
```diff
@@ -33,15 +35,37 @@ def global_costing_parameters(self, year=None):
     self.CE_index = Param(mutable=True, initialize=ce_index_dic[year],
                           doc='Chemical Engineering Plant Cost Index $ year')
 
+    if integer_n_units not in [True, False]:
+        raise ValueError('integer_n_units must be True or False')
```
andrewlee94, line 39: I would suggest making this the following:

`"{} - integer_n_units must be True or False".format(self.name)`

That way, the user gets some information about which model/block has the issue without needing to try to follow the traceback.

idaes/core/util/unit_costing.py
```diff
@@ -86,19 +110,19 @@ def hx_costing(self, hx_type='U-tube',
     _make_vars(self)
 
     self.L_factor = Param(mutable=True, initialize=1.12,
-                          doc='hx tube length correction factor, FL')
+                          doc='Hx tube length correction factor, FL')
```
andrewlee94, line 113: If the "H" is capitalizes, then I think the "X" should be as well - i.e. "HX"

idaes/core/util/unit_costing.py
```diff
@@ -86,19 +110,19 @@ def hx_costing(self, hx_type='U-tube',
...
 
     self.material_factor = Var(initialize=3.5,
                                domain=NonNegativeReals,
-                               doc='construction material correction'
+                               doc='Construction material correction'
                                'factor - Mat_factor')
 
     self.hx_os = Param(mutable=True, initialize=1.1,
-                       doc='hx oversize factor 1.1 to 1.5',
+                       doc='Hx oversize factor 1.1 to 1.5',
```
andrewlee94, line 125: As above.

idaes/core/util/unit_costing.py
```diff
@@ -262,13 +291,14 @@ def pressure_changer_costing(self, Mat_factor="stain_steel",
...
-    work_hp = pyunits.convert(w, to_units=pyunits.hp)
+    work_hp = pyunits.convert(w, to_units=pyunits.hp) / self.number_of_units
 
     # if compressor is == False, that means pressure changer is a Turbine
     if self.parent_block().config.compressor is False:
         #                           -1*work_hp because work is negative
         def CP_rule(self):
-            return self.purchase_cost == 530*(-1/pyunits.hp * work_hp)**0.81
+            return self.purchase_cost == self.number_of_units \
+                * 530*(-1/pyunits.hp * work_hp)**0.81
```
andrewlee94, line 301: A couple of minor formatting notes (that don't need to be fixed now):

1. PEP* standards say the operator should go before a line break rather than after, i.e.:

```
a * \
b
```

rather than 

```
a \
* b
```

2. Parenthesis tend to make things look neater and more readable than the linebreak character `\`.

idaes/core/util/unit_costing.py
```diff
@@ -463,9 +504,9 @@ def cp_cost_rule(self):
         # if thermodynamic_assumption is not = Pump
         # (pressure changer could be a Fan, Blower, or Compressor)
         else:
-            w = self.parent_block().\
-                work_mechanical[self.parent_block().flowsheet().
-                                config.time.first()]
+            # w = self.parent_block().\
+            #     work_mechanical[self.parent_block().flowsheet().
```
andrewlee94, line 508: Why is all of this commented out? Is it still needed?

idaes/core/util/unit_costing.py
```diff
@@ -599,7 +652,7 @@ def CP_rule(self):
                             CE_index/500)*self.base_cost)
                 self.cp_cost_eq = Constraint(rule=CP_rule)
             else:
-                raise Exception('mover type not supported')
+                raise ValueError('mover type not supported')
```
andrewlee94, line 655: Suggest: `"{} - mover type '{}' not supported. Please see documentation for list of supported mover types.".format(self.name, mover_type)`

andrewlee94 (changes requested):
Mostly just typos and grammar, and a few suggestions for improving error messages.

idaes/generic_models/unit_models/heat_exchanger.py
MAZamarripa: added before the initialize method, removed

idaes/generic_models/unit_models/heat_exchanger.py
MAZamarripa: yes, I imported costing. heat_exchanger testing is correct
hx costing test (we assert that purchase cost is same after initialize and after solving the model):
    m.fs.unit.initialize()

    assert m.fs.unit.costing.purchase_cost.value == \
        pytest.approx(529738.6793, 1e-5)

    assert_units_consistent(m.fs.unit.costing)

    results = solver.solve(m)

    # Check for optimal solution
    assert results.solver.termination_condition == TerminationCondition.optimal
    assert results.solver.status == SolverStatus.ok

    assert m.fs.unit.costing.purchase_cost.value == \
        pytest.approx(529738.6793, 1e-5)

eslickj (approved):

andrewlee94 (approved):
Changes look good.

PR #137, opened 2020-12-21
Speed up Helmholtz EoS with units of measure.

### LINKED ISSUE #132: Slow units of measure
Units of measure seem to be causing a lot of slowness.  This is going to be a critical issue to things like the dynamic power plant model where the model is large and there are a lot of complex expressions.  An easy place to see this is the controller test here https://github.com/IDAES/idaes-pse/tree/main/idaes/generic_models/control.

The old version without units ran in 25 seconds and about 15 seconds was spending solving so it took roughly 10 seconds to set up the model.  The test in the new framework took 100 seconds solve time was the same. So construing the model takes 7 or 8 times as long with units.

If were don't use ```convert``` in our model expressions and we are careful, we may be able to allow units to be turned off.  Just as a test, I swapped out the pyomo units object class for this dummy and the test again runs in 25 seconds.  I think the answer could be to write some good guidelines for how to write models such that units are optional.  We can't afford much slow down on the large dynamic models.

```python
class PyomoUnitsContainerDummy(object):

    def __init__(self):
        pass

    def __getattr__(self, item):
        setattr(self, item, 1)
        return 1

    def get_units(self, expr):
        return None

    def convert_temp_K_to_C(self, value_in_K):
        return value_in_K - 273.15

    def convert_temp_C_to_K(self, value_in_C):
        return value_in_C + 273.15

    def convert_temp_R_to_F(self, value_in_R):
        return value_in_R - 458.67

    def convert_temp_F_to_R(self, value_in_F):
        return value_in_F + 458.67

    def convert(self, src, to_units=None):
        return src

    def convert_value(self, num_value, from_units=None, to_units=None):
        return num_value
```

eslickj:
## Fixes #132 

This is a major improvement in speed for units of measure with the Helmholz EoSes.  Should speed up dynamic power plant models by a lot.  For the controller test I was using for testing speed, with this PR, the time went from 130 seconds for about 31.  For comparison without units at all, it took ~25 seconds.

- Move unit conversions factors and generally remove convert() calls from expressions
- Minor improvements to htpx function allowing P,x specification somebody requested this at some point  maybe @jghouse88?

### Legal Acknowledgement


eslickj:
@dangunter, @makaylas, or @perrenyang, this somehow breaks the UI flowsheet serializer test.  Not really sure what's going on there.  I think the rest of the tests pass, so I assume this is something fairly minor.

idaes/generic_models/properties/swco2.py
```diff
@@ -166,21 +164,17 @@ def build(self):
...
         self.tc_d_1 = Param(
-            initialize=2.447164e-2,
-            doc="Residual thermal conductivity parameter",
-            units=pyunits.mW/pyunits.K*pyunits.m**2/pyunits.kg)
+            initialize=2.447164e-5,
+            doc="Residual thermal conductivity parameter")
         self.tc_d_2 = Param(
-            initialize=8.705605e-5,
-            doc="Residual thermal conductivity parameter",
-            units=pyunits.mW/pyunits.K*pyunits.m**5/pyunits.kg**2)
```
andrewlee94, line 175: Should we really be removing units from these? Unless you actually do a convert later, having units here shouldn't slow anything down, and having them allows us to do unit consistency checks which are good for giving people confidence in our code.

idaes/generic_models/properties/swco2.py
```diff
@@ -258,14 +247,14 @@ def rule_tc(b, p):
...
-                (0.475598*pyunits.mW/pyunits.K**(3/2)/pyunits.m *
-                 sqrt(T)*(1 + 2.0/5.0*cint_over_k)/G) +
-                params.tc_d_1*rho[p] +
-                params.tc_d_2*rho[p]**2 +
-                params.tc_d_3*rho[p]**3 +
-                params.tc_d_4*rho[p]**4,
-                to_units=pyunits.W/pyunits.K/pyunits.m)
+            rho_unit = pyunits.kg/pyunits.m**3
+            return (
+                (0.475598/1e3 *
```
andrewlee94, line 252: Looking at this, the expression itself still has a mess of units spread throughout it (notably the cancelation of the density units). Based on that observation, I think there would be no performance loss by letting the parameters have the correct units (W/K.m) in the end.

The important change here is removing the `convert` statement and thus assuming that units are correct at this point (which they should be). This is probably where most of the performance hit was occurring, so making this assumption should solve the observed issue. However, because we made this assumption we would want to do a unit consistency check as part of our tests to confirm (but not as part of building the model in use). Due to that, I think that having units on the parameters would be a good idea, just to add a little more confirmation.

As an additional bonus, this would then allow users to use the `convert` methods to get the correct units if they ever chose to try to change the parameter values.

idaes/generic_models/properties/helmholtz/helmholtz.py
```diff
@@ -58,100 +58,177 @@
...
-    "func_phi0_delta2": {"fname": "phi0_delta2", "units": None, "arg_units": [None, None]},
-    "func_phi0_tau": {"fname": "phi0_tau", "units": None, "arg_units": [None, None]},
-    "func_phi0_tau2": {"fname": "phi0_tau2", "units": None, "arg_units": [None, None]},
-    "func_phir": {"fname": "phir", "units": None, "arg_units": [None, None]},
-    "func_phir_delta": {"fname": "phir_delta", "units": None, "arg_units": [None, None]},
-    "func_phir_delta2": {"fname": "phir_delta2", "units": None, "arg_units": [None, None]},
-    "func_phir_tau": {"fname": "phir_tau", "units": None, "arg_units": [None, None]},
-    "func_phir_tau2": {"fname": "phir_tau2", "units": None, "arg_units": [None, None]},
-    "func_phir_delta_tau": {"fname": "phir_delta_tau", "units": None, "arg_units": [None, None]},
+    "func_p": { # pressure as a function of delta and tau
```
andrewlee94, line 70: I am going to assume that the only changes here were adding the comments explaining what each method does. GitHub marked this entire section as changes with no further detail of what the differences were.

andrewlee94 (commented):
I think this looks good overall - the main aim here appears to be assuming that units are consistent and using predetermined conversions based on this assumption, rathe than using the full `convert` method which does expensive consistency checks.

My only major comment was that based on this approach, my initial feeling was that parameters (e.g. thermal conductivity) should be defined with units where appropriate (not having the source, they might actually be dimensionless empirical constants, in which case I would suggest we mark them explicitly as `units.dimensionless`). Given we are assuming consistency throughout I don't think we would see performance hits, and we would still be able to do unit consistency checks and conversion on the parameters as necessary.

idaes/generic_models/properties/helmholtz/helmholtz.py
eslickj: Some of the external functions were missing units, so I filled in a few things, but nothing much changed.  I think the slight reformatting marked the whole thing changed.

idaes/generic_models/properties/swco2.py
eslickj: Thanks @andrewlee94.  You're right about this.  I'll fix it.  I guess I just got in remove units mode and wasn't thinking clearly.

idaes/generic_models/properties/swco2.py
eslickj: Agree with you will fix.

eslickj:
On the UI failing test, it looks like there is a stored json compared to a new generated one, and it contains doc strings for some model components.  I took units off some doc strings since now they have a unit attribute.  That broke it.  I guess I'll try to undo those changes.

eslickj:
@andrewlee94 and @MAZamarripa this should be good to go now.  I hope.

idaes/generic_models/properties/swco2.py
```diff
@@ -258,14 +257,13 @@ def rule_tc(b, p):
...
-                (0.475598*pyunits.mW/pyunits.K**(3/2)/pyunits.m *
-                 sqrt(T)*(1 + 2.0/5.0*cint_over_k)/G) +
-                params.tc_d_1*rho[p] +
-                params.tc_d_2*rho[p]**2 +
-                params.tc_d_3*rho[p]**3 +
-                params.tc_d_4*rho[p]**4,
-                to_units=pyunits.W/pyunits.K/pyunits.m)
+            return (
+                0.475598/1e3 * pyunits.W/pyunits.K/pyunits.m *
+                   sqrt(T/pyunits.K)*(1 + 2.0/5.0*cint_over_k)/G +
```
jghouse88, line 262: Why is divided by `pyunits.K` ? Should it be multiplied?

jghouse88 (approved):
I had a few comments/questions. Otherwise, LGTM.

andrewlee94:
> On the UI failing test, it looks like there is a stored json compared to a new generated one, and it contains doc strings for some model components. I took units off some doc strings since now they have a unit attribute. That broke it. I guess I'll try to undo those changes.

@eslickj By "undo those changes", did you mean you fixed the json that is used in the tests, or that you added units back to some doc strings? From the changes I can see, I think it is the first which would have been my suggestion.

eslickj:
@andrewlee94 I thought more about how confusing it will be when we display pressure is some other unit like kPa and the doc string says it's in Pa, and just updated the json instead.  Also, I know that I missed a lot doc strings, so a lot still have units.  I will eventually remove them all.

idaes/generic_models/properties/swco2.py
eslickj: I assume you mean the T in the square root.  The /K is right.  T has units, I'm taking them off.  It that or divide by another sqrt(K) where the other units are.

idaes/generic_models/properties/iapws95.py
```diff
@@ -283,13 +283,11 @@ def build(self, *args):
...
             L0 = self.config.parameters.tc_L0
             L1 = self.config.parameters.tc_L1
-            return pyunits.convert(
-                (sqrt(1.0 / tau) / sum(L0[i] * tau ** i for i in L0) *
-                 exp(delta[p] * sum(
-                     (tau - 1)**i * sum(L1[i, j] * (delta[p] - 1)**j
-                                        for j in range(0, 6))
-                     for i in range(0, 5)))),
-                to_units=pyunits.W/pyunits.K/pyunits.m)
+            return (
```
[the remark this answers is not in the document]
eslickj: It's already in the right units.  Originally it was in mW/K/m and converted to W/K/m, which I guess was to match the units of the parameters.  I just changed the parameters, so we don't need convert.

idaes/generic_models/properties/helmholtz/helmholtz.py
```diff
@@ -575,6 +640,20 @@ def build(self):
...
+        # to moles is a simple conversions, since this is only one component
+        self.uc_J_per_mol_to_kJ_per_kg = pyunits.kJ/self.mw/1000/pyunits.J
+        self.uc_kJ_per_kg_to_J_per_mol = self.mw*pyunits.J*1000/pyunits.kJ
+        # Entropy type units are same as energy type units since K is left as is
+        self.uc_J_per_molK_to_kJ_per_kgK = self.uc_J_per_mol_to_kJ_per_kg
+        self.uc_kJ_per_kgK_to_J_per_molK = self.uc_kJ_per_kg_to_J_per_mol
+        # Pressure conversions
+        self.uc_kPa_to_Pa = 1000*pyunits.Pa/pyunits.kPa
+        self.uc_Pa_to_kPa = pyunits.kPa/pyunits.Pa/1000
+
```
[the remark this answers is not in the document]
eslickj: Yeah, that's right.  This is nice because it's fast and you can still check for consistency in the units.  The drawback is that you can't mix units like you could if we use convert.  Everything needs to be in the expected unts.

idaes/generic_models/properties/helmholtz/helmholtz.py
```diff
@@ -195,66 +272,65 @@ class PhaseType(enum.Enum):
...
+    Pmin=1e-3*pyunits.kPa, Pmax=1e6*pyunits.kPa):
     """
     Convenience function to calculate enthalpy from temperature and either
     pressure or vapor fraction. This function can be used for inlet streams and
     initialization where temperature is known instead of enthalpy.
-
-    User must provide values for one (and only one) of arguments P and x.
-
+    User must provide values for one of these sets of values: {T, P}, {T, x},
+    or {P, x}.
```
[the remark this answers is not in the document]
eslickj: Yes.  I remembered you asked for it, but couldn't remember where.

andrewlee94 (approved):
@eslickj Those were pretty much my thoughts as well. I just wanted to make sure I understood which path you took.

PR #142, opened 2020-12-29
Adding support for true and apparent species lists

andrewlee94:
## Summary/Motivation:
Part of building up the infrastructure required for modeling electrolyte systems, this PR adds the code necessary to switch the basic of the component list between true and apparent species. Also included is updating some other modules to make use of the new phase and component list properties to facilitate this switching, plus some basic tests.

-
-

### Legal Acknowledgement


idaes/generic_models/properties/core/generic/tests/test_electrolyte_components.py
```diff
@@ -0,0 +1,420 @@
...
+            "HCO3-", "K+", "H2O", "CO2", "KHCO3", "N2"]
+
+        assert m.fs.props.true_phase_component_set == [
+            ("Liq", "HCO3-"), ("Liq", "K+"), ("Liq", "H2O"), ("Liq", "CO2"),
+            ("Vap", "H2O"), ("Vap", "CO2"), ("Vap", "N2")]
+        assert m.fs.props.apparent_phase_component_set == [
+            ("Liq", "H2O"), ("Liq", "CO2"), ("Liq", "KHCO3"),
+            ("Vap", "H2O"), ("Vap", "CO2"), ("Vap", "KHCO3"), ("Vap", "N2")]
+
+        assert m.fs.state[1].component_list is m.fs.props.true_species_set
```
TimBartholomew, line 380: This assertion confuses me, why is the component list and true species set the same for the state block? The state block will still include the apparent species, right?

idaes/generic_models/properties/core/generic/tests/test_electrolyte_components.py
```diff
@@ -0,0 +1,420 @@
...
+            assert p in ["Liq", "Vap"]
+        assert isinstance(m.fs.state[1].phase_frac, Var)
+        assert len(m.fs.state[1].phase_frac) == 2
+        for p in m.fs.state[1].phase_frac:
+            assert p in ["Liq", "Vap"]
+
+        assert isinstance(m.fs.state[1].mole_frac_comp, Var)
+        assert len(m.fs.state[1].mole_frac_comp) == 5
+        for j in m.fs.state[1].mole_frac_comp:
+            assert j in ["HCO3-", "K+", "H2O", "CO2", "N2"]
```
TimBartholomew, line 404: Ok, so the state block does not explicitly include apparent species, right? How will users access the apparent species?

TimBartholomew (commented):
Looks good. I have an implementation question for apparent species on the state block as follows.

idaes/generic_models/properties/core/generic/tests/test_electrolyte_components.py
andrewlee94: The state block will contain both true and apparent species. However, the main framework expects there to be a `component_list` which it uses to index the component balances, etc.

If you look at the different tests, we have the ability to set the state block to use either the true or apparent species as the master component list (see line 221 for an example); this case happens to be the true component example. Thus, we want `component_list` to be the same as `true_species_list` for this case. There is another test for the reverse case where `component_list` is the same as `apparent_species_list`.

TimBartholomew: Ok makes sense

TimBartholomew (approved):

agarciadiego (approved):
LGTM

PR #150, opened 2021-01-12
Fixed unit label rotating when the unit model rotates

makaylas:
## Summary/Motivation:
In the current version, when a unit model is rotated on right click the label rotates with it. This change makes the label stay in place during rotation.

- Unit label stays in place when unit model rotates

### Legal Acknowledgement


idaes/ui/fsvis/static/js/paper.js
```diff
@@ -63,6 +63,28 @@ export class Paper {
         // context menu
         self._paper.on("element:contextmenu", function(cellView, evt) {
             cellView.model.rotate(90)
+            // This is needed to keep the text labels for the unit models in the correct orientation
+            // x and y were specifically picked to keep the label in the same place 
+            // in relation to the unit model (bottom middle)
+            let x = 0
+            let y = 0
+            if (cellView.model.angle() == 0) {
```
lbianchi-lbl, line 71: Since it is used a few times, would it make sense to call `cellView.model.angle()` once and save the result to a `const`?

idaes/ui/fsvis/static/js/paper.js
```diff
@@ -63,6 +63,28 @@ export class Paper {
...
+                x = 0
+                y = 5
+            }
+            else if (cellView.model.angle() == 90) {
```
lbianchi-lbl, line 75: Is the angle always guaranteed to be positive and "normalized", i.e. in the interval `[0, 360)`? Otherwise, it would require some pre-processing to make sure that e.g. 270 is not missed because it is returned as -90.

idaes/ui/fsvis/static/js/paper.js
lbianchi-lbl, line 71: You could consider a `switch` statements with a `default` case instead of the `if...else if` chain with `let` variables. The two would be equivalent so that'd mostly a matter of style/clarity.

idaes/ui/fsvis/static/js/paper.js
```diff
@@ -63,6 +63,28 @@ export class Paper {
...
+            }
+            else if (cellView.model.angle() == 180) {
+                x = 0
+                y = -72
+            }
+            else if (cellView.model.angle() == 270) {
+                x = -38
+                y = -34
+            }
+            cellView.model.attr("label/transform", "translate(" + x + ", " + y + ") rotate(-" + cellView.model.angle() + ")")
```
lbianchi-lbl, line 87: Using template strings could be a way of making this a bit more readable, i.e. 

```js
 cellView.model.attr("label/transform", `translate(${x}, ${y}) rotate(-${cellView.model.angle()})`)
```

lbianchi-lbl (commented):
I've left a few comments mostly for style/readability, but I don't have strong feelings on this, so if you were not planning to iterate on this feel free to dismiss them and I'll go ahead and approve.

idaes/ui/fsvis/static/js/paper.js
makaylas: Yes the angle is always positive and normalized.

idaes/ui/fsvis/static/js/paper.js
makaylas: Good point

makaylas: Done

idaes/ui/fsvis/static/js/paper.js
makaylas: I didn't know about template strings. Thanks!

lbianchi-lbl (dismissed):
:+1:

idaes/ui/fsvis/static/js/paper.js
```diff
@@ -63,6 +63,34 @@ export class Paper {
...
         self._paper.on("element:contextmenu", function(cellView, evt) {
             cellView.model.rotate(90)
+            // This is needed to keep the text labels for the unit models in the correct orientation
+            // x and y were specifically picked to keep the label in the same place 
+            // in relation to the unit model (bottom middle)
+            let x = 0
+            let y = 0
+            const angle = cellView.model.angle()
+            switch (angle) {
+                case 0:
```
dangunter, line 73: where do these "magic" numbers come from? Are all icons the same width and height? (If so, shouldn't they at least have different widths and heights..?)

dangunter (commented):
curious about the constants

idaes/ui/fsvis/static/js/paper.js
makaylas: The numbers come from me manually changing the numbers bit by bit until the labels stay in the same place.

Yes, all of the icons are the same width and height. That's something that we knew needed to happen but it was put off as low priority.

dangunter (changes requested):
Once the numbers have moved into constants, preferably with a "todo" about doing this more generally once time allows, it looks good.

makaylas:
@dangunter I made the numbers into constants

idaes/ui/fsvis/static/js/paper.js
```diff
@@ -66,6 +66,42 @@ export class Paper {
         // context menu
         self._paper.on("element:contextmenu", function(cellView, evt) {
             cellView.model.rotate(90)
+            // This is needed to keep the text labels for the unit models in the correct orientation
+            // x and y were specifically picked to keep the label in the same place 
+            // in relation to the unit model (bottom middle)
+            // TODO Make this figuring out the x and y positions a function so that we can compute it
+            const angle = cellView.model.angle()
+            switch (angle) {
```
dangunter, line 74: So let me suggest a more data-driven way to approach this.
```
function translate_for_angle(angle, width, height) {
   // TODO: replace with geometry that considers width and height
   const angle_translation = {0: [0, 5], 90: [38, -35], 180: [0, -72], 270: [-38, -34]};
   return angle_translation[angle];
}
// then in this function:
[x,y] = translate_for_angle(angle);
```
This hides your heuristic somewhere that you can change later.
Is the case of a number not in the list normal behavior? That seems like it should be a runtime error (here it's  a stack trace, but you could explicitly catch it).

dangunter (changes requested):
I think this can be condensed. My point wasn't really to make x and y constants but rather to set yourself up for more flexible logic on this in the future.

idaes/ui/fsvis/static/js/paper.js
makaylas: Your implementation makes much more sense. Thank you

No it should always be one of those numbers because that's the only rotation we allow. But I did add a catch and error message for that.

dangunter (approved):
Looks good

lbianchi-lbl (approved):
Looks good!

PR #161, opened 2021-01-14
Add config options for default state var bounds

eslickj:
@MAZamarripa, needed a way to change the default bounds of Helmholtz EOS block, so here it is. 

### Legal Acknowledgement


andrewlee94 (changes requested):
Only one minor request - could you add this to the documentation somewhere. You should also note that the expected inputs are a Pyomo expression with units, which will be important when/if someone tries to use this.

eslickj:
@andrewlee94 these docs should get generated, but you are right I need to add the units info.  Do you happen to know what happens with units when you set the bounds on a variable.  I guess as long as it doesn't break anything they should be fine.  I figured they are either ignored of converted to the variable units.   Either way I should be fine here.  I guess if they were auto converted I could cut out a few lines of code.

andrewlee94:
@eslickj Units and bounds are completely separate as far as the Pyomo components are concerned. I.e. bounds (like values) are just a float, and `Vars` and `Params` have a separate `units` attribute. I.e. the bounds are set directly with just a float with no unit checking.

eslickj:
@andrewlee94 thanks.  I updated the config option docs strings, which should finish this, but I discovered that the Helmholtz EOS documentation doesn't really exist and only the IAPWS-95 class is documented.  I guess I'll be a while finishing up the docs here.

MAZamarripa (dismissed):
lgtm

eslickj:
@andrewlee94 and @MAZamarripa, the documentation should be there now.  Unfortunately it required a major update, since the old docs where from before I generalized the Helmholtz EoS.  Should be ready for review.

docs/technical_specs/model_libraries/power_generation/unit_models/feedwater_heater_0D.rst
```diff
@@ -7,12 +7,12 @@ Feedwater Heater (0D)
...
   :noindex:
 
-The FWH0D model is a 0D feedwater heater model suitable for steady state modeling.  
-It is intended to be used primarily used with the 
-:ref:`IAWPS95 <technical_specs/model_libraries/generic/property_models/water:Water/Steam - IAPWS95>` property package. 
-The feedwater heater is split into three sections the condensing section is required while 
-the desuperheating and drain cooling sections are optional. There is also an optional mixer 
-for adding a drain stream from another feedwater heater to the condensing section.  The figure 
+The FWH0D model is a 0D feedwater heater model suitable for steady state modeling.
+It is intended to be used primarily used with the
```
andrewlee94, line 11: Typo: duplicate "used"

idaes/generic_models/properties/helmholtz/helmholtz.py
```diff
@@ -563,7 +563,9 @@ class HelmholtzParameterBlockData(PhysicalParameterBlock):
             default=None,
             domain=tuple,
             description="Tuple providing default temperature bounds",
-            doc="Tuple providing default temperature bounds",
+            doc="""This is an optional tuple providing default temperature bounds.
+The elements of the tuple should include units of temperature.
```
andrewlee94, line 567: It might just be the speed at which I skimmed the docs, but I didn't see an explicit example of what this tuple should look like i.e. is it `(lower, upper, units)` or something else?

andrewlee94 (changes requested):
Two minor comments.

idaes/generic_models/properties/helmholtz/helmholtz.py
eslickj: Okay, should be all fixed up.  You were right about it being unclear.  I make the documentation of the tuple format explicit and included an example.

andrewlee94 (dismissed):
Looks good.

MAZamarripa (dismissed):
lgtm

eslickj:
@MAZamarripa and @andrewlee94 could you guys take a look again?  I think you both approved, but I had to bump up the test coverage to get the tests to pass the coverage requirements.

andrewlee94 (approved):
Looks Ok.

MAZamarripa (approved):
lgtm

PR #162, opened 2021-01-15
Allow ideas get-extensions to run with macOS

adowling2:
- Replaced exception with a printed warning

### Legal Acknowledgement

### DIFF
--- a/idaes/solvers.py
+++ b/idaes/solvers.py
@@ -100,7 +100,8 @@ def download_binaries(
         _log.debug("URLs \n  {}\n  {}\n  {}".format(url, solvers_from, libs_from))
         _log.debug("Destinations \n  {}\n  {}".format(solvers_tar, libs_tar))
         if platform == 'darwin':
-            raise Exception('Mac OSX currently unsupported')
+            # raise Exception('Mac OSX currently unsupported')
+            print("WARNING: macOS support is experimental.")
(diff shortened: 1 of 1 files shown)


adowling2:
@ksbeattie @eslickj This one line change allows me to test installing the macOS binaries. Please provide feedback if you prefer something instead of a print statement for the warning message. Let's wait to merge until we have the binaries tested for multiple people.

eslickj:
I made PR #163, which is a more permanent update for this.  It also let's you skip the checksum check, which I think will be good for us to test the Mac binaries, since it's a little like a "I know it's not quite right but I want it anyway" option.

idaes/solvers.py
```diff
@@ -100,7 +100,8 @@ def download_binaries(
         _log.debug("URLs \n  {}\n  {}\n  {}".format(url, solvers_from, libs_from))
         _log.debug("Destinations \n  {}\n  {}".format(solvers_tar, libs_tar))
         if platform == 'darwin':
-            raise Exception('Mac OSX currently unsupported')
+            # raise Exception('Mac OSX currently unsupported')
+            print("WARNING: macOS support is experimental.")
```
ksbeattie, line 104: I think a `_log.warn()` might be more appropriate here...

eslickj:
I'll go ahead an close this.  Feature is in #163 now.

PR #163, opened 2021-01-19
Updatedate get-extensions command, prep for OSX

eslickj:
- Does not raise a specific objection to darwin platform
- New exception when the file is not available this works but still a bit ugly as the 404 error produce a bit of extra output
- Option to skip the checksum check
- Option to install the libraries only and skip the solvers.  Currently solvers is: Ipopt, Pynumero, and k_aug.  Libraries contains mostly Helmholtz and Cubic EoS external functions.

### Legal Acknowledgement


eslickj:
Need to add a libraries only option to allow property libraries to be installed without the solver package.

eslickj:
I added a test that basically just downloads and extracts the release to an alternate location.  Not sure it's the best idea, but hopefully it's better than nothing.  It would be nice if we could flag the test CI only.  Do we have a marker like that?

jsiirola (approved):

eslickj:
I added a --no-download option, so I could test the commands without downloading or extracting anything.  The download and extract function is tested separate from for commands.  I'm trying to get my test coverage up to passing level.

lbianchi-lbl:
@eslickj the integration tests are triggered as expected, but they end up failing almost immediately for some reason. As the error happens before the code is even checked out, I believe this is due to a bug or corner case in the CI workflow rather than something having to do with the code. I'm looking into it.

eslickj:
@jsiirola, could you take another look at this?  Since you last approved I had to make some changes to get better test coverage.  I think the changes were to improve testing, and make this more amenable to future simple unit tests.

jsiirola (approved):

adowling2 (approved):
LGTM

adowling2:
@lbianchi-lbl Is it possible to merge this tomorrow AM into master? We are going to test out the macOS compilation instructions. It will simplify the process if this is in master (don't need to add John's fork as a remote).

PR #164, opened 2021-01-19
Supercritical CO2 compressor

minhwvu:
## Summary/Motivation:
Developing a supercritical CO2 compression model 

- Adding a model and test
-

### Legal Acknowledgement


idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,411 @@
...
+    CONFIG.thermodynamic_assumption = ThermodynamicAssumption.isentropic
+    CONFIG.get("thermodynamic_assumption")._default = \
+        ThermodynamicAssumption.isentropic
+    CONFIG.get('thermodynamic_assumption')._domain = \
+        In([ThermodynamicAssumption.isentropic])
+
+    ################################################
+
+    def build(self):
+        super(InletStageData, self).build()
```
andrewlee94, line 65: In Python 3+, you only need to do `super().build()`

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,411 @@
...
+        config = self.config # shorter config pointer
+
+        # The thermodynamic expression writer object, te, writes expressions
+        # including external function calls to calculate thermodynamic quantities
+        # from a set of state variables.
+        te = ThermoExpr(blk=self, parameters=config.property_package)
+
+        if SolverFactory('ipopt').available():
```
andrewlee94, line 73: There should be no reason to have a solver in the `build` method, especially one which i dependent on a specific solver.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,411 @@
...
+        #####################################################
+        # Declare variables for the model
+
+        # Some shorter refernces to property blocks
+        properties_in = self.control_volume.properties_in
+        properties_out = self.control_volume.properties_out
+
+        
+        self.kappaT = Var(self.flowsheet().config.time,
+                            initialize=5.0e-7)
```
andrewlee94, line 90: 1. Check indentation.
2. You should make use of Pyomo `units` here to define the units of measurement on all `Vars` and `Params`.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,411 @@
...
+                            doc="Polytropic head coefficient",
+                            bounds=(0.001,1))
+        self.eff_p_v = Var(self.flowsheet().config.time,
+                            initialize=0.85,
+                            doc="Polytropic stage efficiency vaned diffuser",
+                            bounds=(0.001,1))
+        self.eff_p = Var(self.flowsheet().config.time,
+                            initialize=0.85,
+                            doc="Polytropic stage efficiency",
+                            bounds = (0.001, 1.0))
```
andrewlee94, line 118: Check PEP8 formatting. Indentation is inconsistent, and there should not be spaces around the `=` here.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,411 @@
...
+        self.rspeed = Var(self.flowsheet().config.time,
+                            initialize = 1500,
+                            doc = 'Rotation speed of the impeller',
+                            bounds = (0.5, 8000))      
+        self.r2 = Var(initialize= 0.075, 
+                            doc="Impeller tip radius")  
+        
+        ########################################################     
+        
+        self.dp = Param(default= 1, within=Reals, doc='a pressure difference for finite difference derivative calculation')
```
andrewlee94, line 142: Line too long.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,411 @@
...
+        self.diffuser_type = Param(default=1.0,
```
andrewlee94, line 143: These look like things that would be better defined as Config arguments, where users could use a named Enum rather than an integer to set these.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,411 @@
...
+        self.impeller_type = Param(default=2.0,
+                                    mutable=True,
+                                    doc="1 for covered impeller, 2 for open impeller")
+        self.efficiency_mech = Param(default= 0.97,
+                                    doc="Compressor mechanical efficiency")
+        self.eff_drive = Param(default= 1.0,
+                                doc="Compressor driver efficiency")        
+
+        ############################################################ 
+        # Fix variables
```
andrewlee94, line 155: This is changing the initial value, not fixing the value.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,411 @@
...
+        self.eff_drive = Param(default= 1.0,
+                                doc="Compressor driver efficiency")        
+
+        ############################################################ 
+        # Fix variables 
+        self.ratioP[:] = 2.0   # make sure these have a number value  
+        self.deltaP[:] = 0     #   to avoid an error later in initialize
+    
+        ###########################################################
+        # Speed of sound stuff
```
andrewlee94, line 160: Speed of sound calculations

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,411 @@
...
+        # finite difference derivative for kappa_T calculation
```
andrewlee94, line 161: Can we find a way to avoid using finite differences here? We supposedly have full access to the equations, so we should be able to determine this analytically/automatically.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,411 @@
...
+        @self.Expression(self.flowsheet().config.time, doc="Thermodynamic power")
+        def fluid_pow(b,t):
+            flow = properties_in[t].flow_mol 
+            return b.delta_enth_actual[t]*flow
+
+        @self.Expression(self.flowsheet().config.time, doc="Shaft power")
+        def elec_pow(b,t):
+            return b.fluid_pow[t]/(b.efficiency_mech * b.eff_drive)
+
+    def initialize(self, state_args={}, outlvl=6, solver='ipopt',
```
andrewlee94, line 301: Outlvl and reporting should use the idaes logger and the predefined levels.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,411 @@
...
+            solver (str): Solver to use for initialization
+            optarg (dict): Solver arguments dictionary
+        """
+        init_log = idaeslog.getInitLogger(self.name, outlvl, tag="unit")
+        solve_log = idaeslog.getSolveLogger(self.name, outlvl, tag="unit")
+
+        opt = SolverFactory(solver)
+        opt.options = optarg
+
+        flags = self.control_volume.initialize(outlvl=outlvl-1,
```
andrewlee94, line 320: outlvl=outlvl

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,411 @@
...
+                                       optarg=optarg,
+                                       solver=solver,
+                                       state_args=state_args)
+
+        print('dof for step 1 =', degrees_of_freedom(self))
```
andrewlee94, line 325: Use the logger instead of print method.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,411 @@
...
+        init_log.info_high("Initialization Step 1 Complete.")
+        
+        # sp is what to save to make sure state after init is same as the start
+        #   saves value, fixed, and active state, doesn't load originally free
+        #   values, this makes sure original problem spec is same but initializes
+        #   the values of free vars
+        sp = StoreSpec.value_isfixed_isactive(only_fixed=True)
```
andrewlee94, line 332: Whilst this will guarantee there are no issues, it strikes me that this could just mask underlying issues in the initializations. I would personally prefer to explicitly activate and deactivate constraints and fix and unfix variables, and then check that the state is the same.

idaes/power_generation/carbon_capture/compression_system/tests/test_compressor.py
```diff
@@ -0,0 +1,129 @@
...
+    m = build_unit
+    assert degrees_of_freedom(m) == 0
+    # Check unit config arguments
+    assert len(m.fs.unit.config) == 10
+    # assert m.fs.unit.config.has_heat_transfer
+    assert m.fs.unit.config.thermodynamic_assumption
+    assert m.fs.unit.config.property_package is m.fs.properties_co2
+
+
+# @pytest.mark.integration
```
andrewlee94, line 83: This needs to be uncommented - this will probably require the addition of units in the main model.

idaes/power_generation/carbon_capture/compression_system/tests/test_compressor.py
```diff
@@ -0,0 +1,129 @@
...
+# def test_units(build_unit):
+#     assert_units_consistent(build_unit)
+
+
+# @pytest.mark.skipif(not iapws95.iapws95_available(),
```
andrewlee94, line 88: Remove the commented lines.
You can also remove the `skipif` line below.

idaes/power_generation/carbon_capture/compression_system/tests/test_compressor.py
```diff
@@ -0,0 +1,129 @@
...
+#                     reason="IAPWS not available")
+@pytest.mark.skipif(solver is None, reason="Solver not available")
+@pytest.mark.component
+def test_initialize(build_unit):    
+    initialization_tester(build_unit, dof=0)
+
+
+# @pytest.mark.skipif(not iapws95.iapws95_available(),
```
andrewlee94, line 96: Again, remove commented lines.

andrewlee94 (changes requested):
Mostly just formatting issues, but you should add units to all the `Vars` and there are a few things that I think could be done in better ways.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,411 @@
...
+import math
+
+
+_log = idaeslog.getLogger(__name__)
+
+@declare_process_block_class("InletStage", doc="Inlet stage of compressor model")
+class InletStageData(PressureChangerData):
+    # Pressure changer with isentropic compressor options
+    CONFIG = PressureChangerData.CONFIG()
+    CONFIG.compressor = True
```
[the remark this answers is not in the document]
eslickj: I know it's a bit on the weird side, but I think it may be a good idea to keep these unneeded options.  The model is still a pressure-changer and it's nice that it has the expected pressure changer options for people who may want to inspect pressure changer models for some reason.  Besides, the base class expects them in places.

idaes/power_generation/carbon_capture/compression_system/compressor.py
eslickj: Further, these will depend on the property package.  I'm being a bit lazy, but you should be able to see something like the flow coefficient in the steam turbine outlet stage for an example of how to set the units in a generic way.

idaes/power_generation/carbon_capture/compression_system/compressor.py
eslickj: True, and if that's what you want to do, better off using the initialize argument when you define the Var.

idaes/power_generation/carbon_capture/compression_system/compressor.py
eslickj: should be ```outlvl=idaeslog.NOTSET```.  That will use the default logger settings.  The levels are an enumerated type.  6 won't cause an error, but it will be roughly the same a ```outlvl=idaeslog.DEBUG```

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,411 @@
...
+from pyomo.environ import (ConcreteModel, SolverFactory, Reals, NonNegativeReals,
+    Var, Param, Expression, Constraint, SolverFactory, value, exp, log, sqrt)
+from pyomo.common.config import ConfigBlock, ConfigValue, In
+
+
+# Import IDAES cores
+from idaes.core import (declare_process_block_class,
+                        FlowsheetBlock)
+from idaes.generic_models.unit_models.pressure_changer import (PressureChangerData,
+																ThermodynamicAssumption)
```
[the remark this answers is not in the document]
minhwvu: Thanks Andrew. I have uploaded a revised version. Please help me check it.

idaes/power_generation/carbon_capture/compression_system/compressor.py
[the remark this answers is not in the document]
minhwvu: As John said, I prefer keeping those. Any suggestion?

idaes/power_generation/carbon_capture/compression_system/compressor.py
minhwvu: It is revised.

idaes/power_generation/carbon_capture/compression_system/compressor.py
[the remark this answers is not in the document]
minhwvu: I prefer keeping those codes. Do you have any suggestion for us if we want to remove them?

idaes/power_generation/carbon_capture/compression_system/compressor.py
minhwvu: It has been revised.

idaes/power_generation/carbon_capture/compression_system/compressor.py
minhwvu: Yes, I removed those lines in the revised version.

idaes/power_generation/carbon_capture/compression_system/compressor.py
minhwvu: I added to the Vars and Params in the revised version. Please check them.

idaes/power_generation/carbon_capture/compression_system/compressor.py
minhwvu: It is revised.

idaes/power_generation/carbon_capture/compression_system/compressor.py
minhwvu: It is revised.

idaes/power_generation/carbon_capture/compression_system/compressor.py
minhwvu: Those have been modified in the revised version.

idaes/power_generation/carbon_capture/compression_system/compressor.py
minhwvu: These variables should be changed. I have deleted this line.

idaes/power_generation/carbon_capture/compression_system/compressor.py
minhwvu: It is revised.

andrewlee94:
@minhwvu You do not need to respond to each comment individually - for simple things just hit the "resolve issue" button. Otherwise, each response you make generates another email from GitHub which quickly fills up peoples inboxes (meaning they are likely to be ignored).

idaes/power_generation/carbon_capture/compression_system/compressor.py
minhwvu: Until now, I cannot find any solution. John gives me some comments "I think on the first one that would depend on if the property package provides it. I don’t see what exactly Andrew is referring to there, but I assume it’s a partial derivative of some property.  I’d guess that the answer is that you need to stick with the finite difference until you get this derivative into the property packages that you are using.  It is possible for the property package to provided derivative of properties, but in general they don’t. ". If you have any idea, please let me know.

idaes/power_generation/carbon_capture/compression_system/compressor.py
minhwvu: Hi @andrewlee94 , I am not sure what you suggests here. I supposed that this method allows me to fix all variables if we do not know that they should be fixed and will be unfixed later. If you prefer doing another solution, please let me know. Thanks.

andrewlee94: The issue I see here is that there could be a mistake in the initialization routine where a constraint is accidentally left deactivated when one of the intermediate solves is called, resulting in an unexpected degree of freedom. This could lead to unexpected values in the model variables even though the solver reports an optimal solution.

My suggestion is that rather than restoring the model state as you are doing here, we rely on the initialization routine itself to restore the state as it goes, and then do a confirmation check as a test to ensure this is the case. That way, if we do accidentally introduce a degree of freedom in the initialization, it will hopefully show up in the test that the initial state is not the same as the final state and we can identify that there is an error in the initialization routine. By forcibly restoring the state however, we hide any of these potential issues.

minhwvu: Hi @andrewlee94 : I have changed the initialization approach, deactivate and activate constraints  with fixed/unfixed variables. Can you check the revised version? 
Also, I am not sure how to fix the inlets with this model when it can be connected to an outlet or the another unit. Is there any standard approach for doing this? Thanks.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,662 @@
...
+            optarg (dict): Solver arguments dictionary
+        """
+        init_log = idaeslog.getInitLogger(self.name, outlvl, tag="unit")
+        solve_log = idaeslog.getSolveLogger(self.name, outlvl, tag="unit")
+
+        opt = SolverFactory(solver)
+        opt.options = optarg
+
+        with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
+            res = opt.solve(self, tee=slc.tee)
```
andrewlee94, line 546: Why is the very first thing you are doing a solver call? If this actually works, then there is no point in the rest of this routine as the model is already solved - otherwise this is just going to fail and be a waste of time (or worse, raise a terminal exception).

The first thing you should be doing is calling the initialization routines of the sub-blocks (either control volume or the actual state blocks) and passing in a set of initial guesses. This can also be used to fix the inlet states for you as well, using the `hold_state=True` argument.

andrewlee94 (changes requested):

idaes/power_generation/carbon_capture/compression_system/compressor.py
andrewlee94: @minhwvu You should call the `initialize` method of the contorl volume of state blocks first. As part of this, you can set the `hold_State=True` argument, which will leave the state variables fixed and return a list of things that needs to be unfixed later.

```
flags = blk.control_volume.initialize(state_args=state_args, hold_state=True)

# the rest of the initialization routine

blk.control_volume.release_state(flags=flags)
```

minhwvu: Hi @andrewlee94 : Added those codes. Is there any modification suggested? Thanks.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -522,8 +522,8 @@ def psi_s_stage_eqn(b, t):
                 return b.psi_s[t] == b.coeff_a * b.psi_3[t]**2 + \
                     b.coeff_b * b.psi_3[t] + b.coeff_c
 
-    def initialize(self, state_args={}, outlvl=idaeslog.NOTSET, solver='ipopt',
-                   optarg={'tol': 1e-6, 'max_iter': 500}):
+    def initialize(blk, state_args={}, outlvl=idaeslog.NOTSET, solver='ipopt',
+                   optarg={'tol': 1e-6, 'max_iter': 200}):
```
andrewlee94, line 526: A max iter of 200 is huge for an initialization routine. Normally we do not even set this, but it you do it probably should be only 50 (or less) unless there is a good reason for it to be larger.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -536,127 +536,125 @@ def initialize(self, state_args={}, outlvl=idaeslog.NOTSET, solver='ipopt',
...
 
-        opt = SolverFactory(solver)
-        opt.options = optarg
-
-        with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
-            res = opt.solve(self, tee=slc.tee)
-        init_log.info_high(
-                "Initialization Step 1 {}.".format(idaeslog.condition(res))
-            )
+        flags = blk.control_volume.initialize(state_args=state_args,
```
andrewlee94, line 542: You should also pass `solver=solver` and `optarg=optarg` here

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -536,127 +536,125 @@ def initialize(self, state_args={}, outlvl=idaeslog.NOTSET, solver='ipopt',
...
+                                              hold_state=True)
 
-        for t in self.flowsheet().config.time:
-            for k, v in self.inlet.vars.items():
+        for t in blk.flowsheet().config.time:
```
andrewlee94, line 545: You do not need to do this here - the control volume initialization should have done this for you.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -536,127 +536,125 @@ def initialize(self, state_args={}, outlvl=idaeslog.NOTSET, solver='ipopt',
...
+            for k, v in blk.inlet.vars.items():
                 v[t].fix()
-            for k, v in self.outlet.vars.items():
+            for k, v in blk.outlet.vars.items():
```
andrewlee94, line 548: As above.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -536,127 +536,125 @@ def initialize(self, state_args={}, outlvl=idaeslog.NOTSET, solver='ipopt',
...
             eff.fix(eff.value if value(eff) > 0.3
                     and value(eff) < 1.0 else 0.85)
             # for outlet pressure try outlet pressure, pressure ratio
             # then if none of those look reasonable use a pressure ratio of 2.0
             # to calculate outlet pressure
-            Pout = self.outlet.pressure[t]
-            Pin = self.inlet.pressure[t]
+            Pout = blk.outlet.pressure[t]
+            Pin = blk.inlet.pressure[t]
             prratio = value(Pout/Pin)
```
andrewlee94, line 561: As it is currently written, this will almost always evaluate to 1, as Pout will be the same as Pin unless the user fixed Pout as their design spec. It could be that you are assuming that this will be the case, but that would need to be clearly documented, and we would need to have checks to make sure the users didn't try fixing deltaP or ratioP instead.

What you really need to do is determine which out of Pout, detalP and ratioP is fixed, and then work from there.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -536,127 +536,125 @@ def initialize(self, state_args={}, outlvl=idaeslog.NOTSET, solver='ipopt',
...
                 elif prratio < 4.0 and prratio > 1.0:
                     Pout.fix(value(prratio*Pin))
                 else:
                     Pout.fix(value(Pin * 2.0))
             else:
                 Pout.fix()
 
-        self.deltaP[:] = value(Pout - Pin)
-        self.ratioP[:] = value(Pout / Pin)
+        blk.deltaP[:] = value(Pout - Pin)
```
andrewlee94, line 572: This is not good, as it will overwrite any user provided value for deltaP and ratioP - i.e. what if the user fixed one of htese values instead of Pout?

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -536,127 +536,125 @@ def initialize(self, state_args={}, outlvl=idaeslog.NOTSET, solver='ipopt',
...
+        blk.ratioP[:] = value(Pout / Pin)
 
-        for t in self.flowsheet().config.time:
-            self.properties_isentropic[t].pressure.value = value(
-                self.outlet.pressure[t]
+        for t in blk.flowsheet().config.time:
+            blk.properties_isentropic[t].pressure.value = value(
```
andrewlee94, line 576: This code presumes a specific set of state variables-  this might be OK if this model is limited to a specific property package, but that needs to be clearly documented. Otherwise, you should iterate over the `define_state_vars`.

Also, you should use this information to create a new `isen_state_args` and pass that the `blk.properties_isentropic.initilize()` in order to initialize the isentropic state block (which you are currently not doing).

andrewlee94 (changes requested):

idaes/power_generation/carbon_capture/compression_system/compressor.py
minhwvu: Hi @andrewlee94 : Can you check the model again ? Thanks!

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -542,47 +543,45 @@ def initialize(blk, state_args={}, outlvl=idaeslog.NOTSET, solver='ipopt',
...
+                blk.ratioP[t] = value(
+                    (blk.control_volume.deltaP[t] + blk.inlet.pressure[t]
+                     ) / blk.inlet.pressure[t])
+            elif blk.ratioP[t].fixed:
+                blk.outlet.pressure[t] = value(
+                    blk.ratioP[t] * blk.inlet.pressure[t])
+
+            # initialize the isentropic state block
+            isen_state_args = blk.properties_isentropic[t].define_state_vars()
+            blk.properties_isentropic.initialize(isen_state_args)
```
andrewlee94, line 569: This is not doing what you want it to do - in fact I am not sure what it is doing.

What you want to do here is something similar to:

```
isen_state = {}
t_init = blk.flowsheet().config.time.first()
for i in blk.properties_isentropic[t_init].define_state_vars().keys():
    state_var = getattr(blk.control_volume.properties_out[t_init], i)
    if state_var.is_indexed:
        isen_state[i] = {}
        for j in state_var:
            isen_state[i][j] = state_var[j].value
    else:
        isen_state[i] = state_var.value

blk.properties_isentropic.initialize(state_args=isen_state, solver=solver, optarg=optarg)
```

andrewlee94: Also note, you might want to do something similar for the outlet state. Currently it is initialized at the same conditions as the inlet, but seeing as we at least have a good idea about pressure, you should probably initialize at the expected outlet pressure instead.

To do this, rather than using the `control_volume.initialize`, you should instead call the initialize methods separately on `properties_in` and `properties_out`, providing each with good initial guesses.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -542,47 +543,45 @@ def initialize(blk, state_args={}, outlvl=idaeslog.NOTSET, solver='ipopt',
...
+
+            # initialize outlet state
+            blk.outlet.flow_mol[t] = value(blk.inlet.flow_mol[t])
+            # blk.outlet.enth_mol[t] = value(blk.inlet.enth_mol[t] * 1.2)
+
+            @blk.Expression(blk.flowsheet().config.time,
```
andrewlee94, line 575: Why are you creating an `Expression` in the `initialize` routine? You can achieve the same end by just evaluating the actual expression, without attaching it to the model.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -542,47 +543,45 @@ def initialize(blk, state_args={}, outlvl=idaeslog.NOTSET, solver='ipopt',
...
+                            doc="outlet enthalpy expression")
+            def h_o(blk, t):
+                return blk.inlet.enth_mol[t] + (
+                    blk.properties_isentropic[t].enth_mol -
+                    blk.inlet.enth_mol[t]) / blk.efficiency_isentropic[t]
+            blk.outlet.enth_mol[t] = value(blk.h_o[t])
+            # blk.control_volume.work[t] = value(
```
andrewlee94, line 582: Commented lines should be removed.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -542,47 +543,45 @@ def initialize(blk, state_args={}, outlvl=idaeslog.NOTSET, solver='ipopt',
...
+            # blk.inlet.flow_mol[t]*blk.inlet.enth_mol[t] -
+            # 	blk.outlet.flow_mol[t]*blk.outlet.enth_mol[t])
 
         # Deactivate special constraints
         blk.eff_isen_eqn.deactivate()
```
andrewlee94, line 587: You could save yourself some lines of code by making a list of constraints to deactivate and variables to fix, and then iterate over that list. This saves repeating the names again when you need to undo your changes, and makes it less likely that something will get forgotten.

idaes/power_generation/carbon_capture/compression_system/tests/test_compressor.py
```diff
@@ -52,11 +52,12 @@ def build_unit():
     # compressor first stage
     p = 1.13937 * 1e5  # Pa
     t = 40.0113 + 273.15  # K
-    fin = 1689.31  # mol/s
+    fin = 1548.52  # mol/s 1689.31
```
andrewlee94, line 55: What is the trailing 1689.31 for?

idaes/power_generation/carbon_capture/compression_system/tests/test_compressor.py
```diff
@@ -52,11 +52,12 @@ def build_unit():
...
 
     hin_co2 = swco2.htpx(T=t*pyo.units.K, P=p*pyo.units.Pa)
 
     # inlet stream
+    m.fs.unit.ratioP[:].value = 2.3351
```
andrewlee94, line 60: You don't need the `[:]` here,

idaes/power_generation/carbon_capture/compression_system/tests/test_compressor.py
```diff
@@ -114,10 +115,10 @@ def test_run(build_unit):
                       * m.fs.unit.outlet.enth_mol[0]
                       + m.fs.unit.work_mechanical[0]))
     # pressure change
-    assert (pytest.approx(177991.4247463311, abs=0.1) ==
+    assert (pytest.approx(185822.29121655718, abs=0.1) ==
```
andrewlee94, line 118: If you are setting an absolute tolerance of 0.1, you don't need all the additional significant figures in the test value.

andrewlee94 (changes requested):
I think there area few things that still need work - notably I am not sure what is happening when you initialize the isentropic block, but it is probably not what you want.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -18,7 +18,7 @@
 Control of a Multistage CO2 Compresor System." International Journal
 of Greenhouse Gas Control. v62., page 31-45
 Created: April 2020
-__Author__ = "Quang Minh Le, John Eslick"
+__Author__ = "Quang Minh Le, John Eslick, Andrew Lee"
 """
 
 from __future__ import division
```
eslickj, line 24: We don't support Python 2.x anymore, so don't need this line.

idaes/power_generation/carbon_capture/compression_system/compressor.py
eslickj: Here I think self is more the python standard, so its probably best to stick to self.  The main reason you'll see code with blk rather than self is that a long time ago we we're using a different style less object-oriented style where we had functions that just took Pyomo block as an argument and did something to it.

idaes/power_generation/carbon_capture/compression_system/compressor.py
minhwvu: Hi @andrewlee94 : I got the below error at the outlet enthalpy calculation when I use the initial code for the isentropic state. Can you let me know how to fix it? If I keep the previous syntax, this error does not show up. Thanks!

ERROR    pyomo.core:compressor_v8.py:602 evaluating object as numeric value: fs.unit.control_volume.properties_in[0.0].enth_mol + (fs.unit.properties_isentropic[0.0].enth_mol - fs.unit.control_volume.properties_in[0.0].enth_mol)/fs.unit.efficiency_isentropic[0.0]
    (object: <class 'pyomo.core.expr.numeric_expr.SumExpression'>)
unsupported operand type(s) for +: 'int' and 'dict'

andrewlee94: That would seem to indicate a deeper issue, but first try doing `value(state_var)` instead of `state_var.value` in both places (indexed and unindexed states).

However, this is seemingly saying that one of the state variables (enthalpy) is an `Expression` and not a `Var`, which would seem to indicate a problem.

eslickj: I'm not sure, but it says you are trying to add and int to a dictionary.  I can't really tell where this is coming from though.  Not sure if if has anything to do with something expected to be a variable being a dict.  You may have set the value of a variable to a dictionary.

minhwvu: Hi @andrewlee94 and @eslickj : I try those suggestions but they do not work. I think that is coming from the isentropic enthalpy calculation because it is not a variable (at least I did not see that variable in Pressure Changer model). Do you have any idea to fix this problem? Thanks.

andrewlee94: Well, the simple answer is that you cannot fix the value of an `Expression`, which is what the entropy is - you would have calculate a value for the entropy and then use that to estimate the associated state variables.

However, this is somewhat property package specific, as different property packages use different state variables. The simplest approach, and the one I think we use elsewhere, is to assume isothermal behavior for initialization (or something equivalent).

eslickj:
I'm going to grab this today, run it, and review very carfully.  I may push some changes.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,659 @@
...
+_log = idaeslog.getLogger(__name__)
+
+
+class ImpellerType(Enum):
+    cover_impeller = 1
+    open_impeller = 2
+    custom = 3
+
+
+def _cover_impeller_callback(blk):
```
eslickj, line 49: Unless the user specifies the callback function by directly specifying a function, it's not really a callback.  You could incorporate this directly into build method, or if that is too messy just rename this like _build_covered_impeller_equations(), just to avoid confusion.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,659 @@
...
+    vaneless_diffuser = 2
+    custom = 3
+
+
+def _vane_diffuser_callback(blk):
+    @blk.Constraint(blk.flowsheet().config.time)
+    def polytropic_head_coeff_eqn(b, t):
+        return b.mu_p[t] == b.mu_p_v[t]
+
+    def polytropic_efficiency_eqn(b, t):
```
eslickj, line 100: It looks like you are missing a decorator here.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,659 @@
...
+    blk.mass_flow_coeff = Var(blk.flowsheet().config.time,
+                              initialize=0.0735,
+                              doc="Compressor Flow Coefficient",
+                              bounds=(0.01, 0.15))
+
+    @blk.Constraint(blk.flowsheet().config.time)
+    def polytropic_head_coeff_eqn(b, t):
+        return b.mu_p[t] == b.impeller_work_coeff[t] * b.eff_p[t]
+
+    def polytropic_efficiency_eqn(b, t):
```
eslickj, line 114: Looks like you are missing a decorator here too.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,659 @@
...
+                         within=NonNegativeReals,
+                         units=pyunits.dimensionless)
+        self.psi_s = Var(self.flowsheet().config.time,
+                         initialize=0.01,
+                         doc="Dimensionless Isentropic Head Coefficient",
+                         within=NonNegativeReals,
+                         units=pyunits.dimensionless)
+        ########################################################
+        # Declare parameters for the model
+        self.k_v = Param(default=1.2766,
```
eslickj, line 278: It is probably better just to use all Var rather than Param, and fix the variables.  If you know the parameter values, it doesn't mater, but if you want to use parameter estimation, you'll want these to be variables.  I don't think I would uses parameters unless it's a value you really know, and would never try to estimate for some kind of data.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,659 @@
...
+        self.eff_drive = Param(default=1.0,
+                               doc="Driver efficiency")
+
+        ############################################################
+
+        self.ratioP[:] = 2.0   # make sure these have a number value
+        self.deltaP[:] = 0     # to avoid an error later in initialize
+
+        ###########################################################
+        # Speed of sound calculations
```
eslickj, line 305: If you test this with CO2 we can get the speed of sound from the property package.  I think we can just make speed of sound a requirement that this model has of property packages  that will get you out of needing a finite difference here and move this to the prop pack where it can be handled more appropriately.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,659 @@
...
+
+        for t in blk.flowsheet().config.time:
+            # if there is not a good guess for efficiency or outlet pressure
+            # provide something reasonable.
+            eff = blk.efficiency_isentropic[t]
+            eff.fix(eff.value if value(eff) > 0.3 and value(eff
+                                                            ) < 1.0 else 0.85)
+            # check for alternate pressure specs
+            if blk.outlet.pressure[t].fixed:
+                blk.ratioP[t] = value(
```
eslickj, line 557: May want to throw the third pressure change variable in here (deltaP in this case).  There is a chance someone fixed two things to specify pressure change, and you may want to test for that, and raise and exception explaining it to the user.

idaes/power_generation/carbon_capture/compression_system/compressor.py
eslickj: I think a good approach to initialization here is to just use the pressure changer initialization once you have the inlet, the efficiency and pressure ration reasonable.  Just deactivate all extra constraints added in this model fix the efficiency, call the pressure changer initialization.  Then either reactivate all the constraints and free the efficiency and solve if possible, or reactivate them in a step-wise way.

eslickj (changes requested):

eslickj:
@minhwvu let me know when you get through the comments and I'll do some more testing and maybe help you put the final polish of this.

idaes/power_generation/carbon_capture/compression_system/compressor.py
andrewlee94: I second that - speed of sound is a thermophysical property, and so should be calculated by the property package.

idaes/power_generation/carbon_capture/compression_system/compressor.py
minhwvu: Hi @eslickj , @andrewlee94 : I got the same error when running a test. Can you check the latest one? Thanks!
TypeError: unsupported operand type(s) for *: 'int' and 'dict'

idaes/power_generation/carbon_capture/compression_system/tests/test_compressor.py
minhwvu: It is a wrong value of inlet flowrate. I update the new value for checking the pressure ratio.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,603 @@
...
+                         "Change")
+        def delta_enth_actual(b, t):
+            return properties_out[t].enth_mol - properties_in[t].enth_mol
+
+        @self.Expression(self.flowsheet().config.time, doc="Entropy change in"
+                         "Compressor Stage")
+        def deltaS(b, t):
+            return properties_out[t].entr_mol - properties_in[t].entr_mol
+
+        @self.Constraint(self.flowsheet().config.time, doc="Isentropic"
```
eslickj, line 337: I was having some trouble with convergence and how the initialization worked, and going through the model in more detail, and I found that it seem like the isentropic  efficiency constraint seems to be duplicated here and in the pressure changer model.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,603 @@
...
+    def polytropic_head_coeff_eqn(b, t):
+        return b.mu_p[t] == b.mu_p_v[t]
+
+    @blk.Constraint(blk.flowsheet().config.time)
+    def polytropic_efficiency_eqn(b, t):
+        return b.eff_p[t] == b.eff_p_v[t]
+
+
+def _build_vaneless_diffuser_equations(blk):
+    blk.mass_flow_coeff = Var(blk.flowsheet().config.time,
```
eslickj, line 105: Looks like this variable is duplicated here and in the main build method.

idaes/power_generation/carbon_capture/compression_system/compressor.py
minhwvu: hi @eslickj : I comment out those lines and they give me an extra degree of freedom when I run the test. I try to fix the impeller tip radius variable (r2) but the solver throw me a below error.
pyutilib.common._exceptions.ApplicationError: Solver (ipopt) did not exit normally

eslickj: The extra degree of freedom could be a result of a missing constraint due to the earlier missing constraint decorator.

minhwvu: Hi @eslickj : Thanks. I check the model and adding the missing earlier decorator. Now I still get the same error. Please see the below.
c:\users\minh2\miniconda3\envs\compression\lib\site-packages\pyomo\opt\base\solvers.py:575: in solve
    self._presolve(*args, **kwds)
c:\users\minh2\miniconda3\envs\compression\lib\site-packages\pyomo\opt\solver\shellcmd.py:198: in _presolve
    OptSolver._presolve(self, *args, **kwds)
c:\users\minh2\miniconda3\envs\compression\lib\site-packages\pyomo\opt\base\solvers.py:675: in _presolve
    **kwds)
c:\users\minh2\miniconda3\envs\compression\lib\site-packages\pyomo\opt\base\solvers.py:746: in _convert_problem
    **kwds)
c:\users\minh2\miniconda3\envs\compression\lib\site-packages\pyomo\opt\base\convert.py:105: in convert_problem
    problem_files, symbol_map = converter.apply(*tmp, **tmpkw)
c:\users\minh2\miniconda3\envs\compression\lib\site-packages\pyomo\solvers\plugins\converter\model.py:191: in apply
    io_options=io_options)
c:\users\minh2\miniconda3\envs\compression\lib\site-packages\pyomo\core\base\block.py:1814: in write
    io_options)
pyomo\repn\plugins\ampl\ampl_.pyx:362: in pyomo.repn.plugins.ampl.ampl_.ProblemWriter_nl.__call__
    ???
pyomo\repn\plugins\ampl\ampl_.pyx:363: in pyomo.repn.plugins.ampl.ampl_.ProblemWriter_nl.__call__
    ???
pyomo\repn\plugins\ampl\ampl_.pyx:365: in pyomo.repn.plugins.ampl.ampl_.ProblemWriter_nl.__call__
    ???
pyomo\repn\plugins\ampl\ampl_.pyx:958: in pyomo.repn.plugins.ampl.ampl_.ProblemWriter_nl._print_model_NL
    ???
pyomo\repn\standard_repn.pyx:372: in pyomo.repn.standard_repn.generate_standard_repn
    ???
pyomo\repn\standard_repn.pyx:978: in pyomo.repn.standard_repn._generate_standard_repn
    ???
pyomo\repn\standard_repn.pyx:477: in pyomo.repn.standard_repn._collect_sum
    ???
pyomo\repn\standard_repn.pyx:950: in pyomo.repn.standard_repn._collect_standard_repn
    ???
pyomo\repn\standard_repn.pyx:843: in pyomo.repn.standard_repn._collect_negation
    ???
pyomo\repn\standard_repn.pyx:950: in pyomo.repn.standard_repn._collect_standard_repn
    ???
pyomo\repn\standard_repn.pyx:477: in pyomo.repn.standard_repn._collect_sum
    ???
pyomo\repn\standard_repn.pyx:950: in pyomo.repn.standard_repn._collect_standard_repn
    ???
pyomo\repn\standard_repn.pyx:862: in pyomo.repn.standard_repn._collect_identity
    ???
pyomo\repn\standard_repn.pyx:950: in pyomo.repn.standard_repn._collect_standard_repn
    ???
pyomo\repn\standard_repn.pyx:570: in pyomo.repn.standard_repn._collect_prod
    ???
pyomo\repn\standard_repn.pyx:950: in pyomo.repn.standard_repn._collect_standard_repn
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   TypeError: unsupported operand type(s) for *: 'int' and 'dict'

pyomo\repn\standard_repn.pyx:656: TypeError

idaes/power_generation/carbon_capture/compression_system/compressor.py
minhwvu: Hi @eslickj @andrewlee94 : I uploaded the latest model. I still cannot get the convergence due to the below error. Please help me check the code again. Thanks!
  TypeError: unsupported operand type(s) for *: 'int' and 'dict'

eslickj:
@minhwvu, I was working with your model and I cleaned it up a bit and tried to rework the initialization some.  The attached initialization solves, but it has an extra degree of freedom.  I think there is either a problem with the constraints or the problem specification in your model, but I'm not familiar with it.

Can you do two things.  1) Working from the attached file check all the constraints and 2) provide some more information about what problem you are solving in the test and why you fix what you fix and what you expect to get from the solution?
[compressor.zip](https://github.com/IDAES/idaes-pse/files/6091072/compressor.zip)

andrewlee94:
@minhwvu What is the progress on this PR? It has been sitting for a while and a lot of tests are failing.

minhwvu:
> @minhwvu What is the progress on this PR? It has been sitting for a while and a lot of tests are failing.

Hi @andrewlee94 : I am working on it. The revised model shows a little bit improvement on initialization. I need to compare its performance with the base case before sending it for your review again.

minhwvu:
hi @andrewlee94 , @eslickj : Could you review my revised model again? Thanks!

eslickj:
@minhwvu looks like tests are failing.  Could you take a look and see if you can figure out why?

minhwvu:
Hi @eslickj : I fixed the test failing issue. Could you check the model again? Thanks!

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -0,0 +1,507 @@
...
+            "polytropic_correlation",
+            "delta_enth_polytropic_con",
+            "mass_flow_coeff_eqn",
+            "psi_3_eqn",
+            "psi_s_eqn"]
+
+        # deactivate unit model level constraints
+        for c in self.component_objects(pyo.Constraint, descend_into=True):
+            if c.local_name in constraint_list:
+                c.deactivate()
```
eslickj, line 481: This is kind of a nit pick, but a more straight-forward way would be

```python

for c in constraint_list:
    self.component(c).deactivate()
```
I guess that's assuming the constraints are attached to ```self```, which they probably are but I didn't check.

eslickj (approved):
As far as I can tell by looking at the code, it's good.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -178,27 +162,18 @@ class CompressionStageData(PressureChangerData):
...
 
     def build(self):
         super().build()
 
+        # add ports to compressor
+        self.add_inlet_port(name="vapor_inlet", block=self.control_volume)
+        self.add_outlet_port(name="vapor_outlet", block=self.control_volume)
+
+        # Pyomo arc connect vapor inlet steam with control volume inlet
+        TransformationFactory("network.expand_arcs").apply_to(self)
```
andrewlee94, line 174: What are you trying to do here? As far as I can tell, this is build on a standard compressor model, which already has inlet and outlet ports.

Lines 170 and 171 seem to be just duplicating the existing Ports, and there is no need to expand Arcs, as there are no Arcs in the model as far as I can tell.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -207,25 +182,16 @@ def build(self):
         self.mass_flow_coeff = Var(self.flowsheet().config.time,
                                    initialize=0.0735,
```
andrewlee94, line 183: Just an observation, but these are rather specific values for a general purpose model.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -273,65 +232,35 @@ def build(self):
...
-        self.efficiency_mech = Param(default=0.97,
-                                     doc="Mechanical Efficiency")
-        self.eff_drive = Param(default=1.0,
-                               doc="Driver efficiency")
+        # Declare variables for the model
+        self.z_s = Var(initialize=0.97373,
+                       doc="Compressibility factors at suction")
+        self.z_d1 = Var(initialize=0.88949,
+                        doc="variable used for calculating"
+                        "compressibility factor at discharge pressure")
```
andrewlee94, line 242: Formatting: Comment should start with a capital.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -273,65 +232,35 @@ def build(self):
...
         def Ma_con(b, t):
-            return b.Ma[t] == b.U2[t] / b.c0[t]
-
-        @self.Constraint(self.flowsheet().config.time, doc='Pressure Ratio')
-        def Pratio_con(b, t):
-            return properties_out[t].pressure == b.ratioP[t] * \
-                  properties_in[t].pressure
+            # assume single vapor phase.
+            speed_of_sound = self.control_volume.properties_in[
+                t].speed_sound_phase['Vap']
```
andrewlee94, line 262: Make sure you document that this model assumes a single vapor phase named "Vap" - IDAES lets a user use whatever name they wish.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -417,113 +339,81 @@ def delta_enth_polytropic_con(b, t):
...
-        def psi_s_stage_eqn(b, t):
-            if self.config.first_stage is True:
-                return b.psi_s[t] == b.coeff_a[t] * b.psi_3[t]**2 + \
-                    b.coeff_b[t] * b.psi_3[t] + b.coeff_c[t]
-            else:
-                return b.psi_s[t] == b.coeff_a * b.psi_3[t]**2 + \
-                    b.coeff_b * b.psi_3[t] + b.coeff_c
-
-    def initialize(blk, state_args={}, outlvl=idaeslog.NOTSET, solver='ipopt',
+    def initialize(self, state_args={}, outlvl=idaeslog.NOTSET, solver='ipopt',
```
andrewlee94, line 416: A recent PR changed to standard approach for solver setting in initialization routines to fix some bugs in the general approach.

The standard arguments should now be `solver=None, optarg={}`

Later comments will show how to make this work.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -533,127 +423,84 @@ def initialize(blk, state_args={}, outlvl=idaeslog.NOTSET, solver='ipopt',
...
-        init_log = idaeslog.getInitLogger(blk.name, outlvl, tag="unit")
-        solve_log = idaeslog.getSolveLogger(blk.name, outlvl, tag="unit")
-
-        flags = blk.control_volume.initialize(state_args=state_args,
-                                              hold_state=True)
+        init_log = idaeslog.getInitLogger(self.name, outlvl, tag="unit")
+        solve_log = idaeslog.getSolveLogger(self.name, outlvl, tag="unit")
 
-        opt = SolverFactory(solver)
+        opt = pyo.SolverFactory(solver)
```
andrewlee94, line 433: Replace this with:

```
opt = get_solver(solver, optarg)
```

You can import `get_solver` from `idaes.core.util`.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -533,127 +423,84 @@ def initialize(blk, state_args={}, outlvl=idaeslog.NOTSET, solver='ipopt',
...
+        for c in constraint_list:
+            self.component(c).deactivate()
+
+        state_args = {
+            'flow_mol': self.vapor_inlet.flow_mol[0].value,
+            'pressure': self.vapor_inlet.pressure[0].value,
+            'enth_mol': self.vapor_inlet.enth_mol[0].value
+        }
+
+        flags = self.control_volume.initialize(
```
andrewlee94, line 488: This step seems redundant - the compressor initialization (next step) should do this for you anyway, Also, passing the `state_args` also appears redundant, as you are just using the inlet conditions, which is the default for the compressor initialization anyway.

andrewlee94 (changes requested):
There appear to be a few redundant steps in all of this that should be removed.

idaes/power_generation/carbon_capture/compression_system/compressor.py
eslickj: I wonder if this should be a config option.  Could default to "Vap", but could specify something else.  The phase will always be vapor, but there is a chance the property package doesn't call it "Vap."  That would also serve as partial documentation of it.

andrewlee94: An alternative would be iterate through all phases and check to see it it is an instance of VaporPhase.

minhwvu: @eslickj : I like this idea and it is useful if the property package can handle two-phase inlet stream. Is it correct?
@andrewlee94 : can you point me out how to proceed the iteration? Do I need to declare the VaporPhase and LiquidPhase to iterate it? Please let me know. Thanks.

andrewlee94: Phases are decaled by the property package:

```
for p in self.config.property_package.phase_list:
    pobj = self.config.property_package.get_phase(p)
    if isinstance(pobj, VaporPhase):
        return p
```

minhwvu: Thanks @andrewlee94 . I think this method does not work when I use with the swCO2 property package. Instead, I will save that with another VLE property package. Please check the revised model again. Thanks.

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -113,6 +117,11 @@ def polytropic_efficiency_eqn(b, t):
 class CompressionStageData(CompressorData):
     # Pressure changer with isentropic compressor options
     CONFIG = CompressorData.CONFIG()
+    CONFIG.declare("inlet_phase", ConfigValue(
+        default='Vap',
```
andrewlee94, line 121: This does not really resolve the issue at hand, and in fact could make it worse. All of the calculations are based on the assumption of a single vapour phase, so we probably need to check this and raise an Exception if it is not the case. I.e. add the following to `build`:

```
if self.config.property_package.phase_list != ["Vap"]:
    raise ConfigurationError(
        "{} SCO2 compressor provided with invalid property package. "
        "The SCO2 compressor only support a single vapor phase."
        .format(self.name))
```

As it stands, this would allow a user to provide a property package with an arbitrary number of phases, but would ignore anything other than the indicated phase ("Liq" or "Vap").

andrewlee94 (changes requested):
Still a couple of things that need to be fixed. Notably, the config argument for selecting phase doesn't actually solve the problem at hand.

idaes/power_generation/carbon_capture/compression_system/compressor.py
minhwvu: Hi @andrewlee94 : I upload the revised version. Could you take a look at it again? Thanks!

idaes/power_generation/carbon_capture/compression_system/compressor.py
```diff
@@ -503,7 +492,7 @@ def initialize(self, state_args={}, outlvl=idaeslog.NOTSET, solver=None,
 
         with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
             res = opt.solve(self, tee=slc.tee)
-        self.control_volume.release_state(flags=flags, outlvl=outlvl)
+        self.control_volume.release_state(flags={}, outlvl=outlvl)
```
andrewlee94, line 495: You don't need to call `release_state` here, as you didn't set `hold_state=True` when you called the `initialize` method above.

andrewlee94 (changes requested):
Only one minor comment now.

idaes/power_generation/carbon_capture/compression_system/compressor.py
minhwvu: Hi @andrewlee94 : Could you review the model again? Thanks.

eslickj (approved):

andrewlee94 (approved):
Looks good.

minhwvu:
Hi @andrewlee94 , @eslickj : I have seen the below error when running the tests. Could you let me know if there is any issue? Thanks!
"DeferredImportError: The pysp.scenariotree module (an optional Pyomo dependency) failed to import"

andrewlee94:
@minhwvu That is a known issue with Pyomo dependencies which we are trying to fix.

Issue #169, opened 2021-01-21
Ipopt Scaling Option


eslickj:
When you use the IDAES scaling tools and Ipopt, you should pretty much always use the "user-scaling" option.  I'm not sure what the best way to make people are aware of this is.  Maybe it would be a good idea to somehow make user-scaling the IDAES default?

Another issue with user-scaling is that if it is not the default, we have a lot of the solver objects created for initialization, and those should also use user-scaling.  For myself, when I want to make sure every solve uses user scaling, I still monkey-patching the Ipopt solver class, which works, but probably isn't good.

carldlaird:
@eslickj Sorry, John, I don't understand the issue. Is the problem that you want to use user-scaling for the variables, but then use IPOPT's automatic scaling for the constraints?

eslickj:
No just use normal use user scaling.  Everyone should, but nobody knows it.  I'm suggesting we just somehow make user scaling the default.  I'll edit the issue to make it more clear.

carldlaird:
@eslickj Ah, I understand. Does IDAES have a set of "wrapper" classes for the solvers / solver factory?
@jsiirola Should we consider an approach to change the defaults for specific solvers? Or maybe allow "new" solvers (e.g., "ipopt-idaes" that could essentially be a wrapper with different defaults / reporting?

eslickj:
@carldlaird, I like those ideas.  I'm not totally sure how it works, but maybe I could create as you say an "ipopt-idaes" solver class that inherits the ipopt one, and just have some default solver options that make sense for IDAES, maybe just user scaling.  It seems pretty straight forward, but there is probably a reason I haven't done it already. Then we could slowly migrate everything over to the new solver class.  If we go to user scaling by default, and no one uses the scaling framework, a lot of stuff will break.  Getting everyone on board and everything updated is a little delicate.

eslickj:
Ok PR #180 implements @carldlaird's suggestion.  It also lets you configure other solver options, so later we can adjust other things if needed.  You just need to ```import idaes.core.plugins``` then the "ipopt-solver" is available and the default solver options are part of the idaes configuration.

eslickj:
Changed the approach here now PR #207 maybe?

PR #180, opened 2021-01-22
Add ipopt-idaes solver

eslickj:
- Add an ipopt-idaes solver where we can configure default solver options
- Default configure to use ~~user-scaling~~ gradient-based (as in the plain ipopt solver object), but allows globally switching to user-scaling.
- Addresses #169 

### Legal Acknowledgement


idaes/config.py
```diff
@@ -117,6 +122,24 @@ def new_idaes_config_block():
...
+        ),
+    )
+
+    _config["ipopt-idaes"].declare(
+        "options",
+        pyomo.common.config.ConfigBlock(
+            implicit=True,
+            description="Default solver options for 'ipopt-idaes'",
+            doc="Default solver options for 'ipopt-idaes' solver"
+        ),
```
jsiirola, line 140: You can mix declared options with `implicit=True`.  Should you declare the options that you really want to be present (like `nlp_scaling_method`?

jsiirola (commented):
I think this is reasonable, but have a couple questions.

carldlaird:
@eslickj Should we import idaes.core.plugins when idaes is imported to make them register? Or is it common to import this anyway? @jsiirola Can you discuss pros/cons from doing this in Pyomo?

idaes/core/plugins/solvers.py
```diff
@@ -0,0 +1,12 @@
...
+
+import idaes
+
+@SolverFactory.register('ipopt-idaes', doc='The Ipopt NLP solver, with IDAES options')
+class IpoptIdaes(IPOPT):
+    def __init__(self, **kwds):
+        super().__init__(**kwds)
+        for k, v in idaes.cfg["ipopt-idaes"]["options"].items():
+            if k not in self.options:
+                self.options[k] = v
```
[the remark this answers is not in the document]
eslickj: I guess you are right.  I can remove that check.

idaes/config.py
eslickj: Yeah.  That's a good idea.  Will do.

adowling2:
As part of this, can we allow the user to optionally specify the path to the Ipopt executable? This is really helpful on macOS where someone may have already compiled Ipopt with HSL.

eslickj:
@carldlaird, we used to import plugins automatically, but the pyomo import was slow and there were complaints about the speed of imports and importing unneeded stuff.  This was especially true for command line things like get extensions where waiting 10 seconds for imports was intolerable.  I think that is less of an issue now since the Pyomo import was sped up, but I'm not sure if we should go back to automatically importing plugins.

carldlaird:
@eslickj OK, sounds good to me. I agree with this.

eslickj:
@adowling2, maybe, but it should work fine if the ipopt executable is in your path and not in the idaes bin directory.  If there is an issue finding the correct ipopt, we can figure something out for sure.  We can add a solver path option that I can slap on the front of the PATH environment variable.  Anyway file another issue if you like, and I'll do what needs done.

eslickj:
Okay, addressed comments.  Should be good to go.

jsiirola:
@adowling2: this inherits off the Ipopt solver, so you should be able to set the path to the executable the same as you can with ipopt, either:
```python
solver = SolverFactory('idaes-ipopt', executable='/path/to/your/exec')
```
or
```python
solver = SolverFactory('idaes-ipopt')
solver.set_executable('/path/to/your/exec')
```

jghouse88:
@eslickj and @carldlaird What about the case where my flowsheet does not use the scaling framework from IDAES? In that case, it is better to use the regular ipopt? The reason I ask is, the scaling framework is not entirely deployed across all unit models and prop packs (I think).

eslickj:
@jghouse88, yes it is better not to use user scaling if you don't use the IDAES scaling framework.  Using the ipopt-idaes solver is still fine you just need to set the "nlp_scaling_method" option to "gradient-based".  Or maybe better, set ```idaes.cfg["idaes-ipopt"]["options"]["nlp_scaling_method"] = "gradient-based"``` before creating the solver object.

The tricky part that worries me is initialization.  If we switch to "ipopt-idaes" in all the initialization methods and you don't want to use the IDAES scaling framework, you'd want to do set ```idaes.cfg["idaes-ipopt"]["options"]["nlp_scaling_method"] = "gradient-based"``` right away, and should be fine.  The nice thing about this is it gives you sort of global control over the solver settings.

andrewlee94:
We also have the issue that parts of the automatic scaling are currently missing if I am not mistaken - notably @TimBartholomew's open PR. I do agree that we should be moving to make the scaling transformation part of the default workflow, but I don't think we have quite filled in all the pieces yet.

jghouse88:
@eslickj and @andrewlee94 makes me nervous. But I guess, this PR is only adding an "idaes-ipopt" solver in our code base. This PR is not really going in and changing initialization solvers from ipopt to idaes-ipopt.

carldlaird:
@jghouse88 Understood. I suspect that we should start testing those out as well at some point and switching them over as we are happy - not with this PR. Agreed?

jghouse88:
> 
> 
> @jghouse88 Understood. I suspect that we should start testing those out as well at some point and switching them over as we are happy - not with this PR. Agreed?

@carldlaird Agree but I guess after we have implemented the scaling methods for all unit models and prop packs. For now, folks who are using the "idaes scaled" models can use the "idaes-ipopt".

eslickj:
@jghouse88, @andrewlee94, @carldlaird I guess we won't get the full benefit of this until we swap out the solvers used for initialization, since the point is global control over the solver options.  I think the main issue is what the default solver options should be.  Do we default assume you are using IDAES scaling or not using IDAES scaling?  We should probably pick one a stick to it to avoid messing people up in the future.  Now that @jghouse88 brought this up, I feel like maybe it's better to set the default back to gradient-based, and document that you need to set the option to user-scaling when you are using the IDAES scaling framework, since this is most like what we do now.  The only downside is that at some point setting the user-scaling option will just become boilerplate code that everyone does every time.  That said, you also need to call ```calculate_scaling_factors()``` so nothing is fully automatic and you still need to know at least a little about what's going on.

andrewlee94:
I am not too concerned about having boiler plate code for things like this - that was one of the decisions we made when we did the rebuild of the framework (i.e. be explicit about everything that happens).

jghouse88:
@eslickj My take is that as long as the scaling is deployed across all models and as long as the end user is only using our models, the default to idaes-ipopt is good. However, if the end user developed their own prop pack and are mixing and matching with our unit models, that is a scenario where things could get ugly. I don't really have a good answer what should be the default option.

eslickj:
@andrewlee94 and @jghouse88, it sounds to me like the consensus is to make "ipopt" and "ipopt-idaes" the same by default.  If you use scaling just explicitly set the user scaling option.  This eliminates @jghouse88's concern and makes everything a lot easier on us.  The biggest upside to setting the default scaling to gradient based is that it will let us finish updating all the solvers to "ipopt-idaes" in the initialization and everywhere else, without breaking anything.

I will modify the PR.

carldlaird:
@eslickj Does it make sense to "detect" whether scaling is there or not, and adjust the option? This kind of scares me, but maybe it makes sense? Maybe have an option "scaling=auto-detect"?

eslickj:
@carldlaird, probably too complicated.  I'm not sure how to reliably determine the user's intent.   I'm okay with just requiring it to be set explicitly, even if user-scaling is not the default.

eslickj:
Okay, I updated to default to "gradient-based" nlp scaling.  After this is merged, it should be totally safe to go ahead and update all our ipopt to ipopt-idaes.

eslickj:
This should be ready to go.  I think it's pretty straight forward, and it defaults to behave the same as the plain ipopt, to head off worries that using it will cause problems.

andrewlee94:
I don't understand enough of some of these changes to commnet, but I will ask if we should change the `get_default_solver method inhttps://github.com/IDAES/idaes-pse/blob/main/idaes/core/util/testing.py to point to `idaes-ipopt` instead of `ipopt`.

eslickj:
@andrewlee94, once this PR is in, the idea is to replace every instance of ipopt with ipopt-idaes (they are basically the same).  I figured I'd wait to start that until this is merged.

jghouse88 (dismissed):
LGTM

eslickj:
@jghouse88, I had to fix a merge conflict, so I need a new review but nothing really changed.

jghouse88 (approved):
LGTM

MAZamarripa (approved):
lgtm

PR #183, opened 2021-01-27
Fixing potential recursive loop in StateBlocks

andrewlee94:
## Summary/Motivation:
@agarciadiego identified an unexpected situation where accessing component_lists (and similar) through the new properties on the StateBlockData objects would end up in an infinite loop if the StateBlock was unindexed. Whilst this should never occur in practice, it is probably best to avoid the possibility, and the fix was very minor.

## Changes proposed in this PR:
- Change component_list, phase_list and phase_component_set properties on StateBlockData objects to call the  _return_component_list() (or equivalent) on the StateBlock object.

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #185, opened 2021-01-28
Fix shell escaping issues in GitHub Actions workflows

### LINKED ISSUE #181: Integration tests runs for some PRs failing immediately with shell syntax errors
After the recent updates to the CI workflows, some runs of the integration tests workflow are failing for a (seemingly random) subset of PRs. The failure happens before the PR code is checked out and is most likely entirely due to the CI machinery. It will show up as all jobs failing simultaneously with an error message similar to this:

```txt
/home/runner/work/_temp/d9c90005-4699-44b6-b0d8-02ddc74799a3.sh: line 232: syntax error near unexpected token `(' 
```

My working hypothesis is that the CI runner is not escaping the JSON payload correctly in the step I added to display it (ironically) for debugging. Concretely, if the PR body (or, I presume, title) contains an apostrophe (single quote), the integration tests workflow (again, not the integration tests themselves) will fail. 
For two PRs where this has occurred (#131 and #163), I've edited the text changing "doesn't" to "does not" and that seems to have done the trick, with the integration tests running as expected after that.

I'm working on a proper fix that will address the root cause of this issue. In the meanwhile, **a temporary workaround is to avoid using single quotes in the text (body or title) of a pull request**. Subsequent comments to the PR are fine, as they don't appear in the PR payload received by the CI runner.

lbianchi-lbl:
- Use dedicated actions to display debug info consistently and without causing errors due to shell escaping issues that would occur with the previous strategy

Testing problematic characters: it's this one, `maybe_this_one_too`.

### Legal Acknowledgement

### DIFF
--- a/.github/actions/display-debug-info/action.yml
+++ b/.github/actions/display-debug-info/action.yml
@@ -1,0 +1,17 @@
+name: Display debug info
+description: Display information about the environment and event payload
+runs:
+  using: composite
+  steps:
+    - name: Display environment
+      shell: bash
+      run: |
+        echo "::group::Environment"
+        env
+        echo "::endgroup"
+    - name: Display GitHub event payload
+      shell: bash
+      run: |
+        echo "::group::GitHub event payload"
+        jq . "$GITHUB_EVENT_PATH"
+        echo "::endgroup"
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -48,8 +48,8 @@ jobs:
           - ubuntu-18.04
           - windows-latest
     steps:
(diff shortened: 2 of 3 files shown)


ksbeattie (approved):
LGTM

lbianchi-lbl:
As a way to test that the changes in this PR successfully address the issues in #181 for other PRs, @andrewlee94 merged the changes into his branch in #183 (one of the PRs where the issues occurred), and it seems to have worked.

andrewlee94 (approved):
Not sure what exactly it does, but it fixed the problem.

PR #195, opened 2021-02-02
CLC oxidation property package

chineduobiora:
## Summary/Motivation:
Include an oxidation property package for a chemical looping oxygen carrier in the idaes-pse gas solid contactors model library.

- Adds an oxidation property package and example flowsheet to the idaes-pse gas solid contactors model library
- Provides documentation for the property packages, and flowsheets in the idaes-pse gas solid contactors model library

### Legal Acknowledgement


idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/gas_phase_thermo.py
```diff
@@ -0,0 +1,798 @@
...
+                'therm_cond': {'method': '_therm_cond', 'units': 'kJ/m.K.s'},
+                'diffusion_comp': {'method': '_diffusion_comp',
+                                   'units': 'cm2/s'}})
+
+        obj.add_default_units({'time': 's',
+                               'length': 'm',
+                               'mass': 'kg',
+                               'amount': 'mol',
+                               'temperature': 'K',
+                               'energy': 'kJ',
```
andrewlee94, line 244: 1. Remove energy and holdup from here - they are no longer supported or needed here.
2. We should start moving towards using Pyomo units here. See comment later.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/gas_phase_thermo.py
```diff
@@ -0,0 +1,798 @@
...
+
+        # Object reference for molecular weight if needed by CV1D
+        # Molecular weights
+        add_object_reference(self, "mw_comp",
+                             self.config.parameters.mw_comp)
+
+        """List the necessary state variable objects."""
+        self.flow_mol = Var(initialize=1.0,
+                            domain=Reals,
+                            doc='Component molar flowrate [mol/s]')
```
andrewlee94, line 445: You should start assigning units to variables now that we support them. Note that this will not immediately result in exceptions due to unit inconsistency, as these will only arise if you actually do a conversion or unit check.

However, we should start working towards full units support, and the first step is to start assigning units to `Vars` and `Params`.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/hetero_reactions.py
```diff
@@ -0,0 +1,517 @@
...
+    def build(self):
+        '''
+        Callable method for Block construction.
+        '''
+        super(ReactionParameterBlock, self).build()
+
+        self._reaction_block_class = ReactionBlock
+
+        # Create Phase objects
+        self.Vap = VaporPhase()
```
andrewlee94, line 100: 1. Reaction blocks should generally not build their own phase and component lists - I think it would be better to point to the original sets in the thermo property packages.
2. `Phase` and `Component` objects were not intended for use in reaction packages. Whilst there is no issue at the moment, we might want to be careful in case something arises in the future.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/hetero_reactions.py
```diff
@@ -0,0 +1,517 @@
...
+        self.N2 = Component()
+        self.O2 = Component()
+        self.CO2 = Component()
+        self.H2O = Component()
+        self.Fe2O3 = Component()
+        self.Fe3O4 = Component()
+        self.Al2O3 = Component()
+
+        # Component list subsets
+        self.gas_component_list = Set(initialize=['O2', 'N2', 'CO2', 'H2O'])
```
andrewlee94, line 113: Continuing from above, referring to the component lists in the thermo property packages would have got you these directly.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/hetero_reactions.py
```diff
@@ -0,0 +1,517 @@
...
+        self.rate_reaction_stoichiometry = {("R1", "Vap", "O2"): -1,
+                                            ("R1", "Vap", "N2"): 0,
+                                            ("R1", "Vap", "CO2"): 0,
+                                            ("R1", "Vap", "H2O"): 0,
+                                            ("R1", "Sol", "Fe2O3"): 6,
+                                            ("R1", "Sol", "Fe3O4"): -4,
+                                            ("R1", "Sol", "Al2O3"): 0}
+
+        # Reaction stoichiometric coefficient
+        self.rxn_stoich_coeff = Param(self.rate_reaction_idx,
```
andrewlee94, line 135: Do you need this `Param` component? Why not just refer to the dict you just created?

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/hetero_reactions.py
```diff
@@ -0,0 +1,517 @@
...
+        self.dh_rxn = Param(self.rate_reaction_idx,
+                            initialize=dh_rxn_dict,
+                            doc="Heat of reaction [kJ/mol]")
+
+        # Smoothing factor
+        self.eps = Param(mutable=True,
+                         default=1e-8,
+                         doc='Smoothing Factor')
+        # Reaction rate scale factor
+        self._scale_factor_rxn = Param(mutable=True,
```
andrewlee94, line 152: A note for the future, but we should also get the IDAES scaling tools supported here so that we can remove these (semi-)hard-coded scaling factors.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/hetero_reactions.py
```diff
@@ -0,0 +1,517 @@
...
+                'OC_conv_temp': {'method': "_OC_conv_temp", 'units': None},
+                'reaction_rate': {'method': "_reaction_rate",
+                                  'units': 'mol_rxn/m3.s'}
+                })
+        obj.add_default_units({'time': 's',
+                               'length': 'm',
+                               'mass': 'kg',
+                               'amount': 'mol',
+                               'temperature': 'K',
+                               'energy': 'kJ'})
```
andrewlee94, line 216: Again, energy is no longer required or supported here.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/hetero_reactions.py
```diff
@@ -0,0 +1,517 @@
...
+        init_log.info_high('Starting initialization')
+
+        # TODO - Update in the future as needed
+        # Get a single representative block for getting config arguments
+        for k in blk.keys():
+            break
+
+        # Fix state variables if not already fixed
+        # Fix state variables of the primary (solid) state block
+        state_var_flags = fix_state_vars(blk[k].config.solid_state_block)
```
andrewlee94, line 256: This maybe required here, but we should be careful not to get ourselves in a mess with different things fixing and unfixing state variables. For this case, I would suggest (if possible) that the unit model tells the state block to fix and hold the state variables, then the unit model calls the reaction initialization and finally the unit model tells the state block to unfix the state variables.

I.e. Let the reaction block assume the state is fixed (and initialized) already. At most, maybe the reaction block should check that the state variables are fixed and raise an exception if otherwise?

The reason for this is just to make things easier to follow.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/hetero_reactions.py
```diff
@@ -0,0 +1,517 @@
...
+                            blk[k].gas_state_ref.dens_mol_comp[j].value)
+            if blk[k].solid_state_ref.dens_mass_skeletal.fixed is True:
+                Dflag[k] = True
+            else:
+                Dflag[k] = False
+                blk[k].solid_state_ref.dens_mass_skeletal.fix(
+                        blk[k].solid_state_ref.dens_mass_skeletal.value)
+
+        # Set solver options
+        opt = SolverFactory(solver)
```
andrewlee94, line 280: Minor comment, but to save time you could move the creation of the solver object to within the `if` statement that runs the solver. This way, we only build the solver object if it is actually needed.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/solid_phase_thermo.py
```diff
@@ -0,0 +1,567 @@
...
+# Lawrence Berkeley National Laboratory,  National Technology & Engineering
+# Solutions of Sandia, LLC, Carnegie Mellon University, West Virginia
+# University Research Corporation, et al. All rights reserved.
+#
+# Please see the files COPYRIGHT.txt and LICENSE.txt for full copyright and
+# license information, respectively. Both files are also available online
+# at the URL "https://github.com/IDAES/idaes-pse".
+##############################################################################
+"""
+This package provides the necessary constraints for solid phase properties of
```
andrewlee94, line 14: I'll assume any comments I might have have already been covered in the other property package comments.

andrewlee94 (changes requested):
A few comments, but almost all of them are minor. THe biggest one is I think it would be good to add a little more detail in the docs so a reader has some more information about how each property is calculated.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/hetero_reactions.py
andrewlee94: Also look at the new `get_solver` method.

ksbeattie:
@chineduobiora, any news on this?

chineduobiora:
@ksbeattie I haven't had time to work on this  but will be on it in the upcoming week.

Robbybp:
Just tested this out with my BFB simulation scripts and it works as expected. I'll try to review the code tomorrow.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/hetero_reactions.py
```diff
@@ -0,0 +1,517 @@
...
+        opt.options = optarg
```
Robbybp, line 281: Same comment as above. Use `opt.options.update(optarg)` to not override the type

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/hetero_reactions.py
```diff
@@ -0,0 +1,517 @@
...
+
+        # Solve property block if non-empty
+        free_vars = 0
+        for k in blk.keys():
+            free_vars += number_unfixed_variables_in_activated_equalities(
+                blk[k])
+
+        if free_vars > 0:
+            with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
+                res = solve_indexed_blocks(opt, [blk], tee=slc.tee)
```
Robbybp, line 314: Same comments as in gas phase regarding repeating work and solving.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/hetero_reactions.py
Robbybp: I think I agree with this, but the reaction block requires more variables than just the states to be fixed, right? So you're drawing a distinction between state variables and other variables, where the unit model fixes state vars, then the reaction block does whatever it needs to during initialization to get a square problem, assuming state variables are fixed?

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/solid_phase_thermo.py
```diff
@@ -0,0 +1,567 @@
...
+            # Check when the state vars are fixed already result in dof 0
+            for k in blk.keys():
+                if degrees_of_freedom(blk[k]) != 0:
+                    raise Exception("State vars fixed but degrees of freedom "
+                                    "for state block is not zero during "
+                                    "initialization.")
+
+        # Set solver options
+        opt = SolverFactory(solver)
+        opt.options = optarg
```
Robbybp, line 273: Same comment. `opt.options.update(optarg)`

Robbybp (commented):
Mostly some general questions about initialization. Agree with Andrew about the need to test for particular values of the property variables.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/gas_phase_thermo.py
```diff
@@ -0,0 +1,798 @@
...
+                               'temperature': 'K',
+                               'energy': 'kJ',
+                               'holdup': 'mol'})
+
+
+class _GasPhaseThermoStateBlock(StateBlock):
+    """
+    This Class contains methods which should be applied to Property Blocks as a
+    whole, rather than individual elements of indexed Property Blocks.
+    """
```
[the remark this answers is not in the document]
chineduobiora: Updated. State blocks sounds about right.

idaes/gas_solid_contactors/flowsheets/ss_BFB_OC_oxidation.py
```diff
@@ -0,0 +1,171 @@
...
+                     "gas_property_package": m.fs.gas_properties})
+
+    # Build the BFB in the flowsheet
+    m.fs.BFB = BubblingFluidizedBed(
+            default={
+                    "flow_type": "co_current",
+                    "finite_elements": 5,
+                    "transformation_method": "dae.collocation",
+                    "gas_phase_config":
+                    {"property_package": m.fs.gas_properties},
```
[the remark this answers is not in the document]
chineduobiora: It might be because "property_package" is inline with the terminology used in other process simulators. Not sure of the particular reason though as this has been the standard for majority of the IDAES unit models. Maybe @andrewlee94 has a better idea.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/gas_phase_thermo.py
```diff
@@ -0,0 +1,798 @@
...
+            # Check when the state vars are fixed already result in dof 0
+            for k in blk.keys():
+                if degrees_of_freedom(blk[k]) != 0:
+                    raise Exception("State vars fixed but degrees of freedom "
+                                    "for state block is not zero during "
+                                    "initialization.")
+
+        # Set solver options
+        opt = SolverFactory(solver)
+        opt.options = optarg
```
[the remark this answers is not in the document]
chineduobiora: I have removed this. The current standard is `opt = get_solver(solver, optarg)`

idaes/gas_solid_contactors/flowsheets/ss_BFB_OC_oxidation.py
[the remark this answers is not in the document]
andrewlee94: That is what everything else uses. A better question is why state blocks use "properties".

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/gas_phase_thermo.py
```diff
@@ -0,0 +1,798 @@
...
+                if hasattr(blk[k], "enthalpy_shomate_eqn"):
+                    calculate_variable_from_constraint(
+                            blk[k].enth_mol_comp[j],
+                            blk[k].enthalpy_shomate_eqn[j])
+
+        # Solve property block if non-empty
+        free_vars = 0
+        for k in blk.keys():
+            free_vars += number_unfixed_variables_in_activated_equalities(
+                blk[k])
```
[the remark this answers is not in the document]
chineduobiora: What if the number of free vars in each key are different? Depending on what property methods are called from the unit model, this might be a possibility.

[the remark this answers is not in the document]
andrewlee94: The issue here is that in some cases a state block can end up with no unfixed variables (and thus no constraints), which causes an error in IPOPT.

One alternative solution would be to only search until an unfixed variable is found and then proceed - that would probably be more efficient as you would probably either find an unfixed variable fairly quickly, or not have many variables to iterate over in the case of a model with no unfixed vars.

That said, there might be better solutions than that. Do you have any suggestions @Robbybp?

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/gas_phase_thermo.py
```diff
@@ -0,0 +1,798 @@
...
+
+        if free_vars > 0:
+            with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
+                res = solve_indexed_blocks(opt, [blk], tee=slc.tee)
+        else:
```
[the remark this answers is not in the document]
chineduobiora: Good question. The variables not initialized by `calculate_variables_from_constraint` will take on their `initialize` values which might be arbitrary and not feasible for the problem to be solved. However, a quick check shows that not solving the blocks might shave some time of the solve if the current (prior to solving the block) values of the variables are feasible. I don't use `calculate_variables_from_constraint` for all the variables as, depending on the (nonlinearity of the) constraint, it can be finicky and give infeasible local solutions.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/hetero_reactions.py
chineduobiora: I have removed this. The current standard is opt = get_solver(solver, optarg)

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/gas_phase_thermo.py
```diff
@@ -0,0 +1,798 @@
...
+
+            if hasattr(blk[k], "mw_eqn"):
+                calculate_variable_from_constraint(
+                            blk[k].mw,
+                            blk[k].mw_eqn)
+
+            if hasattr(blk[k], "ideal_gas"):
+                calculate_variable_from_constraint(
+                            blk[k].dens_mol,
+                            blk[k].ideal_gas)
```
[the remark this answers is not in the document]
chineduobiora: That's an interesting idea. Will you like to look into this?

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/gas_phase_thermo.py
```diff
@@ -0,0 +1,798 @@
...
+                                    "for state block is not zero during "
+                                    "initialization.")
+
+        # Set solver options
+        opt = SolverFactory(solver)
+        opt.options = optarg
+
+        # ---------------------------------------------------------------------
+        # Initialise values
+        for k in blk.keys():
```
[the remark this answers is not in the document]
chineduobiora: I'm not sure if we do.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/tests/test_CLC_solid_prop.py
```diff
@@ -0,0 +1,79 @@
...
+
+from idaes.core import FlowsheetBlock
+
+from idaes.core.util.model_statistics import degrees_of_freedom
+
+from idaes.core.util.testing import (get_default_solver,
+                                     initialization_tester)
+
+from idaes.gas_solid_contactors.properties.oxygen_iron_OC_oxidation. \
+    solid_phase_thermo import SolidPhaseThermoParameterBlock
```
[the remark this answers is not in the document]
chineduobiora: "Thermo" can be dropped from the names.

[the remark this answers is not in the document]
chineduobiora: I've removed "Thermo" from the names. Let me know if you have additional suggestions.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/gas_phase_thermo.py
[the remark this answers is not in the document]
Robbybp: Yep, I have a prototype of this functionality in a local file. It's related to some other ideas I'd like to explore, like only performing the full-space nonlinear solve if it is necessary (i.e., if the variables of this state block don't form a DAG).

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/gas_phase_thermo.py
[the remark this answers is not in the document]
Robbybp: Searching only until a single variable is found sounds like a good idea. Really, I would like to omit this full-space nonlinear solve if it is unnecessary, but determining this at run-time would require a block triangularization. I don't think this is prohibitive, and I don't think it should be any more expensive than writing the nl file, but I haven't settled on what I think is the best way to implement this.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/gas_phase_thermo.py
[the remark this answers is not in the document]
Robbybp: If `calculate_variable_from_constraint` is failing, I think that means we are calculating variables in the wrong order. It should never fail if the variable we are solving for are linear in its constraints and the other variables are initialized. Determining the right order at runtime, however, requires a block triangularization, which could be implemented but requires a bit of thought about how to do it right (e.g. how to avoid repeating work, what to do in edge cases,...).

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/gas_phase_thermo.py
chineduobiora: @andrewlee94 This is an aside but how do we add derived units to the `define_metadata` method? For example I would like to add derived units for `power` like `pyunits.kJ * pyunits.s ** -1`. I haven't seen an example in any of the IDAES property packages.

andrewlee94: You cannot. In order to keep the number of unit conversions to a minimum, we force everything to be based on the 5 base quantities. Otherwise, everything single Expression and Constraint would need a unit conversion in it.

In the sense that this results in some poorly scaled variables in the natural forms, that should be addressed using the scaling tools ,not by defining different sets of units for given quantities.

chineduobiora: I see... I understand the challenges but I thought we want users to be able to specify and use any units they want.

chineduobiora: @andrewlee94 do the scaling tools allow us to do variable scaling?

andrewlee94: 1. We are looking at tools to handle the input-output level so that we can report variables in any set of units. However, for the models themselves, we need to keep things to a consistent set (we have already seen that unit conversion can be very expensive).

2. The short answer is yes, the scaling tools can do variable scaling. However, the long answer is a bit more complex. First of all, we have found that constraint scaling alone is often enough, unless you are dealing with variables with very small deviations from bounds. To get variable scaling requires using user-defined scaling factors in IPOPT which disables IPOPT's Jacobian scaling. Thus, we need an extra step (for which we have a tool) to replicate this Jacobian scaling before we send the problem to IPOPT.

ksbeattie:
@chineduobiora, any news?

chineduobiora:
> @chineduobiora, any news on this?

@ksbeattie Almost done. Hopefully by C.O.B today

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/gas_phase_thermo.py
chineduobiora: Yes, that's on my todo list. I won't have time to work on it in the next month. I'll setup a separate PR for it.

Robbybp:
I need to 
1. Test the new now-with-units property package with my simulations and optimizations
2. Implement some ideas I have for improving initialization (using block-triangularization to skip IPOPT solves if not available). This could potentially happen in another PR.

Robbybp:
I tested this code with my scripts this morning and it seems fine. I'm fine to merge. I am slightly concerned about the lack of "regression" testing, and would like to improve the initialize method, but these can be taken care of in a separate PR.

Robbybp (approved):

PR #198, opened 2021-02-04
Address issues with using modules in Config Blocks

### LINKED ISSUE #166: Generic property packages are not copyable
We (@MAZamarripa) recently hit a snag where a model using the generic Cubic EOS was not compatible with GDPOpt because the model was not cleanly clonable.  We tracked it down to the property parameter data definitions, specifically the definitions of:
```python
{ #...
"enth_mol_ig_comp": NIST,
}
```

The problem is that `NIST` is a module, and therefore not deepcopyable.  Talking to @andrewlee94 we think we can create a static class (`NIST`) that should resolve the issue:
```python
class cp_mol_ig_comp():
    @staticmethod
    def build_parameters(cobj):
        # ...

    @staticmethod
    def return_expression(b, cobj, T):
        # ...


class enth_mol_ig_comp():
    @staticmethod
    def build_parameters(cobj):
        # ...

    @staticmethod
    def return_expression(b, cobj, T):
        # ...


class entr_mol_ig_comp():
    @staticmethod
    def build_parameters(cobj):
        # ...

    @staticmethod
    def return_expression(b, cobj, T):
        # ...

class pressure_sat_comp():
    @staticmethod
    def build_parameters(cobj):
        # ...

    @staticmethod
    def return_expression(b, cobj, T, dT=False):
        # ...

    @staticmethod
    def dT_expression(b, cobj, T):
        # ...

class NIST(object):
    cp_mol_ig_comp = cp_mol_ig_comp
    enth_mol_ig_comp = enth_mol_ig_comp
    entr_mol_ig_comp = entr_mol_ig_comp
    pressure_sat_comp = pressure_sat_comp
```
The challenge is that this would need to be propagated to all the property packages.

andrewlee94:
## Summary/Motivation:
The generic property framework currently allows for users to specify modules in the configuration dictionaries when defining property packages. However, this results in issues when cloning/copying models, which was limiting their use in conceptual design tools.

- Update all module based libraries to include static classes and update all examples to use these
- Add deprecation warnings when users use modules in order to remove this in future.

### Legal Acknowledgement


agarciadiego (dismissed):
LGTM

andrewlee94:
@jsiirola Would you have time to look at this and make sure it will fixes the problem with copying/cloning models (and hopefully without creating any other problems).

jsiirola (dismissed):
This looks good.  My only comment is that I am an old-school programmer that does not like implicit base classes, so I would change all your class definitions to:
```python
class FcTP(object):
    #...
```

andrewlee94:
@agarciadiego @jsiirola I addressed the comment about implicit base classes, so this will need to be approved again.

jsiirola (approved):

agarciadiego (approved):

PR #200, opened 2021-02-04
Enable newer Pyomo version in setup.py

lbianchi-lbl:
## Fixes
- Installation errors preventing Pyomo version 6 and up from being installed

- Remove installation constraints requiring `pyomo<6` in `setup.py`
- Add `pytest-dev` job using dev dependencies (`requirements-dev.txt`) and Python 3.9
  - This uses a separate test job matrix without coverage report, and with "experimental mode", i.e. settings to prevent the entire workflow to be marked as failed if any job fails

### Legal Acknowledgement

### DIFF
--- a/.github/actions/setup-idaes/action.yml
+++ b/.github/actions/setup-idaes/action.yml
@@ -28,7 +28,9 @@ runs:
     - name: Install extensions
       shell: bash
       run: |
+        echo '::group::Output of "idaes get-extensions" command'
         idaes get-extensions --verbose
+        echo '::endgroup::'
         # add bin directory to $PATH (only valid for subsequent steps)
         echo "$(idaes bin-directory)" >> $GITHUB_PATH
     - name: Test access to executables
--- a/.github/actions/setup-idaes/action.yml
+++ b/.github/actions/setup-idaes/action.yml
@@ -18,7 +18,12 @@ runs:
     - name: Install idaes and dependencies
       shell: bash
       run: |
+        echo '::group::Output of "pip install" command'
         ${{ inputs.install-command }} ${{ inputs.install-target}}
+        echo '::endgroup::'
+        echo '::group::Output of "pip show pyomo idaes-pse"'
+        pip show pyomo idaes-pse
+        echo '::endgroup::'
         idaes --version
     - name: Install extensions
       shell: bash
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -35,40 +35,9 @@ defaults:
 
 jobs:
   pytest:
+    # description: Run pytest with dev dependencies
     name: pytest (py${{ matrix.python-version }}/${{ matrix.os }})
     runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        python-version:
-          - '3.6'
-          - '3.7'
-          - '3.8'
-        os:
-          - ubuntu-18.04
-          - windows-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: ./.github/actions/display-debug-info
-      - name: Set up Python
-        uses: actions/setup-python@v2
-        with:
-          python-version: ${{ matrix.python-version }}
-      - name: Set up idaes
-        uses: ./.github/actions/setup-idaes
-        with:
-          install-target: '.'
-      - name: Run pytest (not integration) with coverage
-        uses: ./.github/actions/pytest
-        with:
-          markexpr: not integration
-          extra-args: --cov
-      - name: Upload coverage report to Codecov
-        uses: codecov/codecov-action@v1
-  pytest-dev:
-    name: pytest-dev (py${{ matrix.python-version }}/${{ matrix.os }})
-    runs-on: ${{ matrix.os }}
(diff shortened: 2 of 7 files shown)


blnicho:
> Add `pytest-dev` job using dev dependencies (`requirements-dev.txt`) and Python 3.9
> 
>     * This uses a separate test job matrix without coverage report, and with "experimental mode", i.e. settings to prevent the entire workflow to be marked as failed if any job fails

Would it make more sense to make the dev test required and the release test as optional, or "experimental mode"? The whole point of having requirements-dev point to the IDAES branch of Pyomo is to let IDAES development depend on changes in Pyomo that might not be in the latest Pyomo release. So I think our required tests should depend on the IDAES branch of Pyomo and not a Pyomo release. 

Testing against the latest Pyomo release will let us know when it is safe to cut an IDAES release but I think this should be an optional test that can be ignored until we get close to the release date (since we usually cut Pyomo releases in preparation for the IDAES release).

lbianchi-lbl:
> > Add `pytest-dev` job using dev dependencies (`requirements-dev.txt`) and Python 3.9
> > * This uses a separate test job matrix without coverage report, and with "experimental mode", i.e. settings to prevent the entire workflow to be marked as failed if any job fails
> 
> Would it make more sense to make the dev test required and the release test as optional, or "experimental mode"? The whole point of having requirements-dev point to the IDAES branch of Pyomo is to let IDAES development depend on changes in Pyomo that might not be in the latest Pyomo release. So I think our required tests should depend on the IDAES branch of Pyomo and not a Pyomo release.
> 
> Testing against the latest Pyomo release will let us know when it is safe to cut an IDAES release but I think this should be an optional test that can be ignored until we get close to the release date (since we usually cut Pyomo releases in preparation for the IDAES release).

Sure, I think this makes sense. My main motivation for this iteration was to have the least amount of change in the CI setup (mostly motivated by me not being available for further work on this until next week.

If I understand what @blnicho is proposing, I could do the following:

  - Rename the `pytest-dev` job in this PR to `pytest` and mark it as required, i.e. make the default `pytest` job use the dev dependencies
  - Rename the `pytest` job in this PR to `pytest-rel`, so that the `pytest-rel` job will instead install dependencies from `pip install .` and therefore use a released version of Pyomo

Let me know if this is OK. These changes would be simple so I could implement this today by COB.

blnicho:
@lbianchi-lbl your proposed modifications sound great to me

blnicho (approved):

blnicho (changes requested):
After I approved this I double checked the Pyomo version being used by the tests and it looks like there is still an issue, it's still reporting use of Pyomo 5.7.3 and not 6.0.dev0

image

lbianchi-lbl:
@blnicho The Pyomo version being installed shown in your screenshot (5.7.3) matches what I've been seeing during my tests. Do we expect a different one? The version info in the current `IDAES` branch of `IDAES/pyomo` (i.e. as of commit [6c659865](https://github.com/IDAES/pyomo/commit/6c659865849eb49d774e5b107377628c553e08a0)) seems to be set to 5.7.3: [pyomo/version/info.py](https://github.com/IDAES/pyomo/commit/6c659865849eb49d774e5b107377628c553e08a0#diff-da1ece44e0a02641eea89f0ecfaaa2ae02718f60812b9d1e2c3fa06a2a60d0e1).

lbianchi-lbl:
Other than the Pyomo version, this should be ready in the sense that all tests are passing now with both dev and default dependencies.

- There were some pytest failures, which I think were unrelated to the Pyomo version
  - The failing tests are doctests run through `sphinx-build` that were not being run by pytest as the `sphinx-build` command is only available with the dev dependencies
  - Eventually, we could consider looking into running these as a separate CI job rather than being called as a subprocess by pytest, since the current strategy anyway doesn't contribute to pylint metrics such as e.g. coverage (as I understand, but of course I might be wrong)

The errors causing the test failures were straightforward to fix, so I went ahead and implemented the changes. However, I'd really need someone with more domain/IDAES expertise to double-check. I've added @eslickj as a reviewer as the one being `git blame`d, but feel free to chime in if the code in question is general enough. I'm also happy to extract those changes to a separate PR if anyone thinks this would make things neater, as the errors causing the failures are only related to the changes in this PR by coincidence.

blnicho:
> The version info in the current `IDAES` branch of `IDAES/pyomo` (i.e. as of commit [6c659865](https://github.com/IDAES/pyomo/commit/6c659865849eb49d774e5b107377628c553e08a0)) seems to be set to 5.7.3: [pyomo/version/info.py](https://github.com/IDAES/pyomo/commit/6c659865849eb49d774e5b107377628c553e08a0#diff-da1ece44e0a02641eea89f0ecfaaa2ae02718f60812b9d1e2c3fa06a2a60d0e1).

@lbianchi-lbl you're right. I guess we never updated that branch after cutting the Pyomo release. I'm going to update the IDAES branch of Pyomo and then we should relaunch the tests on this PR to make sure it is using Pyomo 6.0.dev0

blnicho (approved):

eslickj (approved):

jsiirola (approved):
Looks good, with one question about 3.9 testing...

PR #201, opened 2021-02-04
fixing molecular weight expression to use mw instead of mw_comp

nareshsusarla:
- Following changes made for computing `mw_phase` at 2 places. Using `self.params.get_component(j).mw` instead of `self.params.get_component(j).mw_comp`
-

### Legal Acknowledgement

``


eslickj (approved):
Looks good to me, so to expedite things I approve.  @andrewlee94 will have to be the final authority on this one.

andrewlee94 (approved):
I am going backwards-and-forwards on the naming convention for this, but I think in this case this is the correct way to go.

Whilst `mw_comp` is the standard naming convention, that is for a variable indexed by component. In this case however, we are looking up a single mw value from a component object, so I believe mw is the correct name.

PR #203, opened 2021-02-05
Reorganising tests of generic properties

andrewlee94:
## Summary/Motivation:
Currently, testing of some core aspects of the generic properties framework is being done in the examples tests rathe than tests of the actual components responsible for creating the required features. This is not good for multiple reasons, not least being that making otherwise simple changes in these core components results in multiple example tests failing which then need to be fixed.

This PR reorganizes the tests so that these core features are tested in the tests for the associated components rather than in the examples.

- Move all testing of components and methods created by the state definition classes to the tests for those modules, and removing them from the examples tests.
- Removing some redundant pytest mark` that should be deprecated.
- Linting of some affected some files.

### Legal Acknowledgement


agarciadiego (approved):
LGTM

eslickj (approved):

Issue #202, opened 2021-02-05
Property initialization routines ignore solver argument


andrewlee94:
In trying to use the property test harness, @TimBartholomew discovered that the `initialize` method of most (if not all) property packages take a `solver` argument, but promptly ignore this by creating their own solver object using "ipopt".

I believe the correct behavior would for the `solver` option to take either an instance of a Pyomo solver or default to `None`. The `initialize` method should then use the solver provided, or if `None`, use the default solver for IDAES (either through the `get_default_solver` method or via the IDAES config.

andrewlee94:
@eslickj What would you suggest as the best way to create the default solver object? The `get_default_solver` method was originally intended to provide a convenient (and centralized) means of doing that, but with the new use of the IDAES Config we were discussing deprecating that method.

The advantage of the centralized method is it means we can change the default solver by modifying the centralized method rather than needing to update every model in the library.

eslickj:
Sorry, not sure how I missed this.  I think the default solver should be "idaes-ipopt," but of course all the initialization methods should take the solver argument and pass it down to any children initialization methods.  I think making it None is also a solid option, but at this point people probably expect it to just work with Ipopt if they don't say otherwise, so the transition could be rough.

andrewlee94:
@eslickj Actually, the transition is fairly straight forward if we use `get_default_solver`. I am basically working my way throguh all the initialize methods and adding something similar to:

```
if solver is None:
    solverobj = get_default_solver()
else:
    solverobj = SolverFactory(solver)
    solverobj.options = optarg
```

I have most of this in place, and it seems to be working fine so far.

This way, we can easily change the default solver by just updating the `get_default_solver` method, which we can point to `idaes-ipopt` at some point (for the first pass I will stick with the current options just so we don't see changes in output values).

eslickj:
@andrewlee94, I guess "idaes-ipopt" serves the purpose of the default solver.  I just wasn't sure if we wanted to set the tolerance to 1e-6 since it would be a change from the plain ipopt defaults, but if you use the ```get_default_solver()``` method everywhere it's the same thing.  The ma27 option we should drop to avoid making trouble for people that don't have it if we are going to use the default solver outside of just testing.  It's a little confusing having two systems to do approximately the same thing.

Sorry I may have been confused before, but when we originally talked about this I thought ```get_default_solver()``` was just for testing, so I didn't give it much thought.

eslickj:
[...]

PR #205, opened 2021-02-08
Moving get_default_solver method

andrewlee94:
## Summary/Motivation:
This is the first part of addressing #202. This moves the existing `get_default_solver` method from util.testing to util.misc to avoid a potential recursive import loop and updates all existing models and tests that rely on the method. This will provide a centralized method for creating a default solver object in the absence of user input. For now, the solver options are hard-coded, but we can look at moving to using the IDAES config file in future (and all models/tests that use the method will automatically be updated0.

- Moved the `get_default_solver` method to `util.misc`, with a wrapper method left in `util.testing` with a deprecation warning for backwards compatibility.
- Updated the default initialize methods in `UnitBlockData` to use the new `get_default_solver` method in place of a hard-coded iopt instance.
- Changed the default values of the `solver` and `optarg` arguments in the `ControlVolume0D` and `ControlVolume1D` initialize methods to not presume any settings.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/misc.py
+++ b/idaes/core/util/misc.py
@@ -30,6 +30,22 @@ _log = idaeslog.getLogger(__name__)
 
 
 # Author: Andrew Lee
+def get_default_solver():
+    """
+    Tries to set-up the default solver for testing, and returns None if not
+    available
+    """
+    if pyo.SolverFactory('ipopt').available(exception_flag=False):
+        solver = pyo.SolverFactory('ipopt')
+        solver.options = {'tol': 1e-6,
+                          'linear_solver': 'ma27'}
+    else:
+        raise RuntimeError("Default solver unavailable.")
+
+    return solver
+
+
+# Author: Andrew Lee
 def add_object_reference(self, local_name, remote_object):
     """
     Method to create a reference in the local model to a remote Pyomo object.
--- a/idaes/core/control_volume0d.py
+++ b/idaes/core/control_volume0d.py
@@ -1265,7 +1265,7 @@ class ControlVolume0DBlockData(ControlVolumeBlockData):
                              'model_check method to the associated '
                              'ReactionBlock class.'.format(blk.name))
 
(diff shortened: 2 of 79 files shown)


idaes/core/util/misc.py
```diff
@@ -29,6 +29,22 @@
...
+# Author: Andrew Lee
+def get_default_solver():
+    """
+    Tries to set-up the default solver for testing, and returns None if not
+    available
+    """
+    if pyo.SolverFactory('ipopt').available(exception_flag=False):
+        solver = pyo.SolverFactory('ipopt')
+        solver.options = {'tol': 1e-6,
+                          'linear_solver': 'ma27'}
```
eslickj, line 41: If you are using Ipopt and ma27 is available, I believe ma27 is the default linear solver.  So the IDAES version should default to ma27.  Specifying this option causes the solve to fail if ma27 isn't available, which I guess could be what you are after.  If someone has Ipopt and no ma27 it is probably a small number of failures if you don't specify this option vs everything failing if you do.

idaes/core/util/misc.py
```diff
@@ -29,6 +29,22 @@
...
 
 
+# Author: Andrew Lee
+def get_default_solver():
+    """
+    Tries to set-up the default solver for testing, and returns None if not
+    available
+    """
+    if pyo.SolverFactory('ipopt').available(exception_flag=False):
+        solver = pyo.SolverFactory('ipopt')
```
eslickj, line 39: It doesn't really mater, but we'll probably want to standardize on "ipopt-idaes" here.  The default settings will probably be the same as plain ipopt forever, but just for future consistency and to set a good example.

andrewlee94:
@eslickj My intent is to standardize the solver and options in another PR. For now, I thought this was getting big enough with just moving things around, so I put it in first.

eslickj (approved):

MAZamarripa (approved):
lgtm

PR #206, opened 2021-02-09
UI visual improvements

makaylas:
## Summary/Motivation:
UI improvements to match Sarah's mockup.

- Lots of visual improvements
- Rappid toolbar is now a bootstrap navbar and a flexbox

### Legal Acknowledgement


makaylas:
@jghouse88 @andrewlee94 @MAZamarripa 

I don't need you to review the code so much as review the look of the UI. Here's a screenshot of it.

UI_after_improvements

ksbeattie (approved):
I ran `python idaes/ui/fsvis/tests/sample_visualization.py` and opened it in Firefox, Chrome and Safari on MacOS and it looks good.

jghouse88:
@makaylas Thanks for the image. My only comment from the figure is the column `Variable` in the table below the flowsheet. The first 8 rows are only showing the index of the `flow_mol_phase_comp` variable but are not showing the actual var name. I think that is a bug...

dangunter:
One minor thing I found is in the stream table, somehow -- and I can't reproduce it, so I'm not sure how --  I got to where the last couple of columns are narrower than the rest, even if you make the window wide enough to easily handle them. See image below. @makaylas if you can reproduce it, you can see if you can fix it.. but I think Refresh fixes it so not a big deal.

image

dangunter:
Another minor thing I found was that the "Help" button doesn't open a new tab, so it essentially closes the UI. You should just add "target=_blank" to the HTML attrs.

dangunter:
re: the stream table I take it back it is a bigger deal. I can reproduce it now. If you run the HDA Tutorial and put the call to visualize() at the *end* after the solve, the initial stream table has the problem of cutting off values.

image

dangunter:
If I zoom out to 30%, say, the two panels are not the same size
image

makaylas:
@dangunter for the zooming problem, what browser are you using? I can't reproduce it on Firefox, Chrome, or Safari

makaylas:
@jghouse88 It's not a bug. That was a conscious choice we made. We removed the `flow_mol_phase_comp` because it didn't look nice and we didn't realize it was necessary.

Should I put that back in?

makaylas:
@dangunter I made the Help button open a new tab.

dangunter (approved):
Looks good

lbianchi-lbl (commented):
I looked at the code and everything seems fine. Should I do anything specific in addition to that for the review? Otherwise, I'm happy to approve.

makaylas:
@lbianchi-lbl no, that's good. Thanks!

lbianchi-lbl (approved):

PR #207, opened 2021-02-09
SolverFactory Wrapper

eslickj:
## Summary/Motivation:
After talking to @andrewlee94 about how best to handle default solver settings, I decided instead of creating solver wrappers in idaes, to set global config settings, it may be better to create a wrapper for SolverFactory.  This PR does that, and gets rid of the "ipopt-idaes" solver.  This may create a little confusion, but I think I got this in before anyone started using it.

The main benefit of this approach is that it lets you set default options for any solvers without having to create new solver classes. 

The idaes version of SolverFactory will also return the default solver defined in the IDAES config if no specific solver is specified.  

Also set the default ipopt tol to 1e-6 to match ```get_default_solver()```
### Legal Acknowledgement


eslickj:
@jsiirola, I just updated this slightly.  Now if you ```import idaes.core.solvers``` the Pyomo SolverFactory object gets replaced by the IDAES one.  That means (I think) it would be used by things like meta-solvers too.  The only purpose of the new idaes.core.solvers module is to replace the solver factory with one that reads default keyword and options from the IDAES config.  One needs to explicitly import this so they should know that the solver factory was modified. 

The only difference between there IDAES solver factory and the Pyomo one is that the IDAES one reads default keyword arguments and options out of the IDAES config if there is a config block for the solver, and if you don't specify a solver it will give you whatever is set for the IDAES.

I know this could cause some problems, so I need an expert opinion.  I've been trying to come up with a nice generic way to globally set solver options, and so far this is the best I can come up with.

jsiirola:
I think this is still rather dangerous.  You are not replacing the `SolverFactory`, but rather the instance that is in `pyomo.environ`.  A solver / package / routine that imports the `SolverFactory` from elsewhere (e.g., `pyomo.opt.base.solvers`, `pyomo.opt.base`, or `pyomo.opt`) would still see the old `SolverFactory`.  Worse, a module that imports the `SolverFactory` using `from pyomo.environ import SolverFactory` may or may not see the new solver factory depending on if that module was originally imported before or after `idaes.core.solvers` was imported.

I think I would recommend one of two approaches:
- Update the *Pyomo* `SolverFactory` to allow for the registration of default solver options
- create a simple class that handles the options, deregister specific solvers from the SolverFactory, then re-register them using your new class.  Something like:

```python
class SolverWrapper(object):
    def __init__(self, name):
        self.name = name
        self.solver = SolverFactory.get_class(name)
        doc = SolverFactory.get_doc(name)
        SolverFactory.unregister(name)
        # We should really make a public method to register a class without using a decorator
        SolverFactory.register(name, doc)(self)

    def __call__(self, *args, **kwargs):
         if self.name in idaes.cfg:
            for k, v in idaes.cfg[self.name].items():
                if k not in kwargs:
                    kwargs[k] = v
                elif k == "options":
                    # options is in ConfigBlock and in kwargs, treat "options"
                    # special so individual options can have defaults not jut
                    # the whole options block
                    for opk, opv in v.items():
                        if opk not in kwargs["options"]:
                            kwargs["options"][opk] = opv
        return self.solver(*args, **kwds)

#...
SolverWrapper('ipopt')
```

andrewlee94:
I will ask the dumb question - do we even need something that fancy?

Why not just have a simple method like the current `get_default_solver` that does the following:

```
def get_default_solver(custom_options):
    solver = SolverFactory("get name from config")
    solver.options = get_default_options
    solver.options.update(custom_options)
```

jsiirola:
@andrewlee94: very good question.  I believe one of the concerns was getting default options into the solvers used by "meta solvers" like GDPOpt.

andrewlee94:
@jsiirola The metasolvers issue is a good question - however that is starting to blur the line between IDAES and Pyomo. How do we normally pass solvers and options to GDPOpt?

eslickj:
@andrewlee94 and @jsiirola, thanks.  @jsiirola I like your idea.  Maybe @andrewlee94 is right though we should probably take a step back I see what we really want out of this.  Part of the issue is that this got mixed up in @andrewlee94's ```get_default_solver()``` conversion, with a few off-hand comments people made, and that probably made it get out of control.

Here's what I need this to do:  IDAES creates solver objects like like mad, especially in the initialization.  I almost exclusively use Ipopt.  For problem scaling, I want a way at the top level to set the options for all the solvers that get created when you set up a model.  I want to set user scaling, maybe tolerances...  I do think something like @andrewlee94's ```get_default_solver()``` function would do what I want.  Then there are other questions about how the solver options you set through the IDAES config should be expected to work.  Should you be able to set options for any solver.  How to they apply to meta solvers...

I'm okay with some new idaes ```get_solver()``` function.  Maybe I was trying too hard for a sort of seamless transition. 

For now I'll take a look at @jsiirola's suggestion and see what I can do with it.

eslickj:
@jsiirola and @andrewlee94, I updated based on @jsiirola's suggestion.  I may still be overreaching, but hopefully we're closing in on a solution.  I added a "default" solver that will get you the default, which is basically a replacement for @andrewlee94's get default solver function.  I also re-registered all the registered solvers.  My thinking was that this would allow a person to configure all the solvers, but maybe that is problematic for reasons I'm unaware of.

eslickj:
@andrewlee94, the test are failing here because I updated the default Ipopt options and that causes the homotopy test to converge in fewer iterations.  Still assuming the approach here is acceptable, I'm not sure if we should stick with setting the default tolerance for ipopt to 1e-6 and update the homotopy test or not set a default ipopt tolerance and leave the test alone.

andrewlee94:
@eslickj I believe almost all our tests run with a tolerance of 1e-6 (the default of 1e-8 is very tight for a PSE application).

eslickj:
@andrewlee94, thanks.  I updated the homotopy solver test.

idaes/core/solvers.py
```diff
@@ -0,0 +1,39 @@
...
+        self.name = name
+        if name == 'default':
+            self.solver = None
+            doc = "IDAES Configured Default Solver"
+        else:
+            self.solver = SolverFactory.get_class(name)
+            doc = SolverFactory.doc(name)
+        SolverFactory.unregister(name)
+        # Re-register the solver (register is a decorator)
+        SolverFactory.register(name, doc)(self)
```
jsiirola, line 15: Should there be a test to make sure we don't override a solver twice?  Something like:
```python
try:
    if name in SolverFactory:
        self.solver = SolverFactory.get_class(name)
        doc = SolverFactory.doc(name)
    else:
        self.solver = None
        if name == 'default':
            doc = "IDAES Configured Default Solver"
        else
            doc = ""
    if type(self.solver) is not SolverWrapper:
        SolverFactory.unregister(name)
        # Re-register the solver (register is a decorator)
        SolverFactory.register(name, doc)(self)

**Review by jsiirola — COMMENTED** [review]
Overall I think this is OK.  That said, I am a little uneasy that importing IDAES will potentially change all the solvers.  That is the kind of side effect that causes hours of debugging pain.  Maybe it would be better to put the act of overriding the solvers in a function that is explicitly called by the model?  I am thinking something like:
```python
def configure_idaes_environment()
    # override solver behavior
    for c in list(SolverFactory):
        SolverWrapper(c)
    SolverWrapper('default')
```

eslickj:
@jsiirola, I totally agree with you about making the solver config biz require some explicit action from the user.  That's why it doesn't happen unless you import idaes.core.solvers.  This is the only think importing that module does, and the only reason to import it is to get this behavior where the solvers will take a default config from the IDAES config.  Do you think that's sufficient?  I could probably use something more clearly named to make sure the user knows what's happening and not just copying off and example not really knowing.

idaes/core/solvers.py
eslickj: I'm not sure.  Since this happens directly when you import a module isn't it already protected from happening twice?

jsiirola: Yes - although if you make a function that adds the wrappers (that the user calls), then there is a distinct possibility that the wrappers could be created twice.

eslickj:
@jsiirola, I added the function wrapper to be more explicit about what's happening, and a guard on re-registering a class multiple times.  I got a little fancy too and added an option to turn reading the IDAES config off and on.

idaes/core/solvers.py
```diff
@@ -0,0 +1,71 @@
...
+                    # the whole options block
+                    for opk, opv in v.items():
+                        if opk not in kwargs["options"]:
+                            kwargs["options"][opk] = opv
+        return solver(*args, **kwargs)
+
+
+def use_idaes_solver_configuration_deafults(b=True):
+    """
+    This function enables (or disables if given False as the argument) solvers
```
jsiirola, line 46: If `b == False`, the wrappers are not actually disabled...

jsiirola (approved):
I like this solution.  One thing I caught (fixable either by correcting the comment or by adding code to disable the wrappers)

idaes/core/solvers.py
jsiirola: My mistake.  I didn't catch that the boolean flag was polled as part of `__call__` above.  One stylistic thing, though: I have a knee-jerk aversion to global variables.  You could get the same behavior without using Globals by making `_use_idaes_config` a class attribute on `SolverWrapper`.

eslickj:
@jsiirola and @andrewlee94, this should be good to go.  I just switched over to a class var from a global as recommended, since @jsiirola approved.

jsiirola (approved):

andrewlee94 (approved):
I will assume that if @jsiirola is OK with this then it is ready to merge.

PR #208, opened 2021-02-09
Updating all models to use get_default_solver method

### LINKED ISSUE #202: Property initialization routines ignore solver argument

andrewlee94:
## Summary/Motivation:
This PR standardizes all unit models to make use of the `get_default_solver` method if the user does not provide a specific solver to use, and standardizes how user-provided solver names are passed between models. This primarily affects `initialize` methods, but was also extended to example and tests so that solver creation is now standard across the code base.

As a benefit of this, we can now change the default solver settings for the entire code base by changing one method instead of having to update all models and tests individually.  This can be seen by the significant reduction in the occurrence of the string `"ipopt"` in the code base.

- Standardize arguments to `initialize` methods to be `solver=None` and `optarg={}`.
- Standardize creation of solver objects in the absence of user-provided arguments
- Cleaning up a few other issues that caught my eye as I went through (notably cases where the logger levels were not being used).
- Removing a lot of redundant pytest marks - I think we can and should expect that anyone running the tests should have the default solver (IPOPT with MA27) and the IDAES binaries available.

### Legal Acknowledgement


andrewlee94:
@Robbybp @mengleCMU @chineduobiora @makaylas @OOAmusat 

I thought I should make you all aware of this change. This PR standardizes how the core models and examples create solver objects. One benefit of this is to reduce the number of hard-coded instances of use calling for IPOPT and replacing these with a centralized method so that we can make changes more easily. I did not touch the surrogate, Caprese or UI parts of the code however, so you may want to consider incorporating this new approach at some point.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
```diff
@@ -289,9 +290,12 @@ def initialize(blk, state_args=None, hold_state=False,
...
 
-        # Set solver options
-        opt = SolverFactory(solver)
-        opt.options = optarg
+        # Create solver
+        if solver is None:
+            opt = get_default_solver()
+        else:
+            opt = SolverFactory(solver)
+            opt.options = optarg
```
eslickj, line 298: This one's going to be problematic.  Since default solver comes with some options set and optarg is an empty dict, this will overwrite the default settings.  I think you want ```opt.options.update(optarg)```

eslickj: I hate to suggest this given how many times this if/else block appears, but with PR #207, I think you can replace this (assuming that PR gets approved) with:

```python
from idaes.core.solvers import SolverFactory
...
opt = SolverFactory(solver, options=optarg)
```

If solver is ```None```, you'll get the default solver specified by the IDAES config, which by default is ipopt with the ```get_default_solver()``` options.

andrewlee94: I had been thinking of ways we could eliminate all the if/else statements, as it is a lot of repeated code. If PR #207 will fix that, then that is a good thing.

TimBartholomew:
**Documentation looks like this:**

image
image
image
image

eslickj (dismissed):
Looks good.

andrewlee94:
@eslickj I added some docs for the get_solver method (and the rest of the misc utility library), so this will need to be re-approved.

eslickj (approved):

TimBartholomew (dismissed):
LGTM

TimBartholomew (approved):

jghouse88 (approved):
LGTM

PR #215, opened 2021-02-12
Enable running examples as part of the integration CI workflow

lbianchi-lbl:
## Summary/Motivation
See #213

- Create GitHub Actions action to encapsulate setup and commands to run examples
- Add job to integration workflow

### Legal Acknowledgement


lbianchi-lbl:
@dangunter @andrewlee94 this should be ready to review, in the sense that all the functionality is there and the test failures are due to actual tests failing rather than CI issues (of which the commit log offers a fun glimpse...).

The source of the build failures seems to be the same as the ones we experiences with @ksbeattie while cutting the last release, i.e. IDAES/examples-pse#23 and IDAES/examples-pse#24. One way to address those is to exclude the offending notebook sources from the build config file (which is now generated as part of the CI job, to have more control over its contents and keep the two-repo coupling to a minimum), but maybe it makes sense to use the momentum to try and fix the errors instead. What would you recommend?

There is also another issue, namely the unexpected success of the lone non-failing test run. As all other parameters are the same, the different outcome and the differences in the log order makes me think of some possible race conditions or other similarly tricky to debug issues in the build process. Compare for instance:

- [Run examples (py3.6/ubuntu)](https://github.com/IDAES/idaes-pse/pull/215/checks?check_run_id=1884688342) (passing)
- [Run examples (py3.6/windows)](https://github.com/IDAES/idaes-pse/pull/215/checks?check_run_id=1884688349) (failing)

The order of the logging (normal and colored part) is different, with the passing run displaying the same error traceback information but reporting "No notebooks build" instead of the 24/26 notebooks being built without errors in the other runs.

andrewlee94:
@lbianchi-lbl I would say we need to push to get those notebooks fixed, as otherwise we have examples that don't work (which is not good). Excluding them is a bad idea, as then they will just be forgotten - if it comes down to it I would rather remove them completely than exclude them from the tests.

From what I could tell, they both appear to be DFM examples, which would mean that @dangunter hopefully has a better idea of what they are and what it would take to fix them.

dangunter (commented):
I would modify build-ci.yml in the top-level directory instead of embedding the config here. This will allow for easier comparison with local builds and make this file smaller. Also, even though it's unlikely to be the source of problems, you should (a) build the index file with a separate command (see extended usage for an example), and (b) build the Sphinx docs, even though they're not going anywhere, so use '-cdr' instead of '--test'. Something I should have done that I would ask you do, is put the CI build commands in a comment at the top of the build-ci.yml file, so it's easy to replicate the entire workflow.

dangunter:
See https://github.com/IDAES/examples-pse/pull/31

The Parmest return value changed and this broke a DMF-related test. If that fix works (it works locally, so it should), then we can merge that in.

lbianchi-lbl:
@dangunter: I think the fix you made in IDAES/examples-pse#31 worked, even though the `run-examples` jobs are still failing on the same notebook.

The error this time is less clear, as it's throwing a `NameError` for the variable `dmf_params` when running the last cell. At first sight this looks strange because `dmf_params` is defined just above, in the previous cell. However, looking into the raw .ipynb source, the cell where `dmf_params` is defined has the `remove_cell` metadata tag, which I imagine could mean that in certain circumstances it doesn't get executed, causing the missing variable error in the the following cells. Do you think this is something that can be fixed quickly in the notebook itself?

About the build config file: I hadn't thought of local builds, and I think it's a great point, so I think that having this configuration in a standalone YAML file is the better option. Just to be sure, when you say to modify `build-ci.yml`, are you referring to the one in the `examples-pse` repository, or a separate copy to be added here in `idaes-pse`?
My original concern with relying on a file stored externally is that it effectively introduces "hidden state" with respect to this repo's tests, i.e. the outcome of the tests will depend on something that can change without any visible sign in this repository. If the `build-ci.yml` config file used for the `idaes-pse` integration tests was defined in this repository, even if it was modified in a separate PR, other open PRs would be notified of the need to be updated through the "this branch is out of date" prompt.

That said, this is more of a general concern: as I'm not familiar with all of the choices behind the current examples build setup, I'll follow your recommendation on this. If you think this is OK, I can go ahead and remove the configuration from the workflow file and open a PR in `examples-pse` implementing the same changes in `build-ci.yml`.

ksbeattie (approved):
LGTM

lbianchi-lbl:
@ksbeattie it seems the example tests run just fine now after merging  in IDAES/examples-pse#32. Give me a few minutes to revert the changes I made to make debugging easier, and then this should be ready to merge.

ksbeattie (approved):
A little worried about duplicating config details here but not enough to stop me from LGTM'ing this.

ksbeattie (approved):
LGTM

dangunter (approved):
LGTM

PR #216, opened 2021-02-12
Adding support for inherent reactions

andrewlee94:
## Summary/Motivation:
When we original designed the IDAES framework, we considered reactions to be isolated to specific reactor units, thus the use of specialized Reaction Blocks to hold reaction constraints. However, this does not work so well for electrolyte systems where many reactions are inherent in the system and need to be considered in all unit operations.

However, many unit models do not support reactions (i.e. most do not), so we need a new approach for dealing with this.

This PR proposes a solution to this - adding a new type of reaction ("inherent reactions") that need to be present in all unit operation and that are contained within the thermophysical property package instead of the reaction package (similar to phase equilibria). This adds a new optional term in the control volume material balances, which is activated automatically by a flag in the property package if inherent reactions are present.

- Add `has_inherent_reactions` property to `PhysicalParameterBlocks`, and methods to access this from the `StateBlocks`.
- Add new terms to control volume material balances and scaling factors
- Tests for new attributes and terms
- No documentation as yet....

### Legal Acknowledgement


andrewlee94:
I am open to suggests on alternative approaches for this. I am a little concerned about adding a new distinction in types of reactions, but I haven't thought of a better approach yet.

eslickj (approved):
It all looks good to me, but should probably get one review from someone with immediate plans to use this if possible to make sure everything makes sense.

andrewlee94:
@eslickj Unfortunately the only person immediately planning to use this is me....

NAWI will need this eventually, but I am not sure when another person will get to this.

eslickj:
@andrewlee94, in that case, I'll take a closer look.

eslickj:
It still seems good to me.  The only other idea I have (which may not be a good one) is that if you used a sort of true and apparent component approach, you could hide this from the framework and handle the inherent reaction business entirely in the property package, since framework would see apparent components, so the reactions wouldn't affect the mass balances there, and the reaction balances could go with the property package.  I guess the down side is things like separation units would need to be specially design to see the true components, so maybe that approach doesn't really get you anything better.

andrewlee94:
@eslickj I am working to support both true and apparent species for the material baalnces, but the issue comes when you also have kinetic reactions. Here you need to have a term in the material balances, but you would need to write these in terms of apparent species too which is not necessarily easy.

andrewlee94:
@eslickj's comments have made me realize that this might need more work, given different behaviour between true and apparent species.

adowling2:
@andrewlee94 Would this framework allow for certain reactions to be added to all unit operations (control volumes)? Would it be possible to use this approach to make the reactor models (PBR, PRF, CSTR, etc.) a special case of this generalized approach? This is more of a long-term big picture question.

andrewlee94:
@adowling2 This uses the same approach that is currently used for other types of reactions, then only difference being the logic used to determine if a given type of reaction should be included.

andrewlee94:
OK, this got a lot bigger quickly, but it now works and has tests to prove it.

andrewlee94:
Additional note: I found the source of the unexpected degree of freedom when testing inherent reactions in 1D systems, however that needs a separate PR (#230), as it has some (hopefully minor) wider reaching effects.

idaes/core/tests/test_control_volume_0d.py
```diff
@@ -729,6 +731,53 @@ def test_add_phase_component_balances_eq_rxns_no_idx():
...
 
 
+@pytest.mark.unit
+def test_add_phase_component_balances_in_rxns():
+    m = ConcreteModel()
+    m.fs = Flowsheet(default={"dynamic": False})
+    m.fs.pp = PhysicalParameterTestBlock()
+
+    # Set property package to contain inherent reactions
+    m.fs.pp._has_inherent_reactions = True
```
TimBartholomew, line 741: In practice, this `_has_inherent_reactions` attribute would be set within the property package, correct?

In other words, the user wouldn't necessarily set this at the flowsheet level, right?

TimBartholomew (approved):
LGTM. The approach looks good, I'm assuming the user testing and needs for additional features will come up with your interactions with the NAWI chemistry subteam.

idaes/core/tests/test_control_volume_0d.py
TimBartholomew: Oh I thought I got rid of this comment. I saw you set the attribute in the build method of the generic property package...

andrewlee94: To answer anyway: Yes, this set by the property package. This was just for generic testing purposes.

eslickj (approved):
Looks good to me too.  This is one of those things I don't think I can get a good grip on without really using it, but I see no problems.

PR #221, opened 2021-02-17
Enable integration tests to run automatically on approved PRs

lbianchi-lbl:
## Summary/Motivation
- Currently, running integration tests on a PR requires manually adding a label to the PR, which is cumbersome and not intuitive
- This PR enables integration tests to be dispatched automatically for approved PRs
- More specifically, integration tests will run:
  - As soon as the second approving review is submitted on a PR
  - When a commit is pushed to the branch of an approved PR without affecting the reviewed code (this is common when merging back changes from `main`, which required repetitive manual intervention)
  - (same as before) When the target branch is not a PR (e.g. when integration tests are run on the `main` branch as part of nightlies)
  - (same as before) When a special trigger label is added on a PR
    - This functionality is being kept, even though the vast majority of cases should be covered by the automatic triggers
    - The criteria used to identify the trigger label have been tweaked, so that it's easier to manipulate in other workflows, such as e.g. automatically removing the label (and thus "resetting" the trigger) as implemented in #222

## Changes proposed in this PR:
- Add functionality to programmatically determine a PR's review approval status
- Improve clarity and robustness when deciding whether integration tests should run or be skipped
- Improve distinguishing between actual CI jobs and utility jobs by prefixing utility jobs names with "(util)"

## Implementation details

- GitHub does not offer a direct way to determine a PR's overall approval status (as opposed to a single review's), so this required a bit of trial and error to get right
- The solution in this PR is relatively straightforward, using (documented) functionality of the issue/PR search API to get a list of approved open PRs, and then searching if the PR from where the workflow was triggered is among them
  - This does not require authentication or particular permissions, and therefore can be run directly on `pull_request` events triggered from forks (which run in CI environment with very limited permissions for security reasons)
  - Using only tools that are already available on the CI runners means that utility jobs to determine whether all trigger conditions are set can be run very quickly (< 5 s)

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #222, opened 2021-02-17
Add utility workflow to remove labels used to trigger CI workflows

lbianchi-lbl:
## Summary/Motivation
- Because of security contraints built into GitHub Actions, there is no direct way to manually dispatch a workflow so that it runs on a PR branch coming from a fork of the main repository
- To work around this limitation without bypassing the security measures, one of the event types allowed for PRs can be used as a makeshift trigger, as long as the workflow is set up to include this event type among its triggers
- An example of this is the integration tests workflow, that is set up to trigger when a specific label is added to the PR
- This current approach works, but has a few shortcomings:
  - The actual triggering event is emitted when the label is added, so if the trigger label is already present on the PR, it must be removed and then added again, which is cumbersome and not very intuitive
  - Additionally, if the label is not manually removed immediately after being applied, it will stay on the PR together with the actual (non-trigger) labels, contributing to visual noise

- Add utility GitHub Actions workflow containing a job to remove labels used to trigger CI workflows

## Implementation details

- The workflow is triggered when a label is added to a PR
- If the label that was added is not identified as a trigger label, the workflow will be skipped
- Otherwise, the label will be removed
  - The label removal event will be associated with the account corresponding to the access token specified 
- This workflow uses the secure [`pull_request_target`](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target) event, which isolates the CI environment from the (potentially arbitrary) code contained in the PR branch
- Both the workflow and the label removal job have their name prefixed by "(util)" to make it easier to distinguish them from the actual workflows in the various GitHub interfaces where workflow runs and checks are displayed

## For reviewers
- The `pull_request_target` event only triggers workflows defined on a repository's default branch, so this will need to be merged into `main` to verify whether it works properly in this repository
  - The workflow has been tested extensively in the separate repository that I'm using for GitHub Actions development
- In any case, this is a standalone workflow that doesn't affect any of the other workflows, so it shouldn't require particular extra precautions before being merged in
- Taking this into account, I've marked this as high priority. This is not particularly urgent per se, but having it available before cutting the release (together with #221) would reduce some of the manual work related to reviewing and merging PRs

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/util-cleanup.yml
+++ b/.github/workflows/util-cleanup.yml
@@ -1,0 +1,19 @@
+name: (util) Cleanup
+
+on:
+  pull_request_target:
+    types: [labeled]
+
+jobs:
+  clean-labels:
+    name: (util) Remove workflow trigger labels
+    runs-on: ubuntu-latest
+    if: contains(github.event.label.description, 'triggers_workflow')
+    env:
+      GITHUB_TOKEN: ${{ secrets.IDAES_BUILD_TOKEN }}
+      pull_number: ${{ github.event.pull_request.number }}
+      label_name: ${{ github.event.label.name }}
+    steps:
+      - name: Issue API call to remove label
+        run:
+          gh api --method DELETE "repos/$GITHUB_REPOSITORY/issues/$pull_number/labels/$label_name"


ksbeattie (approved):
LGTM

eslickj (approved):

PR #223, opened 2021-02-17
Fixing error in NIST property library

andrewlee94:
## Summary/Motivation:
The was an error in the NIST library for the calculation of specific molar enthalpy using the Shomate equation.

- Remove reference temperature terms in Shomate equation for enthalpy
-Update tests and confirmed they match numbers from NIST Webbook
- Removed patch fix in natural gas properties in power generation library

### Legal Acknowledgement


idaes/generic_models/properties/core/pure/NIST.py
```diff
@@ -109,11 +109,11 @@ def return_expression(b, cobj, T):
         H = (cobj.cp_mol_ig_comp_coeff_H if not h_form
              else 0*pyunits.kJ*pyunits.mol**-1)
 
-        h = (cobj.cp_mol_ig_comp_coeff_A*(t-tr) +
```
AlexNoring, line 112: `tr` defined above is now unused so you should delete it too

AlexNoring (changes requested):
tr is defined and unused, but other than that it looks good

AlexNoring (approved):
lgtm

agarciadiego (approved):
LGTM

PR #225, opened 2021-02-19
Improving header heat transfer (boiler hx 2D model)

MAZamarripa:
## Fixes
 updated header heat transfer

## Summary/Motivation:
updating drum1D calc
Fixing tests
updated header heat transfer calc

- updated header heat transfer
- updated tests

### Legal Acknowledgement


idaes/power_generation/unit_models/boiler_heat_exchanger_2D.py
```diff
@@ -2352,6 +2502,21 @@ def initialize(blk, shell_state_args=None, tube_state_args=None,
         blk.heat_tube_eqn.deactivate()
         blk.heat_shell_eqn.deactivate()
 
+        if blk.config.has_header is True:
+            blk.head_heat_conduction_eqn.deactivate()
```
andrewlee94, line 2506: Minor comment: It might be work making this a list of names that you can iterate over to activate and deactivate constraints. This will save repeating all these names again later when you reactivate them, and make it less likely that something will accidentally be forgotten.

idaes/power_generation/unit_models/boiler_heat_exchanger_2D.py
```diff
@@ -400,6 +399,9 @@ def _make_geometry(self):
                                         mutable=False,
                                         doc='Header wall thickness')
 
+            self.head_ri_scaling = Param(initialize=0.1, mutable=True,
+                                         doc='header inner radius scaling')
```
andrewlee94, line 403: For clarity, is this a numerical scaling factor, or or parameter regarding build up of physical scaling? The doc string does not make this clear.

If it is the first case, we really need to get the scaling tools fully deployed so that we can get rid of numerical scaling parameters.

andrewlee94 (commented):
This is a fairly significant change to be pushing through at the last minute. I only had one major comment on clarifying the purpose of a parameter and a couple of minor suggestions that do not need to be addressed now.

However, someone with more knowledge of the equations needs to look this over to make sure it is correct.

idaes/power_generation/flowsheets/subcritical_power_plant/tests/test_subcritical_flowsheets.py
```diff
@@ -48,9 +48,9 @@ def test_subcritical_boiler():
     # FEGT temperature
     assert (pytest.approx(1399.9583, abs=1e-2) ==
             pyo.value(m.fs_main.fs_blr.aBoiler.flue_gas_outlet.temperature[0]))
-    assert (pytest.approx(329148487.5260, abs=1e-1) ==
+    assert (pytest.approx(329148469.4485, abs=1e-1) ==
```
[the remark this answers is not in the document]
eslickj: @andrewlee94 beat me to it.  I agree.

idaes/power_generation/unit_models/boiler_heat_exchanger_2D.py
eslickj: It looks like scaling for the for a differential equation.  Should we try to use the scaling utilities for this, or is there a gap that makes them not work with the DAE stuff?  I'm probably missing something, but just as a matter of style I think we are moving away from scaling parameters like this.

eslickj (approved):
Looks good to me.  Why this change?  Is it for thermal stress calculations?  Do we need to get it in the new release?

eslickj (approved):

andrewlee94 (approved):
Looks OK to me.

minhwvu:
Looks good to me.

PR #230, opened 2021-02-24
Removing unnecessary constraint terms in 1D control volumes

andrewlee94:
## Summary/Motivation:
Note that this PR currently includes the changes from #216 as well.

As part of #216, it was found that the 1D control volume material balances write stoichiometric constraints for rate and equilibrium reactions at the control volume boundary. As the reaction terms material balances are not written at the boundary, these constraints should be redundant.

However, this current creates issues in the gas-solids contactor models, as they try to use these constraints during initialization.

- Add logic to skip stoichiometric constraints at control volume boundaries.

### Legal Acknowledgement


andrewlee94:
@chineduobiora @aostace01 In working on some stuff for NAWI, I found that the 1D control volume was writing some constraints for reaction stoichiometry at the inlet boundary. These should not be necessary, as we don't write material balances at the inlet.

However, this breaks the initialization BFB and MB models, as they do not skip initializing some of these variables (by `calcuate_variable_from_constraint`) at the boundary. I hope this is an easy fix, but it is probably better that you fix it rather than me (as you know the models better than me).

chineduobiora:
@andrewlee94 sure, I'll be happty to take a look.

andrewlee94:
Thanks @chineduobiora. Don't be put of by the number of files changed - the only changes that matter for this PR are in ControlVolume1D, lines 609-614 and 639-644 (the rest come from changes in the PR this stems from).

andrewlee94:
@chineduobiora I merged your changes and got the last few issues ironed out. This PR should be ready for review now.

chineduobiora (approved):
Looks good.

TimBartholomew (approved):
LGTM. From my perspective it looks like just a few if statements were added, then everything else was updated tests.

jghouse88 (approved):
LGTM.

Robbybp:
FWIW my preferred solution would be to just write the material and energy balances at the boundaries.

andrewlee94:
@Robbybp How would you do so without over-specifying the problem? Note that this only applies to the inlet boundary, where the inlet state is fully specified, thus I don't really see how you can write the balance equations.

Robbybp:
> @Robbybp How would you do so without over-specifying the problem? Note that this only applies to the inlet boundary, where the inlet state is fully specified, thus I don't really see how you can write the balance equations.

It applies to the boundary where the derivatives (fluxes) are not used by the discretization equations (except in Legendre/central discretizations, IIRC), which I don't think necessarily has anything to do with the inlets.
I don't think the state being fully specified has anything to do with whether or not you can write the balance equations. The balance equations (differential equations) specify the derivatives, which are used by the discretization equations to calculate the flows. The state is fully specified at this point, which means the discretization equations are not necessary to calculate the flows. If the discretization equations were present, _they_ would be redundant. The derivatives at this point are not used anywhere because they are not in the disc equations, and balances don't exist. But if the balances did exist, the derivatives would be used in them. We would be adding one additional variable and one additional constraint to the model, so the model would not be overspecified. The benefit would be that we could support a wider range of spatial discretizations if we had derivatives at these points, as well as to alleviate some confusion that results from having equations skipped for a non-obvious reason.

andrewlee94:
@Robbybp I am not sure how well that would actually work in practice. First of all, I think it is fairly standard practice to not write balances at the inlet boundary (although as you noted this is more due to the discretizations used historically, where the derivate at the boundary didn't matter.

Next, some terms in the material balance cannot be written at the inlet anyway - notably equilibrium constraints - as the state is defined and may not satisfy these constraints. That means that it is not possible to include generation/consumption terms related to the equilibrium reactions at the boundary as well, meaning the form could be different between the boundary and all other points.

Finally, there is the question of what the remaining generation/consumption terms actually mean at the boundary. E.g. what does it mean to have heat transfer at the inlet? Given the inlet is normally treated as a point (i.e. no volume), can you even have a generation term there?

Robbybp:
> Next, some terms in the material balance cannot be written at the inlet anyway - notably equilibrium constraints - as the state is defined and may not satisfy these constraints. That means that it is not possible to include generation/consumption terms related to the equilibrium reactions at the boundary as well, meaning the form could be different between the boundary and all other points.
Something like this already happens with "sum equations" for mole and mass fractions, right? Presumably the equilibrium equation's job is to fully specify the concentrations. Without the equilibrium equation, the concentrations are fully specified anyway, so there is no need for these equations. Then the concentrations are sufficient to calculate the equilibrium reaction terms in the balance equations? Let me know if there is something I'm missing here.

> Finally, there is the question of what the remaining generation/consumption terms actually mean at the boundary. E.g. what does it mean to have heat transfer at the inlet? Given the inlet is normally treated as a point (i.e. no volume), can you even have a generation term there?
To me, the inlet means the same thing as any other spatial discretization point. What exactly this is depends on your discretization. FWIW, I think of spatial discretization points as each being the center of a grid cell, which I believe fits the typical "shell balance" derivation of these differential equations.
I don't see anything inconsistent about having heat transfer at an inlet because I consider the inlet to still be "in the control volume."

andrewlee94:
@Robbybp Your second point is I think the more important one here, and is the reason for our different approaches. To me, the inlet (point 0), represents the actual inlet of the unit, i.e. a single point at which the conditions are fully known, where to you it represent the centre of the first finite element. In your case, the conditions at point 0 would therefore be different to the inlet conditions, and you would be able to write a set of material balances (indeed, you would probably need to).

Effectively, I think the only difference between the two approaches comes down to the number of finite elements compared to the number of points in the domain. In my approach, the number of finite element is equal to number of points-1, whilst for you they are equal.

Regarding your first point, you are correct that the equilibrium constraints are excluded at the boundary, for the reasons you listed. However, that also means that at that boundary there is no equilibrium reaction occurring, as the concentrations are already defined and not changing. Thus, the extent of reaction for any equilibrium reaction must be 0, and therefore there is no need to include it in the material balances.

Robbybp:
> Effectively, I think the only difference between the two approaches comes down to the number of finite elements compared to the number of points in the domain. In my approach, the number of finite element is equal to number of points-1, whilst for you they are equal.

I believe this is correct. To make matters worse, I think about sets of time points in the way you are describing, where each discretization point represents a discrete instant.

Issue #233, opened 2021-02-25
Add Conda installation steps to installation docs


lbianchi-lbl:

lbianchi-lbl:
Fulfilled by #234.

PR #234, opened 2021-02-26
Add section on installing IDAES using Conda to installation docs

lbianchi-lbl:
## Summary/Motivation
- Now that a Conda package is available for `idaes-pse`, add instructions on how to install it to the general installation docs (issue #233)

- Add "Install IDAES using Conda" section to installation docs

### Legal Acknowledgement

### DIFF
--- a/docs/getting_started/index.rst
+++ b/docs/getting_started/index.rst
@@ -179,6 +179,37 @@ Powershell Prompt.  Regardless of OS and shell, the following steps are the same
    output. You can report problems on the |github-issues|
    (Please try to be specific about the command and the offending output.)
 
+**Install IDAES using Conda**
+
+As an alternative to the ``pip install`` method described above, IDAES can also be installed using the Conda package manager.
+
+1. Create a new Conda environment with a name of your choice (in this example, ``my-idaes-env``)::
+
+    conda create --yes --name my-idaes-env python=3.8
+
+..
+
+    .. note:: The ``--yes`` optional flag can be used to perform the installation without having it pausing to ask for confirmation.
+
+2. Activate the ``my-idaes-env`` Conda environment::
+
+    conda activate my-idaes-env
+
+..
+
+    .. note:: This step is needed when starting a new session or a new console tab/window.
+
+3. Install the IDAES Conda package using the ``conda install`` subcommand::
+
+    conda install --yes -c IDAES-PSE -c conda-forge idaes-pse
+
+..
+
+    .. note:: The most recent stable release will be selected by default.
+              To instead select a particular version, specify the version tag using an ``=`` after the package name, e.g. ``idaes-pse=1.9.0rc0``.
+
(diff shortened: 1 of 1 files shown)


lbianchi-lbl:
Reviewers: you should be able to access a build of the docs generated directly from this PR at https://idaes-pse--234.org.readthedocs.build/en/234/getting_started/index.html#generic-install.

docs/getting_started/index.rst
```diff
@@ -179,6 +179,37 @@ Powershell Prompt.  Regardless of OS and shell, the following steps are the same
...
+
+    conda activate my-idaes-env
+
+..
+
+    .. note:: This step is needed when starting a new session or a new console tab/window.
+
+3. Install the IDAES Conda package using the ``conda install`` subcommand::
+
+    conda install --yes -c IDAES-PSE -c conda-forge idaes-pse
```
jsiirola, line 204: Do you need both the IDAES-PSE and conda-forge channels?  Is there a reason we are distributing through a channel separate from conda-forge?

docs/getting_started/index.rst
```diff
@@ -179,6 +179,37 @@ Powershell Prompt.  Regardless of OS and shell, the following steps are the same
...
+
+..
+
+    .. note:: The most recent stable release will be selected by default.
+              To instead select a particular version, specify the version tag using an ``=`` after the package name, e.g. ``idaes-pse=1.9.0rc0``.
+
+4. To complete the installation, follow the instructions described in the previous section from Step 2 ("Run the ``idaes get-extensions`` command...") onward.
```
jsiirola, line 211: Why don't we include the extensions in the conda install?

jsiirola (approved):
Documentation is good, some questions on the current implementation, though (but if addressed that can happen through other PRs)

docs/getting_started/index.rst
lbianchi-lbl: That's an excellent point. At the moment (i.e. the informal "alpha/beta" phase for the Conda package) the IDAES-PSE channel is needed since we're building and uploading the packages locally, as part of the release process. The next step would be to set up the conda-forge build configuration, at which point only the `-c conda-forge` will be needed (I guess the IDAES-PSE channel serves as a "training wheels" packaging solution, if you allow me the metaphor)

docs/getting_started/index.rst
lbianchi-lbl: This sounds intriguing but I'm not sure I understand precisely what you mean. Could you elaborate a bit on what you have in mind?

jsiirola: Conda packages are platform-dependent and can contain binaries/executables.  There is no (technical) reason why we could not distribute the pre-built binaries from idaes-ext as part of the conda package.  That way, users installing the conda package could just skip the `idaes get-extensions` step.

blnicho (approved):

PR #235, opened 2021-03-05
Updating SCPC steam cycle

Morgan88888888:
## Fixes
1. Replaced condenser with NTU condenser
2. Updated scaling factors
3. Updated pytest
## Summary/Motivation:
1. Replaced condenser with NTU condenser
2. Updated scaling factors
3. Updated pytest

1. Replaced condenser with NTU condenser
2. Updated scaling factors
3. Updated pytest

### Legal Acknowledgement


idaes/power_generation/flowsheets/supercritical_steam_cycle/supercritical_steam_cycle.py
```diff
@@ -487,12 +452,15 @@ def steam_cycle_eff(b, t):
...
     )
+    
     m.fs.condenser_mix_to_condenser = Arc(
-        source=m.fs.condenser_mix.outlet_tpx, destination=m.fs.condenser.inlet_1
+        source=m.fs.condenser_mix.outlet, destination=m.fs.condenser.inlet_1
     )
+    
     m.fs.COND_01 = Arc(
-        source=m.fs.condenser.outlet_1_ph, destination=m.fs.hotwell.condensate
+        source=m.fs.condenser.outlet_1, destination=m.fs.hotwell.condensate
```
eslickj, line 461: While the ```outlet_1``` port name still works for backward compatibility it may be better to use the new more clear name ```hot_outlet```.  Same with the other ports.  Also ```side_1``` -> ```hot_side``` and ```side_2``` -> ```cold_side```.  Also ```tube``` and ```shell``` work.  There are a few options, but the old 1 and 2 always make me have to look up which is which.  I know that the names are in that way because it's an old model, but may be worth updating.

eslickj (approved):
Looks good.  Just one minor suggestion left in the comments on the code.

MAZamarripa (approved):
lgtm

PR #239, opened 2021-03-11
Minor update to copy_port_values(), now can take and arc

eslickj:
This makes it possible to specify an arc for the ```copy_port_values()``` function.  Just saves some typing and make it a little easier to use.  It still works the way it did, just provides an extra option. 

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/misc.py
+++ b/idaes/core/util/misc.py
@@ -213,18 +213,25 @@ def svg_tag(
 
 
 # Author: John Eslick
-def copy_port_values(destination, source):
+def copy_port_values(destination=None, source=None, arc=None):
     """
     Copy the variable values in the source port to the destination port. The
     ports must containt the same variables.
 
     Args:
-        (pyomo.Port): Copy values from this port
-        (pyomo.Port): Copy values to this port
+        destination (Port): Copy values from this port
+        source (Port): Copy values to this port
+        arc (Arc): If arc is provided, use it to get default values of source
+            and destination
 
     Returns:
         None
     """
+    if arc is not None:
+        if source is None:
+            source = arc.source
+        if destination is None:
+            destination = arc.dest
     for k, v in destination.vars.items():
         if isinstance(v, pyo.Var):
             for i in v:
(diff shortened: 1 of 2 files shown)


idaes/core/util/misc.py
```diff
@@ -213,18 +213,25 @@ def svg_tag(
...
+        destination (Port): Copy values from this port
+        source (Port): Copy values to this port
+        arc (Arc): If arc is provided, use it to get default values of source
+            and destination
 
     Returns:
         None
     """
+    if arc is not None:
+        if source is None:
```
andrewlee94, line 231: It seems a bit odd to me to allow a user to specify a source/destination along with an Arc. For one, if a user were to specify all three then the `arc` argument is effectively ignored.

This won't stop the method from working, but it might be confusing to the user. Notably, the doc string says that the `arc` is used to define source and destination if provided, which is not quite correct.

I would be inclined to be restrictive here and make it an either/or approach - i.e. raise an Exception if a user provides both an `arc` and a `source` or `destination`.

andrewlee94 (approved):
Looks good to me.

idaes/core/util/misc.py
```diff
@@ -213,18 +214,48 @@ def svg_tag(
 
 
 # Author: John Eslick
-def copy_port_values(destination, source):
+def copy_port_values(destination=None, source=None, arc=None):
```
jghouse88, line 217: Very minor comment. Can we switch source and destination arguments to be consistent with arc definition? I am fine with this if you prefer this.

jghouse88 (commented):
One comment.

idaes/core/util/misc.py
eslickj: Yeah, I want to, but this method has been around for a long time, and I afraid to break peoples stuff.  I guess it's best to just use named args to avoid confusion.

jghouse88: Okay, fine with me.

jghouse88 (approved):
LGTM

PR #242, opened 2021-03-12
Mark bin download test as integration

eslickj:
## Fixes
Mark bin download test as integration.  That makes it run less often and random network-type errors that causes the download to fail will happen less often.

### Legal Acknowledgement


lbianchi-lbl (approved):
I'm not familiar with this particular test, but in general I think this is a good idea.

andrewlee94 (approved):
Looks good.

PR #251, opened 2021-03-19
Patch for initializing moving bed reaction generation variables

Robbybp:
## Summary/Motivation:
I am working on some code to construct moving bed models where the two phases have discretizations in opposite directions. This requires checking the `transformation_scheme` of the control volumes rather than the unit model. Note that it wouldn't be necessary to branch on `transformation_scheme` at all if we let all constraints be defined at the inlets.

### Legal Acknowledgement


andrewlee94 (approved):
Looks fine to me.

aostace01 (approved):
Looks good.

PR #253, opened 2021-03-19
Add tests for addtional solvers and Bump IDAES-EXT Version

eslickj:

- Add tests for additional solvers
- Bump IDAES-EXT version to 2.4.0

Due to my unfortunate testing setup, I have a bit of a chicken and egg problem.  I need to merge this before I can run the full set of tests all solvers on all platforms.  If it passes here but not in the more comprehensive tests with CentOS and additional Ubuntu releases, I'll quickly make a 2.4.1 release.  I'm 90% sure everything works, and pretty much 100% sure Ipopt at least works everywhere, so this shouldn't break things for anyone.  

I'd like merge this ASAP so I can finish up the testing and stop messing with this.

### Legal Acknowledgement


andrewlee94 (dismissed):
Looks OK to me. Only minor comment was that I didn't see an IPOPT test in there, but that is minor (and it's not like we don't have tests which use it).

eslickj:
@andrewlee94, Ipopt is so important it already has it's very own test file.

lbianchi-lbl (dismissed):

eslickj:
@andrewlee94 and @lbianchi-lbl, sorry I forgot to push the change to increment the binary release version.  Could you re-review.

andrewlee94 (dismissed):
Looks good.

lbianchi-lbl (approved):

eslickj:
@lbianchi-lbl, there is a problem I ran into here with the github action for installing IDAES.  I think we should skip the part that tries to run ipopt after running get extensions.  Using ipopt from an IDAES model does not require updating PATH.  We'll also see that ipopt is missing later in the tests if somehow it wasn't installed without running into an error.  Alternatively we can keep the Ipopt check in the action, but you'll need to update LD_LIBRARY_PATH as well now.

lbianchi-lbl:
@eslickj This makes sense, I added the check without being too aware of the internals, but as I understand from what you're saying, it doesn't add anything that it's not already provided within IDAES.

If you're OK with it, I can get rid of that step in the GH Actions workflows, and push the changes directly on your branch for this PR.

eslickj:
@lbianchi-lbl, sounds good.  Thanks for updating it.

andrewlee94 (approved):
I will assume that everyone knows what they are doing.

lbianchi-lbl (approved):
:+1:

Issue #236, opened 2021-03-06
Pytest fails on brand new source installation WSL2


bernalde:
Environment data
OS (Windows | Mac | Linux distro) and version: 5.4.72-microsoft-standard-WSL2 Description: Ubuntu 20.04.1 LTS Release: 20.04
Python and/or Anaconda version: Python 3.8.5 conda 4.9.2
Type of virtual environment used (N/A | venv | virtualenv | conda | ...): conda
idaes-pse version: idaes, version 1.9.0.dev0
pyomo version: Pyomo 6.0 (devel {master}) (CPython 3.8.5 on Linux 5.4.72-microsoft-standard-WSL2)
[idaes_test.txt](https://github.com/IDAES/idaes-pse/files/6094380/idaes_test.txt)

After doing the installation from source, and calling `idaes get-extensions`, y ran the tests via pytests. Some of them failed, see the logs below:
Full test output attached.
```
=============================================== short test summary info ================================================
FAILED idaes/commands/tests/test_commands.py::test_examples_cli_default_version - assert 0 == -1
FAILED idaes/commands/tests/test_commands.py::test_examples_install_src - FileExistsError: [Errno 17] File exists: '/...
FAILED idaes/core/tests/test_control_volume_0d.py::test_add_total_element_balances_lineraly_dependent - assert 30 == 21
FAILED idaes/core/tests/test_control_volume_1d.py::test_add_total_element_balances_lineraly_dependent - assert 30 == 21
FAILED idaes/core/util/tests/test_phase_equilibria.py::test_Txy_data_no_dew - assert [191.46973758...6254729593885] =...
FAILED idaes/generic_models/unit_models/distillation/tests/test_tray_column.py::TestBTXIdeal::test_solution - assert ...
================================ 6 failed, 3749 passed, 3 skipped in 911.96s (0:15:11) =================================
```
Each detailed test output is:
```
E       assert 0 == -1
E        +  where 0 = <Result okay>.exit_code

idaes/commands/tests/test_commands.py:131: AssertionError
```
```
>       src_dir.mkdir()

idaes/commands/tests/test_commands.py:259:
```
```
E           FileExistsError: [Errno 17] File exists: '/home/bernalde/.idaes/_scratch/examples_install_src/src'

../../../anaconda3/envs/optim/lib/python3.8/pathlib.py:1284: FileExistsError
```
```
E           assert 30 == 21
[...]

PR #257, opened 2021-03-19
Add more debug logging to solver download.

eslickj:
 Add some extra debug logging to diagnose issue #236.  Now it you use the --verbose flag you'll see what we think the Linux distro is.  My guess with issue #236 is that we will see that it included something about WSL.   

### Legal Acknowledgement


andrewlee94 (approved):
Looks OK to me.

dangunter:
I think it would be safer to print out a list of known distributions and stop instead of assuming a default and continuing. To support the workflow where an expert user knows the right one, an option like `--os=<name>" can be added, so they can pick one of the values from the list. After listing out possible values, the error message could include a hint like "If you know that your OS and distribution is equivalent to one of the ones listed, use the '--os=<name>' option to choose the appropriate name and avoid this check."

dangunter (changes requested):
See comment re: setting OS explicitly.

eslickj:
@dangunter and @andrewlee94, updated it, now if a supported platform isn't detected it raises and exception and prompts you to specify one.  It also tells you what was detected so like if it detects ubuntuWSL.18.04 or something weird you know what happened.

Here's and example of asking for "linux" on Windows 10 (seems a little silly because I detect the linux distro as win10 so maybe I should just get windows anyway, but I did say I wanted linux, anyway you get the idea):

```sh
(idaes-pse) C:\Users\EslickJ\git\idaes-pse\idaes\generic_models\properties\core\generic>idaes get-extensions --platform linux --verbose
Getting files... 
2021-03-22 21:18:32 [DEBUG] idaes.util.download_bin: Detected Linux distribution: win10                                                                                                                                                                                                                                                 

Detected platform win10 is not recognized as supported platform.                                                                                                                                                                                                                                                                        

Specify one of the following platforms with --platform <os>:
  auto
  centos6
  centos7
  centos8                                                                                                                                                             
  darwin
  rhel6                                                                                                                                                               
  rhel7
  rhel8                                                                                                                                                               
  ubuntu1804
  ubuntu1910                                                                                                                                                          
  ubuntu2004
  windows                                                                                                                                                                                 
```

jsiirola (dismissed):

eslickj:
@dangunter, could you take another look at this?

dangunter:
@eslickj I think it is good except one minor comment. You added a new command "get-extensions-platforms", but wouldn't it be simpler to just print out the platforms (and stop) if the user gives the "--platforms" option (with an s)?

```
idaes get-extensions --platform foobar  # prints out 'Unknown platform: foobar' and then lists available
idaes get-extensions --platforms # prints out nothing and lists available
```

This would require changes like:
```
@click.option(
    "--platform",
    help="Platform to download binaries for. Use '--platforms'"
         " for a list of available platforms (default=auto)",
    default="auto")
@click.option(
    "--platforms", help="List available platforms for --platform argument",
    is_flag=True)
```

and then add "platforms" to the function args, and refactor the list-platforms function:
```
def _list_platforms(platform=None):
    click.echo(
        "Specify one of the following platforms with --platform <os>:")
    for i in sorted(idaes.config.known_binary_platform):
        if i == platform:
            # auto or linux specified, and it didn't work.
            # will need to be more specific.
            continue
        click.echo(f"  {i}")
```

and add at the top of the function logic like:

```
 if platforms:
        _list_platforms()
        return
```
and, obviously, call _list_platforms() also if there is an UnsupportedPlatformError

eslickj:
@dangunter, unfortunately I have several commands like that, and I agree with you about how they should have been done.  I think about updating them every month or so then let it go.  I guess since you brought it up.  It's a good time to make some updates.

dangunter (dismissed):
I'm sure it's good now.

jsiirola (approved):

dangunter (approved):
Looks good

PR #260, opened 2021-03-20
Add MIME type for returned local JS modules

### LINKED ISSUE #259: Fix UI JavaScript MIME type errors for Windows 10
For at least Firefox/Windows10, there is a problem that the simple http server used for the UI doesn't set the MIME type to an acceptable value for the JavaScript files that are marked as 'type="module"' in the index.html file. This is a very simple fix, and should be harmless to installations that were fine with the text/plain type that was being returned previously by default.

dangunter:
## Summary/Motivation:
- This was breaking the UI on FireFox/Windows 10

- One-liner fix with a comment
- We don't have integration tests for the web UI (which is why this wasn't caught earlier)

### Legal Acknowledgement


makaylas (approved):
LGTM

ksbeattie (approved):
LGTM

PR #232, opened 2021-02-25
NGFC flowsheet example

AlexNoring:
## Summary/Motivation:
This PR includes the NGFC without carbon capture flowsheet and the SOFC ROM needed to run it. A follow-up PR to examples-pse will include a jupyter notebook importing and running the model. 

## Changes proposed in this PR:
- SOFC ROM and kriging coefficients are added to idaes/power_generation/properties/NGFC/ROM
- NGFC flowsheet example, init file, and results visualization files are added to idaes/power_generation/flowsheets/NGFC

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #261, opened 2021-03-22
Fixing bug in control volume scaling

andrewlee94:
## Summary/Motivation:
PR #232 revealed a bug in the control volume scaling methods, where it was assumed that a volume variable would exist whenever when scaling `rate_reaction_extent`. After considering how this term is defined in a unit model, I came to the conclusion that only the unit model can tell how to scale this variable (e.g. whether it will be a fixed yield, or based on reaction rate), so this should be left to the unit model to determine.

- Set scaling factor of `rate_reaction_extent` to 1 if undefined in both 0D and 1D control volumes.
-

### Legal Acknowledgement


idaes/core/control_volume0d.py
```diff
@@ -1630,12 +1630,11 @@ def calculate_scaling_factors(self):
...
-                        self.reactions[t].reaction_rate[r],
-                        default=1,
-                        warning=True)
-                    sf *= iscale.get_scaling_factor(self.volume[t])
-                    iscale.set_scaling_factor(v, sf)
+                    # Control Volume has no way of knowing how best to scale
+                    # reaction extent - e.g. whether it will be calculated from
+                    # reaction rate or a fixed yield. Unit models should
+                    # provide better scaling factors where possible.
+                    iscale.set_scaling_factor(v, 1)
```
eslickj, line 1637: This is okay, but if you just want the scaling factor to be 1 you don't need to set it.  If you don't set it you'll be able to generate a warning based on it not being set.

andrewlee94: I actually did this somewhat deliberately as a way of guiding the user as to what variables existed but were not scaled. This way, the user can display the suffix and see that this variable has a factor of 1, whereas if it were not set it would not appear at all. Given that the control volumes are likely "black boxes" to many users, I thought it would help to populate the suffix as a guide.

eslickj: I know we need a better way to guide users to the variables that they should scale.  Either documentation or some sort of automated tools.  As you say, if we fill the suffixes of variables we expect need to be scaled, we could develop some sort of tool based on that, so I'm good.

eslickj (dismissed):

idaes/core/control_volume0d.py
andrewlee94: Some counter-points:

1. Regarding the warning system, the control volume will generally be one of the last things called - only the unit model and flowsheet should come later. If the unit model is depending on the control volume to calculate scaling factors for reaction extents, then the modeler has a fundamental disconnect in how things should be working that the warning won't help fix. That is - the unit model really should be telling the control volume what the scaling is and not the other way around. This is actually the key point here - the current scaling tools assume information only flows from the bottom up, where as there are cases like this where information has to flow both ways.
2. It is rather hard to document all of this, especially for something as generic as a control volume. Unit models definitely should recommend setting scaling factors for these when appropriate, and that would supersede anything the control volume does.
3. I would argue that displaying the scaling factors is critical - as it shows you exactly what is being scaled and how. Seeing a lot of ones in there should be a suggestion that maybe you should consider updating these.

TimBartholomew: My thoughts on your points:
1. The current scaling implementation allows unit models to assign scaling factors at the control volume level (i.e. the unit model's calculate_scaling_factor method could have a self.control_volume.X assignment). You are right that the current implementation doesn't support the control volume calculating other scaling factors based on the unit model input, but I'd argue that for any control volume variables that need input from the unit model, the scaling should be set at the unit model method (and that the control volume should not set it at all, even if it is just to 1).
2. Yes the control volume is too general for the scaling documentation. Unit model documentation should state whether the user needs to provide a scaling factor for the unit model's control volume.
3. Yes displaying scaling factors is important, but I don't think displaying and looking at the ones should be the way to see whether a variable had its scaling set or not. We have tools to make a generator of the unscaled variables and constraints for users.

An additional thought:
Most of the time when we are setting scaling factors, we have an if statement that checks if it is not None, so that `calculate_scaling_factors` doesn't override user specified scaling factors. But if the control volume puts a 1 in for a scaling factor that needs to updated elsewhere, then that messes up our override feature. The line of code in the unit model will see a scaling factor has been assigned and skip it.

andrewlee94:
@TimBartholomew @eslickj I have removed the setting of scaling factors for the reaction extents.

I think there is still an issue regarding point 1 I made above in the the flow of information is based around the assumption for a bottom up approach. Whilst there are definitely ways around this, it does involve working around the current system rather than with the current system. A good case is the reaction extents, which need to be set by the unit model, but also depend on information from the reaction block. Ideally, the information flow would be:

1. Calculate scaling factors in the state blocks (as the reaction block will probably need this).
2. Calculate scaling factors in the reaction block (there is actually a second issue here in how you can do this for things that are exponential functions of temperature).
3. Calculate scaling factors for the extent of reaction in the unit model and set these in the control volume.
4. Calculate scaling factors in the control volume.
5. Calculate remaining scaling factors in the unit model.

However, the current tools don't really support the inclusion of Step 3 - once `calculate_scaling_factors` is called, it will more-or-less automatically run through steps 1, 2 4 and 5 without the ability to do the intermediate step 3.

TimBartholomew:
> @TimBartholomew @eslickj I have removed the setting of scaling factors for the reaction extents.
> 
> I think there is still an issue regarding point 1 I made above in the the flow of information is based around the assumption for a bottom up approach. Whilst there are definitely ways around this, it does involve working around the current system rather than with the current system. A good case is the reaction extents, which need to be set by the unit model, but also depend on information from the reaction block. Ideally, the information flow would be:
> 
> 1. Calculate scaling factors in the state blocks (as the reaction block will probably need this).
> 2. Calculate scaling factors in the reaction block (there is actually a second issue here in how you can do this for things that are exponential functions of temperature).
> 3. Calculate scaling factors for the extent of reaction in the unit model and set these in the control volume.
> 4. Calculate scaling factors in the control volume.
> 5. Calculate remaining scaling factors in the unit model.
> 
> However, the current tools don't really support the inclusion of Step 3 - once `calculate_scaling_factors` is called, it will more-or-less automatically run through steps 1, 2 4 and 5 without the ability to do the intermediate step 3.

Do you imagine step 4 relying on step 3? 

If step 4 is independent, then its fine for step 3 to be rolled into 5. 

If some calculations in step 4 require scaling factors in step 3, then I think those should also be rolled into step 5 because then we are saying they are unit model specific. This puts the burden on the unit model developer and we'd need to document for them what scaling factors on the control volume they should set, but I think it works out ok because they are setting all unit model specific scaling factors in step 5.

idaes/core/control_volume0d.py
```diff
@@ -1627,15 +1627,9 @@ def calculate_scaling_factors(self):
...
-                if iscale.get_scaling_factor(v) is None:
-                    sf = iscale.get_scaling_factor(
-                        self.reactions[t].reaction_rate[r],
-                        default=1,
-                        warning=True)
-                    sf *= iscale.get_scaling_factor(self.volume[t])
-                    iscale.set_scaling_factor(v, sf)
+        # Control Volume has no way of knowing how best to scale
+        # reaction extents - this is something only the unit model can provide
+        # This also applies ot the phase_equilibrium_generation term.
```
TimBartholomew, line 1632: "to" instead of "ot"

idaes/core/control_volume1d.py
```diff
@@ -1912,15 +1912,9 @@ def calculate_scaling_factors(self):
...
-                if iscale.get_scaling_factor(v) is None:
-                    sf = iscale.get_scaling_factor(
-                        self.reactions[t, x].reaction_rate[r],
-                        default=1,
-                        warning=True)
-                    sf *= iscale.get_scaling_factor(self._area_func(t, x))
-                    iscale.set_scaling_factor(v, sf)
+        # Control Volume has no way of knowing how best to scale
+        # reaction extents - this is something only the unit model can provide
+        # This also applies ot the phase_equilibrium_generation term.
```
TimBartholomew, line 1917: "to" instead of "ot"

eslickj:
@andrewlee94, I was worried this situation would arise.  So the extent of reaction scaling factor is actually calculated by the unit model or just set by the user?  I assume calculated because set by the user is fine.  I guess, unfortunately the solution is tough to figure out.  My best idea is to somehow flag scaling factors as user-provided or calculated.  That would let you recalculate the control volume scaling from the unit model, or just recalculate them in general.

eslickj (dismissed):

idaes/core/tests/test_control_volume_0d_scaling.py
```diff
@@ -147,7 +145,7 @@ def test_full_auto_scaling():
 
     # check that all variables have scaling factors
```
TimBartholomew, line 146: It could be useful to put an inline comment here stating which variables do not have scaling factors.

TimBartholomew (commented):
Noted a typo and I think it would be helpful to state which variables should not have scaling factors in the comments of a test. Or you could explicitly test that the variables in unscaled_var_list are the variables that should not have scaling factors.

andrewlee94:
@TimBartholomew The issue is that steps 3, 4 and 5 are all interdependent. Step 3 needs scaling for the reactions and the unit volume to calculate scaling for extents, step 4 needs those scaling factors to scale generation and the associated constraints, and step 5 may depend on those scaling factors for other things.

Whilst the unit model developer can always work through this themselves, it is working around short combining's in the scaling tools rather than working with the scaling tools. This is not something to fix in this PR, but it is something we need to think about.

andrewlee94:
@TimBartholomew @eslickj I have fixed the typos and added comments on what was not scaled. Test for these would be a bit of effort, as there are a lot of unscaled variables in some cases (even if they are just index elements of the same three `Vars`).

TimBartholomew (approved):
LGTM

eslickj (approved):

PR #262, opened 2021-03-23
Add addtional flow expressions to generic properties

eslickj:
- Add a bunch of new flow expressions to generic properties
- Add test for the expressions.

### Legal Acknowledgement


idaes/generic_models/properties/core/generic/generic_property.py
```diff
@@ -833,11 +833,15 @@ def define_metadata(cls, obj):
         # created by state var methods
         obj.add_properties(
             {'flow_mol': {'method': None},
-             'flow_mol_phase': {'method': None},
```
andrewlee94, line 836: From my opening comments, I would suggest that `flow_mol_phase_comp` and `mole_frac_phase_comp` should have `'method': None` based on the fact that the state definition module is expected to construct these. Thus, they should not have build-on-demand methods available to them (or at least, the state definition module can set up the build-on-demand methods if desired).

However, `flow_mol` should be given a build-on-demand method based on the sum of `flow_mol_phase_comp`.

idaes/generic_models/properties/core/generic/generic_property.py
```diff
@@ -1808,6 +1812,97 @@ def rule_dens_mol_phase(b, p):
...
+                return self.mole_frac_comp[i]*self.flow_mol
+            self.flow_mol_comp = Expression(
+                self.params.component_list,
+                doc="Component molar flow rate",
+                rule=rule_flow_mol_comp)
+        except AttributeError:
+            self.del_component(self.flow_mol_comp)
+            raise
+
+    def _flow_mol_phase_comp(self):
```
andrewlee94, line 1893: I would suggest removing this, and making it a standard expectation of the state definition module.

idaes/generic_models/properties/core/generic/generic_property.py
```diff
@@ -833,11 +833,15 @@ def define_metadata(cls, obj):
...
-             'mole_frac_phase_comp': {'method': None},
+             'flow_vol': {'method': '_flow_vol'},
+             'flow_mass': {'method': '_flow_mass'},
+             'flow_mass_phase': {'method': '_flow_mass_phase'},
+             'flow_vol_phase': {'method': '_flow_vol_phase'},
+             'flow_mol_phase': {'method': '_flow_mol_phase'},
+             'flow_mol_phase_comp': {'method': '_flow_mol_phase_comp'},
+             'flow_mol_comp': {'method': '_flow_mol_comp'},
+             'mole_frac_comp': {'method': '_mole_frac_comp'},
+             'mole_frac_phase_comp': {'method': '_mole_frac_phase_comp'},
```
andrewlee94, line 844: I note that you did not actually add a method for this below anyway.

andrewlee94, line 844: I also wonder if we should add scaling factors for these as well, but that could be a separate PR.

andrewlee94 (changes requested):
I think we need to make sure we have a well-defined standard behind this, as otherwise we cannot guarantee what state variables we can count on to exist. Currently, the the documented standard says that `mole_frac_phase_comp`, `temperature`, `pressure` and `phase_frac` must be constructed by the state definition model. However, we probably should add a  minimum definition of flow to that as well so there is something we can count on existing.

I would suggest we add `flow_mol_phase_comp` to the list, as it is something that more or less has to exist any way (for the `get_material_flow_terms` method). Once we have this variable, we can calculate any other flow basis as needed.

Stemming from that, we should make sure we update the documentation to reflect the final decision: https://github.com/IDAES/idaes-pse/blob/main/docs/user_guide/components/property_package/general/state_definition.rst

idaes/generic_models/properties/core/generic/generic_property.py
eslickj: Forgot what I was doing once I got into it.  Should be fixed.

idaes/generic_models/properties/core/generic/generic_property.py
eslickj: Yeah.  We should do that.  There are so many flows the logic for calculating the missing scale factors will be interesting.

andrewlee94 (approved):
Looks good to me.

TimBartholomew (approved):
LGTM

PR #264, opened 2021-03-23
Replaced numpy nans with string NaN

makaylas:
## Summary/Motivation:
Need to replace numpy NaNs with a string NaN in the serialized flowsheet because ajax can't handle numpy NaNs

- Replace numpy NaNs with "NaN" in serialized flowsheet

### Legal Acknowledgement


idaes/ui/flowsheet.py
```diff
@@ -280,6 +281,7 @@ def _add_unit_model_with_ports(self, unit):
                 stream_df = stream_df.reset_index().rename(
                     columns={"index": "Variable"}
                 )
+                stream_df = stream_df.replace(np.nan, "NaN", regex=True)
```
lbianchi-lbl, line 284: What's the reason to use `regex=True`? As I understand from the docs, this will cause the first argument to be interpreted as a regex pattern, which doesn't seem to be applicable is the argument is `np.nan`

idaes/ui/flowsheet.py
```diff
@@ -292,6 +294,7 @@ def _add_unit_model_with_ports(self, unit):
                 performance_df["Value"] = performance_df["Value"].map(
```
lbianchi-lbl, line 294: Minor remark (unrelated to the changes in question): the `lambda` here looks redundant, and one could use `value()` directly:

```py
performance_df["Value"] = performance_df["Value"].map(value)
```

idaes/ui/flowsheet.py
```diff
@@ -292,6 +294,7 @@ def _add_unit_model_with_ports(self, unit):
...
                     lambda v: value(v)
                 )
+                performance_df = performance_df.replace(np.nan, "NaN", regex=True)
```
lbianchi-lbl, line 297: Since this conversion step is applied more than once (and, more importantly, for the same purpose), we could consider extracting it to an utility function, e.g. `_convert_for_ajax_compatibility(df: pd.DataFrame) -> pd.DataFrame`

lbianchi-lbl (commented):
I left a few comments and questions, but nothing looks like a showstopper.

Playing the role of the "testing advocate", I was wondering: how much effort would it take to add a test to check for this or similar issues with AJAX serialization? From my superficial understanding, the data in `stream_df`/`performance_df` might vary quite a bit, and it would be good to be able to catch future serialization issues early.

dangunter:
@makaylas when you say "ajax can't handle numpy NaNs" what exactly is the error? It would be nice to unambiguously preserve the NaN  in case we need to do calculations later, so we should be as precise as possible. Did you research this on the web and see how other people are handling it?

idaes/ui/flowsheet.py
makaylas: Good call. Done.

idaes/ui/flowsheet.py
makaylas: Removed.

idaes/ui/flowsheet.py
makaylas: I think that was just a copy from stackoverflow. I removed it.

makaylas:
@lbianchi-lbl I modified an existing test to add a nan into the `stream_df`. I don't know how to change the `performance_df`

makaylas:
@dangunter I'm going to put the highlights of our conversation here.

ajax was trying to parse the input as JSON (because we specified `dataType:json`) and NaNs are not a valid part of JSON. If I don't specify the `dataType` I get the same error because the `Content-Type` is `application/json`. When I changed `dataType:txt` jointjs couldn't read it because it was a string and not JSON.

For now this fix is okay but I made a ticket #280 to remind us to look at this problem again and possibly use a custom JSON encoder to fix it.

dangunter:
@makaylas right, I think it *should* be valid JSON so the error makes sense. The thing to revisit is whether to take care of this once and for all with a json.JSONEncoder subclass that isolates the logic from the other places that populate the Python dict we're going to serialize.

idaes/ui/flowsheet.py
makaylas: Okay, the lambda is necessary because I'm applying a function to all of the values in the pandas Series. When I remove it the tests fail.

lbianchi-lbl: Sorry for the confusion. What I meant with "removing the lambda" was to use directly the `value` function (i.e. the function object itself), rather than calling it, as an argument to `Series.map()` (which is how it was used in the commit where the lambda was removed, from what I can see from the errors log). In other words, `my_func` and `my_other_func` below are equivalent, which is what I was referring to with "the lambda looks redundant":

```py
my_func = lambda x: value(x)
my_other_func = value
```

makaylas: Oh okay, that makes more sense. I'm sorry I didn't understand. I removed the lambda and just used `value`

makaylas:
@lbianchi-lbl @dangunter I've modified a test to include a NaN. Can you re-review?

idaes/ui/flowsheet.py
```diff
@@ -259,6 +260,9 @@ def _map_edges(self):
         # Note we're only using the keys from self.ports, {port: parentcomponent}
         return set(self.ports) - used_ports
 
+    def _convert_for_ajax_compatibility(self, df):
```
dangunter, line 263: I would name this more like "_make_valid_json" since AJAX is an implementation detail. Also a docstring saying what it's doing?

idaes/ui/flowsheet.py
```diff
@@ -259,6 +260,9 @@ def _map_edges(self):
...
+        return df.replace(np.nan, "NaN")
```
dangunter, line 264: Why not also to Inf and -Inf while we're at it. i.e. do the same with `np.inf` and `-np.inf`

idaes/ui/tests/test_flowsheet.py
```diff
@@ -142,7 +143,7 @@ def flash_flowsheet():
     m.fs.flash = Flash(default={"property_package": m.fs.properties})
     m.fs.flash.inlet.flow_mol.fix(1)
     m.fs.flash.inlet.temperature.fix(368)
-    m.fs.flash.inlet.pressure.fix(101325)
+    m.fs.flash.inlet.pressure.fix(np.nan)
     m.fs.flash.inlet.mole_frac_comp[0, "benzene"].fix(0.5)
```
dangunter, line 147: set one to np.inf and another to -np.inf to check all three

dangunter (changes requested):
Minor changes requested

idaes/ui/tests/test_flowsheet.py
andrewlee94: As a general question - what is the specific case why this is needed? I am fairly certain Pyomo does not support `numpy` floats (at least not yet), hence I wonder if there would ever be a case where a model output would be involve a numpy object.

makaylas: I'm not sure the exact situation but during our user testing one of the models that a user used ended up with a NaN in the stream contents.

If you want the specific model, I can ask the user if they would be willing to share the model via email.

idaes/ui/flowsheet.py
makaylas: That makes more sense.

Changed the name and added a docstring

idaes/ui/flowsheet.py
makaylas: Done

idaes/ui/tests/test_flowsheet.py
andrewlee94: We might want to look into that, as there should be no way to get an NaN in a valid model anyway (solvers don't like those). This is sounding more like an issue that should be handled at the model level than in the UI.

makaylas: To capture what Andrew and I discussed:
The NaNs are a result of pandas filling in empty values with NaNs.
Andrew's PR #298 is going to take care of the empty values and replace them with "-" 
This PR still stands because we should future proof this in case we do support NaNs, Inf, and -Inf in the future

dangunter (approved):
LGTM

lbianchi-lbl (approved):
Looks good! :+1:

PR #265, opened 2021-03-23
Adding support for electrolytes with FPhx states

andrewlee94:
## Summary/Motivation:
This PR builds of some previous work to extend support for FPhx and FcTP states in electrolytes systems. Also fixed a bug in some flow expressions and added some additional features for switching mass/mol basis.

-
-

### Legal Acknowledgement


idaes/generic_models/properties/core/generic/generic_property.py
```diff
@@ -832,14 +836,16 @@ def define_metadata(cls, obj):
         # TODO : Need to fix to have methods for things that may or may not be
         # created by state var methods
         obj.add_properties(
-            {'flow_mol': {'method': None},
+            {'flow_mol': {'method': 'flow_mol'},
```
eslickj, line 839: Not totally sure, but doesn't the flow\_mol method here create a component with the same name, so would replace itself?

eslickj (changes requested):
Noted one little thing, otherwise looks good as far as I can tell.

TimBartholomew (approved):
LGTM. My only comment is since the tests for all the bases are similar, it may make sense and would be easier to maintain to have a test harness rather than duplicating the tests in each test file.

andrewlee94:
@eslickj Sorry, missed the underscore. I just pushed the fix. @TimBartholomew could you re-review?

eslickj (approved):

TimBartholomew (approved):

PR #266, opened 2021-03-24
Update the binaries version to pick up a recent k_aug change.

eslickj:
There was a minor bug fixed in k_aug recently.  This updates the default binary release to pick up that fix.  Otherwise there shouldn't be any changes from the 2.4.0 binaries. 

### Legal Acknowledgement


andrewlee94 (approved):
I don't see anything wrong here.

JanghoPark-LBL (approved):
That looks good to me.

PR #268, opened 2021-03-24
Adding support for more state definitions with electrolytes

andrewlee94:
## Summary/Motivation:
This should be the final PR in this batch of changes, adding support for the last of the current state definitions in electrolyte systems.

As @TimBartholomew noted, the test for all of these look very similar, but there are enough differences to make it more effort than I want to go to to come up with a standard form for them.

- FpcTP and FcPh forms for electrolyte systems
-

### Legal Acknowledgement

### DIFF
--- a/idaes/generic_models/properties/core/state_definitions/FpcTP.py
+++ b/idaes/generic_models/properties/core/state_definitions/FpcTP.py
@@ -22,6 +22,8 @@ from idaes.core import (MaterialFlowBasis,
                         EnergyBalanceType)
 from idaes.generic_models.properties.core.generic.utility import \
     get_bounds_from_config
+from .electrolyte_states import \
+    define_electrolyte_state, calculate_electrolyte_scaling
 from idaes.core.util.exceptions import ConfigurationError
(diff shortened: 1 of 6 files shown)


eslickj (approved):

idaes/generic_models/properties/core/state_definitions/FpcTP.py
```diff
@@ -22,6 +22,8 @@
                         EnergyBalanceType)
 from idaes.generic_models.properties.core.generic.utility import \
     get_bounds_from_config
+from .electrolyte_states import \
```
TimBartholomew, line 25: Why did you choose not to include the whole path? 

I don't think it really matters, but its also not consistent with the other imports.

TimBartholomew (approved):
LGTM. Only a minor note on an import.

idaes/generic_models/properties/core/state_definitions/FpcTP.py
andrewlee94: @TimBartholomew The only reason is to keep the import line short.

PR #269, opened 2021-03-24
Add performace curve structure to isentropic pressure changers.

eslickj:
Adding some support for performance curves in pressure changers.  Need these for gas turbines among other things.  This seems to work, but we can have some discussion around the design of it.  The main reason I used a sub-block for performance curves is that I didn't put any restriction on the form of the performance curves, so it gives me an easy way to deactivate and activate them.  The bad thing about it is that the variables and expressions that go into the performance cruves are defined in the parent block, so writing the performance curves is a little convoluted.  I can add some docs if everyone agrees to the implementation.  There are two ways to add curves, as seen in the tests.  Either a callback or just adding constraints after constructing the pressure changer.  Either way ends up the same. 

The main reason for adding this and not just slapping a few extra constraints into the existing turbine model is that I want to integrate with the pressure changer initialization, so people don't need to create some custom initialization if they want to use performance curves. 

### Legal Acknowledgement

### DIFF
--- a/idaes/generic_models/unit_models/pressure_changer.py
+++ b/idaes/generic_models/unit_models/pressure_changer.py
@@ -54,16 +54,16 @@ class ThermodynamicAssumption(Enum):
 
 @declare_process_block_class("IsentropicPerformanceCurve")
 class IsentropicPerformanceCurveData(ProcessBlockData):
-    """Block that holds perfomance curves. Typically these are in the form of
+    """Block that holds performance curves. Typically these are in the form of
     constraints that relate head, efficiency, or pressure ratio to volumetric
     or mass flow.  Additional varaibles can be included if needed, such as
     speed. For convenience an option is provided to add head expressions to the
-    block. Perfomance curves, and any additional variables, constraints, or
+    block. performance curves, and any additional variables, constraints, or
     expressions can be added to this block either via callback provided to the
     configuration, or after the model is constructued."""
 
     CONFIG = ProcessBlockData.CONFIG(
-        doc="Configuration dictionary for the performace curve block.")
+        doc="Configuration dictionary for the performance curve block.")
     CONFIG.declare("build_callback", ConfigValue(
         default=None,
         doc="Optional callback to add performance curve constraints"))
@@ -71,7 +71,7 @@ class IsentropicPerformanceCurveData(ProcessBlockData):
         default=True,
         domain=bool,
         doc="If true add expressions for 'head' and 'head_isentropic'."
-            " These expressions can be used in perfomance curve constraints."))
+            " These expressions can be used in performance curve constraints."))
 
     def has_constraints(self):
         for o in self.component_data_objects(Constraint):
@@ -238,16 +238,16 @@ see property package for documentation.}""",
         ),
     )
     CONFIG.declare(
-        "support_isentropic_perfomance_curves",
+        "support_isentropic_performance_curves",
         ConfigValue(
             default=False,
             domain=In([True, False]),
-            doc="Include a block for perfomance curves, configure via"
-                " isentropic_perfomance_curves.",
+            doc="Include a block for performance curves, configure via"
+                " isentropic_performance_curves.",
         ),
     )
     CONFIG.declare(
-        "isentropic_perfomance_curves",
+        "isentropic_performance_curves",
         IsentropicPerformanceCurveData.CONFIG(),
         # doc included in IsentropicPerformanceCurveData
     )
@@ -534,9 +534,9 @@ see property package for documentation.}""",
                     b.work_isentropic[t] * b.efficiency_isentropic[t]
                 )
 
-        if self.config.support_isentropic_perfomance_curves:
+        if self.config.support_isentropic_performance_curves:
             self.performance_curve = IsentropicPerformanceCurve(
-                default=self.config.isentropic_perfomance_curves)
+                default=self.config.isentropic_performance_curves)
 
     def model_check(blk):
         """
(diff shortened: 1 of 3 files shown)


idaes/generic_models/unit_models/pressure_changer.py
```diff
@@ -455,6 +479,38 @@ def actual_work(b, t):
                     b.work_isentropic[t] * b.efficiency_isentropic[t]
                 )
 
+        # Add head expressions (often calcuated from performance curve)
+        try:
+            @self.Expression(self.flowsheet().config.time)
+            def head_isentropic(b, t): # units are energy/mass e.g. J/kg
```
andrewlee94, line 485: We should not be presuming set units here, as the PressureChanger gets it's units from the property package. We probably need to make use of unit conversions for this.

idaes/generic_models/unit_models/pressure_changer.py
```diff
@@ -455,6 +479,38 @@ def actual_work(b, t):
                     b.work_isentropic[t] * b.efficiency_isentropic[t]
                 )
 
+        # Add head expressions (often calcuated from performance curve)
```
andrewlee94, line 482: This comments doesn't seem to match what follows - it mentions performance curves, but they don't appear in the following `try` statement.

idaes/generic_models/unit_models/pressure_changer.py
```diff
@@ -455,6 +479,38 @@ def actual_work(b, t):
...
+                        b.control_volume.properties_in[t].flow_mol /
+                        b.control_volume.properties_in[t].mw)
+
+        except PropertyNotSupportedError:
+            _log.warning(
+                "flow_mass or flow_mol and mw not supported by property package"
+                " but are required for isentropic pressure changer head"
+                " calculation")
+
+        # Add performace curve block.  If this block is empty that's fine
```
andrewlee94, line 510: Should we put this in an `if` statement which checks to see if the user provides a performance curve object? If they don't we probably don't need to build this block.

andrewlee94 (commented):
A few comments to start with.

idaes/generic_models/unit_models/pressure_changer.py
eslickj: I'm not assuming anything J/kg are just and example in the comment not necessarily the expected unit.  Whatever the property package uses is what it is.  I'll fix the comment to be clear.

TimBartholomew (commented):
LGTM. I'll approve after Andrew's comments are addressed.

I appreciate the descriptive comment for the performance curve initialization, it would be nice if we had those for other initialization routines :)

idaes/generic_models/unit_models/pressure_changer.py
eslickj: It's set up so you can just add constraints to the performance_curve block later without the callback.  I thought the option may seem more straight forward to a lot of people, even-though it's not much different.

andrewlee94: Also, just noted a typo: "performace"

andrewlee94:
@eslickj I am still a bit unsure about this implementation, as it adds things to the basic isentropic pressure changer that are not needed in the simple cases - the main example being the properties needed for calculating head. Incidentally, this is what caused the test failures as the isentropic compressor was being used as a test for the properties interrogator tool, and the new properties showed up there which caused the tests to fail.

I think we need some way to only build the performance curve parts if the user requires them so that the simple case still works without needing the additioanl properties and without adding an unnecessary Block to the model.

idaes/generic_models/unit_models/pressure_changer.py
```diff
@@ -479,34 +524,6 @@ def actual_work(b, t):
...
-
-        except PropertyNotSupportedError:
-            _log.warning(
-                "flow_mass or flow_mol and mw not supported by property package"
-                " but are required for isentropic pressure changer head"
-                " calculation")
-
         # Add performace curve block.  If this block is empty that's fine
         self.performance_curve = IsentropicPerformanceCurve(
             default=self.config.isentropic_perfomance_curves)
```
andrewlee94: Typo: perfomance

This appears in a few places, so you might want to do a search and replace.

andrewlee94 (changes requested):
Aside from a typo, I think this looks OK.

I am still not completely sold on always building the performance Block (mainly due to the memory overhead), but that isn't critical.

eslickj:
@andrewlee94, I added an option so the performance curve block only gets added if you want it.  I guess I could simplify this a bit, but I'm still stuck on the idea of creating an empty block and letting the user just add curves to it later.  Otherwise I could probably just have a callback option, which if provided, creates and populates the performance curve block.

Maybe this would be simpler by callback only. Then instead of having the head expression option, I could just handle it by calling like ```build_head_expressions()``` in the callback.  Well now I'm not sure, but I guess it would be simpler to rework this and just require the callback.

andrewlee94:
@eslickj My thought was maybe adding a configuration argument to the pressure changer model itself on whether to build the block or not. Creating an empty block so the user can fill it in later is fine, but I wonder if it would be better to let the user decide whether they want the block or not.

eslickj:
@andrewlee94, I think I did that.  I must have snuck it past you in a recent commit.

andrewlee94:
@eslickj Yes, you did sneak that change past me as I was reviewing the code - the new implementation looks good. All that is left is to document it.

eslickj:
Alright docs are in.  Should be ready to go.

andrewlee94 (approved):
Looks OK to me.

andrewlee94 (dismissed):
LGTM

eslickj:
@andrewlee94 and @TimBartholomew, could you approve again.  The example code in the docs was failing its test just because it produced output.  It's fixed now.

andrewlee94 (approved):
Looks good.

TimBartholomew (approved):

PR #275, opened 2021-03-30
Correcting units in log power law forms

### LINKED ISSUE #273: Unit consistency issue in safe log function
The `safe_log` function has a unit consistency issue regarding the `eps` parameter. As this is used in the smooth max operation with the main function argument, it needs to have the same units as the main argument. However, it is also passed as the `eps` argument for the smoothing function, which currently does not accept Pyomo expressions.

andrewlee94:
## Summary/Motivation:
Log power law forms in the generic reaction framework had unit consistency issues due to the use of unitted expressions/variables within log functions. This is strictly correct, as log functions should be unitless, but log power law forms tend to gloss over this. To correct this, the log power law forms now strict units from the expressions if necessary.

- Updated equilibrium log power law form to strip units for arguments to log functions.
-

### Legal Acknowledgement


eslickj (approved):

TimBartholomew (approved):
LGTM

Andresj89 (approved):
I tested it and it works well.

PR #276, opened 2021-03-30
Handling mismatched true and apparent species in generic reaction framework

andrewlee94:
## Summary/Motivation:
In testing @Andresj89 identified a case where the generic reaction properties failed to work as expected for an electrolyte system. This was due to the Physical Property package being set to use apparent species as the basis for state variables, but the Reaction Package was written in terms of true species. This is difficult to handle in a generic sense, as the unit model will expect reactions to be defined in terms of the component list used in the Physical Property package, so for now we will not support this use case.

- Added an Exception to catch this case and return a useful error message to the user.
-

### Legal Acknowledgement


eslickj (approved):

TimBartholomew (approved):
LGTM

Issue #287, opened 2021-04-13
Issues with calculate_scaling_factors and build-on-demand properties


andrewlee94:
In some cases, calculation of scaling factors for property models may depend upon the scaling factors of other build-on-demand properties which may or may not exist. In the current implementation, one of two things will occur:

1. `calculate_scaling_factors` will trigger `__getattr__` and construct the missing property which will add an otherwise unexpected property to the model.
2. If the property construction context manager is used, for some reason `get_scaling_factor` fails with an `AttributeError` from the context manager, even though it is wrapped in a try except. The expected behavior would be that `get_scaling_factor` would return the default value.

We should think carefully about what the correct behaviour should be in these cases, and make sure we test the interaction between these.

andrewlee94:
My initial thoughts are that `calculate_scaling_factors` for property models should always use the context manager to lock attribute creation, as this is the safest approach.

eslickj:
Ok, I think I can fix this.  Off-hand I think the issue is that the context manager is outside the get scaling factor.  I guess we have some options here.  I'll look harder into what the nicest solution is.  I guess one possibility is for the get_scaling_factor method to use the lock attribute context if available.  That's probably the cleanest.

eslickj:
@andrewlee94 can you checkout PR #289?  It's a fairly simple change.  Hopefully it fixes the issue.

andrewlee94:
Closing for now after #289. We can reopen if issues persist.

PR #289, opened 2021-04-14
Scaling Factors and State Blocks

eslickj:
## Summary/Motivation:
Potential solution for issue #287 

### Legal Acknowledgement


andrewlee94:
After testing, this doesn't actually resolve the issue as a component is required as the first argument to `get_scaling_factor`. Hence, the use of the context manager is unnecessary and the solution to the problem needs to occur in the property package itself.

PR #291, opened 2021-04-15
Fixing bug with creating phase-component set in Generic Properties

andrewlee94:
## Summary/Motivation:
A bug was identified in the generic property framework that caused species which should not be present in a given phase to be inadvertently added to the phase-component set. This bug was only present when using the electrolyte extension.

- Fixed incorrect indentation so that species in non-aqueous phases do not get added to `true_phase_component_set`.

### Legal Acknowledgement


eslickj (approved):

Andresj89 (approved):
LGTM

PR #298, opened 2021-04-20
Filling missing stream table entries with placeholder

andrewlee94:
## Summary/Motivation:
The issue addressed by PR #264 arises due to `pandas` filling missing entries in dataframes with `NaN`s. This PR adds to the code that generates the stream table dataframes to fill missing entries with a placeholder (`"-"`) to avoid this.

- Adds extra step in stream table generator to backfill missing keys
- Updates the comparison json used for testing the UI.

### Legal Acknowledgement


eslickj (dismissed):
Looks good.  Why'd the boiler json file change?

idaes/core/util/tables.py
```diff
@@ -234,8 +235,19 @@ def create_stream_table_dataframe(
             for i in disp_dict[k]:
                 if i is None:
                     stream_attributes[key][k] = value(disp_dict[k][i])
+                    if k not in full_keys:
+                        full_keys.append(k)
                 else:
                     stream_attributes[key][k + " " + str(i)] = value(disp_dict[k][i])
+                    if k + " " + str(i) not in full_keys:
+                        full_keys.append(k + " " + str(i))
```
eslickj, line 243: Doesn't really matter but you could also do ``f"{k} {i}"`` looks slightly neater and may be slightly faster.

andrewlee94:
@eslickj The json file changed a it contains the output from a stream table (I believe).

eslickj:
@andrewlee94, thanks, I just realized it's not the kind of json file I thought it was.

idaes/ui/tests/serialized_boiler_flowsheet.json
```diff
@@ -1,2606 +1 @@
...
...ext": {"text": "s_side_2_inlet_1"}}}], "z": 2}, {"type": "standard.Link", "source": {"id": "side_1_inlet_2", "port": "out"}, "target": {"id": "RH", "port": "in"}, "router": {"name": "orthogonal", "padding": 10}, "connector": {"name": "normal", "attrs": {"line": {"stroke": "#5c9adb"}}}, "id": "s_side_1_inlet_2", "labels": [{"attrs": {"rect": {"fill": "#d7dce0", "stroke": "white", "stroke-width": 0, "fill-opacity": 0}, "text": {"text": "feed info", "fill": "black", "text-anchor": "left", "display": "none"}}, "position": {"distance": 0.66, "offset": -40}}, {"attrs": {"text": {"text": "s_side_1_inlet_2"}}}], "z": 2}, {"type": "standard.Link", "source": {"id": "RH", "port": "out"}, "target": {"id": "side_1_outlet_1", "port": "in"}, "router": {"name": "orthogonal", "padding": 10}, "connector": {"name": "normal", "attrs": {"line": {"stroke": "#5c9adb"}}}, "id": "s_side_1_outlet_1", "labels": [{"attrs": {"rect": {"fill": "#d7dce0", "stroke": "white", "stroke-width": 0, "fill-opacity": 0}, "text": {"text": "product info", "fill": "black", "text-anchor": "left", "display": "none"}}, "position": {"distance": 0.66, "offset": -40}}, {"attrs": {"text": {"text": "s_side_1_outlet_1"}}}], "z": 2}]}
```
makaylas, line 1: Could you format this like the old json was formatted? Either copy and paste this into a json formatter on the web or do `json.dumps(parsed, indent=4, sort_keys=True)`

makaylas (commented):
Looks good. Just one comment

idaes/ui/tests/serialized_boiler_flowsheet.json
andrewlee94: I was wondering how to get the formatting done.

I also need some help with sorting out the issue with Windows tolerance. I have not managed to limit the signficiant figures in the dataframe/json dump yet.

idaes/core/util/tables.py
```diff
@@ -234,8 +235,19 @@ def create_stream_table_dataframe(
...
+
+    # Check for missing rows in any stream, and fill with "-" if needed
+    for k, v in stream_attributes.items():
+        for r in full_keys:
+            if r not in v.keys():
+                # Missing row, fill with placeholder
+                v[r] = "-"
 
     return DataFrame.from_dict(stream_attributes, orient=orient)
```
makaylas, line 252: Try `return DataFrame.from_dict(stream_attributes, orient=orient).round(6)` that should round the whole dataframe to 6 sig figs

andrewlee94: I tried something similar, but didn't see any difference in the json. I will try again when I have some time.

makaylas: Okay this is very weird. It's the stream table with the rounding issue but the dataframe we use to create the stream table has the `round` function applied.

Let me look into this more.

makaylas: Ah okay, pandas will only perform the `round` on columns that are numerical types. Since you replaced the NaNs with `-` it changed the column type from `float64` to `object`.

On line 395 (`.round(self._sig_figs)`) and line 422 (`c = self._serialized_contents[unit_name][content_type].round(self._sig_figs).to_dict("index")`) replace the `.round(self._sig_figs)` with
```
.applymap(lambda x: round(x, self._sig_figs) if isinstance(x, (int, float)) else x)
```

That will apply a round to the numbers and leave the strings alone

idaes/core/util/tables.py
```diff
@@ -224,6 +224,7 @@ def create_stream_table_dataframe(
     """
     stream_attributes = OrderedDict()
     stream_states = stream_states_dict(streams=streams, time_point=time_point)
+    full_keys = []  # List of all rows in dataframe to fill in missing data
```
dangunter, line 227: It doesn't seem you are using this in a way that order matters. I would suggest instead making it a set(). This will change lookups ('if k not in full_keys') from linear time to constant time.

dangunter (approved):
One minor comment, but not enough to stop merge.

idaes/core/util/tables.py
andrewlee94: I am not sure why, but this results in the output being non-deterministic. When  tried doing this, I started seeing the output from the serializer test failing every run because the serialized stream table kept changing.

idaes/core/util/tables.py
andrewlee94: I have implemented the suggestions above, and we will see if the tests pass now.

andrewlee94:
@dangunter @makaylas @eslickj The tests are passing now, so this is ready for re-review.

eslickj (approved):

PR #299, opened 2021-04-20
Adding support for constant properties

Andresj89:
## Summary/Motivation:
Currently, if we want to use properties independent of temperature in the generic framework, we use Perry's method to fix the first coefficient to the value of the property and fix the remaining coefficients to zero. This PR eliminates that workaround by adding support for using constant properties.

- A new module `ConstantProperties` has been created with a class `Constant`
- Currently, the class supports constant Cp `cp_mol_liq_comp_coeff`, constant enthalpy `enth_mol_form_liq_comp_coeff`, constant entropy `entr_mol_form_liq_comp_coeff`, and constant density `dens_mol_liq_comp_coeff`.

### Legal Acknowledgement

### DIFF
--- a/idaes/generic_models/properties/core/pure/ConstantProperties.py
+++ b/idaes/generic_models/properties/core/pure/ConstantProperties.py
@@ -1,0 +1,133 @@
+##############################################################################
+# Institute for the Design of Advanced Energy Systems Process Systems
+# Engineering Framework (IDAES PSE Framework) Copyright (c) 2018-2020, by the
+# software owners: The Regents of the University of California, through
+# Lawrence Berkeley National Laboratory,  National Technology & Engineering
+# Solutions of Sandia, LLC, Carnegie Mellon University, West Virginia
+# University Research Corporation, et al. All rights reserved.
+#
+# Please see the files COPYRIGHT.txt and LICENSE.txt for full copyright and
+# license information, respectively. Both files are also available online
+# at the URL "https://github.com/IDAES/idaes-pse".
+##############################################################################
+"""
+Method to set constant pure component properties:
+
+All parameter indicies and units based on conventions used by the source
+"""
+from pyomo.environ import log, Var, units as pyunits
+
+from idaes.core.util.misc import set_param_from_config
+
+
+# -----------------------------------------------------------------------------
+# Heat capacities, enthalpies and entropies
+class Constant(object):
+    
+    class cp_mol_liq_comp(object):
+
+        @staticmethod
+        def build_parameters(cobj):
+            cobj.cp_mol_liq_comp_coeff = Var(
+                doc="Parameter for liquid phase molar heat capacity",
+                units=pyunits.J*pyunits.kmol**-1*pyunits.K**-1)
+            set_param_from_config(cobj, param="cp_mol_liq_comp_coeff")
+
+        @staticmethod
+        def return_expression(b, cobj, T):
+            # Specific heat capacity
+            cp = (cobj.cp_mol_liq_comp_coeff)
+            units = b.params.get_metadata().derived_units
+            return pyunits.convert(cp, units["heat_capacity_mole"])
+
+
+    class enth_mol_liq_comp(object):
+
+        @staticmethod
+        def build_parameters(cobj):
+            # if not hasattr(cobj, "cp_mol_liq_comp_coeff"):
+            #     cp_mol_liq_comp.build_parameters(cobj)
+
+            if cobj.parent_block().config.include_enthalpy_of_formation:
+                units = cobj.parent_block().get_metadata().derived_units
+
+                cobj.enth_mol_form_liq_comp_ref = Var(
+                        doc="Liquid phase molar heat of formation @ Tref",
+                        units=units["energy_mole"])
+                set_param_from_config(cobj, param="enth_mol_form_liq_comp_ref")
+
+            cobj.enth_mol_form_liq_comp_coeff = Var(
+                doc="Liquid phase molar heat of formation",
+                units=pyunits.kJ/pyunits.mol)
+            set_param_from_config(cobj, param="enth_mol_form_liq_comp_coeff")
+
+        @staticmethod
+        def return_expression(b, cobj, T):
+            # Specific enthalpy
+            T = pyunits.convert(T, to_units=pyunits.K)
+            Tr = pyunits.convert(b.params.temperature_ref, to_units=pyunits.K)
+
+            units = b.params.get_metadata().derived_units
+
+            h_form = (cobj.enth_mol_form_liq_comp_ref if
+                    b.params.config.include_enthalpy_of_formation
+                    else 0*units["energy_mole"])
+
+            h = (pyunits.convert(
+                    (cobj.enth_mol_form_liq_comp_coeff), units["energy_mole"]) +
+                h_form)
+
+            return h
+
+
+    class entr_mol_liq_comp(object):
+
+        @staticmethod
+        def build_parameters(cobj):
+            # if not hasattr(cobj, "cp_mol_liq_comp_coeff_1"):
+            #     cp_mol_liq_comp.build_parameters(cobj)
+
+            units = cobj.parent_block().get_metadata().derived_units
+
+            cobj.entr_mol_form_liq_comp_ref = Var(
+                    doc="Liquid phase molar entropy of formation @ Tref",
+                    units=units["entropy_mole"])
+            set_param_from_config(cobj, param="entr_mol_form_liq_comp_ref")
+           
+            cobj.entr_mol_form_liq_comp_coeff = Var(
+                    doc="Liquid phase molar entropy of formation",
+                    units=units["entropy_mole"])
+            set_param_from_config(cobj, param="entr_mol_form_liq_comp_coeff")
+
+        @staticmethod
+        def return_expression(b, cobj, T):
+            # Specific entropy
+            T = pyunits.convert(T, to_units=pyunits.K)
+            Tr = pyunits.convert(b.params.temperature_ref, to_units=pyunits.K)
+
+            units = b.params.get_metadata().derived_units
+
+            s = (pyunits.convert(
+                    (cobj.entr_mol_form_liq_comp_coeff),
+                    units["entropy_mole"]) +
+                cobj.entr_mol_form_liq_comp_ref)
+
+            return s
+
+    # -----------------------------------------------------------------------------
+    # Densities
+    class dens_mol_liq_comp(object):
+
+        @staticmethod
+        def build_parameters(cobj):
+            cobj.dens_mol_liq_comp_coeff = Var(
+                    doc="Parameter for liquid phase molar density",
+                    units=pyunits.kmol*pyunits.m**-3)
+            set_param_from_config(cobj, param="dens_mol_liq_comp_coeff")
+
+        @staticmethod
+        def return_expression(b, cobj, T):
+
+            rho = (cobj.dens_mol_liq_comp_coeff)
+            units = b.params.get_metadata().derived_units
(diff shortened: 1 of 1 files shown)


idaes/generic_models/properties/core/pure/ConstantProperties.py
```diff
@@ -0,0 +1,133 @@
...
+# University Research Corporation, et al. All rights reserved.
+#
+# Please see the files COPYRIGHT.txt and LICENSE.txt for full copyright and
+# license information, respectively. Both files are also available online
+# at the URL "https://github.com/IDAES/idaes-pse".
+##############################################################################
+"""
+Method to set constant pure component properties:
+
+All parameter indicies and units based on conventions used by the source
```
andrewlee94, line 16: This comment should be removed.

idaes/generic_models/properties/core/pure/ConstantProperties.py
```diff
@@ -0,0 +1,133 @@
...
+            cp = (cobj.cp_mol_liq_comp_coeff)
+            units = b.params.get_metadata().derived_units
+            return pyunits.convert(cp, units["heat_capacity_mole"])
+
+
+    class enth_mol_liq_comp(object):
+
+        @staticmethod
+        def build_parameters(cobj):
+            # if not hasattr(cobj, "cp_mol_liq_comp_coeff"):
```
Andresj89, line 48: @andrewlee94 I'm not sure if this conditional is needed in this case. Same in lines 87-88.

idaes/generic_models/properties/core/pure/ConstantProperties.py
```diff
@@ -0,0 +1,133 @@
...
+# Heat capacities, enthalpies and entropies
+class Constant(object):
+    
+    class cp_mol_liq_comp(object):
+
+        @staticmethod
+        def build_parameters(cobj):
+            cobj.cp_mol_liq_comp_coeff = Var(
+                doc="Parameter for liquid phase molar heat capacity",
+                units=pyunits.J*pyunits.kmol**-1*pyunits.K**-1)
```
andrewlee94, line 33: Should we define the parameter in terms of the base units, rather than setting the to SI? Other modules have defined units from the source text, but this one is a lot more general.

idaes/generic_models/properties/core/pure/ConstantProperties.py
```diff
@@ -0,0 +1,133 @@
...
+            set_param_from_config(cobj, param="cp_mol_liq_comp_coeff")
+
+        @staticmethod
+        def return_expression(b, cobj, T):
+            # Specific heat capacity
+            cp = (cobj.cp_mol_liq_comp_coeff)
```
andrewlee94, line 39: You don't really need the parentheses here.

idaes/generic_models/properties/core/pure/ConstantProperties.py
```diff
@@ -0,0 +1,133 @@
...
+            units = b.params.get_metadata().derived_units
+            return pyunits.convert(cp, units["heat_capacity_mole"])
```
andrewlee94, line 41: If we were to define the parameter in terms of the base units, we would not need the `convert` here. This would end up being more efficient.

This also applies to the other properties.

idaes/generic_models/properties/core/pure/ConstantProperties.py
andrewlee94, line 48: Remove commented code.

idaes/generic_models/properties/core/pure/ConstantProperties.py
```diff
@@ -0,0 +1,133 @@
...
+
+            if cobj.parent_block().config.include_enthalpy_of_formation:
+                units = cobj.parent_block().get_metadata().derived_units
+
+                cobj.enth_mol_form_liq_comp_ref = Var(
+                        doc="Liquid phase molar heat of formation @ Tref",
+                        units=units["energy_mole"])
+                set_param_from_config(cobj, param="enth_mol_form_liq_comp_ref")
+
+            cobj.enth_mol_form_liq_comp_coeff = Var(
```
andrewlee94, line 59: It does not really make sense to have a constant enthalpy. Here I would recommend `h = cp*(T-Tref) + h_ref` as the better approach.

idaes/generic_models/properties/core/pure/ConstantProperties.py
```diff
@@ -0,0 +1,133 @@
...
+                doc="Liquid phase molar heat of formation",
+                units=pyunits.kJ/pyunits.mol)
+            set_param_from_config(cobj, param="enth_mol_form_liq_comp_coeff")
+
+        @staticmethod
+        def return_expression(b, cobj, T):
+            # Specific enthalpy
+            T = pyunits.convert(T, to_units=pyunits.K)
```
andrewlee94, line 67: See above, but right now you don't even use these.

idaes/generic_models/properties/core/pure/ConstantProperties.py
```diff
@@ -0,0 +1,133 @@
...
+                h_form)
+
+            return h
+
+
+    class entr_mol_liq_comp(object):
+
+        @staticmethod
+        def build_parameters(cobj):
+            # if not hasattr(cobj, "cp_mol_liq_comp_coeff_1"):
```
andrewlee94, line 87: Commented code.

idaes/generic_models/properties/core/pure/ConstantProperties.py
```diff
@@ -0,0 +1,133 @@
...
+
+        @staticmethod
+        def return_expression(b, cobj, T):
+            # Specific entropy
+            T = pyunits.convert(T, to_units=pyunits.K)
+            Tr = pyunits.convert(b.params.temperature_ref, to_units=pyunits.K)
+
+            units = b.params.get_metadata().derived_units
+
+            s = (pyunits.convert(
```
andrewlee94, line 110: As above, a constant entropy doesn't make a lot of sense. How about `s = cp*ln(T/Tref) + s_ref`

andrewlee94 (changes requested):
A good start, but I don't think constant h and s make sense - we should instead derive them from constant cp (see notes).

Also, this needs tests and documentation.

idaes/generic_models/properties/core/pure/ConstantProperties.py
andrewlee94: It will be needed if you follow my suggestions.

idaes/generic_models/properties/core/pure/ConstantProperties.py
```diff
@@ -0,0 +1,118 @@
...
+                cobj.enth_mol_form_liq_comp_ref = Var(
+                        doc="Liquid phase molar heat of formation @ Tref",
+                        units=units["energy_mole"])
+                set_param_from_config(cobj, param="enth_mol_form_liq_comp_ref")
+
+        @staticmethod
+        def return_expression(b, cobj, T):
+            # Specific enthalpy
+            units = b.params.get_metadata().derived_units
+            T = pyunits.convert(T, to_units=units["temperature"])
```
andrewlee94, line 61: You do not need these temperature conversions here - the cp parameters should already have consistent units.

idaes/generic_models/properties/core/pure/ConstantProperties.py
```diff
@@ -0,0 +1,118 @@
...
+            cobj.entr_mol_form_liq_comp_ref = Var(
+                    doc="Liquid phase molar entropy of formation @ Tref",
+                    units=units["entropy_mole"])
+            set_param_from_config(cobj, param="entr_mol_form_liq_comp_ref")
+           
+        @staticmethod
+        def return_expression(b, cobj, T):
+            # Specific entropy
+            units = b.params.get_metadata().derived_units
+            T = pyunits.convert(T, to_units=units["temperature"])
```
andrewlee94, line 93: As above - also the T units would cancel out anyway.

idaes/generic_models/properties/core/pure/ConstantProperties.py
```diff
@@ -0,0 +1,118 @@
...
+            # Specific enthalpy
+            units = b.params.get_metadata().derived_units
+            T = pyunits.convert(T, to_units=units["temperature"])
+            Tr = pyunits.convert(b.params.temperature_ref, to_units=units["temperature"])
+
+            h_form = (cobj.enth_mol_form_liq_comp_ref if
+                    b.params.config.include_enthalpy_of_formation
+                    else 0*units["energy_mole"])
+
+            h = (pyunits.convert(
```
andrewlee94, line 68: Again, h should already have the correct units, so no need for a unit conversion here.

idaes/generic_models/properties/core/pure/ConstantProperties.py
```diff
@@ -0,0 +1,118 @@
...
+            set_param_from_config(cobj, param="entr_mol_form_liq_comp_ref")
+           
+        @staticmethod
+        def return_expression(b, cobj, T):
+            # Specific entropy
+            units = b.params.get_metadata().derived_units
+            T = pyunits.convert(T, to_units=units["temperature"])
+            Tr = pyunits.convert(b.params.temperature_ref, to_units=units["temperature"])
+
+            s = (pyunits.convert(
```
andrewlee94, line 96: Same here for unit conversion,

idaes/generic_models/properties/core/pure/ConstantProperties.py
```diff
@@ -0,0 +1,118 @@
...
+"""
+from pyomo.environ import log, Var, Param, units as pyunits
+
+from idaes.core.util.misc import set_param_from_config
+
+# -----------------------------------------------------------------------------
+# Heat capacities, enthalpies and entropies
+class Constant(object):
+    
+    class cp_mol_liq_comp(object):
```
andrewlee94, line 25: You should also add `cp_mol_ig_comp` and the equivalent enthalpy and entropy methods.

docs/user_guide/components/property_package/general/pure/ConstantProperties.rst
```diff
@@ -0,0 +1,58 @@
...
+
+Introduction to Chemical Engineering Thermodynamics, 8th Edition
+J.M. Smith, Hendrick Van Ness, Michael Abbott, and Mark Swihart, 2018, McGraw-Hill
+
+Ideal Liquid Molar Heat Capacity (Constant Pressure and Temperature)
+--------------------------------------------------------------------
+
+The ideal liquid molar heat capacity is defined as follows:
+
+.. math:: c_{\text{p liq}} = Constant
```
andrewlee94, line 20: You should include the actual name of the parameter here, not just `Constant`.

docs/user_guide/components/property_package/general/pure/ConstantProperties.rst
```diff
@@ -0,0 +1,58 @@
...
+
+Source
+------
+
+Methods for calculating pure component properties independent of temperature from:
+
+Introduction to Chemical Engineering Thermodynamics, 8th Edition
+J.M. Smith, Hendrick Van Ness, Michael Abbott, and Mark Swihart, 2018, McGraw-Hill
+
+Ideal Liquid Molar Heat Capacity (Constant Pressure and Temperature)
```
andrewlee94, line 15: Only constant pressure here (i.e. it is `cp` and not `cv`). The fact that it doesn't vary with T is a separate issue.

andrewlee94 (changes requested):
There are a few unit conversions that should be removed, and you should also add methods for ideal gases as well.

andrewlee94 (dismissed):
Looks OK to me.

TimBartholomew (approved):
LGTM

andrewlee94 (approved):
Looks good.

PR #301, opened 2021-04-21
Expand Natural Gas Property Parameters

eslickj:
Add some components and reactions to the natural gas properties.  Need these for NGCC model.  This should also allow for water to condense, which will be handy for water knockout in the CO2 compressor.  I've run the gas turbine in two phase mode and it works, but there is no liquid expected.  Two phase will require some more testing.

### Legal Acknowledgement

### DIFF
--- a/idaes/power_generation/properties/natural_gas_PR.py
+++ b/idaes/power_generation/properties/natural_gas_PR.py
@@ -28,16 +28,16 @@ import logging
 import copy
 
 # Import Pyomo units
+import pyomo.environ as pyo
 from pyomo.environ import units as pyunits
 
 # Import IDAES cores
-from idaes.core import VaporPhase, Component
+from idaes.core import VaporPhase, LiquidPhase, Component, PhaseType
 
 from idaes.generic_models.properties.core.state_definitions import FTPx
 from idaes.generic_models.properties.core.eos.ceos import Cubic, CubicType
 
-from idaes.generic_models.properties.core.pure import NIST
-from idaes.generic_models.properties.core.pure import RPP4
+from idaes.generic_models.properties.core.pure import NIST, RPP4, RPP5
 
 from idaes.generic_models.properties.core.reactions.dh_rxn import \
     constant_dh_rxn
@@ -46,7 +46,9 @@ from idaes.generic_models.properties.core.reactions.rate_constant import \
 from idaes.generic_models.properties.core.reactions.rate_forms import \
     power_law_rate
 from idaes.generic_models.properties.core.generic.generic_reaction import (
+    GenericReactionParameterBlock,
     ConcentrationForm)
+import idaes.generic_models.properties.core.reactions as rxn
 
 # Set up logger
 _log = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


minhwvu:
Hi @eslickj : Looks good. I only have a concern whether we can add calculations for several properties such as speed of sound, isentropic enthalpy in the NG property package?

idaes/power_generation/properties/natural_gas_PR.py
```diff
@@ -62,247 +64,366 @@
...
+                'H': -296.8422},
+            'enth_mol_form_vap_comp_ref': (
+                -296.84e3, pyunits.J/pyunits.mol), # NIST <- Chase 1998
+            'entr_mol_form_vap_comp_ref': (
+                248.21, pyunits.J/pyunits.mol/pyunits.K) # NIST <- Chase 1998
+        }
+    },
+    'C2H4': {
+        'type': Component,
+        'valid_phase_types': [PhaseType.vaporPhase],
```
minhwvu, line 379: Hi @eslickj : Should we add "valid_phase_types" declaration for other components too?

eslickj: No, I don't think so.  In this case, the only thing we expect to liquefy is water.  Only adding liquid phase to water should make the problem a lot simpler.  Most of the vapor pressure curves for these things are for much lower temperature ranges than we expect to have.  @andrewlee94 may want to weigh in since he has a much better understanding of how the property package is implemented, but my feeling at least for now is that we should try to keep it simple.  I guess we'll probably have supercritical CO2, but I think that should be fine.  We can also switch to Span-Wagner once we have almost pure CO2.

eslickj:
@minhwvu, these are just property parameters.  Adding stuff like speed of sound will require adding things to the generic property package.  We can do that seperatly.

idaes/power_generation/properties/natural_gas_PR.py
andrewlee94: The choice is entirely up to you of what components you want to assume are condensable. Fewer components will generally be easier to solve however.

eslickj:
@andrewlee94, I need to get some clarification on the sign for the entropy parameters.  I know some of them have the wrong sign, but I want to make sure I understand what's correct.  I think it's currently inconsistent.

andrewlee94:
@eslickj Yes, there is currently an inconsistency with sign conventions, and at this point I have no idea what is correct. I tried working out what the problem was once, but it was far more complicated than it should have been and affected much more than just the property packages.

eslickj:
@AlexNoring, @MAZamarripa, and @andrewlee94, I took a closer look at whats going on, and found that standard enthalpy of formation and standard entropy at 298.15 K are not used for the NIST standard enthalpy and entropy calculations at all, so it doesn't matter what the numbers are.  I removed the parameters.  That should help the confusion.  Let me know if I'm missing something.  Hopefully I didn't break anything.

For the original RPP4 parameters, all the standard entropies at 298.15 K are all positive, which is what I'd expect, but I haven't tried to test them.

andrewlee94:
@eslickj Yes, the NIST equations (Shomate equation) contain the standard enthalpy and entropy within the other parameters (G and H from memory), thus there is no need to define separate parameters for these.

eslickj:
@MAZamarripa and @AlexNoring, could you guys give this a quick review when you get a chance?  I need this one merged before I add the gas turbine model.

AlexNoring (changes requested):
The serialized model "NGFC_flowsheet_init.json" is not compatible with the changes to the property package. I will send you an updated file to replace it with.

AlexNoring (approved):
lgtm

PR #314, opened 2021-04-29
Condition number/Scaled Jacobian

eslickj:
Add some utilities in scaling to get the condition number and jacobian and scaled jacobian.  Condition number can be calculated on the scaled Jacobian.

### Legal Acknowledgement


eslickj:
Something weird happened with the diffs.  Gonna close and try again.

PR #315, opened 2021-04-29
Add utilities for scaled Jacobian and condition number

eslickj:
Add some utility functions so you can get the scaled and unscaled Jacobian and condition numbers.

### Legal Acknowledgement


idaes/core/util/scaling.py
```diff
@@ -501,6 +501,41 @@ def constraint_autoscale_large_jac(
...
+    jac, jac_scaled, nlp = constraint_autoscale_large_jac(m, no_scale=True)
+    if scaled:
+        return jac_scaled, nlp
+    else:
+        return jac, nlp
+
+
+def jacobian_cond(m, scaled=True, ord=None):
+    """
+    Get the Jacobian matrix at the current model values.
```
andrewlee94, line 524: I think this was copied from the previous method and not updated.

andrewlee94 (changes requested):
One minor comment from me.

idaes/core/util/scaling.py
```diff
@@ -521,15 +529,15 @@ def get_jacobian(m, scaled=True):
 
 def jacobian_cond(m, scaled=True, ord=None):
     """
-    Get the Jacobian matrix at the current model values.
+    Get the condion number of the scaled or unscaled Jacobian matirx of a model.
```
andrewlee94, line 532: Typos: condition, matrix

andrewlee94 (changes requested):
A couple more typos.

andrewlee94 (approved):
Looks good to me.

eslickj:
@nareshsusarla, thanks I didn't know about the warning, but it's probably fine.  The matrix inversion will probably take so much longer than anything else that taking some extra time computing the norm is probably nothing.  I'm a bit worried that this will take a really long time or fail for big and/or poorly conditioned problems, but we'll see how it goes.

nareshsusarla (approved):

PR #319, opened 2021-05-01
Use 'addheader' instead of script in repo

### LINKED ISSUE #308: Move copyright-update script to its own package
The copyright-update script has nothing really to do with this repo. To make it easier to use across IDAES related repos, move it out to its own repo and package with PyPI.

Goal is to reduce the stuff in any other repo to the COPYRIGHT.txt itself, and allow the following sequence for developers

dangunter:
## Summary/Motivation:
- Better to have one nice version of this than several cut and pasted hacky ones

- remove script
- add dependency on the addheader package
- add configuration for addheader package
-

### Legal Acknowledgement


addheader.cfg
```diff
@@ -0,0 +1,8 @@
+---
+root: idaes
+text: COPYRIGHT.md
+patterns:
+   - "*.py"
+   - "~__init__.py"
```
lbianchi-lbl, line 6: Does the leading tilde in the pattern mean something special? If so, it could be useful to add a comment here for quick reference

idaes/tests/test_headers.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    import idaes
+    return Path(idaes.__file__).parent
+
+
+@pytest.fixture
+def patterns(package_root):
+    """Grab glob patterns from config file.
+    """
+    conf_file = package_root.parent / "addheader.cfg"
```
lbianchi-lbl, line 28: As I understand the configuration file format is YAML, but the extension is `.cfg`. I don't think this is a problem (except maybe for confusing text editors/IDEs), but is there a deeper reason why the extension is not `.yml`?

ksbeattie (approved):
I have nothing beyond what @lbianchi-lbl has already asked about.

addheader.cfg
dangunter: yeah it's in the docs but who reads those so will add comment

idaes/tests/test_headers.py
dangunter: that's where the wheel stopped, not my fault. but seriously, changing to .yml

idaes/tests/test_headers.py
```diff
@@ -0,0 +1,52 @@
...
+    if not conf_file.exists():
+        print(f"Cannot load configuration file from '{conf_file}'. Perhaps this is not development mode?")
```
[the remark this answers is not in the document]
dangunter: I am changing them to print

requirements-dev.txt
```diff
@@ -23,6 +23,7 @@ flake8
 jsonschema
 jupyter_contrib_nbextensions
 snowballstemmer==1.2.1
+addheader >= 0.2.2
```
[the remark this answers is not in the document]
dangunter: I mean, it is ok but you are right it's not the standard

lbianchi-lbl (approved):
Looks good! :+1:

ksbeattie (approved):
LGTM

PR #320, opened 2021-05-03
Pinned Variable column to the left of the stream table

makaylas:
## Summary/Motivation:
To make the stream table more readable when there is a large number of streams/columns I pinned the Variable column to the left so that when the user scrolls that column stays on the left.

- Variable column is pinned to the left

### Legal Acknowledgement


lbianchi-lbl (approved):
This one-liner LGTM  :+1:

ksbeattie:
@makaylas can you point me to a an example of a flowsheet with a large number of streams to test this with?

dangunter:
@ksbeattie see this gist for an example that creates enough streams to see the behavior. Note that this is from a chemical engineering perspective pure garbage, but can be visualized and has a stream table.

https://gist.github.com/dangunter/f07f9236d2f1c9e0c546e9714199c765

idaes/ui/fsvis/static/js/stream_table.js
```diff
@@ -35,7 +35,7 @@ export class StreamTable {
             let column_header = columns[col]
             if (column_header !== "") {
                 if (column_header === "Variable") {
-                    column_defs.push({headerName: column_header, field: column_header, filter: 'agTextColumnFilter', sortable: true, resizable: true});
+                    column_defs.push({headerName: column_header, field: column_header, filter: 'agTextColumnFilter', sortable: true, resizable: true, pinned: 'left',});
```
dangunter, line 38: Why not add a comment saying why this if clause exists?

dangunter (changes requested):
Looks good, but a comment would be nice for future generations.

idaes/ui/fsvis/static/js/stream_table.js
makaylas: Added.

dangunter (approved):
Looks great!

elbashandy (approved):
Looks good

ksbeattie (approved):
LGTM

PR #326, opened 2021-05-05
Context manager for temprary config changes

eslickj:
This adds a context manager that allows you to temporarily change the IDAES config and have it revert back after the with-block.  This is mainly for testing.  Config changes for one test can affect others.  For example, you may want to set ipopt to default to user scaling, so all the initialization routines use it.  But that setting will stick around for subsequent tests potentially breaking them.  So the idea is that you can use the new context manager to make sure tests don't make lasting changes to the config.

This also makes some changes so the pointer to the IDAES ConfigDict is only in one place. 

### Legal Acknowledgement

### DIFF
--- a/idaes/tests/test_ipopt.py
+++ b/idaes/tests/test_ipopt.py
@@ -36,9 +36,9 @@ def test_ipopt_idaes_config():
     with idaes.temporary_config_ctx():
         # in this context use idaes default solver options
         isolve.use_idaes_solver_configuration_deafults()
-        idaes.cfg.ipopt.options.nlp_scaling_method = "gradient-based"
+        idaes.cfg.ipopt.options.nlp_scaling_method = "toast-based"
         solver = pyo.SolverFactory('ipopt')
-        assert solver.options["nlp_scaling_method"] == "gradient-based"
+        assert solver.options["nlp_scaling_method"] == "toast-based"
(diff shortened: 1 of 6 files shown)


andrewlee94 (approved):
Looks OK to me, although I don't fully understand some of it.

idaes/tests/test_ipopt.py
```diff
@@ -36,19 +36,24 @@ def test_ipopt_idaes_config():
...
+    #
+    # TODO(JCE): Bring back the rest of this test after the tests
+    # untangled in the GT PR
+    #
     # back to the original config, don't use idaes default option settings
-    solver = pyo.SolverFactory('ipopt')
-    assert "nlp_scaling_method" not in solver.options
-    assert solver.options["tol"] != 1
-
+    #solver = pyo.SolverFactory('ipopt')
```
andrewlee94, line 52: Should this comment code still be here?

eslickj:
@andrewlee94, could you approve again?  I didn't really change anything since you approved, I just had to comment out part of a test, until we get some of the fragile tests fixed.  As it is now some test depend on other tests re-configuring the solvers.  I sorted that out in PR #323, so I'll bring back the rest of the test there.

MAZamarripa (approved):
lgtm

PR #331, opened 2021-05-07
Adding first verification tests for eNRTL model

andrewlee94:
## Summary/Motivation:
This adds the first tests to confirm the eNRTL model is correct, using data from literature.

- Adds verification tests for an NaCl-H2O system from literature
- Multiple error corrections in main model

### Legal Acknowledgement


Andresj89 (approved):
LGTM

andrewlee94:
@adam-a-a The data is just for testing purposes - it is digitized from a fairly low-resolution plot so I think we could find better sources if we need it for something else, but it is sufficient for testing.

PR #292, opened 2021-04-15
Refactor setup.py and requirements-dev.txt to streamline installation workflows

lbianchi-lbl:
## Summary/Motivation:
- Move dependencies needed to install a prerelease version of IDAES to `setup.py` so that it can be installed without having to clone the repository
- Streamline installation process by reducing number of recommended installation workflows to two (for **users** and **developers**)

## Changes proposed in this PR:
- Update `setup.py` defining a `[prerelease]` optional tag (`extras_require`) for `pip install` containing the appropriate versions of Pyomo ~~(and, following #249, currently also PySP)~~
- Move developer dependencies to `requirements-dev.txt`
- Remove unnecessary dependencies (e.g. `mock` which is part of the Python stdlib as `unittest.mock`)
- Update documentation accordingly

### Conceptual constraints (`C`)

- `C1`: Persons who need to install IDAES are divided in **2 categories**, depending on whether they **need to access the source code** (~cloning the repo)
  - **Users**: :x: no access to source needed
  - **Developers**: :white_check_mark: need access to source
- `C2`: Both categories need to be able to install **prerelease versions** of IDAES
- `C3`: The **default install command** (`pip install idaes-pse`) should install the latest released version
- `C4`: When possible, follow _principle of **least privilege**_
- `C5`: When possible, **minimize complexity** and redundancy

### Technical constraints (`T`)

- `T1`: Dependencies need to be *accessible* to `setup.py` to be able to **install IDAES without cloning** the repo
- `T2`: Installing dependencies using **requirement files** (`pip install -r requirements.txt`) only possible from a local repo clone
- `T3`: Installing IDAES in **editable mode** (`pip install -e/--editable .`) only makes sense when installing from a local repo clone
- `T4`: Installing **optional dependencies** without cloning requires an `extras_require` install target in `setup.py`
  - e.g. `pip install 'idaes-pse[my_target]'` would install dependencies defined under the `my_target` key of `extras_require` in `setup.py`

### Derived constraints (`D`)

- `C1 & C4` :arrow_right: `D1`: Installation procedure for **users shall not require cloning** the repo (important when working with protected data)
  - :arrow_right: `D2`: Workflows requiring cloning the repo can be considered *developer workflows*

### Results

- `D1 & C2 & C3` :arrow_right: Use `extras_install` target to install prerelease dependencies
- `D2 & C5` :arrow_right: All developer workflows shall require cloning the repo
- `D2 & T2` :arrow_right: `requirements.txt` files only needed by developer workflows
[...]

PR #335, opened 2021-05-11
Update binary version to fix uncertainty tool with newer Pyomo

eslickj:
This just bumps the binary release number to get the updated version of k_aug the produces a more standard version string.  This should fix the problem with the uncertainty tool several people ran into recently.  There is also a secret solver.  The new Ipopt L1 restoration mode is available in an alternate Ipopt version called ipopt_l1 on all platforms except Cent OS 6 and 7.  It doesn't yet compile with the older compiler versions, but I figured this would at least let us do a little evaluation before officially releasing and announcing it. 

### Legal Acknowledgement

### DIFF
--- a/idaes/config.py
+++ b/idaes/config.py
@@ -19,7 +19,7 @@ import importlib
 
 _log = logging.getLogger(__name__)
 # Default release version if no options provided for get-extensions
-default_binary_release = "2.4.2"
+default_binary_release = "2.4.3"
 # Where to download releases from get-extensions
 release_base_url = "https://github.com/IDAES/idaes-ext/releases/download"
 # Where to get release checksums


eslickj:
Closing, as this doesn't seem to work.  I checked the k_aug repo again, and don't see any relevant changes for the version info recently.  I check a little bit ago when @Robbybp suggested I need a newer version of k_aug and it seemed to have been recently updated.  Not sure if I'm just confused or something weird happened.  Regardless this doesn't do what it's supposed to do.

eslickj:
@Robbybp, looks like I was just wrong about the binaries.  They seem fine the issue is something about pysp.  I guess I probably need to update pyomo or something.

Robbybp:
Okay, interesting. You might need https://github.com/Pyomo/pyomo/pull/1778 ? Or maybe The uncertainty toolbox relies on the prior version of Parmest?

jsiirola:
FWIW, I have a patch (lbianchi-lbl/idaes-pse#8) in to @lbianchi-lbl's PR #292.  After both of those are merged, I think things will significantly stabalize.

eslickj:
I just updated everything and tried again, and I'm back to k_aug not available.  The version string is definitely updated.  I guess I'll wait for things to shake out, and try again later.  Now I'm just going in circles somehow.

PR #336, opened 2021-05-11
Generic unit for surrogate models or custom models

jghouse88:
Thanks to @carldlaird , @eslickj, @andrewlee94 and @jsiirola for helpful feedback. I think this is ready for review now. I renamed the unit to `SkeletonUnitModel` based on comments. If anyone has a better name, I am open to suggestions. 

## Summary/Motivation:

- There is a need for having an IDAES unit model object that can be used to add a surrogate for an entire unit with certain inputs/outputs or to have a simple custom block that allows the user to add any custom implementation for a model as they see fit. Currently, we do not have an off-the-shelf solution from the unit_models library that lets us do this. We need something like this in several projects now. 
- While a Pyomo Block can be used for this, linking with other IDAES models is limited with the absence of inlet and outlet ports. This lets us have an API that can work with other parts of IDAES.
- UnitModel could have been used but there is a lot of infrastructure built in there that is not required for this and hence this inherits from ProcessBlock. Also, the initialization method simply calls the controlvolume intialize method which will not be available for surrogate or custom models. 

- Ability to add multiple inlets and outlets and populate with whatever vars user wants. Linking with other unit models in IDAES will probably need a translator block to link variables in ports or won't be necessary if the surrogate uses same state vars for inputs and outputs. 
- Default initialize method and an ability for user to use a callback if custom initialization is required
- See `test_skeleton_unit_model.py` for an example
- Added documentation

## What this could develop into?
- Better integration with our surrogate modeling tools such as ALAMO/PySMO. Ability to directly pass an output json or lst file as a  config argument to build an entire surrogate rather than user manually adding set of vars and constraints. Note this is different from simply adding an expression for a property calculation. 
- Integrate with ideas proposed by @eslickj to look at constrained regression with mass/energy balance. 
- IDAES 2.0 is looking at integration with ML tools in Python and we can create the API to integrate with such models.

### Legal Acknowledgement


idaes/generic_models/unit_models/__init__.py
```diff
@@ -16,3 +16,4 @@
 from .statejunction import StateJunction
 from .translator import Translator
 from .valve import ValveFunctionType, Valve, ValveData
+from .surrogate_model import SurrogateModel
```
eslickj, line 19: May as well import SurrogateModelData here too if you may want to subclass the surrogate block.  If not disregard.

idaes/generic_models/unit_models/surrogate_model.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def add_ports(self, name=None, member_list=None, doc=None):
+        """
+        This is a method to build Port objects in a surrogate model and
+        populate this with appropriate port members as specified. User can add
+        as many inlet and outlet ports as required.
+
+        Keyword Args:
+            name : name to use for Port object.
+            dict : dictionary containing variables to be added to the port
```
eslickj, line 70: Is dict member_list?  If dict is member_list, them maybe member_dict would be more clear.

idaes/generic_models/unit_models/surrogate_model.py
```diff
@@ -0,0 +1,135 @@
...
+            if not member_list[k].is_indexed():
+                local_name = member_list[k]
+            else:
+                local_name = member_list[k]
+
+            # Add Reference to Port
+            p.add(Reference(local_name), k)
+
+    def initialize(self, custom_initialize=None, outlvl=idaeslog.NOTSET,
+                   solver='ipopt', optarg={'tol': 1e-6}):
```
eslickj, line 98: Probably go with optarg=None as default.  @andrewlee94's get_solver() method will get a solver with the default tolerance of 1e-6.  I think it's the new standard in initialization.

idaes/generic_models/unit_models/surrogate_model.py
```diff
@@ -0,0 +1,135 @@
...
+
+        Returns:
+            None
+        '''
+        # Set solver options
+        init_log = idaeslog.getInitLogger(self.name, outlvl, tag="unit")
+        solve_log = idaeslog.getSolveLogger(self.name, outlvl, tag="unit")
+
+        opt = SolverFactory(solver)
+        opt.options = optarg
```
eslickj, line 120: I think you can replace these two lines with

```python
if optarg is None:
    optarg = {}
opt = get_solver(solver, oparg)
```

get_solver can be imported from idaes.core.util.misc

idaes/generic_models/unit_models/surrogate_model.py
```diff
@@ -0,0 +1,135 @@
...
+
+        if custom_initialize is not None:
+            custom_initialize
```
eslickj, line 123: You'll want to call the callback: ```custom_initialize()```

Daison2102:
Started testing this class to implement a custom DAC model in IDAES and it was simple to use and very helpful.

idaes/generic_models/unit_models/surrogate_model.py
```diff
@@ -0,0 +1,135 @@
...
+                "member_list should be a dictionary "
+                "with the keys being the name assigned(strings) and "
+                "values being the variable objects declared in the model.")
+
+        # Create empty Port
+        p = Port(noruleinit=True, doc=doc)
+        setattr(self, name, p)
+
+        # Create References for port members
+        for k in member_list.keys():
```
andrewlee94, line 88: This will depend a bit on how the surrogate itself is built, but do we need to make `References` here?

This was necessary in the core models to make it look lie the state variables were directly indexed by time (instead of being on a `Block` indexed by time). Here I would have guessed that any variable would be directly indexed by time (as surrogates probably lack the block-hierarchical structure of other models). If that is the case, then we should skip the `References`.

idaes/generic_models/unit_models/surrogate_model.py
andrewlee94: On that line, the default value for `solver` is now `None` as well, as `get_solver()` will return the configured default solver in this case.

idaes/generic_models/unit_models/tests/test_surrogate_model.py
```diff
@@ -0,0 +1,118 @@
...
+"""
+Generic template for a surrogate unit model.
+"""
+
+import pytest
+from pyomo.environ import ConcreteModel, Constraint, Var, value, SolverFactory
+from idaes.core import FlowsheetBlock
+from idaes.generic_models.unit_models import SurrogateModel
+from idaes.core.util.model_statistics import degrees_of_freedom
+from idaes.core.util.testing import get_default_solver
```
andrewlee94, line 22: Should switch to `get_solver` now.

`get_default_solver` just points to that now with a deprecation warning.

idaes/generic_models/unit_models/tests/test_surrogate_model.py
```diff
@@ -0,0 +1,118 @@
...
+
+__author__ = "Jaffer Ghouse"
+
+
+# -----------------------------------------------------------------------------
+# Get default solver for testing
+solver = get_default_solver()
```
andrewlee94, line 29: As above.

idaes/generic_models/unit_models/tests/test_surrogate_model.py
```diff
@@ -0,0 +1,118 @@
...
+# -----------------------------------------------------------------------------
+
+m = ConcreteModel()
```
andrewlee94, line 32: It would be a good idea to wrap this in a fixture so that it isn't run on import.

idaes/generic_models/unit_models/tests/test_surrogate_model.py
```diff
@@ -0,0 +1,118 @@
...
+m.fs.surrogate.add_ports(name="outlet_2", member_list=outlet_2_dict)
+
+
+def my_initialize():
+    # Callback for user provided initialization sequence
+    m.fs.surrogate.c2.deactivate()
+    m.fs.surrogate.x_1.fix(1)
+    m.fs.surrogate.x_2.fix(2)
+
+    opt = SolverFactory("ipopt")
```
andrewlee94, line 61: Why not use the `solver` object created earlier?

idaes/generic_models/unit_models/tests/test_surrogate_model.py
```diff
@@ -0,0 +1,118 @@
...
+
+    opt.solve(m)
```
andrewlee94, line 63: Might it be worth recording the solve status of both these solves and  attaching it to the model so that you can asset solver status during the tests (to make sure things worked as expected).

idaes/generic_models/unit_models/tests/test_surrogate_model.py
```diff
@@ -0,0 +1,118 @@
...
+    m.fs.surrogate.c2.activate()
+    opt.solve(m)
+
+
+def test_ports():
```
andrewlee94, line 68: Need pytest marks.

idaes/generic_models/unit_models/tests/test_surrogate_model.py
```diff
@@ -0,0 +1,118 @@
...
+    assert degrees_of_freedom(m) == 0
+
+    m.fs.surrogate.initialize()
+
+    assert value(m.fs.surrogate.x_1) == pytest.approx(0.5, abs=1e-3)
+    assert value(m.fs.surrogate.x_2) == pytest.approx(0.5, abs=1e-3)
+
+    assert value(m.fs.surrogate.y_1) == pytest.approx(0.5, abs=1e-3)
+    assert value(m.fs.surrogate.y_2) == pytest.approx(0.25, abs=1e-3)
+
```
andrewlee94, line 109: Should test a case for the default initialize with too few degrees of freedom and check that the expected error is raised.

idaes/generic_models/unit_models/surrogate_model.py
andrewlee94: An important thing to remember here is that `Port` members need to be indexed by time in order to link up with other IDAES models.

That means that either the internal surrogate model variables need to be indexed by time, or we need to add a time index to them in the `Ports` (via `References`). We need to decide which approach we are going to use and clearly document it.

andrewlee94 (commented):
Some comments and suggestions.

A big part of this is going to have to be documenting what a surrogate model is expected to look like inside (i.e. what the user needs to add), such as how variables should be indexed (especially regarding time indices and `Port` members).

jghouse88:
@eslickj and @andrewlee94 Thank you for your comments. I will address them and wait for any other feedback from others. 

@Daison2102 Thanks for testing this with the DAC model. @minhwvu you can use the code from my branch until it is made available in IDAES.

minhwvu:
> @eslickj and @andrewlee94 Thank you for your comments. I will address them and wait for any other feedback from others.
> 
> @Daison2102 Thanks for testing this with the DAC model. @minhwvu you can use the code from my branch until it is made available in IDAES.

Hi @jghouse88 : I run a test and it works for my model. Thanks.

idaes/generic_models/unit_models/surrogate_model.py
andrewlee94: `get_solver` now accepts `optarg=None` as an argument, so this can be done a a single line:

```
opt = get_solver(solver, optarg)
```

idaes/generic_models/unit_models/surrogate_model.py
jghouse88: @andrewlee94 Thanks for this comment. I am leaning towards model variables indexed by time and leaving it to user. See test for an example. I have also documented this.

idaes/generic_models/unit_models/surrogate_model.py
```diff
@@ -0,0 +1,135 @@
...
+@declare_process_block_class("SurrogateModel")
+class SurrogateModelData(ProcessBlockData):
+    """
+    This is the class for a surrogate unit model. This will provide a user
+    with a generic skeleton to add a surrogate unit model.
+    """
+
+    # Create Class ConfigBlock
+    CONFIG = ProcessBlockData.CONFIG()
+    CONFIG.declare("dynamic", ConfigValue(
```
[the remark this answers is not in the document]
jghouse88: See next comment on time.

idaes/generic_models/unit_models/surrogate_model.py
andrewlee94: @jghouse88 I think indexing the variables (and surrogate constraints) by time is the best approach as well, and the draft surrogate API supports this. This makes it a lot easier at the `SkeletonModel` level as you can just add variables directly to `Ports` as required.

idaes/generic_models/unit_models/skeleton_model.py
```diff
@@ -0,0 +1,154 @@
...
+    when controlvolumes are not required.
+    """
+
+    # Create Class ConfigBlock
+    CONFIG = ProcessBlockData.CONFIG()
+    CONFIG.declare("dynamic", ConfigValue(
+        default=False,
+        domain=In([False]),
+        description="Dynamic model flag",
+        doc="""Model does not handle dynamics but should be set by user."""))
```
andrewlee94, line 52: Not quite sure what this doc string means. The first part says it doesn't, but the second part says the user should set it?

Also, a related note, the absence of `has_holdup` means that `ProcessBlockData._setup_dynamics` will fail (as it expects `has_holdup` to be present). I do not think this is a big problem here (at least not yet), but we should be a ware of this at least. We might even want to consider overloading `_setup_dynamics` with a `NotSupported` error.

idaes/generic_models/unit_models/skeleton_model.py
```diff
@@ -0,0 +1,154 @@
...
+        p = Port(noruleinit=True, doc=doc)
+
+        # Add port object to model
+        setattr(self, name, p)
+
+        # Populate port and map names to actual variables as defined
+        for k in member_dict.keys():
+            p.add(member_dict[k], name=k)
+
+    def _custom_initialize_method(self):
```
andrewlee94, line 107: This seems to be a round-about way of doing this, and there is not an obvious API for how the user would setthe custom initialize method.

My first instinct is to just put a reference to the custom initialization method in the config block, and call it directly from there. I.e.

```
def initialize(self, *args, **kwargs):
    if self.config.custom_initializer is None:
        self._default_initialize(*args, **kwargs)
    else:
        self.config.custom_initializer(*args, **kwargs)
```

There is also a question of defining the standard arguments a custom initialization methods should expect.

idaes/generic_models/unit_models/skeleton_model.py
```diff
@@ -0,0 +1,154 @@
...
+        """
+        # Set solver options
+        init_log = idaeslog.getInitLogger(self.name, outlvl, tag="unit")
+        solve_log = idaeslog.getSolveLogger(self.name, outlvl, tag="unit")
+
+        opt = get_solver(solver=solver, options=optarg)
+
+        if self.config.custom_initializer:
+            self._custom_initialize_method()
+        elif degrees_of_freedom(self) == 0:
```
andrewlee94, line 144: Should this be a global check? I.e. shouldn't this also apply even if there is a custom initialisation method?

idaes/generic_models/unit_models/tests/test_skeleton_unit_model.py
```diff
@@ -0,0 +1,305 @@
...
+            m.fs.skeleton.pressure_in[0] - 10)
+
+        inlet_dict = {"flow_mol_comp": m.fs.skeleton.flow_comp_in,
+                      "temperature": m.fs.skeleton.temperature_in,
+                      "pressure": m.fs.skeleton.pressure_in}
+        outlet_dict = {"flow_mol_comp": m.fs.skeleton.flow_comp_out,
+                       "temperature": m.fs.skeleton.temperature_out,
+                       "pressure": m.fs.skeleton.pressure_out}
+
+        m.fs.skeleton.add_ports(name="inlet", member_dict=inlet_dict)
```
andrewlee94, line 84: Just from the view point of unit testing methods, I would suggest only calling these in the `test_ports` method. That way the method is being called as part of the test for its effects.

idaes/generic_models/unit_models/tests/test_skeleton_unit_model.py
```diff
@@ -0,0 +1,305 @@
...
+        assert hasattr(skeleton_default.fs.skeleton, "flow_comp_out")
+        assert hasattr(skeleton_default.fs.skeleton, "temperature_out")
+        assert hasattr(skeleton_default.fs.skeleton, "pressure_out")
+
+    @pytest.mark.skipif(solver is None, reason="Solver not available")
+    @pytest.mark.component
+    def test_default_initialize(self, skeleton_default):
+        # Check default initialize method raises ConfigurationError
+
+        with pytest.raises(ConfigurationError):
```
andrewlee94, line 123: It is good practice to check the error message here to make sure you are not getting another unexpected error.

```
with pytest.raises(ConfigurationError, match="expected string"):
```

idaes/generic_models/unit_models/tests/test_skeleton_unit_model.py
```diff
@@ -0,0 +1,305 @@
...
+            skeleton_default.fs.skeleton.initialize()
+
+        skeleton_default.fs.skeleton.inlet.flow_mol_comp[0, "c1"].fix(2)
```
andrewlee94, line 126: I would also break this into two tests, just so failure in the first part does not obscure any failures in the second part.

andrewlee94 (commented):
I didn't read the docs thoroughly at this point, but the rest looks pretty good.

My biggest comment is that the way you set up the custom initialization method seems rather cumbersome and doesn't offer an obvious way for the user to provide the custom method. I think it would be much more straight-forward to provide the custom method as an input argument in the config block (rather than the current `bool`) and call it directly from the config block. This way there is an obvious API for user on how they should provide the method.

idaes/generic_models/unit_models/skeleton_model.py
jsiirola: I think I would go one farther and just call the config entry `initializer`, and have it hold a callable.  The default value would be `SkeletonUnitModelData._default_initializer`.  I think this would work easier if the function signature was `(unit, *args, **kwargs)`, then `_default_initializer` could be a `@staticmethod` (and declared above the `CONFIG` declaration).  That would help with things like cloning / pickling, as the callable is no longer a bound method.

idaes/generic_models/unit_models/skeleton_model.py
jghouse88: Updated the doc string to better explain what I meant. 

Regarding the second comment, actually it fails without the dynamic flag.

idaes/generic_models/unit_models/skeleton_model.py
jghouse88: Thanks for the suggestions @andrewlee94  and @jsiirola. I have updated as recommended. Thanks for @jsiirola for help with the staticmethod. I think it looks nice and clean now.

jghouse88:
This is ready for review.

docs/technical_specs/model_libraries/generic/unit_models/skeleton_unit.rst
```diff
@@ -0,0 +1,67 @@
+Skeleton Unit Model
+===============================
+
+The IDAES Skeleton Unit model represents a bare bones unit (hence, the name skeleton) that lets the user define a unit model with a custom set
+of variables and constraints that best suits their needs but that can be connected to other models from the unit model library. This unit can be used
+to either define a surrogate for a complex unit model or to define a first-principles based model that does not use IDAES
```
andrewlee94, line 6: Minor point, but there is nothing stopping a user from adding ControlVolumes here either.

docs/technical_specs/model_libraries/generic/unit_models/skeleton_unit.rst
```diff
@@ -0,0 +1,67 @@
...
+
+The two methods available for this unit are:
+
+1. ``add_ports``: Allows the user to define the inlet and outlet ports for the model and populate the port with a set of variables.
+2. ``initialize``: A default initialize method that will attempt to solve the model as defined or that can be substituted with a custom callback function provided by the user.
+
+Degrees of Freedom
+------------------
+
+There are no variables or constraints as this is a bare minimum skeleton that will be populated with variables and constraints
```
andrewlee94, line 18: The following three sections aren't really applicable in this case (as everything depends on the user). It would probably be better to give an overview of how to use the skeleton model, and what the user needs to keep in mind.

Some examples:

* Port members need to be indexed by time. That leads into how to access the time domain (I will add a PR for making `time` a property of the flowsheet).
* How to use the methods to add ports and how to define the members of the port.
* A reminder that Ports must be consistent, thus users need to match their Port definitions to the flowsheet they intend to use it in.

docs/technical_specs/model_libraries/generic/unit_models/skeleton_unit.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Custom Callback
+---------------
+
+Note that the default value for the ``initializer`` config argument is set to trigger the ``_default_initialize`` method. To use a custom callback instead, the user needs to first define a function
+and set the ``obj.config.initializer`` to the custom callback function they defined. In the ``initialize`` method, after checking that degrees of freedom is zero, the following is triggered:
+
+.. code-block:: Python
+
+   self.config.initializer(self, opt=opt, init_log=init_log, solve_log=solve_log)
```
andrewlee94, line 41: It might be worth noting the expectation of a static method and the fact that self here will be the skeleton model object. Given the range of users we expect, it might even be worth showing some pseudo-code of how to initialize a variable.

idaes/generic_models/unit_models/skeleton_model.py
```diff
@@ -0,0 +1,156 @@
...
+        # Add port object to model
+        setattr(self, name, p)
+
+        # Populate port and map names to actual variables as defined
+        for k in member_dict.keys():
+            p.add(member_dict[k], name=k)
+
+
+    def initialize(self, outlvl=idaeslog.NOTSET,
+                   solver=None, optarg=None):
```
andrewlee94, line 120: Should we add an optional argument for providing initial guesses? This would probably be passed directly to the initializer method, so users could do what they want with it.

andrewlee94 (dismissed):
Looks really good.

My comments are all minor things on potential (but not necessary) improvements.

docs/technical_specs/model_libraries/generic/unit_models/skeleton_unit.rst
andrewlee94: Two other things that would be worth mentioning:

* unit consistency (if the user wishes to assert unit consistency across the model, then the Skeleton will need to be unit consistent)
* Brief mention of how this could interact with the scaling tools (for now, just say the user would need to do this manually).

docs/technical_specs/model_libraries/generic/unit_models/skeleton_unit.rst
jsiirola: There is no need for a user's custom function to be a static method - a "global" (unbound) function works as well.  I agree that we should clarify semantics in the documentation prototype and not use `self` here.

docs/technical_specs/model_libraries/generic/unit_models/skeleton_unit.rst
jghouse88: Yes, they can but I do not see why..

docs/technical_specs/model_libraries/generic/unit_models/skeleton_unit.rst
jghouse88: I plan to create an example in examples-pse

docs/technical_specs/model_libraries/generic/unit_models/skeleton_unit.rst
andrewlee94: My point being we should not give the impression of restrictions that do not actually exist. I would make it more vague, such as "first principles models that do not adhere to the typical IDAES modeling hierarchy".

docs/technical_specs/model_libraries/generic/unit_models/skeleton_unit.rst
jghouse88: Before the code line, I do mention that "In the ``initialize`` method, after checking that degrees of freedom is zero, the following is triggered:". 
@jsiirola I am not sure I follow why self is not appropriate here.

jsiirola: @jghouse88: good point; you are right.

docs/technical_specs/model_libraries/generic/unit_models/skeleton_unit.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Note that the function signature in the above line expects an instance, ``opt`` which is the solver (default is set to ipopt), ``init_log`` and ``solve_log`` which are the idaes logger objects for initialization
+and solve. The user can either pass these arguments in their custom function or simply do the following:
+
+.. code-block:: Python
+
+   def custom_func(unit, **kwargs):
```
jsiirola, line 48: Should we be recommending `**kwargs`, or should we write out the full prototype that will be used to call this?  That is,
```python
def custom_func(unit,  opt, init_log, solve_log):
    pass
```

jghouse88: Now that you mention it, I should probably show both..

idaes/generic_models/unit_models/skeleton_model.py
jghouse88: Interesting suggestion, however, how do you think we should handle that argument with the _default_initialize() method? Traditionally, we have assumed the initial guess to be applicable for the inlet port members. In the custom model case, the inlet port could be a subset of the input vars required for the model. I can always set `state_args` to `None` and not use it in the default_initialize and this could let the user use that argument in their custom method as they see fit. Let me know your thoughts.

andrewlee94: The simplest approach would to be accept it but then ignore it. Given that we don't know what a user will put inside a SkeletonModel, this is probably the best approach.

An alternative would be to assume these are initial values for variables in the outlet and try to assign to the values of these, but I am not sure that adds much (especially seeing as the main use for these models is for surrogates which are generally simple explicit functions).

andrewlee94: To add a little more - the main goal here is to give users some flexibility to pass extra information to their custom initializers. Obviously, this is not applicable to the default initializer, so it should not expect any additional information, but needs to have a placeholder for the addition information which it just ignores.

jghouse88: Alright.

jghouse88:
@andrewlee94 and @jsiirola can you please take a look? I addressed the comments on the docs and also added `initial_guess` to the initialize method that can potentially be passed to the callback if the user desires.

andrewlee94 (approved):

eslickj (approved):

PR #337, opened 2021-05-12
Fixing boiler heat exchanger initialize function

MAZamarripa:
## Fixes
- boiler heat exchanger initialize method

## Summary/Motivation:
fix bug with boiler heat exchanger initialization method

- 
-

### Legal Acknowledgement


idaes/power_generation/unit_models/helm/splitter.py
```diff
@@ -158,12 +158,12 @@ def sum_split(b, t):
         @self.Constraint(time, self.outlet_list, doc="Pressure constraint")
         def pressure_eqn(b, t, o):
             o_block = getattr(self, "{}_state".format(o))
-            return self.mixed_state[t].pressure == o_block[t].pressure
+            return self.mixed_state[t].pressure*1e-6 == o_block[t].pressure*1e-6
```
eslickj, line 161: We should really use the scaling tools to do stuff like this, but it's okay for now.  Probably should revisit it at some point.

andrewlee94, line 161: Not sure if this is standard for the Helmholtz models, but should we be hard coding these scaling factors, or using the scaling tools?

eslickj (approved):

idaes/power_generation/unit_models/helm/splitter.py
eslickj: Actually now that I looked at the rest of the model, this is in the ```calculate_scale_factors``` method.  That means if someone uses that the equations will be double scaled, so we should probably not make these splitter scaling changes.  Let me know if the splitter scaling isn't working for whatever reason and I'll take a harder look at it.

eslickj (changes requested):

eslickj (commented):
@MAZamarripa if you don't mind, since I need to combine this with the gas turbine, I can make the boiler heat exchanger changes in the gas turbine PR and the splitter changes shouldn't be needed. Would that be ok?  It would make it a bit easier for me to deal with, since the gas turbine PR is also waiting.

andrewlee94:
Is this PR still needed, or has it been merged with #323?

eslickj:
@andrewlee94, I did merge it with the gas turbine one, except for the hard coded scaling.  Unless @MAZamarripa, thinks otherwise, I think we can close it.

andrewlee94:
I am told most of this has been merged into #323. Anything that remains can be done in another PR, so I will close this.

PR #339, opened 2021-05-12
Fixing some issues with indexing sets in generic properties

### LINKED ISSUE #338: Generic properties components not is phase
There is a least one maybe more places where components that don't exist in a phase cause trouble in the general properties module.  Here is one I know of:

``` python
    def _mw(self):
        try:
            self.mw = Expression(
                    doc="Average molecular weight",
                    expr=sum(self.phase_frac[p] *
                             sum(self.mole_frac_phase_comp[p, j] *
                                 self.params.get_component(j).mw
                                 for j in self.params.component_list)
                             for p in self.params.phase_list))
```

The component list needs to be restricted to the components that exist in a phase.  @minhwvu brought this to my attention.

andrewlee94:
## Summary/Motivation:
Cleans up a missing check for non-vapourisable and non-condensable components, plus updating to the new API for indexing sets.

-
-

### Legal Acknowledgement


eslickj (approved):

agarciadiego (approved):
LGTM

PR #341, opened 2021-05-13
Starting to fix tests to work with Pyomo 6

andrewlee94:
## Summary/Motivation:
Changes in Pyomo 6 mean a number of our tests will start breaking. This is the first part of addressing these.

- Fixing tests in generic properties examples
-

### Legal Acknowledgement


eslickj (approved):

Andresj89 (approved):

idaes/generic_models/properties/core/eos/tests/test_enrtl.py
```diff
@@ -289,14 +289,14 @@ def test_common(self, model):
         for k in model.state[1].Liq_x_ref:
             assert k in ["H2O", "C6H12", "Na+", "H+", "Cl-", "OH-"]
             if k in ["H2O", "C6H12"]:
-                assert model.state[1].Liq_x_ref[k].expr == 0
+                assert str(model.state[1].Liq_x_ref[k].expr) == str(0.0)
```
jsiirola, line 292: Is `str()` really the best way to do this test?

idaes/generic_models/properties/core/eos/enrtl.py
```diff
@@ -436,7 +437,7 @@ def rule_tau_expr(b, i, j):
             else:
                 alpha = getattr(b, pname+"_alpha")
                 G = getattr(b, pname+"_G")
-                if G[i, j].expr != 1:
+                if str(G[i, j].expr) != "1.0":
```
jsiirola, line 440: Why `str()` here and not `value()`?

jsiirola (commented):
I'm not going to object to the PR, but I really worry about using `str()` to compare Pyomo expressions / values.  That is likely to be a fragile (and slow) comparison.

idaes/generic_models/properties/core/eos/enrtl.py
andrewlee94: This is because we need to distinguish between the case where the value happens to be 1, and where the expression itself is 1. I.e., I am checking for the equivalent of `Expression(expr=1)`.

eslickj: I'm not too sure, but if  G[i, j] is an expression and just 1, the type should be <class 'pyomo.core.expr.numvalue.NumericConstant'>.  You could check the type maybe?  Not sure the best way. I agree str() seems a little weird.

PR #343, opened 2021-05-14
Removing mutable default arguments in core code and models

andrewlee94:
## Summary/Motivation:
This addresses all cases of mutable default arguments in the core code and model libraries. I have not touched the surrogates and dmf code, which still have some of these.

-
-

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/misc.py
+++ b/idaes/core/util/misc.py
@@ -32,7 +32,7 @@ _log = idaeslog.getLogger(__name__)
 
 
 # Author: Andrew Lee
-def get_solver(solver=None, options={}):
+def get_solver(solver=None, options=None):
     """
     General method for getting a solver object which defaults to the standard
     IDAES solver (defined in the IDAES configuration).
@@ -40,7 +40,7 @@ def get_solver(solver=None, options={}):
     Args:
         solver: string name for desired solver. Default=None, use default solver
         options: dict of solver options to use, overwrites any settings
-                 provided by IDAES configuration.
+                 provided by IDAES configuration. DEfault = None.
 
     Returns:
         A Pyomo solver object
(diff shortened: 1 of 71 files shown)


eslickj (dismissed):
Looks good.  I didn't review it really close, but seems to be all mutable default replacement and formatting, which looks good.

andrewlee94:
OK, the test failures here appear to be because of a change that is coming up in Pyomo 6 but has not yet been merged into the IDAES branch of Pyomo.

idaes/core/util/misc.py
```diff
@@ -32,23 +32,26 @@
...
+def get_solver(solver=None, options=None):
     """
     General method for getting a solver object which defaults to the standard
     IDAES solver (defined in the IDAES configuration).
 
     Args:
         solver: string name for desired solver. Default=None, use default solver
         options: dict of solver options to use, overwrites any settings
-                 provided by IDAES configuration.
+                 provided by IDAES configuration. DEfault = None.
```
ksbeattie, line 43: the extra "default=None" here is either redundant or needs to be explained.

ksbeattie (approved):
I was worried this would look awkward but I think this change actually makes things clearer.

@lbianchi-lbl, can we get an explicit check for these "dangerous" warning put into the pylint check?

andrewlee94:
@ksbeattie I was going to suggest we explicitly add that check. However, we need to fix the surrogate folder first.

idaes/core/util/misc.py
andrewlee94: @ksbeattie I added an explanation here, and in all the other places where a similar statement was made. However, this means I need a review again.

eslickj (approved):

eslickj (approved):

ksbeattie (approved):
LGTM again

PR #347, opened 2021-05-17
Adding initial overwrite option to constraint_scaling_transformation

andrewlee94:
## Summary/Motivation:
A few people have noted that the current `calculate_scaling_factors` method will overwrite any custom constraint scaling that has been applied to a model (rescaling the constraints based on the default rules). This is less than desirable, so this PR add a new optional argument to the `constraint_scaling_transformation` to control whether it will rescale constraint which already have a scaling factor. The `calcualte_scaling_factors` method can then set `overwrite=False` to not update any pre-existing constraint scaling.

The current implementation in this PR probably goes too far in the opposite direction however, as it means constraint scaling can only be done once using the `calculate_scaling_factors` method. We probably want to extend this to support updating of any automatically generated constraint scaling, whilst preserving user-defined scaling.

- Adds `overwrite` argument to `constraint_scaling_trandsformation` method, with default value of `True` (to preserve current API).
- Updated all instances of `constraint_scaling_transformation` in `calculate_scaling_factors` (and related) methods to explicitly state `overwrite=False`

### Legal Acknowledgement


eslickj:
This looks good to me so far.

eslickj (dismissed):

Andresj89 (dismissed):
I tested the PR and it works fine

eslickj (approved):

Andresj89 (approved):
LGTM

PR #348, opened 2021-05-18
Utility minimization

agarciadiego:
## Summary/Motivation:
Creates constraints for Duran-Grossmann formulation

- Adds a function to minimize utilities
- Adds a function to generate composite curves

### Legal Acknowledgement


andrewlee94:
@agarciadiego There is currently a test failure due to an rst file not being part of the ToC tree - this means you didn't link the new file to an index in the docs.

idaes/core/util/utility_minimization.py
```diff
@@ -0,0 +1,621 @@
+##############################################################################
```
andrewlee94, line 1: We need to add the IDAES copyright header.

idaes/core/util/utility_minimization.py
```diff
@@ -0,0 +1,621 @@
...
+        DTmin: HRAT (Heat Recovery Approximation Temperature)
+        eps: Epsilon for smoothing operation
+
+    Returns:
+        Constraint and variable object representing the calculation for hot utility and cold utility
+    """
+    epsT = 1
+
+    # Generate lists out of strings from Args
+    exchanger_list = heating + cooling
```
andrewlee94, line 43: It would be good to have some argument verification, but it doesn't need to hold up the PR.

idaes/core/util/utility_minimization.py
```diff
@@ -0,0 +1,621 @@
...
+
+    for i,el in enumerate(heating):
+        dT[str(el)] = DTmin
+
+    def T_in(blk, i):
+        return pinch_streamsdict[i].control_volume.properties_in[0].temperature
+
+    blk.Tin = Expression(pinch_streamsdict.keys(),
+                    rule=T_in,
+                    doc='Inlet temperature in exchangers [K]')
```
andrewlee94, line 90: It would be good to add Pyomo units here. However, this seems to presume that SI units are being used.

idaes/core/util/utility_minimization.py
```diff
@@ -0,0 +1,621 @@
...
+    Returns:
+        Dictionary for the equipment exchanging heat containing:
+            Tin
+            Tout
+            FCp
+            Q
+    """
+    # Temperature Epsilon to add or subsract 1 degree to avoid division over
+    # 0 in equipment not changing temperature
+    espT = 1
```
andrewlee94, line 197: This looks like it should be a global constant. I recall seeing it in the previous function as well.

idaes/core/util/utility_minimization.py
```diff
@@ -0,0 +1,621 @@
...
+    c = max([max(b),0.0])
+    initQs = c
+
+    # Calculate Qw from largest value of Qs
+    initQw = -sum(value(exchangeData[i]['Q']) for i in exchangerdict.keys()) + initQs
+    initQw = max([initQw,0.0])
+
+    # Fill Class with all the data to initialioze Duran-Grossmann variables
+    PD = PinchDataClass(initQs, initQw)
+    print(initQAh)
```
andrewlee94, line 326: The print statements lack any context - are these left over from debugging?

andrewlee94 (commented):
I don't really know enough about what this is doing to really comment. I did see a couple of things to think about.

ksbeattie:
@MAZamarripa will be meeting with @agarciadiego about this, this week as well.

idaes/core/util/utility_minimization.py
agarciadiego: Delete [K] from doc, so we do not assume what temperature we should use.

idaes/core/util/utility_minimization.py
```diff
@@ -0,0 +1,674 @@
...
+        f_[i] = value(exchangerdict[i].control_volume.properties_out[0].flow_mol)
+        Q_[i] = value(exchangerdict[i].control_volume.heat[0])
+
+    # Print the header
+    print("Heat Exchanger Summary: ")
+
+    # Print Inlet Temperature, Outlet Temperature and Heat
+    for i in exchangerdict.keys():
+        print("Heat exchanger: ", exchangerdict[i])
+        print('{0:>{1}s}'.format("Inlet T: ", 16),
```
andrewlee94, line 585: Pylint is complaining about the formatting of these print statements - too many arguments for `format`. I think the issue is the `3` in the second `format` method - you don't actually use it in the string formatting ad the 3 is hard-coded there.

I also think you could probably make this code easier to read with something similar to:

```
f"Inlet T:{" "*7} {Tin_[i]:0.3f} K/n"
```

idaes/core/util/utility_minimization.py
```diff
@@ -0,0 +1,674 @@
...
+              '{0:0.3f}'.format(Tin_[i], 3), end=' K\n')
+        print('{0:>{1}s}'.format("Outlet T: ", 16),
+              '{0:0.3f}'.format(Tout_[i], 3), end=' K\n')
```
andrewlee94, line 588: As above.

idaes/core/util/utility_minimization.py
```diff
@@ -0,0 +1,674 @@
...
+        print('{0:>{1}s}'.format("Q: ", 16),
+              '{0:0.3f}'.format(Q_[i], 3), end=' J/s\n')
```
andrewlee94, line 590: As above.

andrewlee94 (changes requested):
@agarciadiego There is a test failing due to our code checking tool (pylint) that needs to be resolved. I marked the lines of code that need to be fixed, and I think the fix is easy.

idaes/core/util/utility_minimization.py
agarciadiego: I also changed the hardcoded units to use the ones from pyomo

idaes/core/util/utility_minimization.py
```diff
@@ -569,25 +569,27 @@ def print_HX_results(blk, exchanger_list):
...
     # Print Inlet Temperature, Outlet Temperature and Heat
     for i in exchangerdict.keys():
         print("Heat exchanger: ", exchangerdict[i])
-        print('{0:>{1}s}'.format("Inlet T: ", 16),
-              '{0:0.3f}'.format(Tin_[i], 3), end=' K\n')
-        print('{0:>{1}s}'.format("Outlet T: ", 16),
-              '{0:0.3f}'.format(Tout_[i], 3), end=' K\n')
-        print('{0:>{1}s}'.format("Q: ", 16),
-              '{0:0.3f}'.format(Q_[i], 3), end=' J/s\n')
+        print(f'Inlet T:{" "*3} {Tin_[i]:0.3f}, '+str(T_units))
```
andrewlee94, line 590: Minor comment, but you can include the T_units in the f-string too:

```
f'Inlet T:{" "*3} {Tin_[i]:0.3f {T_units}}'
```

andrewlee94 (approved):

MAZamarripa (approved):
lgtm
in the next PR try to add the full reference for Duran Grossmann Formulation

PR #351, opened 2021-05-19
Command line config tools

eslickj:
Even though nobody asked for it, I put together a little utility for idaes configuration from the command line.   I also sped up the imports for the idaes command, so it doesn't run annoyingly slow.  @jsiirola I tried to move to pyomo ConfigBlocks as much as possible, but there is still a bit of an issue with logger configuration since loggers have periods in their names.  That means I have to use a dict there instead of a ConfigBlock, since ConfigBlock keys can't have periods (and I use logging.dictConfig() directly, in which keys may have periods).  @dangunter, I messed with your imports in get-examples to speed things up.  You'll probably want to take a look at that.  The main issue I had was figuring out how to mange setuptools.  I ended up just moving it in the command function call because I couldn't find another way that worked. 

### Legal Acknowledgement

### DIFF
--- a/idaes/commands/base.py
+++ b/idaes/commands/base.py
@@ -22,7 +22,7 @@ import logging
 # separate command logging from normal IDAES logging
 _log = logging.getLogger("idaes.commands")
 _h = logging.StreamHandler()
-_h.setFormatter(logging.Formatter("%(asctime)s %(levelname)-7s %(name)s: %(message)s"))
+_h.setFormatter(logging.Formatter("%(levelname)-7s %(name)s: %(message)s"))
 _log.addHandler(_h)
 _log.propagate = False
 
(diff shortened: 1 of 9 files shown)


dangunter (dismissed):
Changes to get-examples look good. Thanks for that.
I am agnostic on the other changes, but if they pass tests it's fine with me.

idaes/commands/base.py
```diff
@@ -103,6 +103,10 @@ def copyright():
 """
     )
 
+@command_base.command(help="Show how long it takes to imoprt command modules")
```
jsiirola, line 106: typo: "import"

eslickj:
@dangunter and @jsiirola, I fixed the misspelled import.  If you guys finished your reviews the first time around, it's probably safe to re-approve, if not take your time.

andrewlee94 (approved):
I will assume this is good after @dangunter and @jsiirola's comments.

MAZamarripa (approved):
lgtm + (I will assume this is good after @andrewlee94, @dangunter and @jsiirola's comments)

PR #358, opened 2021-05-22
This adds a utiliy to use WSL to run solvers for Linux on Windows

eslickj:
This helps us out by allowing us to use our Linux solvers on Windows.  There are some solvers that are darn near impossible to make work on native Windows (usually due to some reliance on POSIX standards).  Since we already have Linux binaries, the WSL solution seems better than Cygwin.  The point of this PR is to convert Windows paths into something that the WSL expects.  This is the least invasive approach I could think of.  I also thought of just doing this in the batch file, but it's too hard to work with batch files.

To use this:

1. Install WSL2, https://docs.microsoft.com/en-us/windows/wsl/install-win10
2. Use a batch file for a solver executable like this ipopt_wsl.bat
```
   @echo off
   idaes solver-wsl -d Ubuntu-20.04 -u john -e ~/local/bin/ipopt %*
```
3. Create a Pyomo solver like this 
```python
solver = pyo.SolverFactory(
    "ipopt_wsl",
     executable="ipopt_wsl.bat"
     type='asl')
```

Why have I done this?  Mostly to use the PETSc solver without compiling it on Windows.  The main disadvantage to this is that if you use external functions you'll have the set AMPLFUNC up manually for the WSL part before you run your problem.  That should be easy enough though. 

### Legal Acknowledgement


andrewlee94 (approved):
I am not familiar with wsl, but I don't see any issues here.

idaes/commands/wrap_solver_wsl.py
```diff
@@ -0,0 +1,42 @@
...
+    context_settings={"ignore_unknown_options": True},
+    help="Run a linux solver on Windows using WSL")
+@click.option("--distribution", "-d", help="Linux distribution")
+@click.option("--user", "-u", help="User")
+@click.option("--executable", "-e", help="Executable path on WSL")
+@click.argument('args', nargs=-1)
+def solver_wsl(distribution, user, executable, args):
+    al = [None]*len(args)
+    for i, a in enumerate(args):
+        if a.startswith("C:\\"): # it's a windows path, convert for WSL
```
jsiirola, line 38: This is fragile and assumes that the user is running on the first local disk drive...  I think you want to ignore the drive letter.  Maybe a regex like `r'[a-zA-Z]:\\'`?

idaes/commands/wrap_solver_wsl.py
```diff
@@ -0,0 +1,42 @@
...
+            a = a.replace("\\", "/").replace("C:", "/mnt/c")
```
jsiirola, line 39: Instead of the hard-coded `C:`, I think you need something like:
```python
a = '/mnt/' + a[0].lower() + a[3:].replace('\\', '/')
```

jsiirola (changes requested):
I think this needs to be generalized (and not require hard-coded `C:` drive specifications)

dangunter:
When I run  `idaes solver-wsl` with no args I get a stacktrace that says something like "a string is expected not NoneType" (I can give full error if you need it, but probably easy to reproduce).

dangunter (changes requested):
So I think the minor issue of the stack trace when invoked without -d or -e should be fixed, but when I do run it with a WSL distro and path to ipopt it seems to work (at least for ipopt invoked without args or -v).

eslickj:
> 
> 
> So I think the minor issue of the stack trace when invoked without -d or -e should be fixed, but when I do run it with a WSL distro and path to ipopt it seems to work (at least for ipopt invoked without args or -v).

I made distribution, user, and executable required arguments, so once I push that, you should just get a short message about a required argument being missing.

eslickj:
@jsiirola and @dangunter, should be fixed.

idaes/commands/wrap_solver_wsl.py
```diff
@@ -0,0 +1,47 @@
...
+    context_settings={"ignore_unknown_options": True},
+    help="Run a linux solver on Windows using WSL")
+@click.option("--distribution", "-d", required=True, help="Linux distribution")
+@click.option("--user", "-u", required=True, help="User")
+@click.option("--executable", "-e", required=True, help="Executable path on WSL")
+@click.argument('args', nargs=-1)
+def solver_wsl(distribution, user, executable, args):
+    al = [None]*len(args)
+    for i, a in enumerate(args):
+        r = re.match(r"^([A-Z]):\\(.*$)", a)
```
jsiirola, line 39: Can the drive letter ever be lower case?  I think the RE should be `r"^([A-Za-z]):\\(.*$)"`

eslickj: Oh yeah.  You're right I guess it is case insensitive.  Will fix.

jsiirola (approved):

andrewlee94 (approved):
I don't really know what I am looking at, but I will trust those who do.

dangunter (approved):
Looks good, thanks for addressing comments.

PR #360, opened 2021-05-25
Adding prototype solubility product formulation

andrewlee94:
## Summary/Motivation:
NAWI requires precipitation reactions for some unit operations, so this PR adds support for this to the generic reaction framework. As a phase equilibrium phenomena, precipitation requires a complementarity formulation so this will need some testing.

Note that this formulation uses an `eps` parameter which is tunable, and probably needs some adjustment. We will also need some test cases to demonstrate that this actually works.

Documentation will come one we have a handle on how well this behaves.

- Adds new class for solubility product formulation with basic unit tests.
-

### Legal Acknowledgement


Andresj89 (changes requested):
Just a few typos

idaes/generic_models/properties/core/reactions/tests/test_equilibrium_forms.py
```diff
@@ -324,3 +326,203 @@ def test_log_power_law_equil_with_order():
...
+    solubility_product.build_parameters(
+        m.rparams.reaction_r1,
+        m.rparams.config.equilibrium_reactions["r1"])
+
+    # Check reaction form - should raise exception
+    with pytest.raises(ConfigurationError,
+                       match="did not find a solid phase component for "
+                       "precipitation reaction r1. This is likely due to the "
+                       "reaction configuration."):
+        solubility_product.return_expression(
```
adam-a-a, line 527: Does this mean that for a subsaturated solution, a config error will also be generated, or is this only because solids were omitted from the reaction parameter block config? 

I'm guessing this was written from the perspective of triggering precipitation reactions for pretreatment processes, i.e., with the main focus placed on the presence of solids. I am just thinking about how we could implement this to verify that a solution is still undersaturated with respect to some potential foulant (e.g., in the case where we want to avoid mineral scaling on desalination membranes).

idaes/generic_models/properties/core/reactions/equilibrium_forms.py
```diff
@@ -15,7 +15,8 @@
 """
```
adam-a-a, line 15: typo on line 14: "equilibrium"

idaes/generic_models/properties/core/reactions/equilibrium_forms.py
```diff
@@ -117,3 +118,88 @@ def return_expression(b, rblock, r_idx, T):
...
+
+        * If S >= 0, Q = 0
+        * If S = 0, Q >= 0
+
+    Thus, only one of S and Q can be greater than zero at any time.
+    This can be written in the for of a complementarity constraint as:
+
+        * S - MAX(0, S-Q) == 0
+
+    In this implentation, a smooth maximum approximation is used, and S is
```
adam-a-a, line 145: typo "implementation"

adam-a-a (changes requested):
Only a couple minor typos and a question about using this to verify that a solution is undersaturated with respect to potential foulants.

idaes/generic_models/properties/core/reactions/tests/test_equilibrium_forms.py
andrewlee94: This check is only looking for the definition of a solid phase component, not whether there is any actually present in the system.

The model formulation is designed to support sub-saturated systems, and that is part of why it is as complex as it is.

adam-a-a: From `If S = 0, Q >= 0`, the undersaturated case should be handled, so I guess the user just has to be sure to define the potential solids in the reaction config. Is that correct?

andrewlee94: Yes, that is correct. The error is to just catch the case where the potential solid was not defined, in which case `S=None` which won't work out well in an expression.

andrewlee94:
@adam-a-a @Andresj89 The release code freeze is over now, so I would like to start merging the backlog of PRs. Would you be able to re-review this?

Andresj89 (approved):
LGTM

adam-a-a (approved):
LGTM

PR #361, opened 2021-05-25
Fix missing console log handler

eslickj:
This is an important fix.  I left the handler out of the new default config, so you won't see log messages by default.  This corrects the issue.

### Legal Acknowledgement

### DIFF
--- a/idaes/config.py
+++ b/idaes/config.py
@@ -56,81 +56,6 @@ orig_environ = {
     "DYLD_LIBRARY_PATH": os.environ.get("DYLD_LIBRARY_PATH", ""),
 }
 
-# Default configuration json string.  Store the default config like this because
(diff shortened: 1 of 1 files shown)


idaes/config.py
```diff
@@ -56,81 +56,6 @@
     "DYLD_LIBRARY_PATH": os.environ.get("DYLD_LIBRARY_PATH", ""),
 }
 
-# Default configuration json string.  Store the default config like this because
```
andrewlee94, line 59: Just to be sure, was all this deletion expected?

eslickj: Yeah.  I moved the defaults in the Pyomo config block.

Andresj89 (approved):
LGTM

jghouse88 (approved):
LGTM

Issue #330, opened 2021-05-06
Mutable default arguments


eslickj:
I found a problem with a table utility function and I'm afraid it could be an issue elsewhere in IDAES.  I know in quite a few places we may use an empty list or dict as a default.  The problem with a mutable default argument is that if you change the augment in the function, you also change the default value.  Here's and example:

```python
>>> import copy
>>> def f(k, x={}):
...   x[k] = "Hi"
...   return copy.copy(x)
...
>>> print(f(1))
{1: 'Hi'}
>>> print(f(2))
{1: 'Hi', 2: 'Hi'}
>>> print(f(3))
{1: 'Hi', 2: 'Hi', 3: 'Hi'}
```

I'm not sure if we are making a similar mistake anywhere else, but we should probably check. I know we had a lot of mutable objects as default values, but I don't know if we alter them in the function. I wonder if this is something pylint can do or does.

ksbeattie:
@eslickj, this can be caught by pylint.  Given the way we are running pylint in our tests right now, it won't but here's how to search just for that specific warning:

    pylint --disable=all --enable=dangerous-default-value idaes

I just ran it and found...
<details>
  <summary>A lot of warnings</summary>

```
$ pylint --disable=all --enable=dangerous-default-value idaes
************* Module idaes.core.unit_model
idaes/core/unit_model.py:608:4: W0102: Dangerous default value {} as argument (dangerous-default-value)
************* Module idaes.core.control_volume1d
idaes/core/control_volume1d.py:153:4: W0102: Dangerous default value [] as argument (dangerous-default-value)
idaes/core/control_volume1d.py:1637:4: W0102: Dangerous default value {} as argument (dangerous-default-value)
************* Module idaes.core.control_volume0d
idaes/core/control_volume0d.py:1320:4: W0102: Dangerous default value {} as argument (dangerous-default-value)
************* Module idaes.core.util.scaling
idaes/core/util/scaling.py:529:4: W0102: Dangerous default value [] as argument (dangerous-default-value)
************* Module idaes.core.util.misc
idaes/core/util/misc.py:35:0: W0102: Dangerous default value {} as argument (dangerous-default-value)
idaes/core/util/misc.py:118:0: W0102: Dangerous default value {} as argument (dangerous-default-value)
************* Module idaes.core.util.initialization
idaes/core/util/initialization.py:36:0: W0102: Dangerous default value {} as argument (dangerous-default-value)
************* Module idaes.core.util.tables
[...]

PR #365, opened 2021-05-27
Adding solid properties to ideal eos

andrewlee94:
## Summary/Motivation:
In order to support precipitation reactions (and test #330), we need to support properties of solid phases in the generic framework. This PR extends the existing Ideal EoS and base classes to support solid properties.

-
-

### Legal Acknowledgement


TimBartholomew (approved):
LGTM

idaes/core/components.py
```diff
@@ -54,18 +54,27 @@ class ComponentData(ProcessBlockData):
 
     CONFIG.declare("dens_mol_liq_comp", ConfigValue(
         description="Method to use to calculate liquid phase molar density"))
+    CONFIG.declare("dens_mol_sol_comp", ConfigValue(
+        description="Method to use to calculate solid phase molar density"))
+
     CONFIG.declare("cp_mol_liq_comp", ConfigValue(
         description="Method to calculate liquid component specific heats"))
+    CONFIG.declare("cp_mol_sol_comp", ConfigValue(
```
adam-a-a, line 62: @andrewlee94 Should there also be a test for this? I noticed that you added tests for the other properties that you introduced, with the exception of `cp_mol_sol_comp`. 

I was thinking this would just need modification in `test_cv_mol_phase_comp` in `test_ideal.py`.

andrewlee94: There really isn't much need to test config arguments (especially if you don't set a default or domain). If it is missing, it should show up in a later test that tried to use the argument.

adam-a-a (commented):
I only have one question about including a test for `cp_mol_sol_comp`. Aside from that, everything looks good to me.

idaes/core/components.py
adam-a-a: Ah, OK. I guess I just thought you'd want to add the corresponding lines for `sol` the same way you did for `liq` and `vap` (lines 183-191 in `test_ideal.py`):

```
m.params.get_component(j).config.cp_mol_liq_comp = dummy_call
m.params.get_component(j).config.cp_mol_ig_comp = dummy_call       
 
assert str(Ideal.cv_mol_phase_comp(m.props[1], "Liq", j)) == str(42)        
assert str(Ideal.cv_mol_phase_comp(m.props[1], "Vap", j)) == str(42 - pyunits.convert(const.gas_constant,                to_units=pyunits.kg*pyunits.m**2/pyunits.s**2 /pyunits.mol/pyunits.K))
```

Andresj89 (approved):

PR #376, opened 2021-06-03
Fix separator scale factor calcs

eslickj:
## Fixes

This is a problem reported by @minhwvu.  The scaling factor calculation is broken for separators when the separator doesn't have it's own mixed state block.  Unfortunate, I think we should try to get the fix in this release if possible.   I added a test that demonstrates the problem then fixed it.

### Legal Acknowledgement

### DIFF
--- a/idaes/generic_models/unit_models/separator.py
+++ b/idaes/generic_models/unit_models/separator.py
@@ -456,12 +456,14 @@ objects linked the mixed state and all outlet states,
         """
         # Sanity check to make sure method is not called when arg missing
         if self.config.mixed_state_block is None:
-            raise BurntToast(
-                "{} get_mixed_state_block method called when "
-                "mixed_state_block argument is None. This should "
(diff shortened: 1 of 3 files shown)


andrewlee94 (approved):
It would probably be good to have explicit tests for the two cases, but this looks good.

eslickj:
@andrewlee94, yeah probably should be part of the new testing standards, to make sure the scale factors work right.  Right now they probably have pretty minimal coverage.  Just where we incidentally use them.

minhwvu:
Thanks @eslickj . The error was fixed when I did a test with my model.

eslickj:
@andrewlee94, you were right again.  I should have tested both cases because I broke it the opposite way.  Should be fixed and added more tests.  Hopefully it works this time.

andrewlee94 (approved):
Looks good to me.

jghouse88 (approved):
LGTM

PR #379, opened 2021-06-07
2 solvent verification tests for eNRTL model

andrewlee94:
## Summary/Motivation:
Adding a set of test cases for the eNRTL model for 2 solvent, 1 salt systems. This covers most of the possible binary interactions for the eNRTL model, so hopefully this completes the basic verification tests

- Adds tests for solubility of NaBr in three mixed solvent systems
-

### Legal Acknowledgement


idaes/generic_models/properties/core/eos/tests/test_enrtl_verification_2_solvent.py
```diff
@@ -0,0 +1,359 @@
...
+            ("EtOH", "H2O"): -0.18514,
+            ("MeOH", "EtOH"): -0.04394,
+            ("EtOH", "MeOH"): 0.02147,
+            ("H2O", "Na+, Br-"): 9.527,
+            ("Na+, Br-", "H2O"): -4.790,
+            ("MeOH", "Na+, Br-"): 5.910,
+            ("Na+, Br-", "MeOH"): -3.863,
+            ("EtOH", "Na+, Br-"): 6.118,
+            ("Na+, Br-", "EtOH"): -4.450}}}
+
```
adam-a-a, line 120: I assume these data are correct but wanted to do a quick double-check. Can you direct me to where I can find these values or were they digitized from a figure in the reference?

andrewlee94: All of these come from table 3 in the 2009 paper.

adam-a-a: Should `"temperature_ref": 300` be set to 298.15 instead? It seems like it won't make much of a difference, but is there a particular reason why `"temperature_ref"` is 300 instead of 298.15?

andrewlee94: For this it doesn't matter - Tref is not used in the activity coefficient calculations. It would be important if we were testing things like enthalpy, and then it should be matched to the Href being used.

adam-a-a (approved):
LGTM

Andresj89 (approved):
LGTM

PR #381, opened 2021-06-09
Adding ideal properties to eNRTL model

andrewlee94:
## Summary/Motivation:
The PR make the Ideal property package the base class for the eNRTL model so that it has access to common ideal properties. This should make the eNRTL model usable in unit models, and non-ideal properties can be added gradually as they are developed.

This PR also corrects a bug in the Ideal property package where the pressure dependent term for enthalpy of incompressible materials was not included.

- Lots
-

### Legal Acknowledgement

### DIFF
--- a/idaes/generic_models/properties/core/eos/ideal.py
+++ b/idaes/generic_models/properties/core/eos/ideal.py
@@ -44,14 +44,38 @@ class Ideal(EoSBase):
 
     @staticmethod
     def act_phase_comp(b, p, j):
-        return b.get_mole_frac()[p, j]
+        return b.mole_frac_phase_comp[p, j]
+
+    @staticmethod
+    def act_phase_comp_true(b, p, j):
+        return b.mole_frac_phase_comp_true[p, j]
+
+    @staticmethod
+    def act_phase_comp_appr(b, p, j):
+        return b.mole_frac_phase_comp_apparent[p, j]
 
     @staticmethod
     def log_act_phase_comp(b, p, j):
-        return log(b.get_mole_frac()[p, j])
+        return log(b.mole_frac_phase_comp[p, j])
+
+    @staticmethod
+    def log_act_phase_comp_true(b, p, j):
+        return log(b.mole_frac_phase_comp_true[p, j])
+
+    @staticmethod
+    def log_act_phase_comp_appr(b, p, j):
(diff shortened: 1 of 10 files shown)


Andresj89 (approved):
Tested this with a phosphorus removal example. Everything seems ok.

idaes/generic_models/properties/core/eos/ideal.py
```diff
@@ -125,8 +161,24 @@ def energy_internal_mol_phase_comp(b, p, j):
 
     @staticmethod
     def enth_mol_phase(b, p):
-        return sum(b.get_mole_frac()[p, j]*b.enth_mol_phase_comp[p, j]
-                   for j in b.components_in_phase(p))
+        pobj = b.params.get_phase(p)
+        if pobj.is_vapor_phase():
+            return sum(b.get_mole_frac()[p, j]*b.enth_mol_phase_comp[p, j]
```
adam-a-a, line 166: Just a question on naming convention: Should this be modified at some point from `enth_mol_phase_comp[p, j]` to `enth_mol_vap_comp[j]`? 

I only thought this since `enth_mol_liq_comp` and `enth_mol_sol_comp` both exist.

adam-a-a: Actually, there appears to be an `enth_mol_ig_comp` for vapor phase. I don't think this would change the code's functionality but thought you might want to change for naming consistency.

adam-a-a (approved):
I commented about how `enth_mol_phase_comp` is called for the vapor phase, but it shouldn't hold back the PR. I leave it to you to decide if it should be changed.

idaes/generic_models/properties/core/eos/ideal.py
andrewlee94: No, this is correct. It looks al little odd at first, but this is because the vapour phase does not have a pressure dependent term - thus `enth_mol_phase` is jus the sum of `enth_mol_phase_comp` across all components.

The other phases could do this too, but I thought the expressions would turn out neater if they were recalculated for the ideal properties (basically having `V*(P-Pref)` instead of `sum(x[i]*V*(P-Pref) for all i)`). Functionally it is the same, but I thought it best to avoid the unnecessary sum term.

eslickj (approved):

PR #382, opened 2021-06-10
Add check in pytest to enforce that tests have required markers

### LINKED ISSUE #170: Revisit functionality to enforce all tests to be categorized with a pytest.mark
See also IDAES/idaes-dev#1038.

Adding @dangunter and @andrewlee94 as "persons of interest".

lbianchi-lbl:
## Summary/Motivation
All tests in the pytest test suite should have at least one of the required markers (`unit`, `component`, `integration`).
This PR implements this by adding a check that runs when the tests items are collected and causes any test without a required marker to fail.

- Add function in `conftest.py` to check that test items have required markers
- Refactor existing code running during `pytest_runtest_setup` for consistency

### Legal Acknowledgement

### DIFF
--- a/idaes/conftest.py
+++ b/idaes/conftest.py
@@ -29,62 +29,71 @@ import sys
 ####
 
 ####
-# Only run the tests for your particular platform(s), if it is marked with those platform(s)
-# e.g. to mark tests for linux:
-#
-# import pytest
-# @pytest.mark.linux
-# def test_something():
-#    print("this only runs on linux")
-#
-#
-# In addition, you can use "no<platform>" to exclude
-#
-# @pytest.mark.nowin32
-# def test_something():
-#    print("this will not run on windows")
-#
-# The names of the platforms should match what is returned by `sys.platform`, in particular:
-#    Linux = 'linux'
-#    Windows = 'win32'
-#    macOS = 'darwin'
-
-
-ALL = {"darwin", "linux", "win32"}
-ALL_NO = {"no" + tag for tag in ALL}
 
 
 @pytest.hookimpl
 def pytest_runtest_setup(item):
-    for check in [
-        _validate_required_marks,
-        _skip_for_unsupported_platforms,
-    ]:
-        check(item)
+    _validate_required_markers(
+        item,
+        required_markers={"unit", "component", "integration"},
+        expected_count=1
+    )
+    _skip_for_unsupported_platforms(
+        item,
+        all_platforms={"darwin", "linux", "win32"},
+        negate_tag=(lambda tag: f'no{tag}'),
+    )
 
 
-def _skip_for_unsupported_platforms(item):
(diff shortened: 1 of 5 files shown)


andrewlee94 (approved):
Looks good to me. Even better that it is failing because some tests are missing marks.

idaes/conftest.py
```diff
@@ -29,37 +29,72 @@
...
+@pytest.hookimpl
+def pytest_runtest_setup(item):
+    _validate_required_markers(
+        item,
+        required_markers={"unit", "component", "integration"},
+        expected_count=1
+    )
+    _skip_for_unsupported_platforms(
+        item,
+        all_platforms={"darwin", "linux", "win32"},
```
dangunter, line 43: I prefer to define things like this in constants so you can change them without digging through the code.

idaes/conftest.py
```diff
@@ -29,37 +29,72 @@
...
+def _validate_required_markers(item, required_markers=None, expected_count=1):
+    required_markers = set(required_markers or [])
+    item_markers = {marker.name for marker in item.iter_markers()}
+    required_markers_on_item = item_markers & required_markers
+    required_count = len(required_markers_on_item)
+    reason_to_fail = None
+    if required_count < expected_count:
+        reason_to_fail = 'Too few required markers'
+    if required_count > expected_count:
+        reason = 'Too many required markers'
```
dangunter, line 93: I think this should be `reason_to_fail`, as well

idaes/conftest.py
```diff
@@ -29,37 +29,72 @@
...
+    if reason_to_fail:
+        extra_info = f'Expected: {expected_count} of {required_markers}; found: {item_markers}'
+        msg = f'{reason_to_fail} for test function "{item.name}". {extra_info}'
```
dangunter, line 96: Total nitpick, but appending extra_info to the string `msg` (multi-line) would be simpler than the indirection of defining a variable only to use it once, in the next line

dangunter (changes requested):
One minor bug and a comment about the constants.

idaes/conftest.py
lbianchi-lbl: Makes sense. I'll move it to a top-level module constant.

idaes/conftest.py
lbianchi-lbl: Good catch, thanks! I'll change it to `reason_to_fail`.

idaes/conftest.py
lbianchi-lbl: That's a good point, it'd look a bit cleaner. I'll change it following your suggestion.

ksbeattie (approved):
LGTM

dangunter (approved):
Looks great

andrewlee94 (approved):
Looks OK to me.

PR #385, opened 2021-06-11
Fix model serializer root name issue

eslickj:
## Fixes
Fixes an issue where the root component name of the block or module doesn't match what's in the saved state.  Places where this root name mismatch occur is when you have models with different name, but the same structure, or when you have indexed blocks and want to copy the results for one block to other blocks.  This was possible to do before by providing the root name, but now it should just work automatically.

### Legal Acknowledgement


andrewlee94 (approved):
I don't see any problems.

eslickj:
This also came up here https://github.com/IDAES/workspace/issues/151, for the record.

MAZamarripa (approved):
lgtm

PR #391, opened 2021-06-23
Visualizer fails on some flowsheets

### LINKED ISSUE #390: Bug displaying certain flowsheet components
Soraya Rawlings noticed an error when trying to display her flowsheet in the IFV.
The error ended with:
"AttributeError: 'PressureChangerData' object has no attribute 'base_class_module'".
This is causing some flowsheets to fail to display.

dangunter:
## Summary/Motivation:
IFV fails to visualize some flowsheet(s).
The problem, it turns out, was that the methods needed to get the original class name for a unit (needed to know which symbol to display for it) were only added to ProcessBlock, whereas some units in the flowsheet with the error inherited from ProcessBlockData instead. This seemed to have to do with indexed blocks, but regardless the solution is to leverage the decorator that is used on *all* process blocks.

-  Replace `base_class_name` and `base_class_module` methods on ProcessBlock class with dynamically added class methods in the `declare_process_block_class`  class decorator
- Add some fail-safe exception handling in the UI code to assign a default block type even if the provided unit operation still does not have an identifiable base class name.

### Legal Acknowledgement


dangunter:
No rush, @eslickj or @andrewlee94 , but if one of you could look at the proposed changes in `idaes.core.process_block` and let me know if you find them OK,

andrewlee94:
This is one where @eslickj will need to look at it first. He is the one who understands what is going on there and how it all works.

eslickj:
Yeah, the issue for indexed blocks is that they are ProcessBlockData objects contained in a ProcessBlock.  Non-indexed blocks are both a ProcessBlockData and a ProcessBlock.  For either a ProcessBlockData contained in an indexed block or a scalar ProcessBlock, you can use the parent_component() method to get the ProcessBlock object (for a scalar it just returns itself).  I'm still looking at this, and I don't know exactly what the problem is, but ``parent_component()`` could be another solution.

eslickj:
I can't quite understand how the PR fixes the problem, but maybe I don't quite understand the problem.  I'll go out on a limb and suggest this:

```python
    def _get_unit_model_type(self, unit):
        # Get the unit models type
        return unit.parent_component().base_class_module().split(".")[
            -1
        ]  # TODO look for/create equivalent getter, as getname() above
```
Adding parent component should get the ProcessData class you need for the indexed and scalar cases.

dangunter:
I ended up putting the method into the indexed and scalar metaclasses and that works.

eslickj (dismissed):

dangunter:
Sorry @eslickj and @andrewlee94, I fixed a bug or two and that invalidated your reviews. When you get a chance, could you look again? No changes in core since last time (just in idaes.ui.*)

andrewlee94 (approved):
I am told this is good.

eslickj (approved):

PR #394, opened 2021-06-24
Correcting unsymmetric reference state to be near-zero instead of zero

andrewlee94:
## Summary/Motivation:
The Unsymmetric reference state for the eNRTL model is based on infinite-dilution of ionic species. Previously this was represented using zero mole fractions for ionic species, but this resulted in issues with raising 0 to fractional and negative powers.

- Changes the mole fractions of ionic species at the unsymmetric reference state to an EPS value rather than zero.
-

### Legal Acknowledgement


adam-a-a (approved):

Andresj89 (approved):
LGTM

PR #395, opened 2021-06-24
Fix errors due to missing json.gz and dat files when installing IDAES not in editable mode

### LINKED ISSUE #383: Missing json.gz files cause errors and test failures
(Originally reported by @jghouse88 and @eslickj)

- Following #355, several JSON files that are part of the IDAES codebase had their extension changed from `.json` to `.json.gz`
- This might have caused the files to be skipped when installing the IDAES package, in turn causing the errors and test failures in the code that depends on the files being present
- Assuming this is caused by an issue in `setuptools`/`setup.py` configuration, it should occur only if `idaes-pse` is installed not in editable mode
  - In other words, `pip install idaes-pse` will be affected by this issue, but `pip install -e .` or `pip install -r requirements-dev.txt` from a local clone will not
  - This is also the likely reason why this slipped past our tests, since most of the checks use the latter method to install IDAES.
- The fix should be straightforward (specify the correct extension in `setup.py`)
- We should also add a check that would be able to catch similar issues in the future, e.g. installing and running IDAES not in editable mode

lbianchi-lbl:
## Summary/Motivation

- `package_data` patterns need to be specified for any non-Python file that's needed at runtime
- Recent changes from `json` to `json.gz` and introducing `.dat` files caused some tests to break. To wit, these are:
  - `idaes/power_generation/flowsheets/test/test_NGFC.py`
  - `python -c 'from idaes.power_generation.flowsheets.NGFC.NGFC_flowsheet import main; main()'`
- Packaging issues are often hidden by the fact that most of our testing (both individually and in the CI checks) installs IDAES in developer mode (`pip install --editable`), while `package_data` is relevant within IDAES when using `pip install .`

## Changes proposed in this PR

- Add `package_data` glob patterns for gzipped JSON and DAT files

## For reviewers

A proper check to detect such issues going forward is in the making. In the meanwhile, in the interest of getting this out quickly, run these steps to check:

```sh
cd /tmp
conda create -n idaes-pse-tmp python=3.8 -y && conda activate idaes-pse-tmp
# clone the repo to get the tests, but install IDAES from PyPI to test the current release
git clone https://github.com/IDAES/idaes-pse idaes-pse-just-for-tests && cd idaes-pse-just-for-tests
pip install idaes-pse
pip install pytest addheader
idaes get-extensions
# run these two tests: they should fail
pytest idaes/power_generation/flowsheets/test/test_NGFC.py
python -c 'from idaes.power_generation.flowsheets.NGFC.NGFC_flowsheet import main; main()'
# as an extra check, look at the files being installed in the Conda env
find $CONDA_PREFIX/lib/python3.8/site-packages/idaes -type f -name \*.json.gz
# now reinstall IDAES from this PR branch
pip install 'idaes-pse @ https://github.com/lbianchi-lbl/idaes-pse/archive/383-package-data.zip'
# run the tests again: they should now succeed
pytest idaes-pse-just-for-tests/power_generation/flowsheets/test/test_NGFC.py
python -c 'from idaes.power_generation.flowsheets.NGFC.NGFC_flowsheet import main; main()'
# as an extra check, look at the files being installed in the Conda env
find $CONDA_PREFIX/lib/python3.8/site-packages/idaes -type f -name \*.json.gz
```

### Legal Acknowledgement

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -98,8 +98,8 @@ kwargs = dict(
             "*.js",
             "*.css",
             "*.html",
-            ".json.gz",
-            ".dat"
+            "*.json.gz",
+            "*.dat"
         ]
     },
     include_package_data=True,
--- a/setup.py
+++ b/setup.py
@@ -98,6 +98,8 @@ kwargs = dict(
             "*.js",
             "*.css",
             "*.html",
+            ".json.gz",
+            ".dat"
         ]
     },
     include_package_data=True,


ksbeattie (approved):
yes

setup.py
```diff
@@ -98,6 +98,8 @@ def rglob(path, glob):
             "*.js",
             "*.css",
             "*.html",
+            ".json.gz",
```
ksbeattie, line 101:
```
*.json.gz
*.dat
```

ksbeattie (approved):
yes?

PR #396, opened 2021-06-26
Better import tests

### LINKED ISSUE #332: Uncertainty example runs code on import
One of the examples for the uncertainty propagation tools runs code on import:

idaes/apps/uncertainty_propagation/examples/simple_opt_problem.py

This causes the IDAES imports test (idaes/tests/test_import.py) to fail.

Additionally, some of the test for uncertainty propagation lack pytest marks.

dangunter:
## Summary/Motivation:
We are not testing too well whether modules run and/or do funky things on import

- Add test for whether import fails or runs for too long (for every module)
- Add test for whether there are "too many" lines in the top-level of files, i.e. trying to test directly for #332

### Legal Acknowledgement


dangunter:
I think the new `test_import2.py` should wholesale replace `test_import.py`. The mechanism is more robust and could actually be faster if one of the modules runs something for a long time on import (which is exactly the reason we're testing to begin with, so it's not like it's impossible), though in the normal case it is quite a bit slower (2-3 min. on my laptop).

andrewlee94:
I'll let those with more experience comment on the details of the tests, but the results look good. The failures are all due to tests where code that really should have been in fixtures was put as module level code.

However, I note that the listed failures are all test files - i.e. we don't see any actual code modules failing (such as those in #332). I don't see why it wouldn't be, but we should make sure this is covering all the modules in the repo.

idaes/tests/test_import.py
```diff
@@ -11,81 +11,285 @@
...
+    Since most of the overhead of running this is in starting the interpreter,
+    it is much faster to do, e.g., 10 imports at a time -- as long as failure is
+    rare. The drawback is a little extra complexity if there is an error in an
+    import, since one needs to go back through the batch and see which module
+    actually caused the failure. Note that if failure is *not* rare, we have bigger
+    problems than whether this test is slow.
+    """
+    # Try to import everything in the batch, hope for success
+    batch_max = IMPORT_TIME_MAX * len(module_names) + 1
+    t0 = time.time()
```
jsiirola, line 88: `time()` is not the most accurate timer.  I would recommend `time.perf_counter()`

idaes/tests/test_import.py
```diff
@@ -11,81 +11,285 @@
...
+            errlist = [
+                f"Import error for {module_name}:\n{result.error}"
+                for module_name, result in erred.items()
+            ]
+            pytest.fail("\n".join(errlist))
+
+
+class Result:
+    def __init__(self):
+        self.timing, self.error = -1.0, None
```
jsiirola, line 121: I am generally not a fan of multiple unrelated assignments in a single line.

idaes/tests/test_import.py
```diff
@@ -11,81 +11,285 @@
...
+        # For rows with timings
+        if line.startswith("import time:"):
+            # Save if match with a module in the batch
+            for module in modules:
+                if line.endswith(module):
+                    times = re.findall(r"\d+", line)
+                    results[module].timing = float(times[1]) / 1e6
+                    break
+        elif not allow_output:
+            return None
```
jsiirola, line 171: This is duplicated below in `collect_import_time`; I would recommend pulling out into a function.

idaes/tests/test_import.py
```diff
@@ -11,81 +11,285 @@
...
+    for line in f:
+        lineno += 1
+        is_statement = False
+        # if in a comment block, wait for it to end
+        if comment_block:
+            if re.match(r".*('''|\"\"\")", line):
+                comment_block = False
+        # look for start of comment block
+        elif re.match(r"\s*('''|\"\"\")", line):
+            comment_block = True
```
jsiirola, line 274: This will get confused by:
```python
def foo():
    """"a docstring"""
```

Also, wrt vernacular, `"""` is not a comment block, but a multiline string.

idaes/tests/test_import.py
```diff
@@ -11,81 +11,285 @@
...
+            pass
+        # skip function, class defs or decorators
+        elif line.startswith("def") or line.startswith("class") or line.startswith("@"):
+            pass
+        # match anything that looks like a function call, except some special ones
+        elif re.match(r".*\w+\(", line):
+            is_statement = True
+            # these things *are* allowed at the top-level
+            if re.match(
+                ".*(add_command|CONFIG|list\(|dict\(|attempt_import|\.declare\()", line
```
jsiirola, line 286: The RE is missing the raw string indicator (`r"`)

You should also probably skip "import".

You also might consider "guarding" these patterns with `\b`:
```python
if re.match(
    r".*\b(add_command|CONFIG|list|dict|attempt_import|declare)\b", line
    ):
```

idaes/tests/test_import.py
```diff
@@ -11,81 +11,285 @@
...
-SKIP, OK, BAD = "skipped", "success", "failed"
+__author__ = "Dan Gunter (LBNL)"
 
 
-_term = ColorTerm()
+# The root of the source code tree
+idaes_path = Path(idaes.__file__).parent
 
+# Constants controlling the expected behavior
+IMPORT_TIME_MAX = 5     # Seconds to import a file
```
jsiirola, line 47: 5 seconds seems awful long...

jsiirola (changes requested):
This is really nice.  In addition to the comments below, I would recommend making this file "runable": that is, add the `if __name__ == '__main__':` that runs the timing tests and outputs all the results.  I did that with Pyomo's import tester and that has proven invaluable over time.

idaes/tests/test_import.py
```diff
@@ -11,81 +11,285 @@
...
 
 
 @pytest.mark.unit
-def test_import():
-    root_dir = pathlib.Path(idaes.__file__).parent
-    failures, total = importr(root_dir)
-    n = len(failures)
-    assert n == 0, f"{n:d} failures in {total:d} tests: {failures}"
+@pytest.mark.parametrize("module_path", list(idaes_path.rglob("*.py")))
+def test_toplevel_statements(module_path):
```
lbianchi-lbl, line 241: This is more of a general suggestion rather than a comment of this test as it is currently implemented in this PR: should we consider using pylint to perform this check, instead? I can see the following reasons:

- Robustness: I'm personally very wary when it comes to parsing anything, and doubly so if it's source code. The current implementation looks perfectly fine, but I can (conservatively/pessimistically) imagine that there would be edge cases that anything short of a full-blown Python parser wouldn't catch, and that could cause headaches down the line once this is part of the IDAES test suite. Essentially, the functionality here is to extract and count the number of statements in a Python file, which is something that it would IMO be safer and more robust to do at the Abstract Syntax Tree (level). The Python stdlib has an `ast` module, which is the same base of the (`astroid`) module used by pylint
- The way to implement this in pylint would be to create a custom checker. The process is [surprisingly well-documented](http://pylint.pycqa.org/en/latest/how_tos/custom_checkers.html) and, considering we already have all the infrastructure in place for custom Pylint code, it looks like it would be straightforward to do (I'd be happy to take a stab at this)
- Conceptually, having the checks on IDAES modules divided between pytest looking at the *import times* (runtime analysis) and pylint looking at the *code within the module* (static analysis) would better fit the current runtime/static analysis separation of concerns between pytest and pylint which, all else being equal, I think it would make sense to maintain
- As a bonus consideration, building off of a dedicated third-party tool such as pylint (as opposed to adding the module-parsing functionality, which is not specific to IDAES, somewhere within the IDAES test suite) would be more efficient in term of development (and especially maintenance) efforts

Thoughts on this?

lbianchi-lbl (changes requested):
In general, I like this: the import time check is a useful one to have, and the new functionality looks more robust. I can also see the potential of building on these features for more advanced performance checks down the line, which I think would be cool to have at some point. I've added some questions about potential corner cases and a suggestion (ditto-ed from @jsiirola) to refactor some code to a function.

For the check on the import-level statement, I don't see anything wrong with the proposed solution, but I've added a few comments on why I think we should consider using pylint to perform this check instead. This is more in the spirit of a discussion, so please consider it definitely non-binding in terms of this PR.

idaes/tests/test_import.py
dangunter: I'm sure you're right, but does a few milliseconds matter here?

idaes/tests/test_import.py
dangunter: I mean, yes, but who would do that? I guess it could be more robust though.

idaes/tests/test_import.py
dangunter: import is skipped because it doesn't look like a function. but ok on the other stuff.

dangunter:
@lbianchi-lbl  I think you have enough ideas on how to implement this  that you should just run with it, I'll stop doing anything and leave the rest of the PR to you.

idaes/tests/test_import.py
dangunter: It's not exactly duplicated, but kinda sorta duplicated. It felt like abstracting might actually be more rather than less confusing. Maybe when @lbianchi-lbl takes out the batching in favor of `.communicate` this becomes moot anyways.

idaes/tests/test_import.py
jsiirola: No - probably not (at least not when the threshold is 5 seconds - although that should probably be significantly reduced, but that is a different matter).

That said, timing in Python is noisy enough as it is, and `time()` can behave significant;y differently on different platforms.  In my experience, `perf_counter()` seems to have more stable behavior.

idaes/tests/test_import.py
jsiirola: Oh, about 144 files in IDAES, including
- `./idaes/__init__.py`
- `./idaes/dmf/magics.py`
- ...

jsiirola: I think the following would be more robust:
```python
comment_block ^= bool(len(re.findall(r".*('''|\"\"\")", line)) % 2)
if comment_block:
    continue
```
...it will still get confused by blockquotes that contain blockquotes (i.e., putting `'''` inside a blockquote delimited by `"""`)... but the logic to catch that starts to get rather challenging.

andrewlee94:
@dangunter Is this PR still in progress? It doesn't look like it has been updated in almost a year.

lbianchi-lbl:
> @dangunter Is this PR still in progress? It doesn't look like it has been updated in almost a year.

@andrewlee94 The current plan is to implement this kind of checks using a general-purpose tool that I've been developing. The tool itself is mostly complete but testing it and deploying it with IDAES has been pushed to the backlog because of lack of time. I propose we close this and open a new PR when the time comes.

PR #399, opened 2021-06-28
Making time a property of flowsheets

andrewlee94:
## Summary/Motivation:
Currently, there is some inconsistency in how the time domain is handled between top level and sub-flowsheets. Notably, `flowsheet.time` works for top-level flowsheets, but `flowsheet.config.time` is required for sub-flowsheets. This PR makes `time` a property of all `FlowsheetBlocks` so that `flowsheet.time` will work for all flowsheets.

- Make `time` a property of `FlowsheetBlocks`
- Store the time domain or a reference to in in `_time`
- Update `_setup_dynamics` to define the `_time` attribute.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/flowsheet_model.py
+++ b/idaes/core/flowsheet_model.py
@@ -26,6 +26,7 @@ from idaes.core import (ProcessBlockData, declare_process_block_class,
 from idaes.core.util.config import (is_physical_parameter_block,
                                     is_time_domain,
                                     list_of_floats)
+from idaes.core.util.misc import add_object_reference
 from idaes.core.util.exceptions import DynamicError, ConfigurationError
 from idaes.core.util.tables import create_stream_table_dataframe
 from idaes.ui.fsvis.fsvis import visualize
@@ -119,6 +120,11 @@ within this flowsheet if not otherwise specified,
 
         # Set up dynamic flag and time domain
         self._setup_dynamics()
+
+    @property
+    def time(self):
+        # _time will be created by the _setup_dynamics method
+        return self._time
 
     @property
     def time_units(self):
@@ -280,6 +286,7 @@ within this flowsheet if not otherwise specified,
                 raise DynamicError(
                         '{} was set as a dynamic flowsheet, but time domain '
                         'provided was not a ContinuousSet.'.format(self.name))
+            add_object_reference(self, "_time", self.config.time)
             self._time_units = self.config.time_units
         else:
             # If no parent flowsheet, set up time domain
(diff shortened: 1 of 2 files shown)


andrewlee94:
I just had a look at how many times `config.time` appears in the code base, and there are lots. To keep this PR manageable, I'll leave it as just creating the property, and we can change all the references later.

eslickj (approved):
Looks good to me.

jsiirola (approved):
This looks reasonable, but out of curiosity, why are you duplicating time as `._time` instead of just returning `fs.config.time` in the `time` property getter?

andrewlee94:
@jsiirola I did consider doing that and almost went down that path. However, I did see a few minor advantages to doing it this way:

1. Commonality between top-level and nested flowsheets (they both have `_time`).
2. To prevent user mistakes - a config blocks are fairly public so a user might go and change the config block at some point which could result in a mismatch. `_time` is explicitly private and somewhat hidden so there is less change of messing things up.

jsiirola:
>     2. To prevent user mistakes - a config blocks are fairly public so a user might go and change the config block at some point which could result in a mismatch. `_time` is explicitly private and somewhat hidden so there is less change of messing things up.

This is exactly my concern (in reverse): I never like storing the same info in two places, as that is begging for them to drift out of sync.

...An alternative is to leverage the `MarkImmutable` context manager that @michaelbynum added to ConfigValues a while back to "freeze" key Config values (after the FS was built?).

andrewlee94:
I do agree about avoid storing the same information in two places. That actually raises the question of why we are storing the time domain in the config block and not as a reference in the first place (the reason being to get around Pyomo's single owner restriction, but `add_object_reference` does that).

On the topic of `MarkImmutable`, should we be marking our Config blocks as immutable by default?

jsiirola:
I don't think we need to overuse `MarkImmutable`.

andrewlee94:
@jsiirola I have had a look at how we could switch to using `_time` in most cases and avoid storing the time domain in the config block (which I was never happy with). However, this breaks backwards compatibility, as we generally have looked to the the config block for the time domain. Whilst it is an easy fix, we probably don't want to go down this path to avoid breaking user's models.

So, the alternative then is to use the config block to hold the time reference, and only have `_time` in the top-level flowsheet. Not my preferred solution as I would like the time reference to be a private attribute, and I dislike have code change the config blocks (I prefer them to retain the original user values). However this maintains backwards compatibility and we could look at marking the time reference as Immutable to avoid user mistakes.

andrewlee94:
Also, I would note that under the implementation currently in this PR, the `_time` attribute would be considered the canonical place to store the time domain, with the config block being a legacy storage for backwards compatibility in most cases.

PR #404, opened 2021-06-30
Update all references to time domain

andrewlee94:
## Summary/Motivation:
Now that `time` is a property of flowsheets, this PR updates all references to the time domain in the code base to use the property instead of the config block.

- Replace all instances of `config.time` with `.time`
-

### Legal Acknowledgement

### DIFF
--- a/idaes/power_generation/unit_models/waterpipe.py
+++ b/idaes/power_generation/unit_models/waterpipe.py
@@ -217,7 +217,7 @@ mixed phase not supported'''))
                 doc="Cross section area ratio of exit end to pipe")
 
         # Volume constraint
-        @self.Constraint(self.flowsheet().config.time,
+        @self.Constraint(self.flowsheet().time,
                          doc="Total volume of all pipes")
         def volume_eqn(b, t):
             return b.volume[t] == 0.25 * const.pi * b.diameter**2 \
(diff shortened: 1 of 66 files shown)


idaes/power_generation/unit_models/waterpipe.py
```diff
@@ -391,7 +391,7 @@ def initialize(blk, state_args=None, outlvl=idaeslog.NOTSET,
             )
 
         # Unfix outlet enthalpy and pressure
-        for t in blk.flowsheet().config.time:
+        for t in blk.flowsheet().time:
             blk.control_volume.properties_out[t].pressure.unfix()
```
jsiirola, line 395: This idiom shows up a lot, and could potentially be made simpler (in a future PR) with just
```python
blk.control_volume.properties_out[:].pressure.unfix()
```

jsiirola (approved):

idaes/power_generation/unit_models/waterpipe.py
andrewlee94: @jsiirola I didn't realize that would actually work for fixing and unfixing. However, that applies mostly to moved developers rather than core developers, so getting them to adopt that will take some work.

jsiirola: It actually works for any method (as well as property setters), and expands to effectively be the loop you would have written.  It is potentially slightly slower, as it uses an `IndexedComponent_slice` iterator, which has some overhead relative to the simple loop.  BUT, the syntax is pretty.  ;)

eslickj (approved):

PR #408, opened 2021-06-30
Pin pylint version used in CI checks to avoid unexpected check failures

### LINKED ISSUE #407: pylint PR checks failing for previously passing parts of the codebase
Since yesterday, @jghouse88 and @andrewlee94 have reported failures in the pylint check for their PRs. The reported errors are unrelated to the changes in the PR, and instead originate from parts of the codebase that were previously passing pylint's checks.

I've looked into this and I believe this is due to a pylint update to version 2.9.1, which seems to have introduced regressions with respect to the previous version (2.8.3). Comparing the latest passing pylint check and the earliest failing one for the same PR by looking at the installed pylint version (using `pip show pylint` if running locally, or looking at the `pip install` output in the GitHub Actions workflow logs) seems to agree with this hypothesis.

I have a PR with a fix ready, but I wanted to open an issue to leave a reference in case someone encounters the same or related issues.

lbianchi-lbl:
## Summary/Motivation
- From experience in the past months, pylint checks are highly sensitive to changes made in pylint
- Having a pinned version allows us IDAES maintainers to review and possibly update it *in a proactive way*, as opposed to having to be *reactive* when a pylint update is released
- Conceptually, as we're using pylint to check for changes *within IDAES*, it makes sense to treat the pylint version as a "test constant"

- Pin pylint version in `requirements-dev.txt`
- Add version compatibility check to IDAES transform plugin
  - The checks compares the pylint version used at runtime with a "reference version" defined in the plugin, and displays a warning if the two don't match (specifying that this is not necessarily a problem to reduce confusion/"warning fatigue" for the IDAES developer running the check locally)

### Legal Acknowledgement

### DIFF
--- a/.github/actions/pylint/action.yml
+++ b/.github/actions/pylint/action.yml
@@ -10,11 +10,13 @@ runs:
         PIP_INSTALL="pip --no-cache-dir install --progress-bar off"
         echo "::group::Output of pip install commands"
         $PIP_INSTALL --upgrade pip setuptools wheel
-        # TODO the pylint version will have to be pinned
         # TODO installing idaes is necessary in our case because we import some idaes code in pylint plugins
+        # don't think we need to install the extensions, though
         $PIP_INSTALL -r requirements-dev.txt
         echo "::endgroup::"
-        # don't think we need to install the extensions, though
+        echo "::group::Display pylint version"
+        pip show pylint astroid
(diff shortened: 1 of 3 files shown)


andrewlee94 (approved):
Looks good

jghouse88 (approved):
LGTM

PR #410, opened 2021-07-01
Stop using deprecated ConfigEnum class in Caprese

Robbybp:
## Summary/Motivation:
Was using an Enum subclass for config arguments that has been deprecated. Have removed this class, replacing it with enum.Enum to get rid of deprecation warnings.

### Legal Acknowledgement

### DIFF
--- a/idaes/apps/caprese/common/config.py
+++ b/idaes/apps/caprese/common/config.py
@@ -17,43 +17,42 @@ A module of functions and classes for configuring NMPC/MHE problems
 import enum
 from pyomo.environ import SolverFactory
 from pyomo.core.base.var import _GeneralVarData
-from pyomo.common.config import ConfigEnum
 
 
-class ControlInitOption(ConfigEnum):
+class ControlInitOption(enum.Enum):
     FROM_PREVIOUS = 11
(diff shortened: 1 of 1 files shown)


andrewlee94 (approved):
Looks good.

jghouse88 (approved):
LGTM

dangunter:
There seems to be a desire to keep the integer values unique, but there are two classes that use the 20's. Is this intentional? 

A more compact way of doing all this that also guaranteed unique ranges would be to use a base number and invoke Enum as a callable with the 'start' parameter, which you could wrap up into a little function, e.g.:
```
from enum import Enum

_base = 1

def make_enum(name, values):
    global _base
    e = Enum(name, values, start=_base, module=__name__)
    _base += 10
    return e

ControlInitOption = make_enum("ControlInitOption",
    "FROM_PREVIOUS, BY_TIME_ELEMENT, FROM_INITIAL_CONDITIONS, SETPOINT")

ElementInitializationInputOption = make_enum("ElementInitializationInputOption",
    "SETPOINT, INITIAL, CURRENT_VALUES")
# etc.
```

PR #415, opened 2021-07-06
Improvements to Properties Interrogator

andrewlee94:
## Summary/Motivation:
A couple of short comings have been identified in the properties interrogator tools:

1. Tools do not work with Unit Models which define specific phase lists that are not ["Liq", "Vap"]
2. Interrogator tools need to be updated to use the Pyomo UnitsContainer

- Updates units in interrogator tools
- Adds ability for users to (manually) define custom phase and components sets

### Legal Acknowledgement

### DIFF
--- a/docs/technical_specs/model_libraries/generic/property_models/interrogator.rst
+++ b/docs/technical_specs/model_libraries/generic/property_models/interrogator.rst
@@ -18,7 +18,7 @@ An example of how Property Interrogator tool is used is shown below:
     from idaes.generic_models.properties.interrogator import PropertyInterrogatorBlock, ReactionInterrogatorBlock
 
     m = pyo.ConcreteModel()
-    m.fs = FlowsheetBlock(default={"dynamic": True})
+    m.fs = FlowsheetBlock(default={"dynamic": True, "time_units": pyo.units.s})
 
     m.fs.params = PropertyInterrogatorBlock()
     m.fs.rxn_params = ReactionInterrogatorBlock(
--- a/idaes/core/components.py
+++ b/idaes/core/components.py
@@ -233,7 +233,7 @@ class ComponentData(ProcessBlockData):
 
 # TODO : What about LLE systems where a species is a solvent in one liquid
 # phase, but a solute in another?
-@declare_process_block_class("Solute")
+@declare_process_block_class("Solute", block_class=Component)
 class SoluteData(ComponentData):
     """
     Component type for species which should be considered as solutes in
@@ -262,7 +262,7 @@ class SoluteData(ComponentData):
 
 # TODO : What about LLE systems where a species is a solvent in one liquid
 # phase, but a solute in another?
-@declare_process_block_class("Solvent")
+@declare_process_block_class("Solvent", block_class=Component)
 class SolventData(ComponentData):
     """
     Component type for species which should be considered as solvents in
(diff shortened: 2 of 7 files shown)


TimBartholomew (approved):
LGTM

adam-a-a:
I tried using the tool for the RO model again, but the error that I was getting in the aforementioned [issue](https://github.com/nawi-hub/proteuslib/issues/153) remains. 

The RO model gets its units from a property model with the following line: 
`        units_meta = self.config.property_package.get_metadata().get_derived_units
`
From there, units for each variable are defined. The first error occurs due to the line that specifies units in the following snippet:
```
        self.A_comp = Var(
            self.flowsheet().config.time,
            self.solvent_list,
            initialize=1e-12,
            bounds=(1e-18, 1e-6),
            domain=NonNegativeReals,
            units=units_meta('length')*units_meta('pressure')**-1*units_meta('time')**-1,
            doc="""Solvent permeability coeff.""")
```
If I change the way units are called from the above to `units=pyunits.m * pyunits.pa**-1 * pyunits.s`, I get past the error (which then reoccurs when the units are defined for the next variable). I'm not sure why this is the case.

adam-a-a:
Also, it seems like defining the phase_list and component_list myself works partially (at least for the RO model). That is, due to the way the config checks in the RO model were written, a ConfigurationError is raised from the following, which checks whether components are solutes or solvents:
```
          for c in self.config.property_package.component_list:
            comp = self.config.property_package.get_component(c)
            try:
                if comp.is_solvent():
                    self.solvent_list.add(c)
                if comp.is_solute():
                    self.solute_list.add(c)
            except TypeError:
                raise ConfigurationError("RO model only supports one solvent and one or more solutes,"
                                         "the provided property package has specified a component '{}' "
                                         "that is not a solvent or solute".format(c))
```

adam-a-a:
I think your changes look good, so if you want to save any remaining issues for a future PR, I can approve.

adam-a-a:
I think your changes look good, so if you want to save any of the aforementioned issues for a future PR, I can approve.

andrewlee94:
@adam-a-a I am unable to reproduce your issue with the units - I just checked the units metadata for the interrogator and it returns the expected values.

For the Component type issue, we can fix that too, but this is starting to get to the point where the user already needs to be familiar with the unit model anyway.

adam-a-a:
@andrewlee94 I think that we might want to try running the Interrogator on every unit model we have to (1) identify any other unit models that are currently incompatible with the tool and (2) assess why they fail. From there, maybe there'd be common issues between unit models that could be mended at once, while other issues might need a bit more tailoring of the Interrogator tool (e.g., for the RO model). Or maybe the RO model is the only problematic model and we should tweak the unit model itself to help with compatibility. 

Either way, I think all of that would be a big job and is probably beyond the scope of this PR.

andrewlee94:
@adam-a-a I would like to make sure the units issue is resolved before merging this PR. I can see no reason why it shouldn't work, so I want to make sure we understand what the problem is.

As for the case of unit models with specific requirements (such as the RO model), then that is an issue for each model on a case by case basis as you mentioned.

adam-a-a:
The issue with the units seems to have been a fluke. I started a new terminal today and no longer get the error.

adam-a-a (approved):

andrewlee94:
@adam-a-a You should no be able to set phase and component types with custom lists. The format is now a dict `{name: Type}`.

adam-a-a (approved):
I tried using the tool with the RO model again and it worked like a charm.

TimBartholomew (approved):
LGTM

PR #416, opened 2021-07-08
Test property constraints in CLC oxidation property package

Robbybp:
## Summary/Motivation:
Property calculation equations are more or less functions that calculate property variables based on state variables, but in the CLC oxidation property package, only the construction of these constraints was tested, not the values resulting from solves. This PR attempts to remedy this by solving the property package state blocks for a small range of state variables to make sure reasonable values of property variables are obtained, and that these values don't change.

- Add tests to gas, solid, and reaction property package tests

Note that this relies on recent Pyomo PRs, including the `pyomo.util.subsystems` module and the addition of `solve_strongly_connected_components` to the `pyomo.contrib.incidence_analysis` sub-package.

### Legal Acknowledgement


idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/gas_phase_thermo.py
```diff
@@ -489,8 +490,12 @@ def _dens_mol(self):
                             units=pyunits.mol/pyunits.m**3)
 
         def ideal_gas(b):
-            return (b.dens_mol*b._params.gas_const*b.temperature*1e-2 ==
-                    b.pressure)
+            return (
+                    b.dens_mol*
+                    b._params.gas_const*(1e3*pyunits.J/pyunits.kJ)
```
andrewlee94, line 495: You should use the IDAES Constants class for this, along with a unit conversion if required.

`from idaes.core.util.constants import Constants`

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/gas_phase_thermo.py
```diff
@@ -194,6 +194,7 @@ def build(self):
 
         # Component diffusion volumes:
         # Ref: (1) Prop gas & liquids (2) Fuller et al. IECR, 58(5), 19, 1966
+        # TODO: What are the units of these parameters?
```
chineduobiora, line 197: The diffusion volumes are dimensionless

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/gas_phase_thermo.py
```diff
@@ -576,18 +581,21 @@ def _diffusion_comp(self):
...
                                   domain=Reals,
-                                  initialize=1e-5,
+                                  initialize=1.0,
                                   doc='Component diffusion in a gas mixture'
                                   '[cm2/s]',
                                   units=pyunits.cm**2/pyunits.s)
 
         def D_bin(i, j):
             # 1e3 used to multiply MW to convert from kg/mol to kg/kmol
+            # TODO: Does this factor of 1.43e-3 have units?
```
chineduobiora, line 591: The equation is empirical. I've tried to work out the units and it looks nasty. Something like **kg**^2.5 **s**^-5 **K**^-1.75 **kgmol**^-0.5

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/tests/test_CLC_gas_prop.py
```diff
@@ -111,3 +128,851 @@ def test_solution(gas_prop):
...
+        # Construct dens_mol and all prerequisites
+        state.dens_mol
+
+        param_sweeper = ParamSweeper(n_scen, state_values,
+                output_values=target_values)
+        with param_sweeper:
+            for inputs, target in param_sweeper:
+                solve_strongly_connected_components(state)
+
+                # Check that state block equations have been convered
```
chineduobiora, line 409: Typo "converged"

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/tests/test_CLC_gas_prop.py
```diff
@@ -111,3 +128,851 @@ def test_solution(gas_prop):
...
+                    9.248259058533336*u,
+                    19.241674269713887*u,
+                    29.760175857866656*u,
+                    ],
+                "enth_mol_comp[N2]": [
+                    28.1081423656*u,
+                    28.1081423656*u,
+                    28.1081423656*u,
+                    28.1081423656*u,
+                    -0.021818752399997976*u,
```
chineduobiora, line 910: This value looks off. Enthalpy shouldn't be negative.

chineduobiora (changes requested):
I made a few minor comments but looks good in general.

lbianchi-lbl:
This depends on very recent Pyomo changes and will need to have those merged in before the tests can pass.

ksbeattie:
I've resolve a conflict here, in setup.py about which pyomo version to use, so the tests should start looking better, but this PR is on both the Aug and Nov release boards.  Which release is this expected to go out with?

andrewlee94:
@ksbeattie This is for August - I updated the project boards.

andrewlee94:
@Robbybp @ksbeattie I meant to write *not* for August.

ksbeattie:
@Robbybp, any news on this?

Robbybp:
@ksbeattie This is waiting for me to have a little time to figure out why enthalpy is negative in one of the gas phase property calculations. Because this PR is just shoring up some tests, and not adding any functionality, it is not a huge priority.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/tests/test_CLC_gas_prop.py
Robbybp: I believe enthalpy is negative here because coefficients are used that are valid for 500 K - 2000 K, per NIST webbook: https://webbook.nist.gov/cgi/cbook.cgi?ID=C7727379&Mask=1#Thermo-Gas. However, I am testing them at 300 K as gas inlet into the reactor is near room temperature. I'm not sure what to do about this. The right answer seems to be to code enthalpy as a piecewise function where parameters are only used in temperature ranges where they are valid. I left a comment to this effect in the property package file, but have not made any further changes.

Robbybp: Note that I am testing enthalpies also at 600 K, 900 K, and 1200 K, not _just_ at 300 K.

Robbybp:
@andrewlee94 @chineduobiora This is ready for re-review. I believe I addressed all minor comments, and see my comment above on why enthalpy is negative.

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/tests/test_CLC_gas_prop.py
andrewlee94: This is probably due to a detail of the Shomate equation that is generally overlooked (and it looks like it was missed here too). Note that the base form of the Shomate equation gives `H - Href` and not `H`, and that it is based on a piecewise function (the `F` term is the integral of `cp` across all temperatures from Tref to the lower limit of the given Shomate function).

So, the first thing we find here is that as 300K is almost equal to Tref (298.15), H should be almost 0 (but not quite). However, due to accumulated errors in the integral from Tref to 500K, we actually see a negative number here.

Secondly, we really should not be including the `H` term in the Shomate expression, as that is subtracting the reference enthalpy from the equation (although for N2 that doesn't matter as `H` is 0 for an elemental gas). This will matter for other species however.

Robbybp: > So, the first thing we find here is that as 300K is almost equal to Tref (298.15), H should be almost 0 (but not quite). However, due to accumulated errors in the integral from Tref to 500K, we actually see a negative number here.

Is the problem accumulated errors in the integral or just the fact that the particular Shomate equation we use is not meant to predict enthalpy at 300 K, and therefore not required/guaranteed to give a reasonable result there?

> Secondly, we really should not be including the `H` term in the Shomate expression, as that is subtracting the reference enthalpy from the equation (although for N2 that doesn't matter as the `H` term is 0 for an elemental gas). This will matter for other species however.

Isn't this okay? We want to calculate enthalpy with respect to reference enthalpy, right? We can't do anything else, right?

ksbeattie:
@chineduobiora, can you re-review?

idaes/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/tests/test_CLC_gas_prop.py
chineduobiora: > 

Okay that makes sense. Should we then be writing the test at a higher temperature since we know that inside the reactor the temperature is always higher that 500 K?

chineduobiora: <I believe enthalpy is negative here because coefficients are used that are valid for 500 K - 2000 K, per NIST webbook: https://webbook.nist.gov/cgi/cbook.cgi?ID=C7727379&Mask=1#Thermo-Gas. However, I am testing them at 300 K as gas inlet into the reactor is near room temperature. I'm not sure what to do about this>
 
Okay that makes sense. Should we then be writing the test at a higher temperature since we know that inside the reactor the temperature is always higher that 500 K?

chineduobiora: I had to take some time to think through this and review the literature and my past notes. There are two main ways to do energy balances on reactive processes (reference: Chapter 10, section 5 of elemental principles of chemical processes by Felder and Rousseau). Both of them differ based on what the reference conditions are. Back when I was setting up the models in the old IDAES framework I recall spending a lot of time comparing the two methods, and validating that I would get the same results from both of them. I have summarized the details below:

Method 1: reference conditions are the elements that constitute the reactants and products at the reference temperature (usually 25 C).
a. Enthalpy of a stream is the sum of the heat of formation of the stream components at the reference temperature (usually 25 C) and any sensible and latent heats required to bring the species from the reference temperature to its inlet or outlet state.
b. In this method, the heat of reaction is not included in the energy balance as it is implicitly calculated when the heats of formation of the reactants are subtracted from those of the products.
c. This is equivalent to not including the ‘H’ term in the shomate equation for enthalpy when calculating the stream enthalpy.
d. Energy balance of the control volume of interest is then:
`∆H= H ̂_outlet-H ̂_inlet`

Method 2: reference conditions are the enthalpy of the reactant and products in their state of aggregation (i.e. not from elemental form) at a reference temperature for which the heat of reaction is known (usually 25 C).
a. This method includes the heat of reaction which we usually calculate in standard from (i.e. using a reference temperature of 25C)
b. The species and stream enthalpy in this method should include the ‘H’ term in the shomate equation which acts to ‘remove’ the heat of formation for the species enthalpy calculation, and thus the reference state enthalpy is zero.
c. Energy balance of the control volume of interest is then:
`∆H=∆H ̂_r^°+ H ̂_outlet-H ̂_inlet`

Control volume 1D in IDAES has been setup to compute enthalpy with method 2 (has_heat_of_reaction = True) that’s why the gas solid contactor property packages were setup to include the ‘H’ term of the shomate equations for enthalpy calculations.

andrewlee94: @Robbybp 
1. It is primarily the user of the this equation below the lower bound it was fitted for. The `F` term is basically an error correction for the integral up to the lower bound. @chineduobiora The temperature you test at isn't really an issue so long as you understand what the expression is doing.
2. It depends on how you want to write the model - if you use `H-Href` (the base Shomate form) for the enthalpy term, then you need to explicitly include heat of reaction in the energy balances. If you use `H` instead (leaving out the `H` term from the Shomate equation), then heat of reaction is captured implicitly in the difference between enthalpies of products and reactants. Generally speaking, the second approach works best for chemical engineering applications, with the first form being used more by mechanical engineers (where they don't need to worry about chemical reactions).

chineduobiora: @andrewlee94 regarding the second point, is it worthwhile if we update the CV0D and CV1D documentation to explain this? I can imagine a number of users might have challenges figuring this out without writing out the equations themselves.

andrewlee94: I think that is explained somewhere already, and it is not so much documentation of the control volumes but understanding fundamental thermodynamics.

chineduobiora: @andrewlee94 
If it’s already explained somewhere great, though I did a quick check and couldn’t find it. However, I brought up that suggestion because an IDAES process model has a number of different modules, and a little extra help in documentation can be helpful for users (even those who already have a good grasp of fundamental thermodynamics), to get a better picture of how things interconnect.
My quick suggestion is something along the following lines:
has_heat_of_reaction: set to **True** if calculating enthalpy balance of a reactive process with reference conditions of species set at their state of aggregation.
has_heat_of_reaction: set to **False** if no reactions, or if calculating enthalpy balance of a reactive process with reference conditions of species set at the enthalpy of formation.

chineduobiora (approved):
The explanation for the enthalpy test clarified things for me.

Robbybp:
@chineduobiora I went ahead with your suggestion to comment out the test for enthalpy at 300K. I think it is misleading to test at a value where our equation is not technically valid.

andrewlee94 (approved):
Looks good to me.

chineduobiora (approved):
Looks good.

Robbybp:
Test failures appear unrelated to this PR.

andrewlee94:
@Robbybp Yes, the test failures are a know problem due to being halfway through the release process.

PR #419, opened 2021-07-09
Adding support for osmotic pressure in generic properties

andrewlee94:
## Summary/Motivation:
NAWI models will require support for calculating osmotic pressure (incl. multiple solvent systems). This requires being able to identify solvent and solute species, thus the way component sets are created was also revised to ensue solvent and solute sets are created as required,

- Adds methods for calculating osmotic pressure in Ideal and eNRTL property modules
- Reworks creation of component sets to ensure solvent and solute sets are created by default

### Legal Acknowledgement


idaes/core/components.py
```diff
@@ -315,6 +347,14 @@ def _is_phase_valid(self, phase):
     def _is_aqueous_phase_valid(self):
         return True
 
+    def _add_to_component_list(self):
+        """
+        Ions should not be used outside of electorlyte property methods
```
adam-a-a, line 352:
```suggestion
        Ions should not be used outside of electrolyte property methods
```

idaes/core/components.py
```diff
@@ -410,6 +450,14 @@ class ApparentData(SoluteData):
...
```
adam-a-a, line 455:
```suggestion
        Ions should not be used outside of electrolyte property methods
```

idaes/generic_models/properties/core/eos/tests/test_enrtl.py
```diff
@@ -832,3 +833,34 @@ def test_tau(self, model):
...
+        model.state[1].pressure_osmotic_phase.display()
+        model.state[1].act_phase_comp.display()
+
+        assert_units_equivalent(model.state[1].pressure_osmotic_phase["Liq"],
+                                pyunits.Pa)
+        assert len(model.state[1].pressure_osmotic_phase) == 1
+        assert pytest.approx(value(
+            -Constants.gas_constant*300*log(0.1670306)*55e3),
+            rel=1e-6) == value(
+                model.state[1].pressure_osmotic_phase["Liq"])
```
TimBartholomew, line 866: The naming standard I've been using for osmotic pressure was "pressure_osm". I shortened osmotic to osm because saturation pressure in IDAES conventions is "pressure_sat" and I also dropped the phase because I think its only applicable for liquids.

TimBartholomew (commented):
A few comments

idaes/generic_models/properties/core/eos/tests/test_enrtl.py
```diff
@@ -832,3 +833,34 @@ def test_tau(self, model):
...
+        return m
+
+    @pytest.mark.unit
+    def test_pressure_osmotic_phase(self, model):
+        model.state[1].dens_mol_phase = Var(model.params.phase_list,
+                                            initialize=55e3,
+                                            units=pyunits.mol/pyunits.m**3)
+
+        model.state[1].pressure_osmotic_phase.display()
+        model.state[1].act_phase_comp.display()
```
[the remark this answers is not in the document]
andrewlee94: Yes, they are left over. I will clean them up.

idaes/generic_models/properties/core/eos/ideal.py
```diff
@@ -299,6 +300,20 @@ def gibbs_mol_phase_comp(b, p, j):
...
+    @staticmethod
+    def pressure_osmotic_phase(b, p):
+        try:
+            solvent_set = b.params.solvent_set
+        except AttributeError:
+            raise ConfigurationError(
+                f"{b.name} called for pressure_osmotic, but no solvents were "
+                f"defined. Osmotic pressure requires at least one component "
+                f"to be declared as a solvent.")
+        C = sum(b.conc_mol_phase_comp[p, j] for j in b.component_list
```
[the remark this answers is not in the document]
andrewlee94: The issue here is how the user represents ions. This is the sum across all non-solvent species, so if the user declared the ions in some form then they will be implicitly included. However, if you just declare NaCl and expect that to work... I am not even sure how we can catch that condition without adding very case specific arguments.

idaes/generic_models/properties/core/eos/tests/test_enrtl.py
andrewlee94: Two things there:

1. You didn't update the naming standards, so I didn't know that was your standard.
2. The phase index is still needed because there could be multiple phases.

andrewlee94: Also as a counter-point, bubble pressure is `pressure_bubble`.

adam-a-a: Should we be updating IDAES naming conventions for any new names we create for Proteuslib? Or should we have our own section on naming conventions wherever we build on IDAES' naming conventions (or both)?

idaes/generic_models/properties/core/eos/ideal.py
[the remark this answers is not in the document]
adam-a-a: Should we think about making changes to the NaCl property model in Proteuslib based on what you have here, or do you both think the NaCl model can remain independent of this added functionality?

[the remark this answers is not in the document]
TimBartholomew: @andrewlee94 what if we add an attribute to the solute component that notes the number of dissociated ions?

idaes/generic_models/properties/core/eos/tests/test_enrtl.py
andrewlee94: We should try to maintain a single set of standards where possible, so the best approach would be to propose additions to the IDAES standards.

idaes/generic_models/properties/core/eos/ideal.py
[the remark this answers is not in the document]
TimBartholomew: @adam-a-a the NaCl property model won't use these methods, its an independent property package.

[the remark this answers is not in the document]
andrewlee94: @TimBartholomew That is only applicable if the Solute is a salt however. Remember that the Ion components derive from Solute, and number of dissociated ions makes no sense there. Seeing as we probably want to support this use case, I see two options:

1. Make `Apparent` work for non-electrolyte systems and make users define salts using this component (which makes sense),
2. Define a new `Salt` component which handles this, but which only works for non-electrolyte systems.

Writing that out, 1 would seem to be the only sane answer.

@adam-a-a The NaCL and seawater property packages are custom-coded so they can do what they want. However, it would make sense to make this a new standard for aqueous systems and have them define the `solute_set` and `solvent_set` as well.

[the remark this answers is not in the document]
andrewlee94: I've updated `Apparent` so that it should work with non-electrolyte systems and include the `i` term when calculating osmotic pressure.

idaes/generic_models/properties/core/eos/tests/test_enrtl.py
andrewlee94: I've updated the naming convention and added it to the docs.

TimBartholomew (approved):
LGTM

idaes/core/tests/test_components.py
```diff
@@ -419,55 +429,47 @@ def get_metadata(self):
             return m.meta_object
         m.get_metadata = types.MethodType(get_metadata, m)
 
-        m.comp = Ion()
-
-        return m
+        with pytest.raises(
+                NotImplementedError,
+                match="comp The IonData component class is intended as a base "
```
adam-a-a, line 434: Is "comp" supposed to be here or was it left here unintentionally?

adam-a-a: I am guessing it links to line 437 but not sure since I've never used this functionality.

andrewlee94: This is meant to be here - this is checking that the expected error message is returned, and `comp` is the name of the `Component` where the issue arose.

idaes/core/tests/test_components.py
```diff
@@ -419,55 +429,47 @@ def get_metadata(self):
...
-        assert not m.comp._is_phase_valid(m.Vap)
-        assert m.comp._is_phase_valid(m.Aqu)
-        assert not m.comp._is_phase_valid(m.Phase)
+        def get_metadata(self):
+            return m.meta_object
+        m.get_metadata = types.MethodType(get_metadata, m)
 
+        with pytest.raises(
+                PropertyPackageError,
+                match="comp Ion Component types should only be used with "
```
adam-a-a, line 469: I see now. "comp" _is_ tied to `m.comp`. Nevermind my comments above.

adam-a-a (approved):
LGTM!

PR #426, opened 2021-07-16
Fix execute on load with uncertainty_propagation examples

### LINKED ISSUE #332: Uncertainty example runs code on import

adowling2:
- Add missing `if __name__ == "__main__":` in `uncertainty_propagation` examples

### Legal Acknowledgement


andrewlee94:
I think this addresses the main concern. However, it crossed my mind that we should probably have tests for these as well - at least something to make sure that they run and given the expected final result.

lbianchi-lbl:
> I think this addresses the main concern. However, it crossed my mind that we should probably have tests for these as well - at least something to make sure that they run and given the expected final result.

I agree with @andrewlee94 on both points. One solution for this would be to move the code to a `main()` function, and then call it in the `if __name__  ...` block:

```py
from idaes import foo, bar

def main():
    # all code goes here

if __name__ == '__main__':
    main()
```

Then tests would be able to do `from idaes.mymodule import main()` and call it (at the most basic level, checking that no exceptions are raised).

ksbeattie:
> > I think this addresses the main concern. However, it crossed my mind that we should probably have tests for these as well - at least something to make sure that they run and given the expected final result.
> 
> I agree with @andrewlee94 on both points. One solution for this would be to move the code to a `main()` function, and then call it in the `if __name__ ...` block:
> 
> ```python
> from idaes import foo, bar
> 
> def main():
>     # all code goes here
> 
> if __name__ == '__main__':
>     main()
> ```
> 
> Then tests would be able to do `from idaes.mymodule import main` and call it (at the most basic level, checking that no exceptions are raised).

Is there any reason why this code that would be in `main()` should not just be moved into a separate test file?

ksbeattie (approved):
I'm approving this, even though I think the executable code here would better done as a test or as an example.  Hopefully that will be done?

andrewlee94 (approved):

adowling2:
@ksbeattie @andrewlee94 This is in examples mainly because it is intended as the first example for a new user to look at. It is the simple math problem in the documentation.

andrewlee94:
@adowling2 Our main comment here is that these end up buried in the code, and not easily found by users. They would be better put in the docs, or examples-pse repo as notebooks, so that users could easily find them.

That said, they also serve as useful test cases for the test suite, and we should make sure they are included there as well.

PR #433, opened 2021-07-21
Expanding Generic Property Framework

andrewlee94:
## Summary/Motivation:
This PR adds support for a number of properties required by IDAES and NAWI work.

- Adds support for molar volume and updates molar density to be calculated from molar volume
- Adds general mixing rules for molar volume to all equations of state and activity coefficient models
- Adds support for defining models for diffusivity and dynamic viscosity
- Adds tests for utility methods that were missing

### Legal Acknowledgement

### DIFF
--- a/idaes/core/components.py
+++ b/idaes/core/components.py
@@ -52,10 +52,22 @@ class ComponentData(ProcessBlockData):
             doc="Dict indicating phases in which component follows Henry's "
                 "Law (keys) with values indicating form of law."))
 
+    CONFIG.declare("vol_mol_liq_comp", ConfigValue(
+        description="Method to use to calculate liquid phase molar volume",
+        doc="Method to use to calculate liquid phase molar volume. Users "
+        "need only define one of vol_mol_liq_comp and dens_mol_liq_comp."))
+    CONFIG.declare("vol_mol_sol_comp", ConfigValue(
+        description="Method to use to calculate solid phase molar volume",
+        doc="Method to use to calculate solid phase molar volume. Users "
+        "need only define one of vol_mol_sol_comp and dens_mol_sol_comp."))
     CONFIG.declare("dens_mol_liq_comp", ConfigValue(
-        description="Method to use to calculate liquid phase molar density"))
+        description="Method to use to calculate liquid phase molar density",
+        doc="Method to use to calculate liquid phase molar density. Users "
+        "need only define one of vol_mol_liq_comp and dens_mol_liq_comp."))
     CONFIG.declare("dens_mol_sol_comp", ConfigValue(
-        description="Method to use to calculate solid phase molar density"))
+        description="Method to use to calculate solid phase molar density",
+        doc="Method to use to calculate solid phase molar density. Users "
+        "need only define one of vol_mol_sol_comp and dens_mol_sol_comp."))
 
     CONFIG.declare("cp_mol_liq_comp", ConfigValue(
         description="Method to calculate liquid component specific heats"))
--- a/idaes/core/components.py
+++ b/idaes/core/components.py
@@ -55,7 +55,7 @@ class ComponentData(ProcessBlockData):
     CONFIG.declare("vol_mol_liq_comp", ConfigValue(
         description="Method to use to calculate liquid phase molar volume",
         doc="Method to use to calculate liquid phase molar volume. Users "
-        "need only define one of vol_mol_liq_comp and dens_mol_liq_comp."))
+        "need only define either vol_mol_liq_comp or dens_mol_liq_comp."))
     CONFIG.declare("vol_mol_sol_comp", ConfigValue(
         description="Method to use to calculate solid phase molar volume",
         doc="Method to use to calculate solid phase molar volume. Users "
--- a/idaes/core/components.py
+++ b/idaes/core/components.py
@@ -59,7 +59,7 @@ class ComponentData(ProcessBlockData):
     CONFIG.declare("vol_mol_sol_comp", ConfigValue(
         description="Method to use to calculate solid phase molar volume",
         doc="Method to use to calculate solid phase molar volume. Users "
-        "need only define one of vol_mol_sol_comp and dens_mol_sol_comp."))
+        "need only define either vol_mol_sol_comp or dens_mol_sol_comp."))
     CONFIG.declare("dens_mol_liq_comp", ConfigValue(
         description="Method to use to calculate liquid phase molar density",
         doc="Method to use to calculate liquid phase molar density. Users "
--- a/idaes/core/components.py
+++ b/idaes/core/components.py
@@ -63,11 +63,11 @@ class ComponentData(ProcessBlockData):
     CONFIG.declare("dens_mol_liq_comp", ConfigValue(
         description="Method to use to calculate liquid phase molar density",
         doc="Method to use to calculate liquid phase molar density. Users "
-        "need only define one of vol_mol_liq_comp and dens_mol_liq_comp."))
+        "need only define either vol_mol_liq_comp or dens_mol_liq_comp."))
     CONFIG.declare("dens_mol_sol_comp", ConfigValue(
         description="Method to use to calculate solid phase molar density",
         doc="Method to use to calculate solid phase molar density. Users "
-        "need only define one of vol_mol_sol_comp and dens_mol_sol_comp."))
+        "need only define either vol_mol_sol_comp or dens_mol_sol_comp."))
 
     CONFIG.declare("cp_mol_liq_comp", ConfigValue(
         description="Method to calculate liquid component specific heats"))
--- a/idaes/generic_models/properties/core/eos/ceos.py
+++ b/idaes/generic_models/properties/core/eos/ceos.py
@@ -812,6 +812,16 @@ class Cubic(EoSBase):
                 b.entr_mol_phase_comp[p, j] *
                 b.temperature)
 
+    @staticmethod
+    def vol_mol_phase(b, p):
+        pobj = b.params.get_phase(p)
+        if pobj.is_vapor_phase() or pobj.is_liquid_phase():
+            return (Cubic.gas_constant(b)*b.temperature *
+                    b.compress_fact_phase[p] /
+                    b.pressure)
+        else:
+            raise PropertyNotSupportedError(_invalid_phase_msg(b.name, p))
+
 
 def _invalid_phase_msg(name, phase):
     return ("{} received unrecognized phase name {}. Ideal property "
--- a/idaes/generic_models/properties/core/eos/eos_base.py
+++ b/idaes/generic_models/properties/core/eos/eos_base.py
@@ -52,6 +52,27 @@ class EoSBase():
         raise NotImplementedError(_msg(b, "build_parameters"))
 
     @staticmethod
+    def get_vol_mol_pure(b, phase, comp, temperature):
+        try:
+            vol_mol = get_method(b, "vol_mol_"+phase+"_comp", comp)(
+                                 b, cobj(b, comp), temperature)
+        except (AttributeError, ConfigurationError):
+            # vol_mol not defined, try for dens_mol instead
+            try:
+                vol_mol = 1/get_method(b, "dens_mol_"+phase+"_comp", comp)(
+                                       b, cobj(b, comp), temperature)
+            except (AttributeError, ConfigurationError):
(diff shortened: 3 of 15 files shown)


andrewlee94:
@anujad95 This PR should fix the issue you observed with the mixing rules for density.

@adam-a-a @TimBartholomew This PR adds support for `diffus_phase_comp` and `visc_d_phase`, which I think should be the last properties required for the RO models.

idaes/core/components.py
```diff
@@ -52,10 +52,22 @@ class ComponentData(ProcessBlockData):
             doc="Dict indicating phases in which component follows Henry's "
                 "Law (keys) with values indicating form of law."))
 
+    CONFIG.declare("vol_mol_liq_comp", ConfigValue(
+        description="Method to use to calculate liquid phase molar volume",
+        doc="Method to use to calculate liquid phase molar volume. Users "
+        "need only define one of vol_mol_liq_comp and dens_mol_liq_comp."))
```
adam-a-a, line 58: Do you mean that users only need to define either (a) molar volume or (b) molar density, but not both?

idaes/generic_models/properties/core/generic/tests/test_generic_property.py
```diff
@@ -1071,3 +1071,27 @@ def set_flow_basis():
...
+        frame.params.p1.config.diffus_phase_comp = \
+            TestGenericStateBlock.dummy_prop
+        frame.params.p2.config.diffus_phase_comp = \
+            TestGenericStateBlock.dummy_prop
+
+        for p, j in frame.props[1].phase_component_set:
+            assert value(frame.props[1].diffus_phase_comp[p, j]) == 4
+
+    @pytest.mark.unit
+    def test_vics_d_phase(self, frame):
```
adam-a-a, line 1090:
```suggestion
    def test_visc_d_phase(self, frame):
```

idaes/core/components.py
andrewlee94: Yes - dens_mol = 1/vol_mol, so you only need one of them. Do you think this could be more clear (suggestions welcome, but that might also be something better suited to the docs rather than just a doc string).

idaes/generic_models/properties/core/eos/ceos.py
```diff
@@ -812,6 +812,16 @@ def gibbs_mol_phase_comp(b, p, j):
                 b.entr_mol_phase_comp[p, j] *
                 b.temperature)
 
+    @staticmethod
+    def vol_mol_phase(b, p):
+        pobj = b.params.get_phase(p)
+        if pobj.is_vapor_phase() or pobj.is_liquid_phase():
+            return (Cubic.gas_constant(b)*b.temperature *
+                    b.compress_fact_phase[p] /
+                    b.pressure)
```
TimBartholomew, line 821: What is compress_fact_phase?

I'm a little confused how this works for incompressible liquids. I would have thought that it would be calculated just based on the dens_mass and molecular weight. If compress_fact_phase is defined for an incompressible liquid in such a way that it cancels out the gas_constant, temperature, and pressure, and it basically just comes down to dens_mass and molecular weight isn't this just making two relatively complicated constraints and an additional variable?

idaes/generic_models/properties/core/eos/eos_base.py
```diff
@@ -51,6 +51,27 @@ def calculate_scaling_factors(b, pobj):
...
 
+    @staticmethod
+    def get_vol_mol_pure(b, phase, comp, temperature):
+        try:
+            vol_mol = get_method(b, "vol_mol_"+phase+"_comp", comp)(
+                                 b, cobj(b, comp), temperature)
+        except (AttributeError, ConfigurationError):
+            # vol_mol not defined, try for dens_mol instead
+            try:
+                vol_mol = 1/get_method(b, "dens_mol_"+phase+"_comp", comp)(
```
TimBartholomew, line 62: does dens_mol have the phase in the name? i.e. dens_mol_liq_comp? Why doesn't it just follow the other naming convention like dens_mol_phase_comp['liq', j]?

idaes/generic_models/properties/core/eos/tests/test_enrtl_verification_NaCl.py
```diff
@@ -46,21 +47,28 @@
...
+
+    def return_expression(b, cobj, T):
+        return cobj.vol_mol_pure
 
 
 configuration = {
     "components": {
         "H2O": {"type": Solvent,
-                "dens_mol_liq_comp": dummy_method,
+                "vol_mol_liq_comp": ConstantVolMol,
```
TimBartholomew, line 63: Oh ok this answers my previous question on the naming. How bad is it to change this so _phase is in the name and the variable is indexed by the phase?

idaes/generic_models/properties/core/eos/ceos.py
andrewlee94: It is the compressibility factor, Z, which is calculated by the equation of state: `PV=ZRT`. The whole purpose of the equation of state is to calculate `Z` for the current state.

idaes/generic_models/properties/core/eos/eos_base.py
andrewlee94: This is a pure component property, not a mixture property. I.e. we are looking up the config dict for how to calculate dens_mol for a given component and phase type. This is then used to calculate `dens_mol_phase_comp` for the mixture.

idaes/generic_models/properties/core/eos/tests/test_enrtl_verification_NaCl.py
andrewlee94: Again - these are pure component properties, not mixture properties, and the name is different to make that clear. Also, these apply to *any* liquid phase, i.e. there will only be one of these no matter how many liquid phases there are in the mixture.

idaes/generic_models/properties/core/eos/eos_base.py
TimBartholomew: After our discussion, I learned that these are pure component parameters. 👍

TimBartholomew (approved):
LGTM

idaes/core/components.py
adam-a-a, line 58:
```suggestion
        "need only define either vol_mol_liq_comp or dens_mol_liq_comp."))
```

adam-a-a: Although I understood your meaning, this seems a tad bit clearer.

TimBartholomew (dismissed):

idaes/generic_models/properties/core/eos/tests/test_ideal.py
```diff
@@ -226,8 +226,9 @@ def test_dens_mol_phase_liq(m):
         m.params.get_component(j).config.dens_mol_liq_comp = dummy_call
 
     assert str(Ideal.dens_mol_phase(m.props[1], "Liq")) == str(
-        sum(m.props[1].mole_frac_phase_comp["Liq", j]*42
-            for j in m.params.component_list))
+                1/(1/42*m.props[1].mole_frac_phase_comp["Liq", "a"] +
```
adam-a-a, line 229: Just curious why this revision was reformulated in this way.

andrewlee94: That is what the expression actually looks like now: `rho = 1/(sum x_i*1/v_mol_i for all i)`

idaes/generic_models/properties/core/eos/tests/test_ideal.py
```diff
@@ -555,3 +558,61 @@ def test_pressure_osm_phase_w_apparent_component():
...
+    for p in m.params.phase_list:
+        if p == "Vap":
+            assert str(Ideal.vol_mol_phase(m.props[1], p)) == (
+                "kg*m**2/J/s**2*(8.314462618*(J)/mol/K)*"
+                "props[1].temperature/props[1].pressure")
+        else:
+            assert str(Ideal.vol_mol_phase(m.props[1], p)) == str(
+                1/42*m.props[1].mole_frac_phase_comp["Liq", "a"] +
+                1/42*m.props[1].mole_frac_phase_comp["Liq", "b"] +
+                42*m.props[1].mole_frac_phase_comp["Liq", "c"])
```
adam-a-a, line 618: Just making sure, but is this supposed to be ` 42*m.props[1].mole_frac_phase_comp["Liq", "c"])` (as you have it now) or should it be `1/42*m.props[1].mole_frac_phase_comp["Liq", "c"])`? Following the logic of previous sections similar to this, I would assume that it should be the latter.

andrewlee94: In this case, the current form is correct - I am doing double duty in this test and making sure we can used a mixture of `dens_mol_liq_comp` and `vol_mol_liq_comp`.

For `a` and `b`, I defined `dens_mol_liq_comp`, hence the `1/42`. However, for component `c` I defined `vol_mol_liq_comp` instead, hence it is just `42`.

adam-a-a: Ah! I missed that detail. Makes sense!

idaes/generic_models/properties/core/generic/generic_property.py
```diff
@@ -2666,6 +2681,31 @@ def rule_pressure_sat_comp(b, j):
...
+        except AttributeError:
+            self.del_component(self.visc_d_phase)
+            raise
+
+    def _vol_mol_phase(self):
+        try:
+            def rule_vol_mol_phase(b, p):
+                p_config = b.params.get_phase(p).config
+                return p_config.equation_of_state.vol_mol_phase(b, p)
+            self.vol_mol_phase = Expression(
```
adam-a-a, line 2701: Just for my own reference, why declare these as expressions (i.e., `self.vol_mol_phase`, `self.visc_d_phase`, 'self.diffus_phase_comp`) as opposed to variables with constraints? I am just comparing this to how we defined properties in the NaCl and seawater property models.

idaes/core/components.py
```diff
@@ -52,10 +52,22 @@ class ComponentData(ProcessBlockData):
...
                 "Law (keys) with values indicating form of law."))
 
+    CONFIG.declare("vol_mol_liq_comp", ConfigValue(
+        description="Method to use to calculate liquid phase molar volume",
+        doc="Method to use to calculate liquid phase molar volume. Users "
+        "need only define either vol_mol_liq_comp or dens_mol_liq_comp."))
+    CONFIG.declare("vol_mol_sol_comp", ConfigValue(
+        description="Method to use to calculate solid phase molar volume",
+        doc="Method to use to calculate solid phase molar volume. Users "
+        "need only define one of vol_mol_sol_comp and dens_mol_sol_comp."))
```
adam-a-a, line 62:
```suggestion
        "need only define either vol_mol_sol_comp or dens_mol_sol_comp."))
```

idaes/generic_models/properties/core/generic/generic_property.py
andrewlee94: Mostly just a design decision - either approach works in the end.

idaes/core/components.py
```diff
@@ -52,10 +52,22 @@ class ComponentData(ProcessBlockData):
...
     CONFIG.declare("dens_mol_liq_comp", ConfigValue(
-        description="Method to use to calculate liquid phase molar density"))
+        description="Method to use to calculate liquid phase molar density",
+        doc="Method to use to calculate liquid phase molar density. Users "
+        "need only define one of vol_mol_liq_comp and dens_mol_liq_comp."))
```
adam-a-a, line 66:
```suggestion
        "need only define either vol_mol_liq_comp or dens_mol_liq_comp."))
```

idaes/core/components.py
```diff
@@ -52,10 +52,22 @@ class ComponentData(ProcessBlockData):
...
     CONFIG.declare("dens_mol_sol_comp", ConfigValue(
-        description="Method to use to calculate solid phase molar density"))
+        description="Method to use to calculate solid phase molar density",
+        doc="Method to use to calculate solid phase molar density. Users "
+        "need only define one of vol_mol_sol_comp and dens_mol_sol_comp."))
```
adam-a-a, line 70:
```suggestion
        "need only define either vol_mol_sol_comp or dens_mol_sol_comp."))
```

adam-a-a (approved):
LGTM

TimBartholomew (approved):

PR #437, opened 2021-07-23
Degeneracy Hunter: minor bug fix

adowling2:
- Fixes printing mistake in Degeneracy Hunter (indentation error)

### Legal Acknowledgement


adowling2:
This is needed for https://github.com/IDAES/examples-pse/pull/21

andrewlee94 (dismissed):
A couple of things to think about for the future (don't need to hold up this PR):

1. Should `print_level` use an `Enum` or similar to help guide users towards valid inputs. Alternatively, do you have a check elsewhere to screen for invalid inputs?
2. Rather than a simple print statement, it would be good to have some control over the output destination (e.g. a file or some other output channel), but defaulting to `stdout` if no input is provided.

idaes/core/util/model_diagnostics.py
```diff
@@ -145,20 +145,20 @@ def check_residuals(self, tol=1e-5, print_level=2, sort=True):
...
+                        print("\ncount =",i,"\t|residual| =",r)
+                        c.pprint()
+        
+                    if print_level == 2:
+                        # print values and bounds for each variable in the constraint
+                        print("variable\tlower\tvalue\tupper")
+                        for v in identify_variables(c.body):
+                            self.print_variable_bounds(v)
             else:
                 print("No constraints with residuals larger than",tol,"!")
```
blnicho, line 163: @adowling2 you might want to double check this change. If I'm following your code I think the indentation on the for loop was correct and the problem is with the `else` statement on line 162. I think that needs to line up with the `if` on line 140. With this change nothing will be printed if `sort=False` which I don't think is what you intended.

ksbeattie: I'd think a unit test for this would help here?

adowling2:
@blnicho Thank you for catching this. I fixed the indent. I still need to test this on the example.

adowling2:
I verified the example runs correctly with this fix. @blnicho @andrewlee94 

@lbianchi-lbl This is ready to merge with the appropriate reviews.

jsiirola (approved):

andrewlee94 (approved):
Looks good to me.

PR #443, opened 2021-08-04
Robsutness Updates for Boiler HX

eslickj:
I made some minor changes to the boiler heat exchanger.  Unfortunately I also ran black and reformatted the code.  Also added a bunch of tests for the deltaT callbacks.

- Use named expression for temperature difference
- Rewrite Q = UAdeltaT equation to be naturally better scaled as Q/(UA) = deltaT
- Add config option for various deltaT callbacks

### Legal Acknowledgement

### DIFF
--- a/idaes/power_generation/unit_models/boiler_heat_exchanger.py
+++ b/idaes/power_generation/unit_models/boiler_heat_exchanger.py
@@ -39,20 +39,34 @@ from enum import Enum
 
 # Import Pyomo libraries
 from pyomo.common.config import ConfigBlock, ConfigValue, In
+
 # Additional import for the unit operation
-from pyomo.environ import SolverFactory, value, Var, Param, exp, sqrt,\
-    log, PositiveReals, NonNegativeReals, units as pyunits
+from pyomo.environ import (
+    SolverFactory,
+    value,
+    Var,
+    Param,
+    exp,
+    sqrt,
+    log,
+    PositiveReals,
+    NonNegativeReals,
+    ExternalFunction,
+    units as pyunits,
+)
 from pyomo.opt import TerminationCondition
 
 # Import IDAES cores
-from idaes.core import (ControlVolume0DBlock,
-                        declare_process_block_class,
-                        MaterialBalanceType,
-                        EnergyBalanceType,
-                        MomentumBalanceType,
-                        UnitModelBlockData,
-                        useDefault)
-
+from idaes.core import (
+    ControlVolume0DBlock,
+    declare_process_block_class,
+    MaterialBalanceType,
+    EnergyBalanceType,
+    MomentumBalanceType,
+    UnitModelBlockData,
+    useDefault,
+)
+from idaes.core.util.functions import functions_lib
 from idaes.core.util.config import is_physical_parameter_block
 from idaes.core.util.misc import add_object_reference
 from idaes.core.util.constants import Constants as c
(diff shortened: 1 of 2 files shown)


MAZamarripa (approved):
lgtm, note that some tests are failing.

idaes/power_generation/unit_models/boiler_heat_exchanger.py
```diff
@@ -78,135 +92,279 @@ class DeltaTMethod(Enum):
...
+    b.cbrt = ExternalFunction(
+        library=functions_lib(), function="cbrt", arg_units=[temp_units]
+    )
+
+    @b.Expression(b.flowsheet().time)
+    def delta_temperature(b, t):
+        return ((b.cbrt(dT1[t]) + b.cbrt(dT2[t])) / 2.0) ** 3 * temp_units
+
+
+def delta_temperature_underwood_tune_callback(b):
```
andrewlee94, line 149: First - do we actually have a use case for this method? The original paper that came up with this "tuned" Underwood approximation misunderstood the original purpose of the function and the meaning of he parameters. And their "tuned" parameters were barely any different from the original Underwood parameters making me fairly certain that their tuning was an overfit to their specific data.

My first impression would be that we should either not support this, or if we do, that the default parameter values should be those of Underwood (and definitely not the "tuned" paper).

idaes/power_generation/unit_models/boiler_heat_exchanger.py
```diff
@@ -78,135 +92,279 @@ class DeltaTMethod(Enum):
...
+        ) * temp_units
+
+
 @declare_process_block_class("BoilerHeatExchanger")
 class BoilerHeatExchangerData(UnitModelBlockData):
     """
     Standard Heat Exchanger Unit Model Class
     """
+
     CONFIG = ConfigBlock()
```
andrewlee94, line 181: Is there any reason you are not inheriting from the Config Block in the base class here? Notably, the base Config already has `dynamic` and `has_holdup` defined for you (as they are expected arguments).

andrewlee94 (commented):
A couple of minor comments from looking over the code. The reformatting might mean I missed something though.

One comment on the reformulation of `Q = UAdeltaT` - the new form is better scaled, but it does leave open the possibility of a singularity if the user is not careful. `A` is unlikely to be 0 (short of user error), but there is the possibility of `U` doing strange things if a heat transfer correlation is supplied. I am not sure which is better, but the scaling could be addressed using the scaling tools.

idaes/power_generation/unit_models/boiler_heat_exchanger.py
eslickj: Really the only reason I kept it was that was what was originally in this model, so I didn't want to mess anything else up.  We probably don't want to change the default here until we search out and update other models that use it.

idaes/power_generation/unit_models/boiler_heat_exchanger.py
eslickj: I didn't write this model, and I didn't want to get into rewriting it.  Overall it would be nice if it inherited the latest 0D heat exchanger model, but I'm not up for that much change right now.  In the future we should probably make some of these heat transfer and pressure drop correlations available in the generic model and do a serious update of the boiler heat exchanger model.

eslickj:
> 
> 
> A couple of minor comments from looking over the code. The reformatting might mean I missed something though.
> 
> One comment on the reformulation of `Q = UAdeltaT` - the new form is better scaled, but it does leave open the possibility of a singularity if the user is not careful. `A` is unlikely to be 0 (short of user error), but there is the possibility of `U` doing strange things if a heat transfer correlation is supplied. I am not sure which is better, but the scaling could be addressed using the scaling tools.

Yeah, I was really conflicted about this, but settled on the easy option.  I'll update to use the scaling tools.  My thinking was that U was unlikely to be 0, but you never know I guess.

MAZamarripa (approved):
lgtm

eslickj:
I'll just work on this in the NGCC branch and merge it with that when it's ready.

PR #446, opened 2021-08-05
Allow changing parameters for TEL curve

eslickj:
This let's you change the TEL curve for the steam turbine.  This will change with turbine size, but we were stuck with a fixed curve. 

### Legal Acknowledgement

### DIFF
--- a/idaes/power_generation/unit_models/helm/turbine_outlet.py
+++ b/idaes/power_generation/unit_models/helm/turbine_outlet.py
@@ -64,33 +64,33 @@ class HelmTurbineOutletStageData(HelmIsentropicTurbineData):
         self.tel_c0 = Var(
             initialize=0.0064*1e6,
             units=pyunits.J/pyunits.mol,
-            doc="c0 in c0 + c1*fr + c2*fr**2 + ... + c5*fr**5 (fr is the ratio "
-                "of exhaust volumetric flow to design exhaust volumetric flow)")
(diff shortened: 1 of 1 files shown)


idaes/power_generation/unit_models/helm/turbine_outlet.py
```diff
@@ -60,16 +60,29 @@ def build(self):
         self.flow_coeff.fix()
         self.efficiency_mech.fix()
 
+
+        self.tel_c0 = Var(initialize=0.0064)
```
andrewlee94, line 64: Minor nitpicks, but a doc string and explicit units (`units=pyunits.dimensionless`) would be nice.

andrewlee94 (dismissed):
Looks fine - my only comment is a minor thing that doesn't need to be addressed.

MAZamarripa (dismissed):
lgtm

idaes/power_generation/unit_models/helm/turbine_outlet.py
eslickj: Thanks.  I was kind of lazy with the units there.  I moved the units to the coefficients where they belong, and added doc strings.  @andrewlee94 and @MAZamarripa should be okay to re-review now.

andrewlee94 (approved):

MAZamarripa (approved):
lgtm

PR #452, opened 2021-08-10
Get-extensions add "extra" option and bump to 2.4.4 release 

eslickj:
## Fixes
This cleans up the get-extensions download function, and lets you specify extra packages.  It also adds a tool to calculate the binary release hashes.  This is to prep for adding the Petsc solver extra.  It's big, and also not something a lot of people want.  So I don't want to add it to solvers.  The size of it may also make it hard to use GitHub actions to calculate the hashes like I have been doing.  It will probably complain about the amount of stuff I download.

You'd use something like:

```sh
idaes get-extensions --extra petsc
```

You can use the ```--extra``` options multiple times to install multiple extras, but at this point we have only petsc

### Legal Acknowledgement

### DIFF
--- a/idaes/commands/extensions.py
+++ b/idaes/commands/extensions.py
@@ -89,6 +89,18 @@ def print_extensions_version(library_only=False):
     "--show-platforms",
     is_flag=True,
     help="Show the platform options")
+@click.option(
+    "--show-extras",
+    is_flag=True,
+    help="Show list of binary extras")
+@click.option(
+    "--extra",
+    multiple=True,
+    help="Install extras")
+@click.option(
+    "--extras-only",
+    is_flag=True,
+    help="Only install extras")
 @click.option("--verbose", help="Show details", is_flag=True)
 def get_extensions(
     release,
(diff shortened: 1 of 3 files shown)


idaes/commands/extensions.py
```diff
@@ -89,6 +89,22 @@ def print_extensions_version(library_only=False):
...
+    multiple=True,
+    help="Install extras")
+@click.option(
+    "--extras-only",
+    is_flag=True,
+    help="Only install extras")
+@click.option(
+    "--to",
+    default=None,
+    help="Put extrensions in a alternate location")
```
dangunter, line 107: typo: "extrensions"

idaes/commands/extensions.py
```diff
@@ -101,18 +117,28 @@ def get_extensions(
...
+    extra,
+    to):
 
     if show_platforms:
-        click.echo("\nBuild platforms for IDAES binary Extensions.  Most Linux")
-        click.echo("platforms are interchangeable.")
-        for key, mes in idaes.config.known_binary_platform.items():
+        click.echo("\nSupported platforms for IDAES binary extensions.")
+        for key, mes in sorted(idaes.config.known_binary_platform.items()):
             click.echo("    {}: {}".format(key, mes))
```
dangunter, line 129: very minor point, but f-strings are supported by all out Python versions now, so you could write: `click.echo(f"     {key}:{mes}")` instead

idaes/commands/extensions.py
```diff
@@ -152,3 +181,41 @@ def get_extensions(
...
+    "--path",
+    help="Directory of release files",
+    default=None,
+    required=True)
+def hash_extensions(release, path):
+    hfile = f"sha256sum_{release}.txt"
+    if path is not None:
+        hfile = os.path.join(path, hfile)
+
+    def _w(fp, pack, plat):
```
dangunter, line 202: maybe give this a longer name to make it easier to follow this code, `_write_hashes`?

idaes/commands/extensions.py
```diff
@@ -152,3 +181,41 @@ def get_extensions(
...
+        f = f"idaes-{pack}-{plat}-64.tar.gz"
+        if path is not None:
+            h = idaes.util.download_bin.hash_file_sha256(os.path.join(path, f))
+        else:
+            h = idaes.util.download_bin.hash_file_sha256(f)
+        fp.write(h)
```
dangunter, line 208: Is this a format that will be read elsewhere? Maybe it would be easier to just write it as JSON? I may be misunderstanding the purpose of this function.

eslickj: You are right.  The problem is previous to this I used a bash script to do these, and I just used the simple space separated format.  I'd definitely use json if not for the old files.

dangunter (approved):
This looks good to me. All my comments are very minor.

jsiirola (approved):
Looks good

PR #454, opened 2021-08-11
TGA model files

bpaul4:
## Summary/Motivation:
- TGA model files for the 0D fixed bed reactor unit and dynamic example

- Adding unit model package for 0D fixed bed reactor
- Adding dynamic example script for TGA model
- Updated __init__.py in unit_models folder for new files
- Future: add property packages (need to fix deprecated unit definitions first), documentation and tests

### Legal Acknowledgement


idaes/gas_solid_contactors/flowsheets/dyn_TGA_example.py
```diff
@@ -0,0 +1,92 @@
...
+import time
+
+from pyomo.environ import ConcreteModel, SolverFactory, TransformationFactory
+
+from idaes.core import FlowsheetBlock, EnergyBalanceType
+from idaes.core.util.initialization import initialize_by_time_element
+from idaes.core.util import get_solver
+
+# Access parent directory (chemical_looping) of the current directory
+sys.path.append(os.path.join(os.path.dirname(__file__), os.pardir))
```
chineduobiora, line 19: These lines can be deleted as they are no longer needed.

idaes/gas_solid_contactors/flowsheets/dyn_TGA_example.py
```diff
@@ -0,0 +1,92 @@
...
+
+t_initialize = time.time()  # Initialization time
+
+blk = m.fs.TGA
+
+solver = get_solver('ipopt', optarg) # create solver
+
+initialize_by_time_element(m.fs, m.fs.time, solver=solver)
+solver.solve(m, tee=True)
+# write_violated_equalities(m.fs)
```
chineduobiora, line 88: This line can be deleted.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,355 @@
...
+from idaes.core.util import get_solver
+__author__ = "Chinedu Okoli, Andrew Lee"
+
+
+# Assumptions:
+# Only Vap and Sol phases, which are explicitly named as such
+# Perfect mixing in Vap phase
+# Static solid phase
+
+# Need to have build-on-demand properties in ractions for this to work
```
chineduobiora, line 40: Typo - "reactions"

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,355 @@
...
+#         doc="""Indicates what type of mass balance should be constructed,
+# **default** - MaterialBalanceType.useDefault.
+# **Valid values:** {
+# **MaterialBalanceType.useDefault - refer to property package for default
+# balance type
+# **MaterialBalanceType.none** - exclude material balances,
+# **MaterialBalanceType.componentPhase** - use phase component balances,
+# **MaterialBalanceType.componentTotal** - use total component balances,
+# **MaterialBalanceType.elementTotal** - use total element balances,
+# **MaterialBalanceType.total** - use total material balance.}"""))
```
chineduobiora, line 74: This block of code can be deleted.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,355 @@
...
+#                      .format(condition(results)))
+
+        # ---------------------------------------------------------------------
+        # Release Inlet state
+        blk.gas.release_state(gas_phase_flags, outlvl+1)
+        blk.solids.release_state(solid_phase_flags, outlvl+1)
+
+        print('dof', degrees_of_freedom(blk))
+        # init_log.log(5, 'Initialization Complete: {}'
+        #              .format(condition(results)))
```
chineduobiora, line 355: These lines of code can be deleted.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,355 @@
...
+
+        # ---------------------------------------------------------------------
+        # Solve unit
+#         with solver_log(solve_log, DEBUG) as slc:
+#             results = opt.solve(blk, tee=slc.tee)
+# #        results = opt.solve(blk, tee=init_tee(init_log))
+
+#         init_log.log(4, "Initialization Step 2 {}."
+#                      .format(condition(results)))
+
```
chineduobiora, line 347: These lines of code can be deleted.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,355 @@
...
+                 * 6 = No output
+            optarg : solver options dictionary object (default={'tol': 1e-6})
+            solver : str indicating which solver to use during
+                     initialization (default = 'ipopt')
+        Returns:
+            None
+        '''
+        # Set solver options
+        init_log = getInitLogger(blk.name, outlvl)
+#        solve_log = getSolveLogger(blk.name, outlvl, tag="unit")
```
chineduobiora, line 304: This line of code can be deleted.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,355 @@
...
+        opt = get_solver(solver, optarg) # create solver
+        opt.options = optarg
```
chineduobiora, line 306: Take a look at the initialize methods in the bubbling fluidized bed and moving bed and update accordingly. 
The initialize method here should look like:
def initialize(blk, state_args={}, outlvl=idaeslog.NOTSET,
                   solver=None, optarg={}):

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,355 @@
...
+
+        # ---------------------------------------------------------------------
+        # Initialize control volume block
+        print()
+        print('Initialize thermophysical properties')
```
chineduobiora, line 311: Replace these lines with:
init_log.info('Initialize Thermophysical Properties')

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,355 @@
...
+                                 optarg=optarg,
+                                 solver=solver)
+
+        # TODO - maybe set this up in a nicer way
+        # Fix the solids temperature to initial value if isothermal conditions
+        if blk.config.energy_balance_type == EnergyBalanceType.none:
+            for t in blk.flowsheet().config.time:
+                blk.solids[t].temperature.fix(blk.solids[0].temperature.value)
+
+        init_log.log(4, 'Initialization Step 1 Complete.')
```
chineduobiora, line 337: Replace this line with:
init_log.info_high("Initialization Step 1 Complete.")

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
andrewlee94: A minor correction (or more recent change in IDAES): `state_args=None` and `optarg=None` should be the default.

There is an underlying Python design decision that means that passing empty dicts (`{}`) as default arguments is a bad idea - basically the empty dict only gets created when the code is first important, not each time the code is used. Thus that empty dict can often becomes non-empty the first time the code is called.

Note that this means you need to handle this later, generally by doing something like:

```
if state_args is None:
    state_args = {}
```

In the case of `optargs`, the (new) IDAES `get_solver` mehtod accepts `optargs=None` and handles it correctly for you.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,355 @@
...
+        opt = get_solver(solver, optarg) # create solver
+        opt.options = optarg
+
+        # ---------------------------------------------------------------------
+        # Initialize control volume block
+        print()
+        print('Initialize thermophysical properties')
+        # Initialize gas_phase block
+        gas_phase_flags = blk.gas.initialize(
+                                outlvl=outlvl - 1,
```
chineduobiora, line 314: Replace "outlvl - 1" with "outlvl"

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,355 @@
...
+                                optarg=optarg,
+                                solver=solver)
+
+        # Initialize solid_phase properties block
+        solid_phase_flags = blk.solids.initialize(
+                                outlvl=outlvl - 1,
```
chineduobiora, line 320: Replace "outlvl - 1" with "outlvl"

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,355 @@
...
+                                optarg=optarg,
+                                solver=solver)
+
+        print()
+        print('Initialize reaction properties')
+        # Initialize reactions
+        blk.reactions.initialize(outlvl=outlvl - 1,
```
chineduobiora, line 327: Replace "outlvl - 1" with "outlvl"

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,355 @@
...
+
+# Import IDAES cores
+from idaes.core import (declare_process_block_class,
+                        EnergyBalanceType,
+                        UnitModelBlockData,
+                        useDefault)
+from idaes.core.util.config import (is_physical_parameter_block)
+from idaes.logger import (getInitLogger, condition, solver_log,
+                          DEBUG, getSolveLogger)
+from idaes.core.util.model_statistics import (degrees_of_freedom)
```
chineduobiora, line 30: This line can be deleted

chineduobiora (commented):
@bpaul4 this looks good. I suggested a few changes to better streamline the code with idaes-pse standards.

bpaul4:
Thank you @chineduobiora and @andrewlee94 for your suggestions, I'll push the new versions after editing the files locally.

chineduobiora (commented):
The changes look good. We can start thinking about making the documentation.

bpaul4:
The documentation is in-progress (fixed_bed_0D.rst still needs significant work in the model description sections), and this latest commit is to preserve the files as I work with other branches.

docs/technical_specs/model_libraries/gas_solid_contactors/unit_models/fixed_bed_0D.rst
```diff
@@ -0,0 +1,168 @@
...
+* The reactor is assumed to be isothermal, and the reaction is one-way (no equilibrium).
+
+**Requirements:**
+
+* Property packages contain temperature and pressure variables.
+
+The FixedBed0D model equations are derived from spacially simplying an existing model of similar geometry and adding time-variant behavior:
+
+A. Ostace, A. Lee, C.O. Okoli, A.P. Burgard, D.C. Miller, D. Bhattacharyya, Mathematical modeling of a moving-bed reactor for chemical looping combustion of methane, in: M.R. Eden, M. Ierapetritou, G.P. Towler (Eds.),13th Int. Symp. Process Syst. Eng. (PSE 2018), Computer-Aided Chemical Engineering 2018, pp. 325–330 , San Diego, CA.
+
```
chineduobiora, line 22: This can be deleted. The 0DFixedBed model isn't derived from the MovingBed model.

docs/technical_specs/model_libraries/gas_solid_contactors/unit_models/fixed_bed_0D.rst
```diff
@@ -0,0 +1,168 @@
...
+  - EnergyBalanceType.enthalpyTotal - single enthalpy balance for material
+  - EnergyBalanceType.enthalpyPhase - enthalpy balances for each phase
+  - EnergyBalanceType.energyTotal - single energy balance for material
+  - EnergyBalanceType.energyPhase - energy balances for each phase
+
+**Arguments that are applicable to a specific region:**
+
+* gas_property_package - property package to use when calculating properties for the gas phase (default = 'useDefault'). 
+  This is provided as a Physical Parameter Block by the Flowsheet when creating the model. 
+  If a value is not provided, the ControlVolume Block will try to use the default property package if one is defined.
```
chineduobiora, line 55: This line can be removed. The 0DFixedBed model does not use control volumes.

docs/technical_specs/model_libraries/gas_solid_contactors/unit_models/fixed_bed_0D.rst
```diff
@@ -0,0 +1,168 @@
...
+* gas_property_package_args - set of arguments to be passed to the gas phase Property Blocks when they are created (default = 'None').
+* solid_property_package - property package to use when calculating properties for the solid phase (default = 'useDefault'). 
+  This is provided as a Physical Parameter Block by the Flowsheet when creating the model. 
+  If a value is not provided, the ControlVolume Block will try to use the default property package if one is defined.
```
chineduobiora, line 59: This line can be removed. The 0DFixedBed model does not use control volumes.

docs/technical_specs/model_libraries/gas_solid_contactors/unit_models/fixed_bed_0D.rst
```diff
@@ -0,0 +1,168 @@
...
+* solid_property_package_args - set of arguments to be passed to the solid phase Property Blocks when they are created (default = 'None').
+* reaction_package - reaction package to use when calculating properties for solid phase reaction (default = None). 
+  This is provided as a Reaction Parameter Block by the Flowsheet when creating the model. 
+  If a value is not provided, the ControlVolume Block will try to use the default property package if one is defined.
```
chineduobiora, line 63: This line can be removed. The 0DFixedBed model does not use control volumes.

docs/technical_specs/model_libraries/gas_solid_contactors/unit_models/fixed_bed_0D.rst
```diff
@@ -0,0 +1,168 @@
...
+
+In the following, the subscripts :math:`g` and :math:`s` refer to the gas and solid phases, respectively. 
+FixedBed0D units write the following Constraints:
+
+Reaction and Mass/Heat Transfer Constraints
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Solid material holdup:
+
+.. math:: h_{mass,s,t,j} = V_{s} \rho_{mass,s,t} x_{mass,s,t,j}
```
chineduobiora, line 77: "h" is usually associated with enthalpy. Maybe replace "h with "q"? "q" is a bit more standard for solid fluxes.

docs/technical_specs/model_libraries/gas_solid_contactors/unit_models/fixed_bed_0D.rst
```diff
@@ -0,0 +1,168 @@
...
+
+Solid material accumulation:
+
+.. math:: {\dot{h}}_{mass,s,t,j} = V_{s} {MW}_{s,j} {\Sigma}_{r} {r_{s,t,r} \nu_{s,j,r}}
```
chineduobiora, line 81: "h" is usually associated with enthalpy. Maybe replace "h with "q"? "q" is a bit more standard for solid fluxes.

docs/technical_specs/model_libraries/gas_solid_contactors/unit_models/fixed_bed_0D.rst
```diff
@@ -0,0 +1,168 @@
...
+
+Total mass of solids
+
+.. math:: S_{s,t} = V_{s} {\Sigma_{j} {\rho_{mass,s,t,j}}}
```
chineduobiora, line 85: What are your thoughts on using "M" for total mass of solids?

docs/technical_specs/model_libraries/gas_solid_contactors/unit_models/fixed_bed_0D.rst
```diff
@@ -0,0 +1,168 @@
...
+
+Sum of component mass fractions
+
+.. math:: {\Sigma_{j}} {x_{mass,s,t,j}} = 1 \space \space \space \forall \space t
+
+Solid energy holdup:
```
chineduobiora, line 91: This statement should be included before the energy balance equations `if self.config.energy_balance_type != EnergyBalanceType.none:`
The energy balance equations shouldn't be built if the energy balance type is set to none by the user.

docs/technical_specs/model_libraries/gas_solid_contactors/unit_models/fixed_bed_0D.rst
```diff
@@ -0,0 +1,168 @@
...
+List of Variables
+-----------------
+
+.. csv-table::
+   :header: "Variable", "Description", "Reference to"
+
+   ":math:`H_{mass,s,t}`", "Solid phase mass enthalpy", "``solids.enth_mass``"
+   ":math:`H_{rxn,s,t,r}`", "Solid phase reaction enthalpy", "``solids.reactions.dh_rxn``"
+   ":math:`h_{energy,s,t}`", "Energy holdup, solid phase", "``solids.energy_material_holdup``"
+   ":math:`h_{mass,s,t,j}`", "Material holdup, solid phase", "``solids.solids_material_holdup``"
```
chineduobiora, line 108: See prior comment on naming convention.

docs/technical_specs/model_libraries/gas_solid_contactors/unit_models/fixed_bed_0D.rst
```diff
@@ -0,0 +1,168 @@
...
+   ":math:`{\dot{h}}_{energy,s,t}`", "Energy accumulation, solid phase", "``solids.solids_energy_accumulation``"
+   ":math:`{\dot{h}}_{mass,s,t,j}`", "Material accumulation, solid phase", "``solids.solids_material_accumulation``"
```
chineduobiora, line 110: See prior comment on naming convention.

docs/technical_specs/model_libraries/gas_solid_contactors/unit_models/fixed_bed_0D.rst
```diff
@@ -0,0 +1,168 @@
...
+   ":math:`r_{s,t,r}`", "Solid phase reaction rate of reaction r", "``solids.reactions.reaction_rate``"
+   ":math:`S_{s,t}`", "Total mass of solids", "``solids.mass_solids``"
```
chineduobiora, line 112: See prior comment on naming convention.

docs/technical_specs/model_libraries/gas_solid_contactors/unit_models/fixed_bed_0D.rst
bpaul4: "M" is already used for the gas phase mass transfer term in the Moving Bed model, I picked another letter that was not already in use so we are consistent going forwards with more complex models. S tends to be used for solid loading in reactor beds in literature.

docs/technical_specs/model_libraries/gas_solid_contactors/unit_models/fixed_bed_0D.rst
andrewlee94: For solid fluxed I would use J - this is fairly common from what I have seen. Q is generally heat.

docs/technical_specs/model_libraries/gas_solid_contactors/unit_models/fixed_bed_0D.rst
```diff
@@ -0,0 +1,168 @@
...
+
+The initialization method for this model will save the current state of the model before
+commencing initialization and reloads it afterwards. The state of the model will be the same after
+initialization, only the initial guesses for unfixed variables will be changed. For the FixedBed0D
+model, the initialization method is a general purpose routine for simple unit models where a single
+control volume is initalized followed by attempting to solve the entire unit.
+
+The model allows for the passing of a dictionary of values of the state variables of the gas and
+solid phases that can be used as initial guesses for the state variables throughout the time and
+spatial domains of the model (in this model just time). This is optional but recommended. A typical
```
chineduobiora, line 140: "and spatial" can be removed.

docs/technical_specs/model_libraries/gas_solid_contactors/unit_models/fixed_bed_0D.rst
```diff
@@ -0,0 +1,168 @@
...
+   ":math:`h_{mass,s,t,j}`", "Material holdup, solid phase", "``solids.solids_material_holdup``"
+   ":math:`{\dot{h}}_{energy,s,t}`", "Energy accumulation, solid phase", "``solids.solids_energy_accumulation``"
+   ":math:`{\dot{h}}_{mass,s,t,j}`", "Material accumulation, solid phase", "``solids.solids_material_accumulation``"
+   ":math:`r_{s,t,r}`", "Solid phase reaction rate of reaction r", "``solids.reactions.reaction_rate``"
+   ":math:`S_{s,t}`", "Total mass of solids", "``solids.mass_solids``"
+   ":math:`T_{s,t,x}`", "Solid phase temperature", "``solids.temperature``"
+   ":math:`x_{mass,s,t,j}`", "Mass fraction of component j, solid phase", "``solids.mass_frac_comp``"
+   ":math:`V_{s}`", "Total volume of solids", "``solids.volume_solid``"
+   "*Greek letters*", " ", " "
+   ":math:`\varepsilon`", "Reactor bed porosity", "``particle_porosity``"
```
chineduobiora, line 117: `varepsilon` is the particle porosity. The reactor bed porosity is not used in this model

idaes/gas_solid_contactors/flowsheets/dyn_TGA_example.py
```diff
@@ -0,0 +1,100 @@
...
+# Lawrence Berkeley National Laboratory,  National Technology & Engineering
+# Solutions of Sandia, LLC, Carnegie Mellon University, West Virginia University
+# Research Corporation, et al.  All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and
+# license information.
+#################################################################################
+"""
+Flowsheet example of the 2-region bubbling fluidized bed (BFB) model
+for a methane combustion with iron-oxide case study
```
chineduobiora, line 15: Change this description to something on the model i.e. "flowsheet example of the 0DFixedBed model for an iron-oxide reduction with methane case study."

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,330 @@
...
+**default** = True. Fixed beds must be dynamic."""))
+    CONFIG.declare("has_holdup", ConfigValue(
+        default=True,
+        domain=In([True]),
+        description="Holdup construction flag - must be True",
+        doc="""Indicates whether holdup terms should be constructed or not.
+**default** - True. Fixed bed reactors must be dynamic, thus this must be
+True."""))
+    CONFIG.declare("energy_balance_type", ConfigValue(
+        default=EnergyBalanceType.useDefault,
```
chineduobiora, line 61: Update default `EnergyBalanceType.useDefault` to `EnergyBalanceType.enthalpyTotal`.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,330 @@
...
+        domain=In(EnergyBalanceType),
+        description="Energy balance construction flag",
+        doc="""Indicates what type of energy balance should be constructed,
+**default** - EnergyBalanceType.useDefault.
```
chineduobiora, line 65: Update default `EnergyBalanceType.useDefault` to `EnergyBalanceType.enthalpyTotal`.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,330 @@
...
+**Valid values:** {
+**EnergyBalanceType.useDefault - refer to property package for default
+balance type
```
chineduobiora, line 68: Delete

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,330 @@
...
+**PhysicalParameterObject** - a PhysicalParameterBlock object.}"""))
+    CONFIG.declare("gas_property_package_args", ConfigBlock(
+        implicit=True,
+        description="Arguments to use for constructing gas phase "
+        "property packages",
+        doc="""A ConfigBlock with arguments to be passed to a gas phase
+property block(s) and used when constructing these,
+**default** - None.
+**Valid values:** {
+see property package for documentation.}"""))
```
chineduobiora, line 91: Update to `ConfigValue` with `default={}`, and `domain=dict`. Take a look at the equivalent config argument in the MB for `property_package_args` as an example.

chineduobiora (commented):
I made a few minor comments.

chineduobiora (dismissed):

idaes/gas_solid_contactors/unit_models/moving_bed.py
```diff
@@ -1,15 +1,15 @@
-#################################################################################
+###############################################################################
```
andrewlee94, line 1: A general observation (so don't go fixing it), but you generally shouldn't change this from the standard form. For one, there is a test that checks the header is present so you don't want to accidentally break that test.

idaes/gas_solid_contactors/unit_models/tests/test_FB0D.py
```diff
@@ -0,0 +1,481 @@
...
+# Import FixedBed0D unit model
+from idaes.gas_solid_contactors.unit_models.fixed_bed_0D import FixedBed0D
+
+# Import property packages
+from idaes.gas_solid_contactors.properties.methane_iron_OC_reduction. \
+    gas_phase_thermo import GasPhaseParameterBlock
+from idaes.gas_solid_contactors.properties.methane_iron_OC_reduction. \
+    solid_phase_thermo import SolidPhaseParameterBlock
+from idaes.gas_solid_contactors.properties.methane_iron_OC_reduction. \
+    hetero_reactions import HeteroReactionParameterBlock
```
andrewlee94, line 48: You should have a test for unit consistency, and I do not see you importing the Pyomo tools for that.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,361 @@
...
+        description="Arguments to use for constructing reaction packages",
+        doc="""A ConfigBlock with arguments to be passed to a reaction block(s)
+and used when constructing these,
+**default** - None.
+**Valid values:** {
+see reaction package for documentation.}"""))
+
+    def build(self):
+        # Call UnitModel.build to setup dynamics
+        super(FixedBed0DData, self).build()
```
andrewlee94, line 129: This can be `super().build()` - we no longer support older Python versions which require the longer form.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,361 @@
...
+        def volume_bed_constraint(b):
+            return b.volume_bed == (constants.pi * b.bed_height *
+                                    (0.5 * b.bed_diameter) ** 2
+                                    )
+
+        # Solid phase material balance
+        # Volume of solid
+        self.volume_solid = Var(
+                    self.flowsheet().config.time,
+                    initialize=1.0,
```
andrewlee94, line 185: This should have units declared.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,361 @@
...
+
+        # Add solid mass variable and constraint for TGA tracking
+        self.mass_solids = Var(self.flowsheet().config.time,
+                               doc="Total mass of solids",
+                               units=pyunits.kg)
+
+        @self.Constraint(self.flowsheet().config.time,
+                         doc="Calculating total mass of solids")
+        def mass_solids_constraint(b, t):
+            return 1e2*b.mass_solids[t] == 1e2*b.volume_solid[t] * sum(
```
andrewlee94, line 237: We really should be switching to use the IDAES scaling tools rather than hard coding scaling factors in our constraints.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,361 @@
...
+                self.config.reaction_package.
+                reaction_block_class(
+                    self.flowsheet().time,
+                    doc="Reaction properties in control volume",
+                    default=tmp_dict))
+
+        # Volume of reactor
+        self.bed_diameter = Var(initialize=1,
+                                doc='Reactor diameter [m]',
+                                units=pyunits.m)
```
andrewlee94, line 167: Units for all these variables should be derived from the units defined in the property package, otherwise there are going to be unit consistency issues when someone tries to use a property package with different units.

There should also be unit conversion operations wherever two different property packages interact as there is no guarantee that both property packages use the same units).

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,361 @@
...
+                            b.solids[0].temperature)
+
+    def initialize(blk, state_args=None, outlvl=idaeslog.NOTSET,
+                   solver=None, optarg=None):
+        """
+        This is a general purpose initialization routine for simple unit
+        models. This method assumes a single ControlVolume block called
+        controlVolume, and first initializes this and then attempts to solve
+        the entire unit.
+        More complex models should overload this method with their own
```
andrewlee94, line 304: This comment should be updated, as it is for a general unit model.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,361 @@
...
+                    doc="Isothermal solid phase constraint")
+            def isothermal_solid_phase(b, t):
+                if t == b.flowsheet().config.time.first():
+                    return Constraint.Skip
+                else:
+                    return (
+                            b.solids[t].temperature ==
+                            b.solids[0].temperature)
+
+    def initialize(blk, state_args=None, outlvl=idaeslog.NOTSET,
```
andrewlee94, line 297: Should there be separate state args for the gas and solids phases?

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,361 @@
...
+        the entire unit.
+        More complex models should overload this method with their own
+        initialization routines.
+
+        Keyword Arguments:
+            state_args : a dict of arguments to be passed to the property
+                           package(s) to provide an initial state for
+                           initialization (see documentation of the specific
+                           property package) (default = {}).
+            outlvl : sets output level of initialization routine
```
andrewlee94, line 312: These logger levels are all out of date.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,361 @@
...
+        init_log = idaeslog.getInitLogger(blk.name, outlvl)
+        opt = get_solver(solver, optarg)  # create solver
+        opt.options = optarg
+
+        # ---------------------------------------------------------------------
+        # Initialize control volume block
+        init_log.info('Initialize Thermophysical Properties')
+
+        # Initialize gas_phase block
+        gas_phase_flags = blk.gas.initialize(
```
andrewlee94, line 338: You do not have a way to pass initial guesses to the property packages during initialization.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,361 @@
...
+                                optarg=optarg,
+                                solver=solver)
+
+        # Initialize solid_phase properties block
+        solid_phase_flags = blk.solids.initialize(
+                                outlvl=outlvl,
+                                optarg=optarg,
+                                solver=solver)
+
+        print()
```
andrewlee94, line 349: Print statements should be replaced by the logger.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,361 @@
...
+        # Initialize reactions
+        blk.reactions.initialize(outlvl=outlvl,
+                                 optarg=optarg,
+                                 solver=solver)
+
+        init_log.info_high("Initialization Step 1 Complete.")
+
+        # ---------------------------------------------------------------------
+        # Release Inlet state
+        blk.gas.release_state(gas_phase_flags, outlvl+1)
```
andrewlee94, line 360: The `+1` should be removed from the output level now that we are using the logger to manage output.

andrewlee94 (changes requested):
A number of comments, but the big ones are:

1. Units of measurement should be generalized and based on those from the property packages being used.
2. Unit consistency checks need to be added to the tests.
3. The IDAES scaling tools should be used instead of hard-coding scaling factors in constraints,

idaes/gas_solid_contactors/unit_models/moving_bed.py
bpaul4: In general, three of the header lines run long and the linting check flags it. I'll leave them alone from now on.

andrewlee94: Yes, the linter does complain, but that means the header script should be updated instead.

lbianchi-lbl: Oof, sorry for being caught in the middle of the two checkers squabbling with each other... I'll take a look (possibly together with @dangunter for the headers part)

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
bpaul4: After importing the scaling tools as `sc`, would adding `sc.__scale_constraint(mass_solids_constraint(b, t), 100)` on line 240 be the correct way to do this?

andrewlee94: No - you should add a separate method named `calculate_scaling_factors` where you use the `constraint_scaling_transformation` method. Look at Line 1024 in https://github.com/IDAES/idaes-pse/blob/main/idaes/generic_models/unit_models/pressure_changer.py.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
bpaul4: @chineduobiora please comment on this, but I don't believe we need to keep `state_args` as an argument here. It doesn't appear to be used by the initialization routine, as the gas and solid phase flags are pre-defined earlier in the `build()` method using separate state args as inputs.

bpaul4: I see that the other two CLC models pass `gas_phase_state_args` and `solid_phase_state_args` to the initialization routines for the gas and solid inlet blocks. I could add them to the flags statements as arguments, although the gas phase is only required for the reactions and has fixed properties.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
bpaul4: See my comments above.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
chineduobiora: @bpaul4 yes there should be state args for the gas and solid phases just like for the other two gas solid contactor models. At the moment, the gas phase has fixed properties and is only required for the reactions, however the gas density/composition still need to be initialized from the fixed state properties.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
chineduobiora: @bpaul4 take a look at how things are setup in the MB to make the updates. Let me know if you need any help.

bpaul4: I updated the initialization routine, thank you @andrewlee94 and @chineduobiora for your suggestions.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
bpaul4: @andrewlee94 I defined variable units from the appropriate property package, in this case `units_meta_solid` which contains information from the solid properties. However, I'm running into a unit consistency error when running the tests - it seems that internally an added gas property is pressure in bar, while other properties use SI units (kJ, kg, m, s):

`E pyomo.core.base.units_container.InconsistentUnitsError: Error in units found in expression: s.gas_props.gas_const*fs.unit.gas 0.0].dens_mol*fs.unit.gas[0.0].temperature*0.01 - fs.unit.gas 0.0].pressure: kilojoule / meter ** 3 not compatible with bar.`

How would you recommend adding the unit conversion option to the FB0D model? The only interaction between gas/solid properties is the reaction package, and this error seems to be solely a gas-phase issue.

andrewlee94: I see a few things that could be wrong there.

1) The gas_constant might need to be converted from SI to be consistent with everything else (depends on how the gas constant is defined in the properties).
2. What is the `0.01` term, and does it have units as well?

You need to use `pyunits.convert(var_or_expr, to_units=units_of_properties)` wherever there is a possibility of units being different.

bpaul4: The issues seem to be with the CLC property packages and not the unit model files. All of the regular IDAES tests pass with the updates to the FB0D initialization and unit definitions, and all four CLC flowsheet examples run and solve. However, all four flowsheets throw InconsistentUnitsError messages when queried with `assert_units_consistent(m)`. Since the FB0D model is working and all other tests pass, I propose finishing the FB0D files without the unit consistency test for now (MB and BFB don't have them either) and opening a new PR to add units to the property packages, and update the scaling and variable units in the MB and BFB models.

Some more detail on the specific errors I encountered:

When I check unit consistency for the examples, the reaction rate expression has a unit mismatch from a non-integer reaction order exponent. When I add an assert units test in the FB0D test file, I run into errors in the gas and solid properties. The one from above in the gas properties uses a dimensionless scaling factor instead of a `pyunit.convert` statement (kJ to J, then Pa to bar). After fixing the gas properties issue, the solid properties enthalpy Shomate expression runs into issues since the coefficients have no unit containers. I'll go through the gas, solid and reaction property packages and update the units to clear the issues.

chineduobiora: @andrewlee94 I'm onboard with @bpaul4 suggestion of completing this PR without the unit consistency test and subsequently opening a separate PR to add units to the property packages and update the scaling and variable units in the gas solid contactor models.

andrewlee94: Seeing as it sounds like there is some work required at the properties level first, that is probably the best idea. I just want to push this issue as I believe the gas-solids contactors are the only models that aren't using units and scaling at this point.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -0,0 +1,371 @@
...
+                self.flowsheet().config.time,
+                doc="Isothermal solid phase constraint")
+            def isothermal_solid_phase(b, t):
+                if t == b.flowsheet().config.time.first():
+                    return Constraint.Skip
+                else:
+                    return (
+                        b.solids[t].temperature ==
+                        b.solids[0].temperature)
+
```
chineduobiora, line 301: I missed this previously but please can you update the documentation with this.

bpaul4: Yes, I'll update the documentation.

chineduobiora (approved):

chineduobiora (approved):

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -163,14 +167,14 @@ def build(self):
...
         self.bed_height = Var(initialize=1,
-                              doc='Bed length [m]',
-                              units=pyunits.m)
+                              doc='Bed length',
+                              units=units_meta_solid('length'))
         self.volume_bed = Var(initialize=1.0,
-                              doc="Volume of the reactor bed [m**3]",
-                              units=pyunits.m**3)
+                              doc="Volume of the reactor bed",
+                              units=units_meta_solid('length')**3)
```
andrewlee94, line 177: For reference (don't fix now), you can do `derived_units("volume")` - a lot of common unit combinations are declared in the derived units metadata.

idaes/gas_solid_contactors/unit_models/tests/test_FB0D.py
```diff
@@ -205,6 +206,10 @@ def test_solution(self, iron_oc):  # need to update these values
         assert (pytest.approx(1255.59, abs=1e-2) ==
                 iron_oc.fs.unit.solids[3600].temperature.value)
 
+    @pytest.mark.component
```
andrewlee94, line 209: This test should be commented out entirely for now.

idaes/gas_solid_contactors/unit_models/tests/test_FB0D.py
```diff
@@ -417,6 +422,10 @@ def test_solution(self, iron_oc):  # need to update these values
         assert (pytest.approx(1273.15, abs=1e-2) ==
                 iron_oc.fs.unit.solids[3600].temperature.value)
 
+    @pytest.mark.component
+    def test_units_consistent(self, iron_oc):
```
andrewlee94, line 426: As above.

andrewlee94 (approved):
A couple of minor comments, and whilst two really should be fixed I won't hold things up any longer.

chineduobiora:
@bpaul4 Great job pushing this through to completion.

PR #455, opened 2021-08-11
Improved Surrogates API

andrewlee94:
## Summary/Motivation:
This PR add the core code for the new Surrogates API along with a rewritten wapper for Alamo that is consistent with this more consistent with the draft SurrogateModelBuilder API. This will also serve as a test case for the new API.

The new Alamopy class will primarily consist of a definition of supported options for ALAMO, a method to write the necessary .alm file, a method to call the ALAMO executable and a method to read the ALAMO results back in and put them into a form that supports the IDAES SurrogateModelObject API. All other data manipulation (initial sample, plotting, etc.) will be handled by the general SurrogateModelTrainer class methods.

- Adds new base classes for surrogate models trainers and objects which define the new surrogates API
- Adds new metrics module for calculating fitting metrics of surrogate models, which is compatible with all surrogate models that use the new API.
- Adds some new sampling utiltiies for manipulating datasets
- Adds initial plotting tools for visualising surrogate models (functional but need improvement)
- Renamed old `alamopy` folder to `alamopy_depr` - Helmet and Riper depend on this old interface still so it cannot be completely deprecated yet.
- Adds new `alamopy` module with the rebuilt Alamo wrapper.
- Lots of tests

### Legal Acknowledgement


idaes/surrogate/alamopy_new.py
```diff
@@ -0,0 +1,504 @@
...
+                    af.write(f"{o} {self.config[o].value}\n")
+                elif isinstance(self.config[o], bool):
+                    # Cast bool to int
+                    af.write(f"{o} {int(self.config[o])}\n")
+                elif isinstance(self.config[o], (str, float, int)):
+                    # Write value to file
+                    af.write(f"{o} {self.config[o]}\n")
+                else:
+                    # Assume the argument is a list
+                    af.write("XMIN {0}\n".format(
```
eslickj, line 488: Should XMIN be {o}

idaes/surrogate/alamopy_new.py
```diff
@@ -0,0 +1,504 @@
...
+            if self.config.monomialpower is not None:
+                af.write("MONO {0}\n".format(len(self.config.monomialpower)))
+            if self.config.multi2power is not None:
+                af.write("MULTI2 {0}\n".format(len(self.config.multi2power)))
+            if self.config.multi3power is not None:
+                af.write("MULTI3 {0}\n".format(len(self.config.multi3power)))
+            if self.config.ratiopower is not None:
+                af.write("RATIOS {0}\n".format(len(self.config.ratiopower)))
+
+            for o in supported_options:
```
eslickj, line 473: Do you really need supported_options?  Why not iterate through self.config?

idaes/surrogate/alamopy_new.py
```diff
@@ -0,0 +1,504 @@
...
+    MADp = 8
+
+
+class Screener(Enum):
+    none = 0
+    lasso = 1
+    SIS = 2
+
+
+supported_options = [
```
bpaul4, line 42: Should trace and tracefname be included in this list? They are additional ALAMO options to create and name a trace .trc file, and should be supported.

idaes/surrogate/alamopy_new.py
```diff
@@ -0,0 +1,504 @@
...
+        default=None,
+        domain=list_of_ints,
+        description="Vector of three variable combinations of powers to be "
+        "considered as basis functions."))
+    CONFIG.declare('ratiopower', ConfigValue(
+        default=None,
+        domain=list_of_ints,
+        description="Vector of ratio combinations of powers to be considered "
+        "in the basis functions."))
+    CONFIG.declare('constant', ConfigValue(
```
bpaul4, line 158: I like that linear and constant terms are assumed, similar to the GUI. This should be noted in the documentation, though, since it's not immediately obvious to new users.

idaes/surrogate/alamopy_new.py
andrewlee94: Because the config has more options that should not be included, and seeing as it inherits from the base class we do not have full control of the config list.

idaes/surrogate/alamopy_new.py
andrewlee94: No - it is explicitly included separately.

idaes/surrogate/alamopy_new.py
```diff
@@ -0,0 +1,504 @@
...
+
+from idaes.surrogate.my_surrogate_base import Surrogate
+from idaes.core.util.config import list_of_ints, list_of_floats
+from pyomo.common.config import ConfigValue, In
+
+
+TEMPPATH = os.path.dirname(__file__)
+
+
+# The values associated with these must match tose expected in the .alm file
```
bpaul4, line 24: typo, 'tose'

idaes/surrogate/alamopy_new.py
```diff
@@ -0,0 +1,504 @@
...
+
+    This mainly defines a set of configuration options for ALAMO along with
+    methods to read and write the ALAMO input and poutput files and to call
+    the ALAMO executable.
+
+    Generally, options default to None to indicate that no entry will be
+    written in the ALAMO input file. In this case, the default ALAMO settings
+    will be used.
+    """
+    # The follwoing ALAMO options are not (yet) supported
```
bpaul4, line 112: typo, 'follwoing'

idaes/surrogate/alamopy_new.py
```diff
@@ -0,0 +1,504 @@
...
+    # Returning model prediction, as we can do that in IDAES
+    # Ininital sampling of data, as SurrogateModelBuilder should do that
```
bpaul4, line 114: typo, 'Ininital'

idaes/surrogate/alamopy_new.py
```diff
@@ -0,0 +1,504 @@
...
+    # Single valdiation set, due to limitations of current API
```
bpaul4, line 115: typo, 'valdiation'

idaes/surrogate/alamopy_new.py
eslickj: Not really a big deal, but you could probably use a sub-block for options that should be included.  Like alamo_config maybe?  Save the work of maintaining the supported list.  I don't know the details of how to use this, so if that doesn't make sense disregard.

eslickj: Yeah, I was worried about that.  I guess the right answer is whatever is most intuitive for the user.

idaes/surrogate/alamopy_new.py
```diff
@@ -0,0 +1,504 @@
...
+NCUSTOMBAS
+"""
+
+
+class Alamopy(Surrogate):
+    """
+    Standard SurrogateModelTrainer for ALAMO.
+
+    This mainly defines a set of configuration options for ALAMO along with
+    methods to read and write the ALAMO input and poutput files and to call
```
bpaul4, line 105: Didn't catch this before, 'poutput'

idaes/core/util/config.py
```diff
@@ -107,6 +107,33 @@ def list_of_floats(arg):
     return lst
 
 
+def list_of_ints(arg):
+    '''Domain validator for lists of integers
```
dangunter, line 111: note that 'black' may correct these to triple-double quotes

andrewlee94: At some point ,we need to just bite the bullet and run `black` over the code based and then add a test to enforce that standard from now on,

As it is, I am hesitant to run `black` over code in an actual development PR as it will obscure the real changes.

idaes/core/util/config.py
```diff
@@ -107,6 +107,33 @@ def list_of_floats(arg):
...
+    Returns:
+        List of ints
+    '''
+    lst = []
+    try:
+        iter(arg)
+    except TypeError:
+        arg = [arg]
+
+    for v in arg:
```
dangunter, line 125: my preferred strategy is to try to convert to a int and raise an error if it fails. This way anything that implements `__int__` would be ok.
e.g. 
```
try:
    i = int(v)
except (ValueError, TypeError) as e:
    try:
        i = int(float(v))
    except (ValueError, TypeError) as e:
        raise RuntimeError(f"fuggedaboutit with that {v}")
```

idaes/surrogate/tests/test_alamopy.py
```diff
@@ -0,0 +1,350 @@
...
+    alm_obj.config.mipoptcr = 13.6
+    alm_obj.config.linearerror = False
+    alm_obj.config.GAMS = "foo"
+    alm_obj.config.GAMSSOLVER = "bar"
+    alm_obj.config.solvemip = False
+
+    stream = io.StringIO()
+    alm_obj.write_alm_to_stream(stream=stream)
+
+    assert stream.getvalue() == (
```
dangunter, line 207: This test should be not sensitive to things that Alamo also doesn't care about, like the order of keywords in the file. Suggest doing a comparison between `set(stream.getvalue().split("\n")` and a set constant with the strings ` {""# IDAES Alamopy input file", "NINPUTS 2", "NOUTPUTS 1", ...}`. You may want to do a partial match that just the strings in the set constant are in the Alamopy result, so later additions don't break the test.

idaes/core/util/config.py
andrewlee94: I believe I did it this way to explicitly exclude `bools` (which can be cast to `ints`). I am not sure if that is a good idea or not, but the idea was that if you set a `bool` where an `int` was expected it was more than likely that you got something wrong (conceptually at least).

idaes/surrogate/tests/test_alamopy.py
andrewlee94: I am on the fence with that one - whilst ALAMO doesn't care about the order (with a couple of minor exceptions), the order *is* indicative of how the writer is working, I.e. if the order changes then something changed in the writer that maybe we should know about.

idaes/surrogate/alamopy_new.py
```diff
@@ -0,0 +1,673 @@
...
+        description="Whether to use an optimizer to solve MIP."))
+    # CONFIG.declare('log_output', ConfigValue(
+    #     default=False,
+    #     domain=In([True, False]),
+    #     description="Whether to log ALAMO output."))
+
+    # I/O file options
+    CONFIG.declare("temp_path", ConfigValue(
+        default=DEFAULTPATH,
+        domain=str,
```
jsiirola, line 419: If you are going to do this here, I recommend the `Path` domain and not just `str`)

idaes/surrogate/alamopy_new.py
```diff
@@ -0,0 +1,673 @@
...
+        trcname = almname.split(".")[0] + ".trc"
+
+        return almname, trcname
+
+    def write_alm_to_stream(
+            self, stream=None, trace_fname=None, x_reg=None,
+            z_reg=None, x_val=None, z_val=None):
+        """Write the input file for the ALAMO executable to a stream."""
+        if stream is None:
+            raise TypeError("No stream provided to ALAMO writer.")
```
jsiirola, line 481: Why provide a default value for `stream` if it is required?

idaes/core/util/config.py
jsiirola: I actually prefer the following:
```
i = int(v)
if i != float(v):
    raise ValueError(f"Value {v} is not an integer")
return i
```

idaes/surrogate/alamopy_new.py
```diff
@@ -0,0 +1,673 @@
...
+    # CONFIG.declare('log_output', ConfigValue(
+    #     default=False,
+    #     domain=In([True, False]),
+    #     description="Whether to log ALAMO output."))
+
+    # I/O file options
+    CONFIG.declare("temp_path", ConfigValue(
+        default=DEFAULTPATH,
+        domain=str,
+        description="Path for directory to store temp files."))
```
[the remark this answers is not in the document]
andrewlee94: That is on the TODO list, along with a more meaningful default name for the temp files (probably something including data).

idaes/surrogate/alamopy_new.py
```diff
@@ -0,0 +1,673 @@
...
+        # Read back results
+        trace_dict = self.read_trace_file(trcname)
+
+        # Populate results and SurrogateModel object
+        self.populate_results(trace_dict)
+        self.build_surrogate_model_object()
+
+        # Clean up temporary files if required
+        if not self.config.keep_files:
+            self.remove_temp_files(almname, trcname)
```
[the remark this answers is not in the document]
andrewlee94: I assume you mean put most of the `build_surrogate` method in the `try` (i.e. the part which creates the files) with the `finally` part being the `remove_temp_files` (plus, the remove files should make sure that an error in removing the first file doesn't prevent the removal of the following files).

idaes/surrogate/alamopy_new.py
```diff
@@ -0,0 +1,839 @@
...
+        """
+        Method to read the results of an ALAMO run from a trace (.trc) file.
+        The name location of the trace file is tored on the AlamoModelTrainer
+        object and is generally set automatically.
+
+        Args:
+            None
+
+        Returns:
+            trace_dict: contents ofr trace file as a dict
```
bpaul4, line 688: Typo, 'ofr'

bpaul4:
@andrewlee94 An example I'm testing using the old Alamopy on three inputs and two outputs failed because x3 is a list of 0's (XMAX - XMIN = 0 is the specific ALAMO error). However, if there is a prior input file present ALAMO will skip printing the error and continue the run with the previously loaded data so that the surrogate returned matches the previous run, not the current dataset. I believe the failsafe here should default to a print message and aborting the run, since as a new user it took me a while to understand why my fit and training data did not match.

andrewlee94:
@bpaul4 I am not quite sure I understand your problem. Is this involving the old ALAMOpy or the new version? If this is about the old version, then this is not the place for that issue, and ultimately I would not file an issue as that code will be going away rathe than being fixed.

If it is about the new version, then could you provide more information:

First, did you set the bounds on the inputs, as you need to provide that information.
Second, what do you mean by "if another input file is present"? There should be no other input file involved.

bpaul4:
It is involving the old Alamopy, but the error comes from the ALAMO solver itself so the new version may run into the same issue. I did set the bounds on the inputs. The wrapper generates a new input file on each run, either a custom name or 'temp.alm' and will overwrite a previous .alm with the same name. If an older file exists in the folder already (e.g. calling the wrapper on multiple data sets or output variables), and the XMIN - XMAX = 0 error occurs, it will continue with the existing input file instead of breaking.

andrewlee94:
@bpaul4 Where exactly is that XMAX-XMIN error occurring? If that is part of ALAMO itself then that issue needs to go to the ALAMO developers. If that was a check in the old ALAMOpy code (which I am suspecting), then that is a bug in that code that we won't bother to fix).

bpaul4:
The error comes from ALAMO itself.

andrewlee94:
@bpaul4 Then there isn't really a lot we can do about it, short of catching the error ourselves first, but I suppose we could add a check for it. To be clear however, what did you do to trigger that error? Was it a matter of having XMAX == XMIN in the set up?

andrewlee94:
@bpaul4 Either way - you should contact the ALAMO developers with an example of this bug so they can fix the underlying problem. ALAMO should not be continuing after encountering that error.

bpaul4:
The error triggered for me when I passed the data, so it should have detected the issue and stopped running. I'll contact the ALAMO developers about this behavior.

idaes/surrogate/alamopy_new.py
```diff
@@ -0,0 +1,1069 @@
...
+            stream.write(f"TRACEFNAME {trace_fname}\n")
+
+        stream.write("\nBEGIN_DATA\n")
+        # Columns will be writen in order in input and output lists
+        training_data.to_string(
+            buf=stream,
+            columns=self._input_labels + self._output_labels,
+            header=False,
+            index=False,
+            float_format=lambda x: str(x).format(":g"))
```
lbianchi-lbl, line 671: @andrewlee94 re: the tests failing on Python 3.6 because of extra spaces, I was thinking that it's maybe worth a quick try if something like this would address those (also applied to other places where `DataFrame.to_string()` is used:

```suggestion
            float_format=lambda x: str(x).format(":g").strip())

```

jsiirola (approved):

andrewlee94:
@dangunter Not for this PR. We already have people clamoring to use this (although I don't see them rushing to review it) so we want to get it merged ASAP.

eslickj (approved):

andrewlee94:
The test failures are due to examples failing thanks to the deprecation of the old ALAMO wrapper. There is a PR open on the examples repo to fix this, but this will need to be merged first.

eslickj (approved):

jsiirola (approved):

PR #457, opened 2021-08-12
Fixing issue with IndexedArcs in propagate_state

bknueven:
## Summary/Motivation:
`idaes.core.util.initialization.propagate_state` currently handles indexed arcs incorrectly. This PR corrects this issue.

- Deprecate and move the functionality of `idaes.core.util.misc.copy_port_values` to `propagate_state`. `propagate_state` should be able to replace all instances of `copy_port_values` without issue.
- More forgiving duck-typing in `propagate_state` to handle indexed Arcs/Ports
- Tests now check indexed arcs and ports

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/initialization.py
+++ b/idaes/core/util/initialization.py
@@ -15,14 +15,14 @@
 This module contains utility functions for initialization of IDAES models.
 """
 
-from pyomo.environ import Block, Var, TerminationCondition, Constraint
-from pyomo.network.arc import _ArcData
+from pyomo.environ import (
+    Block, Constraint, Param, TerminationCondition, Var, value)
+from pyomo.network import Arc
 from pyomo.dae import ContinuousSet
 from pyomo.core.expr.visitor import identify_variables
 
 from idaes.core import FlowsheetBlock
 from idaes.core.util.exceptions import ConfigurationError
-from idaes.core.util.misc import copy_port_values
 from idaes.core.util.model_statistics import degrees_of_freedom
 from idaes.core.util.dyn_utils import (
     get_activity_dict,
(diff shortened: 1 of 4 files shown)


andrewlee94 (dismissed):
I don't see any issues with this.

eslickj (approved):
I think it's good.  I guess it also lets me know we have multiples of this method, since there is also a copy_port_values.

andrewlee94:
@eslickj Yes - we do have a duplicate of this method, which I discovered when compiling the list of utility functions in IDAES.

bknueven:
It seems like `copy_port_values` makes a similar error. I could make the same change to `copy_port_values` in this PR.

andrewlee94:
@bknueven I would redirect `copy_port_values` to the `propagate_state` method and add a deprecation warning. We shouldn't keep maintaining two versions of the same method.

bknueven:
> @bknueven I would redirect `copy_port_values` to the `propagate_state` method and add a deprecation warning. We shouldn't keep maintaining two versions of the same method.

Actually, this is no so simple. `copy_port_values` has a more general function handle than `propagate_stage`, the later of which only allows the user to specify an arc.
https://github.com/IDAES/idaes-pse/blob/b2d79aba3f6580393687dde51d7c5992dc28040f/idaes/core/util/misc.py#L230

andrewlee94:
@bknueven Then maybe we should move the `copy_port_values` method over to replace `propagate_state`, then replace it with a redirect and deprecation?

bknueven:
My most recent commit is an attempt to eliminate redundant code between the two functions. Because `propagate_state` also has the `direction` kwarg it's not a strict subset of the `copy_port_values` function handle either.

eslickj:
I'm guilty of using the _ArcData like this too.  I've done it a lot, but it always kind on bothered me.  Should we be checking ctype here?  Like find a way that we aren't using protected things.

Like for example:

```py
import pyomo.environ as pyo
m = pyo.ConcreteModel()
m.x = Var([1,2,3])
issubclass(m.x[1].ctype, pyo.Var)
```

@jsiirola, Pyomo must have a right way of doing this.  I'm too lazy to look through the code at the moment.

eslickj:
I think we could add direction to copy port values to cover everything.  Which function name we end up deprecating doesn't matter.

andrewlee94:
Functionally it doesn't matter which is deprecated, but I think this method is mostly used during initialization and thus to me makes most sense to live in `util.initialization`.

eslickj:
@andrewlee94 yeah that makes sense.  So we should make sure ```propagate_states``` covers everything, and deprecate ```copy_port_values```.  Then we only need to deal with updating uses of  ```copy_port_values```.

andrewlee94:
@eslickj That was my thought at least.

jsiirola:
@eslickj: I think the idiom I would recommend depends on the context.  For the most part, I recommend the Python duck-typing idiom: just use it, and if it fails, catch and reraise the exception with a meaningful log message.  Sometimes - if it is particularly critical to check an incoming model component, the preferred thing is to check the ctype.

```diff
diff --git a/idaes/core/util/initialization.py b/idaes/core/util/initialization.py
index 7fd4871..7c5f7c6 100644
--- a/idaes/core/util/initialization.py
+++ b/idaes/core/util/initialization.py
@@ -1,4 +1,4 @@
-# -*- coding: UTF-8 -*-
+# -*- coding: utf-8 -*-
 #################################################################################
 # The Institute for the Design of Advanced Energy Systems Integrated Platform
 # Framework (IDAES IP) was produced under the DOE Institute for the
@@ -135,7 +135,7 @@ def propagate_state(stream, direction="forward"):
     Returns:
         None
     """
-    if not isinstance(stream, Arc):
+    if stream.ctype is not Arc:
         raise TypeError("Unexpected type of stream argument. Value must be "
                         "a Pyomo Arc.")
 
```
```diff
diff --git a/idaes/core/util/misc.py b/idaes/core/util/misc.py
index d0aa3a9..a22e6a6 100644
--- a/idaes/core/util/misc.py
+++ b/idaes/core/util/misc.py
@@ -241,7 +241,7 @@ def copy_port_values(destination=None, source=None, arc=None):
         None
     """
     # Allow an arc to be passed as a positional arg
-    if destination is not None and source is None and isinstance(destination, Arc):
+    if destination is not None and source is None and destination.ctype is Arc:
         arc = destination
         destination = None
     # Check that only arc or source and destination are passed
@@ -249,17 +249,17 @@ def copy_port_values(destination=None, source=None, arc=None):
         raise RuntimeError(
             "In copy_port_values(), must provide source and destination or arc")
     if arc is not None:
-        if not isinstance(arc, Arc):
+        if arc.ctype is not Arc:
             raise RuntimeError(
                 "In copy_port_values(), arc argument is not an instance of an Arc")
         if destination is not None or source is not None:
             raise RuntimeError(
                 "In copy_port_values(), provide only arc or source and destination")
-    if destination is not None:
-        if not isinstance(destination, Port):
+    elif destination is not None:
+        if destination.ctype is not Port:
             raise RuntimeError(
                 "In copy_port_values(), destination is not an instance of Port")
-        if not isinstance(source, Port):
+        if source.ctype is not Port:
             raise RuntimeError(
                 "In copy_port_values(), source is not an instance of Port")
 
@@ -270,10 +270,18 @@ def copy_port_values(destination=None, source=None, arc=None):
 
     # Copy values
     for k, v in destination.vars.items():
-        if isinstance(v, pyo.Var):
+        try:
             for i in v:
                 v[i].value = pyo.value(source.vars[k][i])
-
+        except AttributeError:
+            if getattr(v, 'ctype', None) not in {Var, Param}:
+                _log.error("Port (%s) contains a member (%s) that "
+                           "is not a Var / mutable Param. copy_port_values() "
+                           "works by assigning to the 'value' attribute, "
+                           "thus can only be used when Port members are "
+                           "Pyomo Vars or mutable Params."
+                           % (destination.name, k)
+            raise
 
 def set_param_from_config(b, param, config=None, index=None):
     """
```

eslickj:
@jsiirola, thanks!

bknueven:
I believe I have addressed @eslickj and @andrewlee94's thoughts around how to deprecate `copy_port_values` and replace it with `propagate_state`.

I also switched most of the typing checks in `propagate_state` to simply use duck-typing, except in one place where the `ctype` is used by necessity.

andrewlee94 (approved):
Looks good to me.

eslickj (approved):

PR #460, opened 2021-08-12
Added Cp, Cv, speed of sound equations to the generic CEOS

radhakrishnatg:
- Added equations for computing cp_mol_phase, cv_mol_phase, heat_capacity_ratio_phase
- Added new methods for computing isentropic_speed_sound_phase and isothermal_speed_sound_phase

### Legal Acknowledgement

### DIFF
--- a/idaes/generic_models/properties/core/eos/ceos.py
+++ b/idaes/generic_models/properties/core/eos/ceos.py
@@ -349,7 +349,92 @@ class Cubic(EoSBase):
         else:
             raise PropertyNotSupportedError(_invalid_phase_msg(b.name, p))
 
-    # TODO: Need to add functions to calculate cp and cv
+
+    @staticmethod
+    def cp_mol_phase(blk, p):
+        pobj = blk.params.get_phase(p)
+        if not pobj.is_vapor_phase():
+            raise PropertyNotSupportedError(_invalid_phase_msg(blk.name, p))
+
+        cname = pobj._cubic_type.name
+        am = getattr(blk, cname+"_am")[p]
+        bm = getattr(blk, cname+"_bm")[p]
+        A = getattr(blk, cname+"_A")[p]
+        B = getattr(blk, cname+"_B")[p]
+        fw = getattr(blk, cname+"_fw")
+        kappa = getattr(blk.params, cname+"_kappa")
+        dadT = getattr(blk, cname+"_dadT")[p]
+        Z = blk.compress_fact_phase[p]
+
+        EoS_u = EoS_param[pobj._cubic_type]['u']
+        EoS_w = EoS_param[pobj._cubic_type]['w']
+        omegaA = EoS_param[pobj._cubic_type]['omegaA'] 
+        EoS_p = sqrt(EoS_u**2 - 4*EoS_w)
+
+        d2adT2 = - (0.5 / blk.temperature) * dadT + \
+                 ((Cubic.gas_constant(blk)**2 * omegaA) / (2*blk.temperature)) * (
+                     sum(sum(
+                         blk.mole_frac_phase_comp[p, i] * blk.mole_frac_phase_comp[p, j]
+                         * (1 - kappa[i, j]) * fw[i] * fw[j] * sqrt(
+                             (blk.params.get_component(i).temperature_crit * blk.params.get_component(j).temperature_crit) /
+                             (blk.params.get_component(i).pressure_crit * blk.params.get_component(j).pressure_crit)
+                         )
+                         for j in blk.components_in_phase(p))
+                         for i in blk.components_in_phase(p))
+                 )
+
+        dBdT = -B / blk.pressure
+        dAdT = (A/am) * dadT - (2*A/blk.temperature)
+        K2 = (EoS_u - 1) * B - 1
+        K3 = A - EoS_u * B - (EoS_u - EoS_w) * B**2
+        K4 = - (A * B + EoS_w * B**2 + EoS_w * B**3)
+        dK2dT = (EoS_u - 1) * dBdT
+        dK3dT = dAdT - EoS_u * dBdT - 2 * (EoS_u - EoS_w) * B * dBdT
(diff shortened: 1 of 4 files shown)


idaes/generic_models/properties/core/eos/ceos.py
```diff
@@ -349,7 +349,92 @@ def dens_mol_phase(b, p):
         else:
             raise PropertyNotSupportedError(_invalid_phase_msg(b.name, p))
 
-    # TODO: Need to add functions to calculate cp and cv
+
+    @staticmethod
+    def cp_mol_phase(blk, p):
+        pobj = blk.params.get_phase(p)
+        if not pobj.is_vapor_phase():
```
andrewlee94, line 356: This error is incorrect - cubic EoS works for both liquid and vapour phase (even though we use the Ideal Gas cp value)

idaes/generic_models/properties/core/eos/ceos.py
```diff
@@ -349,7 +349,92 @@ def dens_mol_phase(b, p):
...
+        dK4dT = - (A * dBdT + B * dAdT + 2 * EoS_w * B * dBdT + 3 * EoS_w * B**2 * dBdT) 
+        dZdT = - (Z**2 * dK2dT + Z*dK3dT + dK4dT) / (3*Z**2 + 2*K2*Z + K3)
+
+        expression1 = 2 * Z + (EoS_u + EoS_p) * B
+        expression2 = 2 * Z + (EoS_u - EoS_p) * B
+        expression3 = (2 * dZdT + (EoS_u - EoS_p) * dBdT) / expression2
+        expression4 = (2 * dZdT + (EoS_u + EoS_p) * dBdT) / expression1
+
+
+        # Derived from equation on pg. 120 in Properties of Gases and Liquids
```
andrewlee94, line 402: This comment is not correct for this method - can you give a reference to the source you used instead?

idaes/generic_models/properties/core/eos/ceos.py
```diff
@@ -349,7 +349,92 @@ def dens_mol_phase(b, p):
...
+            sum(blk.mole_frac_phase_comp[p, j] * 
+                get_method(blk, "cp_mol_ig_comp", j)(blk, cobj(blk, j), blk.temperature)
+                for j in blk.components_in_phase(p))
+        )
+
+
+    @staticmethod
+    def cv_mol_phase(blk, p):
+        pobj = blk.params.get_phase(p)
+        if not pobj.is_vapor_phase():
```
andrewlee94, line 416: As above.

idaes/generic_models/properties/core/eos/ceos.py
```diff
@@ -772,6 +857,40 @@ def gibbs_mol_phase_comp(b, p, j):
                 b.entr_mol_phase_comp[p, j] *
                 b.temperature)
 
+    @staticmethod
+    def heat_capacity_ratio_phase(blk, p):
+        pobj = blk.params.get_phase(p)
+        if not (pobj.is_vapor_phase() or pobj.is_liquid_phase()):
```
andrewlee94, line 863: Do we need these checks at all? I don't think any of the other methods in the ceos module use them.

idaes/generic_models/properties/core/generic/generic_property.py
```diff
@@ -905,6 +905,8 @@ def define_metadata(cls, obj):
              'gibbs_mol': {'method': '_gibbs_mol'},
              'gibbs_mol_phase': {'method': '_gibbs_mol_phase'},
              'gibbs_mol_phase_comp': {'method': '_gibbs_mol_phase_comp'},
+             'isentropic_speed_sound_phase': {'method': '_isentropic_speed_sound_phase'},
```
andrewlee94, line 908: Well done working out how to add new properties to the framework. Did you manage to understand what each of these steps is doing?

andrewlee94 (changes requested):
This looks really good. I only had a couple of comments on the code itself where there were some tests for phase type that I don't think are needed.

However, there are two important things which are missing:

1. The new properties need to be added to the documentation. The main place to be updated is https://idaes-pse.readthedocs.io/en/stable/user_guide/components/property_package/general/eos/cubic.html
2. We need some tests for the new properties (and really the old ones too but that is not your problem). What I would like to see is a comparison of your model predictions to some literature data (ideally experimental, not another model). Note that we cannot use Aspen for this due to licensing conditions.

radhakrishnatg:
@andrewlee94 I made all the changes you suggested. Now, I just need to look for data in the literature for tests. Also, I have never written tests before, so I need sometime to understand how to write them. Regarding your comments:
(1) 'This error is incorrect - cubic..': I put the check, because the title of the section in the book was heat capacities of real gases and I assumed they hold only for gases. I have removed it now.
(2) I put references at the beginning of the code now.

idaes/generic_models/properties/core/generic/generic_property.py
radhakrishnatg: Thanks, Andrew! I just followed the instructions you gave to Minh. Overall, I understood what each step was doing. There was one thing I was not sure how it worked. It was not crucial, so I thought I would ask you later about that.

idaes/generic_models/properties/core/eos/ceos.py
radhakrishnatg: I followed the structure used in the enthalpy method. It had the checks, so I put them here as well. I have now removed them.

radhakrishnatg:
@andrewlee94 I have resolved the conflicts. But, for some reason, several tests are failing. The heat capacity ratio is the one that seems to be causing the issue. I don't understand why it is making the tests fail though. Could you please help me with this issue? Thank you!

andrewlee94:
@radhakrishnatg Did you run the tests locally before pushing these changes? That failure should have shown up there as well.

The issue is that you didn't add heat capacity ratio to EoS base - there needs to be a placeholder method there (which raises a NotImplementedError) for all supported properties.

radhakrishnatg:
I did not run the tests locally. Is there any documentation or instructions on how to run tests?

jghouse88:
@radhakrishnatg: @dallan-keylogic may be the better person to review this than me.

PR #465, opened 2021-08-18
Adding units to stream_table

### LINKED ISSUE #322: Add units column to stream table in IFV
Add a column to the flowsheet viewer stream table that shows the units for a given row.

There is a possibility that the row may contain different units (ex. lb/ft and kg/m) so you may need to get the units of the first entry of the row and then convert the entire row to those units via `pyunits.convert(var[index], to_units=output_units)`

To get the units:
```
from pyomo.environ import units as pyunits

pyunits.get_units(var[index])
```

Note: `get_units` only works on an element of an indexed variable so you shouldn't use it on a variable container.

elbashandy:
Fixes #322 

## Summary/Motivation:
Adding units to the stream table view that shows up on the UI page. The units are generated randomly using the pyomo library to read the `derived_units` units from the FlowSheetBlock.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/tables.py
+++ b/idaes/core/util/tables.py
@@ -250,18 +250,21 @@ def create_stream_table_dataframe(
             disp_dict = sb.define_display_vars()
         for k in disp_dict:
             for i in disp_dict[k]:
-                if i is None:
-                    stream_attributes[key][k] = value(disp_dict[k][i])
-                    if add_units:
-                        stream_attributes['Units'][k] = str(units.get_units(disp_dict[k][i]))
-                    if k not in full_keys:
-                        full_keys.append(k)
-                else:
-                    stream_attributes[key][f"{k} {i}"] = value(disp_dict[k][i])
-                    if add_units:
-                        stream_attributes['Units'][f"{k} {i}"] = str(units.get_units(disp_dict[k][i]))
-                    if f"{k} {i}" not in full_keys:
-                        full_keys.append(f"{k} {i}")
+                stream_key = k if i is None else f"{k} {i}"
+                stream_attributes[key][stream_key] = value(disp_dict[k][i])
+                if add_units:
+                    pyomo_unit = units.get_units(disp_dict[k][i])
+                    if pyomo_unit:
+                        pint_unit = pyomo_unit._get_pint_unit()
+                        stream_attributes['Units'][stream_key] = {
+                            'raw': str(pyomo_unit),
+                            'html': '{:~H}'.format(pint_unit),
+                            'latex': '{:~L}'.format(pint_unit)
+                        }
+                    else:
+                        stream_attributes['Units'][stream_key] = None
+                if stream_key not in full_keys:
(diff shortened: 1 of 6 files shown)


dangunter:
This PR should be changed to be adding units, even if it is starting as mock units. A PR for adding mock units would never get merged.

idaes/ui/flowsheet.py
```diff
@@ -449,6 +449,26 @@ def _construct_model_json(self):
             )
         )
 
+        if ('Units' in self._stream_table_df and
+            'Variable' in self._stream_table_df):
+            # Styling the units for display
+            def apply_html(value, style=''):
```
dangunter, line 455: Wouldn't this make more sense in the JS client? Seems wrong to hardcode display styles in the server.

idaes/ui/fsvis/static/js/stream_table.js
```diff
@@ -36,7 +36,17 @@ export class StreamTable {
             if (column_header !== "") {
```
dangunter, line 36: could reduce indentation by doing a continue here instead

idaes/ui/fsvis/static/js/stream_table.js
```diff
@@ -36,7 +36,17 @@ export class StreamTable {
...
                 // If the column_header is Variable then we don't want the column to be right-aligned and we want the column to be pinned to the left so when the user scrolls the column scrolls with them
                 if (column_header === "Variable") {
```
dangunter, line 38: This is where you could take the plain text sent from the server and make it into appropriate HTML, e.g. if the server sent "VariableName,m/s^2" you could split on the comma and format.

idaes/ui/fsvis/static/css/main.css
```diff
@@ -335,4 +335,13 @@ input:checked + .slider:before {
 
 .dropdown-item:active {
   background-color: white;
+}
+
```
dangunter, line 339: Please add some comments here saying what these affect and why. I would suggest more specific class names like "streamtable-variable"

dangunter (changes requested):

idaes/core/util/tables.py
```diff
@@ -245,10 +252,14 @@ def create_stream_table_dataframe(
             for i in disp_dict[k]:
                 if i is None:
```
dangunter, line 253: from here to line 264 seems like duplicated code except for the difference in the key. why not do that up front and use the same logic, i.e. `units_key = k if is None else f"{k} {i}"` then use 'units_key' in place of 'k' and not require the `if i is None` decision

dangunter (approved):
LGTM

lbianchi-lbl (approved):
For testing, I called `idaes.ui.fsvis.visualize()` on the flowsheet created by `idaes.ui.tests.test_flowsheet.demo_flowsheet()`, and checked the stream table in the web UI. The units (and therefore this PR) LGTM!

image

PR #472, opened 2021-08-20
Fix mixer and spe scale when components not in a phase

eslickj:
## Fixes
An issue with mixer and separator scaling when components are not in a phase.

### Legal Acknowledgement

### DIFF
--- a/idaes/generic_models/properties/core/generic/generic_property.py
+++ b/idaes/generic_models/properties/core/generic/generic_property.py
@@ -1646,13 +1646,16 @@ class GenericStateBlockData(StateBlockData):
                 pe_form_config[pp].calculate_scaling_factors(self, pp)
 
             for k in self.equilibrium_constraint:
-                sf_fug = self.params.get_component(
-                    k[2]).config.phase_equilibrium_form[
-                        (k[0], k[1])].calculate_scaling_factors(
-                            self, k[0], k[1], k[2])
+                try:
+                    sf_fug = self.params.get_component(
+                        k[2]).config.phase_equilibrium_form[
+                            (k[0], k[1])].calculate_scaling_factors(
+                                self, k[0], k[1], k[2])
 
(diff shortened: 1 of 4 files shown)


chineduobiora (approved):
Looks good to me.

andrewlee94:
@eslickj Do you have a case where this was a problem that I can look at? I cannot see how this became a problem as all the balance equations should be indexed by the phase-component set, thus only valid phase-component pairs should be appearing.

eslickj:
@andrewlee94, I don't have a good example handy.  I changed the model where I originally observed the problem.  It was using the  generic property package with the parameter dict from https://github.com/IDAES/idaes-pse/blob/main/idaes/power_generation/properties/natural_gas_PR.py.  As it is now only water can be liquid.  So I had H2 and water and saw this.  I'll tak a look at the mass balance, but I assumed since I saw this, probably used Constraint.skip in there.

eslickj:
@andrewlee94, after looking into this more. I found and unneeded "if" with an unused constraint skip in Mixer.  That's not the problem though.  The problem is that calculate scale factors iterates through the constraint, and inside that iterates through the phases, when really the phase index is part of the constraint index.  So while I guess this fix works,  it's probably not the right fix, which would be to fix the loop.

andrewlee94:
@eslickj I think I see the problem now. In the Mixer at least, there is an issue when doing a total component balance approach (lines 1010-1013). I think the other branches should be fine with out the `try/except` however as they are already iterating over the phase-component set and so should exclude any infeasible phase-component combinations.

For the total component basis, we could alternatively use and `if` to handle the issue:

```
for i, p in enumerate(self.mixed_state.phase_list):
    if (p, j) not in self.mixed_state.phase_component_set:
        continue
```

The `try/except` for might be faster however, as we can generally assume most phase-component pairs exist.

PR #483, opened 2021-08-27
Use consistent randomness sources in PySMO sampling methods

eyoung55:
## Fixes
This is a small change to use `np.random.rand()` in place of `random.rand()` in the Latin Hypercube method.  This change makes the random source used across all the sampling methods consistent.

## Summary/Motivation:
Being consistent with the random source allows a user to set the RNG seed in a systematic way outside the sampling methods and produce repeatable outputs for testing purposes.

- Use `np.random.rand()` for consistency
- Remove `random` import

### Legal Acknowledgement


andrewlee94 (approved):
Looks good to me.

OOAmusat (approved):
LGTM

PR #489, opened 2021-09-01
Reimplement _get_state_from_port

### LINKED ISSUE #480: Function to get state block from port
We have a function to get a state block from a port in the tables utility module.  Right now it's private, but it seems like something we may want to make public.  It's probably just a matter of removing the underscore and making sure there are tests and documentation.

dallan-keylogic:
## Fixes
-Fixes issue with stream tables for the distillation column described in #96

## Summary/Motivation:
As suggested by #480, I took a look at `_get_state_from_port` to see whether we could make it public. Because it fails in instances where a naïve user might expect it to succeed (such as on the outlet ports of the Flash unit), I elected to not make it public at this time.  In the course of the investigation, however, I found that the previous implementation did not work when a port was extended through Pyomo (such as in the `TrayColumn` feed). Therefore I reimplemented it along the lines @jsiirola suggested in #480 . We can then close #480 as inadvisable at the moment.

- Reimplement `_get_state_from_port`
- Remove the `_state_block` attribute from ports generated through `add_port` functions
- Add additional tests to ensure that the function works reliably for systems with dynamics, `1DControlVolume`s, and other edge cases.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/tables.py
+++ b/idaes/core/util/tables.py
@@ -269,35 +269,38 @@ def stream_table_dataframe_to_string(stream_table, **kwargs):
     )
 
 def _get_state_from_port(port,time_point):
+    # To prevent circular import
     from idaes.core.property_base import StateBlockData
     states = list()
-    v = None # To make PyLint stop complaining
+    vlist = list()
     for v in port.iter_vars():
         if not v.parent_block().parent_component().is_indexed():
             raise TypeError(
                 f"No state block could be retrieved from Port {port.name} "
-                f"because variable {v.name} does not belong to a state block."
+                f"because Component {v.parent_block().parent_component()} "
+                "is not indexed."
                 )
         if not isinstance(v.parent_block(),StateBlockData):
             raise TypeError(
                 f"No state block could be retrieved from Port {port.name} "
-                f"because variable {v.name} does not belong to a state block."
+                f"because Component {v.name} does not belong to a state block."
                 )
         states.append(v.parent_block().parent_component())
+        vlist.append(v)
         
-    if len(states) == 0:
+    if len(vlist) == 0:
         raise AttributeError(
             f"No state block could be retrieved from Port {port.name} "
-            f"because it contains no variables."
+            f"because it contains no components."
             )
     # Check the number of indices of the parent property block. If its indexed
     # both in space and time, keep the second, spatial index and throw out the
     # first, temporal index. If that ordering is changed, this method will
     # need to be changed as well.
-    idx = v.parent_block().index()
+    idx = vlist[0].parent_block().index()
     if isinstance(idx,tuple):
-        if len(idx) ==2:
-            idx = (time_point,v.parent_block().index()[1])
+        if len(idx) == 2:
+            idx = (time_point,vlist[0].parent_block().index()[1])
         elif len(idx) >2:
             raise NotImplementedError(
                 "State block retrieval is supported for only 0D and 1D control "
@@ -311,7 +314,7 @@ def _get_state_from_port(port,time_point):
         return states[0][idx]
     raise AttributeError(
         f"No state block could be retrieved from Port {port.name} "
-        f"because variables are derived from multiple state blocks."
+        f"because components are derived from multiple state blocks."
         )
 
 
--- a/idaes/core/util/tables.py
+++ b/idaes/core/util/tables.py
@@ -89,24 +89,26 @@ def stream_states_dict(streams, time_point=0):
             key = "{}[{}]".format(n, i)
         stream_dict[key] = sb
 
-    for n in streams.keys():
-        try:
-            if isinstance(streams[n], Arc):
-                for i, a in streams[n].items():
-                    sb = _get_state_from_port(a.ports[1], time_point)
-                    _stream_dict_add(sb, n, i)
-            elif isinstance(streams[n], Port):
-                sb = _get_state_from_port(streams[n], time_point)
-                _stream_dict_add(sb, n)
-            else:
+    for n in streams.keys():        
+        if isinstance(streams[n], Arc):
+            for i, a in streams[n].items():
+                sb = _get_state_from_port(a.ports[1], time_point)
+                _stream_dict_add(sb, n, i)
+        elif isinstance(streams[n], Port):
+            sb = _get_state_from_port(streams[n], time_point)
+            _stream_dict_add(sb, n)
+        else:
+            # _IndexedStateBlock is a private class, so cannot directly test
+            # whether  streams[n] is one or not. 
+            try:
                 sb = streams[n][time_point]
-                _stream_dict_add(sb, n)
-        except (AttributeError, KeyError):
-            raise TypeError(
-                f"Unrecognised component type for stream argument {streams[n]}."
-                f" The stream_states_dict function only supports Arcs, "
-                f"Ports or StateBlocks."
-            )
+            except (AttributeError, KeyError):
+                raise TypeError(
+                    f"Unrecognised component type for stream argument {streams[n]}."
+                    f" The stream_states_dict function only supports Arcs, "
+                    f"Ports or StateBlocks."
+                )
+            _stream_dict_add(sb, n)
     return stream_dict
 
 
@@ -266,22 +268,44 @@ def stream_table_dataframe_to_string(stream_table, **kwargs):
         na_rep=na_rep, justify=justify, float_format=float_format, **kwargs
     )
 
-
-def _get_state_from_port(port, time_point):
-    # Check port for _state_block attribute
-    try:
-        if len(port._state_block) == 1:
-            return port._state_block[0][time_point]
-        else:
-            return port._state_block[0][time_point, port._state_block[1]]
-    except AttributeError:
-        # Port was not created by IDAES add_port methods. Return exception for
-        # the user to fix.
-        raise ConfigurationError(
-            f"Port {port.name} does not have a _state_block attribute, "
-            f"thus cannot determine StateBlock to use for collecting data."
-            f" Please provide the associated StateBlock instead, or use "
-            f"the IDAES add_port methods to create the Port."
+def _get_state_from_port(port,time_point):
+    from idaes.core.property_base import StateBlockData
+    states = list()
+    for v in port.iter_vars():
+        if not v.parent_block().parent_component().is_indexed():
+            raise TypeError(
+                f"No state block could be retrieved from Port {port.name} "
+                f"because variable {v.name} does not belong to a state block."
+                )
+        if not isinstance(v.parent_block(),StateBlockData):
+            raise TypeError(
+                f"No state block could be retrieved from Port {port.name} "
+                f"because variable {v.name} does not belong to a state block."
+                )
+        states.append(v.parent_block().parent_component())
+        
+    # Check the number of indices of the parent property block. If its indexed
+    # both in space and time, keep the second, spatial index and throw out the
+    # first, temporal index. If that ordering is changed, this method will
+    # need to be changed as well. 
+    idx = v.parent_block().index()
+    if isinstance(idx,tuple):
+        if len(idx) ==2:
+            idx = (time_point,v.parent_block().index()[1])
+        elif len(idx) >2:
+            raise NotImplementedError(
+                "State block retrieval is supported for only 0D and 1D control "
+                "volumes."
+            )
+    else:
+        idx = (time_point,)
+    # This method also assumes that ports with different spatial indices won't
+    # end up at the same port. Otherwise this check is insufficient.
+    if all(states[0] is s for s in states):
+        return states[0][idx]
+    raise AttributeError(
+        f"No state block could be retrieved from Port {port.name} "
+        f"because variables are derived from multiple state blocks."
         )
 
 
--- a/idaes/core/util/tables.py
+++ b/idaes/core/util/tables.py
@@ -99,15 +99,15 @@ def stream_states_dict(streams, time_point=0):
             _stream_dict_add(sb, n)
         else:
             # _IndexedStateBlock is a private class, so cannot directly test
-            # whether  streams[n] is one or not. 
+            # whether  streams[n] is one or not.
             try:
                 sb = streams[n][time_point]
-            except (AttributeError, KeyError):
+            except (AttributeError, KeyError) as err:
                 raise TypeError(
                     f"Unrecognised component type for stream argument {streams[n]}."
                     f" The stream_states_dict function only supports Arcs, "
                     f"Ports or StateBlocks."
-                )
+                ) from err
             _stream_dict_add(sb, n)
     return stream_dict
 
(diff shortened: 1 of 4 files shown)


idaes/core/util/tables.py
```diff
@@ -266,22 +268,50 @@ def stream_table_dataframe_to_string(stream_table, **kwargs):
...
-    except AttributeError:
-        # Port was not created by IDAES add_port methods. Return exception for
-        # the user to fix.
-        raise ConfigurationError(
-            f"Port {port.name} does not have a _state_block attribute, "
-            f"thus cannot determine StateBlock to use for collecting data."
-            f" Please provide the associated StateBlock instead, or use "
-            f"the IDAES add_port methods to create the Port."
+def _get_state_from_port(port,time_point):
+    from idaes.core.property_base import StateBlockData
```
eslickj, line 272: The usual style is to put imports at the top, unless there is a really good reason not to.

idaes/core/util/tables.py
```diff
@@ -266,22 +268,50 @@ def stream_table_dataframe_to_string(stream_table, **kwargs):
...
+    if len(states) == 0:
+        raise AttributeError(
+            f"No state block could be retrieved from Port {port.name} "
+            f"because it contains no variables."
+            )
+    # Check the number of indices of the parent property block. If its indexed
+    # both in space and time, keep the second, spatial index and throw out the
+    # first, temporal index. If that ordering is changed, this method will
+    # need to be changed as well.
+    idx = v.parent_block().index()
```
eslickj, line 297: I think you should generally not use the for loop variable outside the loop.  As an alternative, and to make pylint happy you could do (I assume all states are the same):

```
idx = state[0].index()
```

idaes/core/util/tables.py
```diff
@@ -266,22 +268,50 @@ def stream_table_dataframe_to_string(stream_table, **kwargs):
...
-        # the user to fix.
-        raise ConfigurationError(
-            f"Port {port.name} does not have a _state_block attribute, "
-            f"thus cannot determine StateBlock to use for collecting data."
-            f" Please provide the associated StateBlock instead, or use "
-            f"the IDAES add_port methods to create the Port."
+def _get_state_from_port(port,time_point):
+    from idaes.core.property_base import StateBlockData
+    states = list()
+    v = None # To make PyLint stop complaining
```
eslickj, line 274: Pylint probably has a point see comment later where you use v outside the loop.

idaes/core/util/tables.py
dallan-keylogic: When I put it at the top, it errors out, probably due to a circular import.

andrewlee94: We should work out why why there is an issue with an import at the top and resolve that, rather than delay the import (this should be a last resort).

idaes/core/util/tables.py
```diff
@@ -266,22 +268,50 @@ def stream_table_dataframe_to_string(stream_table, **kwargs):
...
+            raise TypeError(
+                f"No state block could be retrieved from Port {port.name} "
+                f"because variable {v.name} does not belong to a state block."
+                )
+        states.append(v.parent_block().parent_component())
+        
+    if len(states) == 0:
+        raise AttributeError(
+            f"No state block could be retrieved from Port {port.name} "
+            f"because it contains no variables."
```
andrewlee94, line 291: Minor technicality, but I would use "Components" rather than "variables" (as Ports can contain Expression and other sufficiently "var-like" components).

idaes/core/util/tables.py
andrewlee94: That doesn't make this good practice thought. You do not directly control `v` so you don't know for sure what it is is. You also needed to add a hack `v = None` initially to satisfy pylint which is a good indication you should do something differently (i.e. you should fix the problem, not just throw something in to hide the error).

idaes/core/util/tables.py
```diff
@@ -266,22 +268,50 @@ def stream_table_dataframe_to_string(stream_table, **kwargs):
...
+            )
+    else:
+        idx = (time_point,)
+    # This method also assumes that ports with different spatial indices won't
+    # end up at the same port. Otherwise this check is insufficient.
+    if all(states[0] is s for s in states):
+        return states[0][idx]
+    raise AttributeError(
+        f"No state block could be retrieved from Port {port.name} "
+        f"because variables are derived from multiple state blocks."
```
andrewlee94, line 314: Again, I would say "components" rather than "variables".

idaes/core/util/tables.py
```diff
@@ -266,22 +268,50 @@ def stream_table_dataframe_to_string(stream_table, **kwargs):
...
-            f"the IDAES add_port methods to create the Port."
+def _get_state_from_port(port,time_point):
+    from idaes.core.property_base import StateBlockData
+    states = list()
+    v = None # To make PyLint stop complaining
+    for v in port.iter_vars():
+        if not v.parent_block().parent_component().is_indexed():
+            raise TypeError(
+                f"No state block could be retrieved from Port {port.name} "
+                f"because variable {v.name} does not belong to a state block."
```
andrewlee94, line 279: First, the error message here needs to be updated to be more clear about the specific issue - i.e. that the parent component is expected to be indexed.

Second, whilst this condition should be true for any core IDAES model, it might be a bit restrictive for special cases (e.g. surrogate models, custom codes models). Of course, allowing for this would mean having to catch that possibility later on too.

idaes/core/util/tables.py
```diff
@@ -266,22 +268,50 @@ def stream_table_dataframe_to_string(stream_table, **kwargs):
...
+                )
+        if not isinstance(v.parent_block(),StateBlockData):
+            raise TypeError(
```
andrewlee94, line 282: Do we really need to insist upon state blocks here as well? For a general purpose method, would it not be sufficient to check to see that all members of the Port simply have the same parent component and return that if true? I.e. At this stage do we actually care what the parent component is and whether it is indexed?

Later tools might care, but it should be up to them to make that decision.

andrewlee94 (changes requested):
A few additional comments. Mostly, I wonder if you are being a bit too restrictive in this method in insisting that the object returned is an IndexedStateBlock. As a general method, I think it might be sufficient to check that all Port members have the same parent and return that. It is then up to tools using this method to check to see if that component has the necessary attributes they require.

idaes/core/util/tables.py
andrewlee94: If the parent component is indexed, then yes we need to deal with that and return the correct IndexedBlock. However, I am not sure that it is necessary to demand that the parent Block *always* be indexed. I.e. I think we should support the case where the parent happens to be a SimpleBlock as well. We do not want to return a slicer however.

Also note that many Pyomo methods only work in the IndexedBlock or SimpleBlock, and do not exist on BlockData objects (and sometime the reverse), so it is up to the tool looking for those methods to decide how to handle that.

idaes/core/util/tables.py
eslickj: I think I agree with @andrewlee94 here.  I could imagine adding a block to mimic a state block, but is not a state block.  We could leave it to other things down the line to complain if some expected thing is missing.

dallan-keylogic: In principle, I would like to support duck typing. However, I'm not sure whether that's worth the effort to support at this stage. Previously, the function would work only for `Port`s on `UnitModelBlock`s, so no functionality is being lost by type checking here.  Right now, the functionality is being used only to create stream tables, for which the function is arguably overkill (I think we could just read the variables/expressions directly from the port). For my global balance functionality, I might use this function during prototyping, but there's a significant chance that functionality to get the relevant info is added on some other object.

idaes/core/util/tables.py
dallan-keylogic: Type checking has been removed, this issue is no longer applicable.

idaes/core/util/tables.py
dallan-keylogic: Type checking has been removed at this point.

idaes/core/util/tables.py
dallan-keylogic: The function no longer checks whether the parent component is indexed, but just tries to retrieve an index. It appears even unindexed Pyomo objects have an index of None, which can then be used to index them, so the function will probably work on `SimpleBlock`s. If you could direct me to a particular example, I would be happy to write a test to ensure this functionality.

eslickj (dismissed):

idaes/core/util/tables.py
```diff
@@ -89,24 +88,26 @@ def _stream_dict_add(sb, n, i=None):
...
+            try:
                 sb = streams[n][time_point]
-                _stream_dict_add(sb, n)
-        except (AttributeError, KeyError):
-            raise TypeError(
-                f"Unrecognised component type for stream argument {streams[n]}."
-                f" The stream_states_dict function only supports Arcs, "
-                f"Ports or StateBlocks."
-            )
+            except (AttributeError, KeyError) as err:
```
andrewlee94, line 104: Is AttributeError still needed here? At this point I think only a KeyError could be raised.

idaes/core/util/tables.py
```diff
@@ -266,22 +267,47 @@ def stream_table_dataframe_to_string(stream_table, **kwargs):
...
-        raise ConfigurationError(
-            f"Port {port.name} does not have a _state_block attribute, "
-            f"thus cannot determine StateBlock to use for collecting data."
-            f" Please provide the associated StateBlock instead, or use "
-            f"the IDAES add_port methods to create the Port."
+        idx = vlist[0].parent_block().index()
+    except AttributeError as err:
+        raise AttributeError(
+                f"No block could be retrieved from Port {port.name} "
+                f"because block {vlist[0].parent_block()} has no index."
```
andrewlee94, line 295: Minor comment, but I would make this `vlist[0].parent_block().name` - at the moment you are getting the actual object rather than the name of the object - in the end this returns the same thing when the object is cast to a string, but to avoid possible issues in future it is probably best to get the name specifically.

idaes/core/util/tables.py
```diff
@@ -266,22 +267,47 @@ def stream_table_dataframe_to_string(stream_table, **kwargs):
...
     )
 
+def _get_state_from_port(port,time_point):
+    """
+    Attempt to get a state-block-like object connected to a Port. If different
+    variables on the port appear to be connected to different state blocks,
+    raise an exception. This process is complicated by the fact that the object
+    may be indexed by space as well as time. Therefore, we assume that, if it
+    is, the time index comes first.
+    """
```
andrewlee94, line 277: Another minor point, but it might be good to document the arguments, returns and raises characteristics of this method. Whilst it is currently private, I could see this being promoted to a public method in the not too distant future.

Either way, it would be useful for future developers to know what errors are raised by different conditions so that they can add error handling to their methods as necessary.

idaes/core/util/tables.py
```diff
@@ -266,22 +267,47 @@ def stream_table_dataframe_to_string(stream_table, **kwargs):
...
+    vlist = [v for v in port.iter_vars()]
+    states = [v.parent_block().parent_component() for v in vlist]
 
-def _get_state_from_port(port, time_point):
-    # Check port for _state_block attribute
+    if len(vlist) == 0:
+        raise AttributeError(
```
andrewlee94, line 282: I would make this a ValueError - it is occurring due to a lack of "values" in the Port. This will also help distinguish the different types of failures that can occur, so that users of this method can decide how to handle the different cases if needed.

idaes/core/util/tables.py
```diff
@@ -266,22 +267,47 @@ def stream_table_dataframe_to_string(stream_table, **kwargs):
...
-        # Port was not created by IDAES add_port methods. Return exception for
-        # the user to fix.
-        raise ConfigurationError(
-            f"Port {port.name} does not have a _state_block attribute, "
-            f"thus cannot determine StateBlock to use for collecting data."
-            f" Please provide the associated StateBlock instead, or use "
-            f"the IDAES add_port methods to create the Port."
+        idx = vlist[0].parent_block().index()
+    except AttributeError as err:
+        raise AttributeError(
```
andrewlee94, line 293: I would make this a `TypeError` instead - the attribute is not an IndexedComponent as expected.

idaes/core/util/tables.py
```diff
@@ -266,22 +267,47 @@ def stream_table_dataframe_to_string(stream_table, **kwargs):
...
+    if isinstance(idx,tuple):
+        idx = (time_point,vlist[0].parent_block().index()[1:])
+
+    else:
+        idx = (time_point,)
+    # This method also assumes that ports with different spatial indices won't
+    # end up at the same port. Otherwise this check is insufficient.
+    if all(states[0] is s for s in states):
+        return states[0][idx]
+    raise AttributeError(
```
andrewlee94, line 308: I think RunTimeError probably describes this best, although AttributeError isn't too far off the mark. However, here we are saying that we couldn't find an object which met our criteria, and not that a specific attribute was missing from an object.

andrewlee94 (changes requested):
Looks really good. I had a few comments, but they are all minor changes. The biggest one is to have each of the failure conditions return a different error type to be more descriptive, so that other methods relying on this can catch edge cases if needed.

idaes/core/util/tables.py
dallan-keylogic: I'm not sure why it was there to begin with. Removing it doesn't cause any tests to fail, which probably means we can get rid of it. 

Unfortunately, pulling on this thread caused some other parts of the code to unravel. Presently, the exception message indicates that the problem is that a StateBlock wasn't passed, when the real error is that it isn't able to be indexed by `time_point`. As a result, this `except` block handles errors for which the real problem was that the user passed in some `time_point` outside of `flowsheet.time`, but doesn't prevent other IndexedBlocks to pass through even though they are the wrong type.  Our options appear to be:

1. Add in a subfunction that crawls up the Pyomo object hierarchy to find the flowsheet, then test `time_point in flowsheet.time`, and make sure it's called in every function that accepts `time_point` as an option. Then explicitly add in testing that `isinstance(sb,StateBlock)` as part of each case.
2. Remove the `except` block and make no pretense that we're protecting the user from themselves.

At this point, I'd rather go with option (2).

idaes/core/util/tables.py
dallan-keylogic: In the spirit of duck typing, I think `AttributeError` is most descriptive. It's failing because `vlist[0].parent_block()` has no attribute `.index()`, not because it isn't an IndexedComponent. In fact, even ScalarComponents have the `.index()` attribute. It is set to `None`, but `x[None] is x` if `x` is a ScalarComponent.

idaes/core/util/tables.py
andrewlee94: I would go with 3:

Raise an exception that says the component was either unindexed or the index was not a member of the indexing set. Then let the user work out the problem.

andrewlee94 (approved):

eslickj (approved):

PR #498, opened 2021-09-08
Update Tagged SVG Output to use ModelTagGroups

eslickj:
This mainly updates updates the svg_tag function to use the new model tag classes.  It also move some feature (conditional formatting, division by zero trapping, and no value trapping) to model tags, as it was previously part of the svg output function.

I moved the svg_tag function and updated the API.  The new function support the old API and old import location.  At some future point we can add deprecation warnings.  I don't want to bother anyone with it yet though. 

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/misc.py
+++ b/idaes/core/util/misc.py
@@ -25,7 +25,7 @@ from pyomo.common.config import ConfigBlock
 
 import idaes.logger as idaeslog
 import idaes.core.solvers
-from idaes.core.util.tags import svg_tag
+from idaes.core.util.tags import svg_tag as svg_tag_new
 
 _log = idaeslog.getLogger(__name__)
 
@@ -117,7 +117,6 @@ def TagReference(s, description=""):
     return r
 
 
-# Author: John Eslick
 def copy_port_values(destination=None, source=None, arc=None,
         direction="forward"):
     """
@@ -130,6 +129,16 @@ def copy_port_values(destination=None, source=None, arc=None,
     from idaes.core.util.initialization import propagate_state
     propagate_state(destination=destination, source=source, arc=arc,
             direction=direction)
+
+
+def svg_tag(*args, **kwargs):
+    """
+    Moved to idaes.core.util.tags.svg_tag
+    Leaving redirection function here for deprecation warning.
+    """
+    _log.warning("DEPRECATED: idaes.core.util.misc.svg_tag has been deprecated. "
+        "It can now be imported from idaes.core.util.tags.svg_tag")
+    return svg_tag_new(*args, **kwargs)
 
 
 def set_param_from_config(b, param, config=None, index=None):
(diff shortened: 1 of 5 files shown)


idaes/core/util/misc.py
```diff
@@ -117,114 +117,6 @@ def TagReference(s, description=""):
     return r
 
 
-# Author John Eslick
```
andrewlee94, line 120: I think we should leave this in for now with a deprecation warning. You never know who might be using this.

Edit: It looks like you moved this to the `tags` module - you should add a deprecation warning with a redirect here then.

idaes/core/util/tags.py
```diff
@@ -223,12 +242,6 @@ def get_unit_str(self, index=None):
     def is_var(self):
         """Whether the tagged expression is a Pyomo Var. Tagged variables
         can be fixed or set, while expressions cannot.
-
```
andrewlee94, line 226: Why did you delete the documentation here? It is rare for more documentation to be bad thing.

andrewlee94 (changes requested):
Only a couple of minor comments, the main one being to add a deprecation warning and redirect for the old svg tag method.

idaes/core/util/tags.py
eslickj: It's a property.  It used to be a function, but I forgot to fix the docstring.

andrewlee94: Ah - Github conveniently cut off the display right at the `def` line so I didn't see that.

idaes/core/util/misc.py
eslickj: Added deprecation warnings for both the old location and the old API.

eslickj:
Ok, should have comments fixed and I fixed the doc example problem that was breaking tests.

idaes/core/util/misc.py
```diff
@@ -132,6 +131,16 @@ def copy_port_values(destination=None, source=None, arc=None,
             direction=direction)
 
 
+def svg_tag(*args, **kwargs):
+    """
+    Moved to idaes.core.util.tags.svg_tag
+    Leaving redirection function here for deprecation warning.
+    """
+    _log.warning("DEPRECATED: idaes.core.util.misc.svg_tag has been deprecated. "
```
andrewlee94, line 139: I would suggest using the Pyomo deprecation warning here: from pyomo.common.deprecation import deprecation_warning

This lets you tag the version that is was deprecated in and the version it will be removed in.

Otherwise, I think you should make a note of when this deprecation was added so we can judge when to finally remove it.

andrewlee94 (changes requested):
A couple of minor typos (mostly so minor I almost ignored them), but I think we need to be more rigorous with deprecations and record when they were added (and if possible, when they will be removed).

eslickj:
Fixed typos and now use Pyomo deprecation stuff.

andrewlee94 (approved):

dallan-keylogic (approved):

PR #502, opened 2021-09-13
Fix minor tag and tables issues.

eslickj:
## Fixes

1.)  Fixes a tag issue displaying variables that don't have values
2.)  In tables, in the function to get a state block for an arc, if the fist port doesn't have a state block try the second before giving up. 

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/tables.py
+++ b/idaes/core/util/tables.py
@@ -91,7 +91,14 @@ def stream_states_dict(streams, time_point=0):
     for n in streams.keys():
         if isinstance(streams[n], Arc):
             for i, a in streams[n].items():
-                try:
+                try: 
+                    # if getting the StateBlock from the destination port
+                    # fails for any reason try the source port. This could
+                    # happen if a port does not have an associated 
+                    # StateBlock. For example a surrogate model may not
+                    # use state blocks, unit models may handle physical
+                    # properties without state blocks, or the port could
+                    # be used to serve the purpose of a translator block.
                     sb = _get_state_from_port(a.ports[1], time_point)
                 except:
                     sb = _get_state_from_port(a.ports[0], time_point)
--- a/idaes/core/util/tables.py
+++ b/idaes/core/util/tables.py
@@ -91,7 +91,10 @@ def stream_states_dict(streams, time_point=0):
     for n in streams.keys():
         if isinstance(streams[n], Arc):
             for i, a in streams[n].items():
-                sb = _get_state_from_port(a.ports[1], time_point)
+                try:
+                    sb = _get_state_from_port(a.ports[1], time_point)
+                except:
+                    sb = _get_state_from_port(a.ports[0], time_point)
                 _stream_dict_add(sb, n, i)
         elif isinstance(streams[n], Port):
             sb = _get_state_from_port(streams[n], time_point)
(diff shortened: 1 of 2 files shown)


idaes/core/util/tables.py
```diff
@@ -91,7 +91,10 @@ def _stream_dict_add(sb, n, i=None):
     for n in streams.keys():
         if isinstance(streams[n], Arc):
             for i, a in streams[n].items():
-                sb = _get_state_from_port(a.ports[1], time_point)
+                try:
+                    sb = _get_state_from_port(a.ports[1], time_point)
+                except:
+                    sb = _get_state_from_port(a.ports[0], time_point)
```
dallan-keylogic, line 97: Does this try-except block exist to deal with input ports to surrogate models? The general philosophy has been "input ports have well-defined states, output ports may not", so putting in a comment explaining what the use-case is would be good.

eslickj: I'll add a comment.  In this case, it deals with me making an occasional translator port or having a unit without state blocks.  It just makes it more likely I'll actually find a state block.  Yeah, I guess the same would apply to surrogate models.

andrewlee94: So, a general comment on surrogates, but with the proposed new API, whilst there will not be a `StateBlock` present, all the variables *should* be part of the same `Block`. So, `_get_state_from_port` should return a block for most surrogates, but it would be a `SurrogateBlock` or `SkeletonModel` instead of a `StateBlock`.

eslickj: @andrewlee94, I agree with that.  It may require a little more from the modeler, but any block with the standard property names for whatever is available would be good.  Then we just attach it as a new attribute onto the port maybe.  So IDAES ports should be expected to have a state_block attribute.  For the future though.

dallan-keylogic: It would be good, eventually, to add in code such that specific exceptions are being caught and if `_get_state_from_port` fails for both ports an error message would say so, rather than that it just couldn't find a state block on the outlet port. Probably beyond the scope of this changeset, though.

dallan-keylogic (approved):

idaes/core/util/tables.py
```diff
@@ -91,7 +91,17 @@ def _stream_dict_add(sb, n, i=None):
...
+                try: 
+                    # if getting the StateBlock from the destination port
+                    # fails for any reason try the source port. This could
+                    # happen if a port does not have an associated 
+                    # StateBlock. For example a surrogate model may not
+                    # use state blocks, unit models may handle physical
+                    # properties without state blocks, or the port could
+                    # be used to serve the purpose of a translator block.
+                    sb = _get_state_from_port(a.ports[1], time_point)
+                except:
```
andrewlee94, line 103: Use of bare excepts is generally discouraged, but I won't insist on changing it.

andrewlee94 (approved):

idaes/core/util/tables.py
eslickj: Yeah.  I know.  I wasn't sure of exactly all the ways this could fail, and if there is a mistake in the method the exception would get raised on the second try anyway.

PR #511, opened 2021-09-20
Adding set_value method to GeneralVarLikeExpression

andrewlee94:
## Summary/Motivation:
@dallan-keylogic identified an issue where using `set_value()` on the IDAES `GeneralVarLikeExpression` component would overwrite the expression with the new value, which was undesired behaviour. This PR overloads `set_value` to raise an `Exception` in normal use, but retains a `force` option to allow users to change the expression if desired.

- Overloads `set_value()`
- Adds new tests for `set_value`
- Fixes a case where `value()` should have been used in a test.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/misc.py
+++ b/idaes/core/util/misc.py
@@ -254,14 +254,39 @@ class _GeneralVarLikeExpressionData(_GeneralExpressionData):
 
     # Define methods for common APIs on Vars in case user mistakes
     # an Expression for a Var
+    def set_value(self, value, force=False):
+        """
+        Overload set_value method to provide meaningful error if user attempts
+        to set the value of the Expression. In order to support changing the
+        expression (and setting it originally), if self._expr is None or
+        force=True, the value of the expression will be updated, otherwise a
+        TypeError will be raised.
+
+        Args:
+            value: value to set for _expr
+            force: force updating of _expr if True (default = False)
+
+        Returns:
+            None
+
+        Raises:
+            TypeError if _expr is not None and force=False
+        """
+        if self._expr is None or force:
+            super().set_value(value)
+        else:
+            raise TypeError(
+                f"{self.name} is an Expression and does not have a value "
+                f"which can be set.")
+
     @property
     def value(self):
-        # Overload value so it behaves like a Var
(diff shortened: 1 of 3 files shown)


idaes/core/util/misc.py
```diff
@@ -254,14 +254,39 @@ class _GeneralVarLikeExpressionData(_GeneralExpressionData):
...
+                f"{self.name} is an Expression and does not have a value "
+                f"which can be set.")
+
     @property
     def value(self):
-        # Overload value so it behaves like a Var
-        return pyo.value(self.expr)
+        raise TypeError(
+            f"{self.name} is an Expression and does not have a value "
+            f"attribute. Use the 'value()' method instead.")
```
jsiirola, line 286: I like this: `.value` on `Expression` was deprecated a long time ago...  Given #504, I think making ths louder is a good thing.

jsiirola (approved):
Looks good, pending successful tests.

idaes/core/util/misc.py
andrewlee94: That was the reason I maintained it.

eslickj (approved):
This looks fine as far as I can tell, but I have a few questions.

1) Should simple be scalar?
2) Isn't this a problem that could hit Pyomo too?  I mean for simple model or models you are very familiar with it's probably not an issue.  But for complex models you didn't write or modular models where there is no guarantee that a particular quantity is an expression or variable, won't there be a high risk of user error?
3) Should I not be using Pyomo Expressions (because I use them a lot probably going around this check)?

andrewlee94:
@eslickj WE deliberately made this IDAES-specific for the particular case of the outlet of Separators (i.e. where we cheat an populate the Ports with Expressions rather than Vars).

So:

1. We could make the transition to Scalar, but I am not sure how much it matters.
2. As I said, we made this IDAES specific for a reason. Pyomo didn't want to make Expressions look like Vars, and they have a good use for `set_Value` in `Expressions`.
3. No, you should use Pyomo `Expressions` in most cases. These are intended for very specific cases where there is a high likelihood of a user confusing the `Expression` for a `Var`.

PR #519, opened 2021-09-23
[WIP] Solvent column model improvements

### LINKED ISSUE #244: MEA model TODO list
Tracking TODOs from PR #140 and feature branch solvent_column:

- [ ] Tests need to be standardized.
    - [ ] Test marks are inconsistent
    - [ ] Need unit consistency checks
    - [ ] Check how pressure drop is handled. At one point, the initialization routine was changing the model state regarding pressure
    - [ ] Need to use initialization tester (this may depend on handling pressure above)
    - [ ] Need more verification checks (material and energy conservation, more asserts for model results)
    - [ ] Need more modular testing of build, initialization and solve steps
- [x] Rename `phe` model to `PlateHeatExchanger`
- [ ] Check for unit consistency and consider standardizing units. Property units are currently mixed (e.g. some using cm**3)
- [ ] Standardize naming of some variables - two I noted were `lnk_eq_conc` (consider if there is or should be a standard name) and `enth_liq_dens` (should be `enth_vol_liq`).
- [ ] Reaction package as a config argument and integrate with model
- [x] Remove reference to the time domain, and just make a local alias instead
- [x] More descriptive name for the var ` self.N_v`
- [ ] Create indexed vars for `yi`
- [x] Switch to using IDAES logger instead of Python `warnings` module
- [x] Change logger messages from all caps to normal
- [ ] Check vars fixed in the model (either convert to params or remove fixing in code) @andrewlee94 - I think I prefer `Vars` for most of these.
- [ ] Column pressure and options. The model currently provides config argument for options which do not exist. Also, column pressure is fixed via a config argument with a somewhat arbitrary default value.
- [ ] Review initialization routine. Logging should be standardized, and we should check if homotopy is required when using MA27 as the linear solver.
- [ ] Plate Heat Exchanger model directly references parameters from property package

Less Important:
- [ ] Consider whether plotting functions should live in a separate utility module
- [ ] Model report functions

andrewlee94:
## Summary/Motivation:
Addressing issues and improvements to the Gen 1 solvent column to make it generic.

-
-

### Legal Acknowledgement

### DIFF
--- a/docs/technical_specs/model_libraries/power_generation/carbon_capture/mea_solvent_system/unit_models/phe.rst
+++ b/docs/technical_specs/model_libraries/power_generation/carbon_capture/mea_solvent_system/unit_models/phe.rst
@@ -98,15 +98,15 @@ post-combustion carbon capture using MEA solvent.
 PHE Class
 ----------
 
-.. module:: idaes.power_generation.carbon_capture.mea_solvent_system.unit_models.phe
+.. module:: idaes.power_generation.carbon_capture.mea_solvent_system.unit_models.plate_heat_exchanger
 
-.. autoclass:: PHE
+.. autoclass:: PlateHeatExchanger
   :members:
 
 PHEData Class
 --------------
 
-.. autoclass:: PHEData
+.. autoclass:: PlateHeatExchangerData
   :members:
 
 References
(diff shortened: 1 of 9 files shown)


andrewlee94:
Times have moved on, and this is no longer necessary. Changes will be done in separate PRs to the main branch.

Issue #522, opened 2021-09-25
propagate_state behaviour


eslickj:
I ran into an issue with propagate_state.  Copy_port_values would overwrite fixed values, so the move to propagate state broke some of my stuff.  I'm not sure what the right behavior is, but I assumed that if you wanted to copy over fixed values you could.  I think we could add an overwrite_fixed option.  Don't know what the default should be but either way we risk breaking stuff.

andrewlee94:
@eslickj My general rule is that you never automatically overwrite a fixed variable - if it is fixed, it implies it was fixed for a reason and that the user needs to decide what to do.

We could add an `overwrite_fixed` option, but I would say the default should be `False`.

eslickj:
@andrewlee94 I figured that was the safest way to go.  It's done in PR #523.  My only worry is that there are things out there that may mysteriously break.  I can't see any way around that though.  Either way we go we may break something, and I don't want to keep both methods forever.  I guess in the deprecated function we could make the default True since that was the old behavior and add some more info to the deprecation warning.  Hopefully people will notice and carefully check there stuff when the update, but probably not.

PR #523, opened 2021-09-25
Add overwrite option to propagate state

eslickj:
## Fixes issue #522 

This adds an options to propagate state to overwrite fixed values.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/initialization.py
+++ b/idaes/core/util/initialization.py
@@ -124,7 +124,7 @@ def revert_state_vars(blk, flags):
 
 
 def propagate_state(destination=None, source=None, arc=None,
-        direction="forward"):
+        direction="forward", overwrite_fixed=False):
     """
     This method propagates values between Ports along Arcs. Values can be
     propagated in either direction using the direction argument.
@@ -135,6 +135,8 @@ def propagate_state(destination=None, source=None, arc=None,
         arc (Arc): If arc is provided, use arc to define source and destination
         direction (str): Direction in which to propagate values.
                 Default = 'forward' Valid values: 'forward', 'backward'.
+        overwrite_fixed (bool): If True overwrite fixed values, otherwise do not
+            overwrite fixed values.
 
     Returns:
         None
@@ -189,7 +191,7 @@ def propagate_state(destination=None, source=None, arc=None,
     for k, v in destination_vars.items():
         try:
             for i in v:
-                if not (v[i].is_variable_type() and v[i].fixed):
+                if v[i].is_variable_type() and ((not v[i].fixed) or overwrite_fixed):
                     v[i].value = value(source_vars[k][i])
 
         except (AttributeError, ValueError):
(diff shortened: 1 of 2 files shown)


andrewlee94 (approved):
Looks good to me.

idaes/core/util/initialization.py
```diff
@@ -189,7 +191,7 @@ def propagate_state(destination=None, source=None, arc=None,
     for k, v in destination_vars.items():
         try:
             for i in v:
-                if not (v[i].is_variable_type() and v[i].fixed):
+                if v[i].is_variable_type() and ((not v[i].fixed) or overwrite_fixed):
```
dallan-keylogic, line 194: The behavior got changed for when `v[i].is_variable_type()` is false. Are there things for which `v[i].is_variable_type()` evaluates to false for which we still want to set a value @andrewlee94  @bknueven ?

bknueven: The error message below speaks about mutable Params, which have a `value` attribute but no `fixed` attribute. To maintain the existing behavior when `overwrite_fixed` is `False` we should do something like
```suggestion
                if overwrite_fixed or not(v[i].is_variable_type() and v[i].fixed):
```

dallan-keylogic: Very good. Hopefully @eslickj will add a test to ensure this behavior is maintained.

andrewlee94: From an IDAES specific stand-point, I have to ask why a `Param` would appear in a `Port` - this would tend to indicate a (very big) modeling error on the part of the user. That said, that is something that should be caught well before this point, which would mean when the `Port` is constructed.

eslickj: I guess my feeling (and I was too lazy to test this) is that named expressions have a value and it can be set.  It is deprecated and produces a warning message, but it still works.  I'll a little worried that if we don't explicitly limit this to variables or variables and parameters if that makes sense, someone could run into random weirdness at some point.

eslickj: I should say I do use expressions in ports and probably the reason I'm not aware of this causing trouble is that the arrangement I have is probably always from expression to variable.

andrewlee94: @eslickj If you are using `Expressions` in `Ports`, that is the one time to use the `VarLikeExpression` component from IDAES as it was built specifically to address these issues. E.g. it overloads the `set_value` method to raise an `Exception` to prevent this from occurring. Also, `Expression` (or `VarLikeExpression` should (probably?) only be used in outlets, and not inlets because of this.

eslickj: @andrewlee94 I do always use them in outlets, and I'm sure that's why I don't seem to have trouble with this.  Most properties aside from enthalpy, pressure, flow and composition in the Helmholtz EoS module are expressions.  I don't know what a person may put in a port or the direction they may think of their arcs going.  I guess a person could do something reasonable that we may think is weird.  That may argue for using VarLikeExpression for all expressions at least.  Maybe guarding against overwriting expressions here.

bknueven: If we abandon the duck-typing with this change (as it is) the exception below will only be raised if the port has some non-Pyomo like object on it or the object incorrectly reports itself as a variable type. If the object implements `is_variable_type` and it returns `False`, it'll simply be ignored.

So if we expect the destinations to be variables I'd move the check `is_variable_type` up raise an exception if it's `False`, to avoid a perhaps unexpected silent failure.

eslickj:
Ok, I updated this to raise a TypeError if propagate_state tries to set anything that is not a variable.  It also raises a KeyError i guess if the variable you are setting is not indexed by a subset of the variable it's coming from.  I added a bunch of tests to to test pretty much all the combinations.

dallan-keylogic (approved):

PR #527, opened 2021-09-29
Generalize Packed Column Model

anujad95:
## Fixes
NA

## Summary/Motivation:
Generalizing the packed column model in terms of its parameters, variables, and constraints, such that it is not hard coded for individual solute/solvent components. 

In progress

### Legal Acknowledgement


idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -486,12 +491,22 @@ def _make_performance(self):
                                       initialize=160,
                                       doc='Enhancement factor')
 
-        self.yi_MEA = Var(self.flowsheet().time,
+        # self.yi_MEA = Var(self.flowsheet().time,
```
andrewlee94, line 494: Remove commented code.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -486,12 +491,22 @@ def _make_performance(self):
...
+        #                   self.liquid_phase.length_domain,
+        #                   domain=NonNegativeReals,
+        #                   initialize=0.5,
+        #                   units=None,
+        #                   doc='''Dimensionless concentration of MEA
+        #                             at interface ''')
+        
+        self.yi_solvent = Var(self.flowsheet().time,
                           self.liquid_phase.length_domain,
+                          solvent_comp_list,
```
andrewlee94, line 504: 1. Why not include H2O in here as well for simplicity? It might not get used elsewhere, but most property packages will not specify this.
2. As there is an equivalent term for solutes as well, I would just make a single component for this, and have the rule decide whether to use the solvent or solute approach.
3. We should call this `conc_interface` or something similar.
4. See also the comment below about `self.equilibrium_components` - these might be the only components that need to appear here, so we might be able to use that as the indexing set here.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -661,11 +676,7 @@ def eq_velocity_liq(blk, t, x):
 
         # vapor mass transfer coeff. for diffusing components [mol/m2.s.Pa]
         def rule_mass_transfer_coeff_vap(blk, t, x, j):
-            if x == self.vapor_phase.length_domain.first():
-                return Expression.Skip
-            elif j == "MEA":
-                return Expression.Skip
-            else:
+            if j in liq_comp_diffus:
```
andrewlee94, line 679: This needs some work - at the moment the indexing set is based on liquid components, but the expression is based on vapor components.

First, the unit model needs to create a new indexing set: `self.equilibrium_components = self.congif.vapor_property_package & self.config.liquid_property_package`

This will give us the intersection of the two component lists - i.e. the components that cross the interface.

Then, this Expression can be indexed by `self.equilibrium_components` and there should be no need for `Expression.Skip`.

We will also be able to use this later on in other constraints and expressions.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -677,6 +688,10 @@ def rule_mass_transfer_coeff_vap(blk, t, x, j):
                       blk.vapor_phase.properties[t, x].dens_mass) /
                      (blk.a_ref *
                       blk.vapor_phase.properties[t, x].visc_d))**(3/4))
+            elif x == self.vapor_phase.length_domain.first():
+                return Expression.Skip
+            else:
+                return Expression.Skip
 
         self.k_v = Expression(self.flowsheet().time,
                               self.vapor_phase.length_domain,
```
andrewlee94, line 697: Further down we have a CO2 specific term - we will need to clean that up too (GitHub won't let me put a comment on the actual line as it wasn't edited).

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -746,17 +761,17 @@ def pressure_at_interface(blk, t, x, j):
 
         # mass transfer of  diffusing components
         def rule_mass_transfer(blk, t, x, j):
```
andrewlee94, line 763: As previously mentions with `self.equilibrium_components`.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -677,6 +688,10 @@ def rule_mass_transfer_coeff_vap(blk, t, x, j):
                       blk.vapor_phase.properties[t, x].dens_mass) /
                      (blk.a_ref *
                       blk.vapor_phase.properties[t, x].visc_d))**(3/4))
+            elif x == self.vapor_phase.length_domain.first():
```
andrewlee94, line 691: This check should (generally) always go first - we don't need to write terms at the inlet, so we should check first. Otherwise, the above condition will be met for at least some species and thus terms will be written at the boundary for them.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -746,17 +761,17 @@ def pressure_at_interface(blk, t, x, j):
...
-                return blk.interphase_mass_transfer[t, x, j] == 0.0
-            else:
+            if j in liq_comp_diffus:
                 return blk.interphase_mass_transfer[t, x, j] == (
                     blk.k_v[t, x, j] *
                     blk.area_interfacial[t, x] * blk.area_column *
                     (blk.vapor_phase.properties[t, x].mole_frac_comp[j] *
                      blk.vapor_phase.properties[t, x].pressure -
                      blk.pressure_equil[t, x, j])) * blk._homotopy_par_m
+            elif x == self.vapor_phase.length_domain.first():
```
andrewlee94, line 771: Again, this check should go first.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -775,13 +790,18 @@ def liquid_phase_mass_transfer_handle(blk, t, x, p, j):
             if x == self.liquid_phase.length_domain.last():
                 return blk.liquid_phase.mass_transfer_term[t, x, p, j] == 0.0
             else:
+<<<<<<< HEAD
```
andrewlee94, line 793: Conflict that needs to be addressed.

andrewlee94 (changes requested):
@anujad95 A good start - I added a number of comments. Notably, we need to rethink how indexing sets are done, as we cannot rely on the property package to define non-standard indexing sets for us.

However, all the information we need should be there - the property package does have separate solvent and solute sets (`solvent_set` and `solute_set`) we can use, and we can get the intersection of the vapour and liquid component lists to determine what species are in equilibrium.

anujad95:
Thank you, @andrewlee94 I will work on your suggestions.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -351,7 +351,12 @@ def _make_performance(self):
         # ======================================================================
         # Aliases for  Sets
         vap_comp = self.config.vapor_side.property_package.component_list
-        liq_comp = self.config.liquid_side.property_package.component_list
+        liq_comp = self.config.liquid_side.property_package.solvent_set | \
```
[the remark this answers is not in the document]
anujad95: I was not sure whether solute and solvent objects get included in the component_list. If they do, then we can keep this as it was before.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
anujad95: Yes, I agree with this.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -351,12 +351,12 @@ def _make_performance(self):
...
-        liq_comp = self.config.liquid_side.property_package.solvent_set | \
-            self.config.liquid_side.property_package.solute_set
-        liq_comp_diffus = \
-            self.config.liquid_side.property_package.component_list_diffus_apparent
+        liq_comp = self.config.liquid_side.property_package.component_list
+        equilibrium_comp = vap_comp & liq_comp
         solvent_comp_list = \
-            self.config.liquid_side.property_package.component_list_solvent_apparent
+            self.config.liquid_side.property_package.solvent_set
+        electrolyte_comp_diffus = self.config.liquid_side.property_package.component_list_diffus\
```
andrewlee94, line 358: We are going to need to change this - the diffusing species set is not standard so will not be present in most property packages. At first glance, it appears this is only used for the concentrations at the interface, and I don't think we actually need those right now (a quick look made it seem like they were only used to calculate the enhancement factor).

Moving forward, I think we will hopefully be able to replace this with the full true species set.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -734,7 +724,7 @@ def rule_phi(blk, t, x):
         # Equilibruim partial pressure of diffusing components at interface
         @self.Constraint(self.flowsheet().time,
                          self.vapor_phase.length_domain,
-                         liq_comp,
+                         equilibrium_comp,
                          doc='''Equilibruim partial pressure of diffusing
```
andrewlee94, line 728: Typo: equilibrium

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -975,83 +960,43 @@ def rule_yb_CO2(blk, t, x):
...
-                         t, x].conc_mol_comp_true['MEA+']))
+                         t, x].conc_mol_comp_true[k]))
 
-        self.yi_MEAH = Expression(self.flowsheet().time,
+        self.yi_electrolyte_diffus = Expression(self.flowsheet().time,
                                   self.liquid_phase.length_domain,
                                   solvent_comp_list,
-                                  rule=rule_yi_MEAH,
-                                  doc='Dimensionless concentration of MEA+')
+                                  electrolyte_comp_diffus,
```
andrewlee94, line 980: This is going to need some more thought and effort - for now I don't think these are actually needed as they are only used to calculate the enhancement factor.

Moving forwards, I think there should only be a single interfacial concentration (i.e. not separate ones for solvents and electrolytes), with different `if/else` branches based on component type. However, we also need to look into exactly how these concentrations are calculated - this might just need to be part of the "enhancement factor" sub-model that users might need to provide.

andrewlee94 (commented):
Making good progress. @anujad95 could you look at updating the tests - a number are failing because you removed the `pressure_drop` config argument, but it is still being used in the tests (which is triggering failures). We want to make sure the only test failures are the unit consistency (for now) to make sure we aren't changing the model results.

anujad95:
Thanks for the revised suggestions @andrewlee94, I will work on them. I will also look at updating the tests accordingly.

andrewlee94:
This has been replaced by #582.

PR #543, opened 2021-10-06
Update Perrys method for Equation Form 2

bpaul4:
## Fixes

Related to: https://github.com/IDAES/examples-pse/pull/67

## Summary/Motivation:
- Add switch to use alternate equation form for specific components

- Modified dens_mol_liq_comp parameter building and expression methods
- Water, o-terphenyl and sulfuric acid will automatically use second equation form when Perrys method is used for liquid molar density calculations

### Legal Acknowledgement


idaes/generic_models/properties/core/pure/Perrys.py
```diff
@@ -146,32 +146,59 @@ class dens_mol_liq_comp():
...
-        cobj.dens_mol_liq_comp_coeff_3 = Var(
-                doc="Parameter 3 for liquid phase molar density",
-                units=pyunits.K)
-        set_param_from_config(cobj, param="dens_mol_liq_comp_coeff", index="3")
-
-        cobj.dens_mol_liq_comp_coeff_4 = Var(
-                doc="Parameter 4 for liquid phase molar density",
-                units=pyunits.dimensionless)
-        set_param_from_config(cobj, param="dens_mol_liq_comp_coeff", index="4")
+        if cobj._name in ("water", "sulfuric_acid", "o-terphenyl"):
```
andrewlee94, line 149: I don't think using specific component names is the correct way to do this. First of all, it is too fragile (e.g. you used "water" where we generally use "H2O" so this would break).

I think we need a flag of some form as an argument to the method which users set in the config dict. E.g.:

```
try:
   flag = cobj.config.parameter_data["dens_mol_liq_comp_form"]
except KeyError:
   flag = DEFAULT_VALUE
```

andrewlee94 (changes requested):

bpaul4:
@andrewlee94 I appreciate your comments, and have pushed updated test and documentation files. Please let me know if you have further suggestions, thank you!

bpaul4:
Thank you for your reviews. The pure component properties tests ran successfully, but the changes cause tests in other property directories to fail. I will update the issues and check that the entirety of the method tests run without errors before pushing again.

idaes/generic_models/properties/core/pure/Perrys.py
```diff
@@ -228,33 +228,35 @@ def return_expression(b, cobj, T):
 
 class dens_mol_liq_comp():  # dens_mol_liq_comp_eqn_1, dens_mol_liq_comp_eqn_2
```
andrewlee94, line 229: I don't think this comment is needed - if fact it might just confuse readers.

idaes/generic_models/properties/core/pure/tests/test_Perrys.py
```diff
@@ -227,3 +229,87 @@ def test_dens_mol_liq_comp_eqn_2(frame):
...
+
+    m.props[1].temperature = Var(initialize=273.16, units=pyunits.K)
+
+    return m
+
+
+@pytest.mark.unit
+def test_dens_mol_liq_comp_deprecated(frame_deprecated):
+    dens_mol_liq_comp.build_parameters(frame_deprecated.params)
+
```
andrewlee94, line 287: You should also add a check here for the deprecation message to make sure it is being recorded. To do this, you can (I think):

1. Add `caplog` as an argument after the fixture in the test declaration
2. Then check the contents of `caplog.txt` to make sure the correct message is present.

See https://docs.pytest.org/en/6.2.x/logging.html

idaes/generic_models/properties/core/pure/tests/test_Perrys.py
```diff
@@ -227,3 +229,87 @@ def test_dens_mol_liq_comp_eqn_2(frame):
     assert value(expr) == pytest.approx(7.48194e10, rel=1e-4)
 
     assert_units_equivalent(expr, pyunits.mol/pyunits.m**3)
+
+
+@pytest.fixture()
```
andrewlee94, line 234: A minor observation, but I would only make a fixture if I intend to use the same code multiple times. In this case, this code is only used in one test, so I would normal just write it in the test directly.

idaes/generic_models/properties/core/pure/Perrys.py
```diff
@@ -228,33 +228,35 @@ def return_expression(b, cobj, T):
...
+            cobj.dens_mol_liq_comp_coeff_eqn_type.value = 1
 
-        if eqn_type == 1:
+        if cobj.dens_mol_liq_comp_coeff_eqn_type.value == 1:
             dens_mol_liq_comp_eqn_1.build_parameters(cobj)
-        elif eqn_type == 2:
+        elif cobj.dens_mol_liq_comp_coeff_eqn_type.value == 2:
             dens_mol_liq_comp_eqn_2.build_parameters(cobj)
         else:
             raise BurntToast(f"{cobj.name} unrecognized value for "
```
andrewlee94, line 251: Looking at how this is written, this should be a `ConfigurationError`, as we do not have nay checks to make sure that the value of this parameter is in the expected range (sorry, I know I suggested `BurntToast` last time).

andrewlee94 (changes requested):
A couple of minor comments.

idaes/generic_models/properties/core/pure/tests/test_Perrys.py
bpaul4: I'll match the structure to existing caplog checks in other test files, thanks for the suggestion @andrewlee94

andrewlee94 (approved):

bpaul4:
@jghouse88 this is ready for your review, and should address the current pytest errors on https://github.com/IDAES/examples-pse/pull/67.

jghouse88 (approved):
LGTM

PR #551, opened 2021-10-13
Fix TypeError introduced by PyYAML 6.0

lbianchi-lbl:
## Summary/Motivation:

- An error in `test_headers.py` related to `yaml.load()` is causing the tests to fail
- The probable cause is that a major version of PyYAML has recently been released and the error was previously a warning that's now being enforced
- The solution proposed in this PR is to use `yaml.safe_load()` instead of `yaml.load()`

- Use `yaml.safe_load()` instead of `yaml.load()`

### Legal Acknowledgement


lbianchi-lbl:
@andrewlee94 @ksbeattie @dangunter This error is showing up in other PRs and it looks like it's fixed by these changes, so feel free to approve this PR to have it get merged.

PR #552, opened 2021-10-14
Petsc DAE Solver Utils

eslickj:
## Summary/Motivation:

This adds utilities mainly for the PETSc time stepping (TS) solvers.  I'll attach examples for now and write some clean official examples for the examples repo.

This depends on https://github.com/Pyomo/pyomo/pull/2152 (which has been merged, but you will need a recent development version of Pyomo).

## Examples
* [car_example1.zip](https://github.com/IDAES/idaes-pse/files/7342039/car_example1.zip) modified version of the car example from Pyomo DAE
* [chemakzo.zip](https://github.com/IDAES/idaes-pse/files/7506716/chemakzo.zip) from: https://archimede.dm.uniba.it/~testset/report/chemakzo.pdf, this one is a little more interesting.  Some stuff happens very fast and it would be harder to solve in the fully discretized form without some careful tuning. 
* [pid.zip](https://github.com/IDAES/idaes-pse/files/7506724/pid.zip) IDAES model, this is the test for the PID controller solved with the PETSc solver.  Full IDAES model with a few units, controller, and external function calls for steam properties.
* [pendulum.zip](https://github.com/IDAES/idaes-pse/files/7749552/pendulum.zip)
Used to test constraint and variable scaling in DAEs.

## Legal Acknowledgement


eslickj:
A added another example based on the steam tank PID controller test.  So IDAES models seem to work.  I'm ready to try some more, if anyone has something to try.

Robbybp:
@eslickj Is the example on the steam tank part of this PR? I don't see it. I'd like an example/test I can run so I don't break anything when I update the "set_dae_suffixes" method.

eslickj:
@Robbybp, it's not in the PR yet, but it's attached to the PR description (the pid.zip link).

eslickj:
@blnicho, @jsiirola, @Robbybp, and @dallan-keylogic, I think this thing is ready to review.  We've been using it for the dynamic SOFC/SOEC modeling, so I'd like to get it merged pretty soon.  With some or all of the attached examples, I can put together some examples for the examples repo, once this is in.

eslickj:
The tests work for me on Ubuntu 20.04.  I'm working on getting Ubuntu 18.04 to see if I can reproduce the failure here.

Robbybp (changes requested):
In addition to running the examples you provided, I tried this out on two of my DAE models.

1. Dynamic simulation of a natural gas pipeline. This converged, with the results I expected, with default options. If I set `ts_options={"--ts_type": "alpha"}`, as I saw in one of the examples, it fails with TSConvergedReason -2.  What is the `--ts_type` option?
2. Dynamic simulation of a moving bed CLC reactor. This failed with the message due to different numbers of differential and algebraic variables. I'm pretty sure this has something to do with whether my differential and derivative variables both exist at the inlets to the reactor. Have you encountered a case like this (where boundary conditions interfere with the definition of differential/derivative variables) in your testing?

I also made some minor changes, mostly to the code for associating derivative and differential variables that I suggested. I opened a PR into this branch: https://github.com/eslickj/idaes-pse/pull/1 that you can merge if these changes look okay.

I would like to figure out exactly why I get an error with my CLC model, but given that this works on several other examples, I don't think that needs to hold up this PR, so I'll approve once the changes in my PR are merged or addressed.

idaes/core/solvers/petsc.py
```diff
@@ -0,0 +1,575 @@
...
+
+    Args:
+        m: model search for varaibles and write suffixes to
+        variables (list): List of time indexed variables at a sepcific time point
+
+    Returns:
+        None
+    """
+    m.dae_suffix = pyo.Suffix(direction=pyo.Suffix.EXPORT, datatype=pyo.Suffix.INT)
+    m.dae_link = pyo.Suffix(direction=pyo.Suffix.EXPORT, datatype=pyo.Suffix.INT)
```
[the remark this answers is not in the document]
Robbybp: I think I figured out what it is for, but a comment would still be nice.

idaes/core/solvers/petsc.py
```diff
@@ -0,0 +1,575 @@
...
+                continue
+            constraints = [con[t] for con in time_cons if t in con]
+            variables = [var[t] for var in time_vars]
+            # Create a temporary block with references to original constraints
+            # and variables so we can integrate this "subsystem" without altering
+            # the rest of the model.
+            t_block = create_subsystem_block(constraints, variables)
+            differential_vars = _set_dae_suffixes_from_variables(t_block, variables)
+            if timevar is not None:
+                t_block.dae_suffix[timevar[t]] = int(DaeVarTypes.TIME)
```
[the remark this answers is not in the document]
eslickj: I just finished an example for the examples repo that uses it, and I've used it in other models.  I'll add a test for it.

eslickj:
@Robbybp, https://petsc.org/main/docs/manualpages/TS/TSType.html.  Currently just implicit methods are supported.  Lately I've been using either beuler or cn (crank-nicolson).  I guess we may want to take a close look at what we can use and what's appropriate, and summarize that in the docs.  It's probably not clear, but most of the solver options are standard PETSc.

eslickj:
@Robbybp, on the CLC model that's failing, does the DAE solve fail or is it the initial condition solve that just plain nonlinear equation?  If you CLC model is in an okay state, I can try to figure out why it isn't working.

idaes/core/solvers/petsc.py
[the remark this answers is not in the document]
eslickj: Yes, I will add a comment.   dae_link is used to associate the differential variables with their time derivatives by assigning them the same unique integer suffix.

idaes/core/solvers/petsc.py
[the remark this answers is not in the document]
eslickj: Ok, I added an explicit time test.  I have some nice Jupyter notebooks I'll put in the examples repo once this is done too.

Robbybp:
In the failing model, the SNES solver gives the error: SNES_DIVERGED_LINE_SEARCH, but the function norm is 7.5e-9. If I'm remembering correctly, the initial conditions are converged (by solving a steady state problem) before I start the time integration. The program seems to enter the TS solver, exits in the first iteration, and triggers an ApplicationError in Pyomo.

eslickj:
@Robbybp I suspected it’s a degree of freedom problem with the TS solve, but I’m not totally sure.  If the initial solve failed maybe the nonlinear solve in the first time step also failed and it couldn’t take a step.  Maybe we try to adjust some solver tolerances?

eslickj:
@Robbybp, I couldn't get the CLC problem to converge, but at least I got it to run.  The fix is to the solver over on IDAES-EXT (https://github.com/IDAES/idaes-ext/pull/138).  The reason it was failing was that there are a different number of variables labeled differential than labeled derivative.  While I think that's a little weird, it shouldn't cause problems.  I did have a check that would force that to fail, but I removed that and checked that I didn't assume that derivatives exist (I did this in one place, setting scaling factors, but it should be fixed).

Robbybp:
@eslickj I'm a little torn about what we should do right now. On one hand, it seems like we should be able to identify "differential variables" that don't have derivatives in active constraints and refrain from labeling them differential. On the other, if the "differential" label doesn't have any impact on the TS solver, then maybe this is fine.

At the very least, seeing output that shows a mismatch between numbers of differential and derivative variables is somewhat confusing, so I would lean towards trying to be better about how we label things.

eslickj:
@Robbybp it occurred to me there is really no need to distinguish between differential and algebraic variables.  Maybe we should leave it to the solver to figure out rather than setting the suffixes.  I’m still not quite sure why this is happening.  It’s like there are derivative variables that don’t appear in the constraints?  Maybe fixed derivatives or something?

dallan-keylogic:
@eslickj For work I would like to do in the future, it's necessary to distinguish differential from algebraic variables. In order to linearize the nonlinear model, we need to eliminate the algebraic variables. Not sure if we need them in this context, though.

eslickj:
@dallan-keylogic and @Robbybp okay I think I know why the count was off.  Discussing with @Robbybp, it seems that there are inactive differential equations.  I'll look into fixing the suffix labeling here, and still keep the change to allow the solver to tolerate some mislabeling of differential variables. There is a more serious potential issue related to this where a differential variable from an inactive differential equation could not appear in forced into the problem.  That would mess up the degrees of freedom check at least.  So this needs to be addressed.

dallan-keylogic:
> I think I fixed the differential variable labeling. It now makes sure the derivatives are in active constraints and not fixed. I could see a person fixing some derivatives to 0.

There's a corner case in which somebody fixes a derivative variable to a non-zero value, perhaps trying to create a ramp. Right now that would result in the derivative being ignored, and hopefully the degrees of freedom inform the person their model is busted. But if they also failed in their model specification, it could result in some extremely confusing errors.

idaes/core/solvers/petsc.py
```diff
@@ -0,0 +1,626 @@
...
+        Returns:
+            (list)
+        """
+        dt = [None]*(len(self.time) - 1)
+        for i in range(len(self.time)-1):
+            dt[i] = self.time[i + 1] - self.time[i]
+        return dt
+
+    def interpolate_vecs(self, times):
+        """Create a new vector dictionary interpolated at times.
```
dallan-keylogic, line 591: Maybe mention that this method does in fact extrapolate if a time is outside the range of times in self.times?

idaes/core/solvers/petsc.py
```diff
@@ -0,0 +1,626 @@
...
+                j = 1
+            dt = self.time[j] - self.time[j - 1]
+            w1 = 1 - (t - self.time[j - 1])/dt
+            w2 = (t - self.time[j - 1])/dt
+            for k, v in self.vecs.items():
+                vecs[k][i] = w1*v[j - 1] + w2*v[j]
+        return vecs
+
+    def plot_trajectory_original(self, vars):
+        """Plot function privided by PETSc.
```
dallan-keylogic: privided -> provided

idaes/core/solvers/petsc.py
```diff
@@ -0,0 +1,626 @@
...
+
+        Args:
+            vars (list): list of variables or variable names to plot.  The
+                varibles include the time index of the final time.
```
dallan-keylogic: varibles -> variables

idaes/core/solvers/petsc.py
```diff
@@ -0,0 +1,626 @@
...
+        Returns:
+            None
+        """
+        vars = self._var_name_list(vars)
+        PetscBinaryIOTrajectory.PlotTrajectories(
+            self.time, self.vecs_by_time, self.vars, self._var_name_list(vars)
+        )
+
+    def delete_files(self):
+        """Delete the trajectory data and varible information files.
```
dallan-keylogic, line 642: varible -> variable

idaes/core/solvers/petsc.py
```diff
@@ -0,0 +1,626 @@
...
+        Returns:
+            None
+        """
+        shutil.rmtree(self.vis_dir)
+        os.remove(f'{self.stub}.col')
+        os.remove(f'{self.stub}.typ')
+
+    def to_json(self, pth):
+        """Dump the trajectory data to a json file in the form of a dictionary
+        with varaible name keys and '_time' with time vectors values.
```
dallan-keylogic, line 656: varible -> variable

idaes/core/solvers/tests/test_petsc.py
```diff
@@ -0,0 +1,139 @@
...
+    petsc._copy_time([m.x, m.y], 1, 2)
+
+    assert pyo.value(m.x[2]) == 1
+    assert pyo.value(m.y[2]) == 3
+
+@pytest.mark.unit
+def test_gen_time_disc_eqns():
+    m, y1, y2, y3, y4, y5, y6 = dae_with_non_time_indexed_constraint()
+
+    # The model has one time element and drivatives for y[1] to y[5]
```
dallan-keylogic, line 163: drivatives -> derivatives

idaes/core/solvers/tests/test_petsc_pid.py
```diff
@@ -0,0 +1,267 @@
...
+            "--ts_type":"beuler",
+            "--ts_dt":0.1,
+            "--ts_monitor":"", # set initial step to 0.1
+            "--ts_save_trajectory":1,
+            "--ts_trajectory_type":"visualization",
+        },
+        vars_stub="tj_vars"
+    )
+
+    # read the trajectory data, and make it easy by interplating a time point
```
dallan-keylogic, line 239: interplating -> interpolating

dallan-keylogic (dismissed):
Very good overall, mostly just typos.

idaes/core/solvers/petsc.py
```diff
@@ -0,0 +1,688 @@
...
+
+        Returns:
+            (list)
+        """
+        dt = [None]*(len(self.time) - 1)
+        for i in range(len(self.time)-1):
+            dt[i] = self.time[i + 1] - self.time[i]
+        return dt
+
+    def interpolate_vecs(self, times):
```
blnicho, line 590: There is an easy way to do this using Numpy's interpolation function which I expect would greatly improve the performance of this method. See: https://github.com/Pyomo/pyomo/blob/main/pyomo/dae/simulator.py#L1058

idaes/core/solvers/petsc.py
```diff
@@ -0,0 +1,688 @@
...
+            dt = self.time[j] - self.time[j - 1]
+            w1 = 1 - (t - self.time[j - 1])/dt
+            w2 = (t - self.time[j - 1])/dt
+            for k, v in self.vecs.items():
+                vecs[k][i] = w1*v[j - 1] + w2*v[j]
+        return vecs
+
+    def _unscale(self, m):
+        """If variable scale factors are used, the solver will see scaled
+        varaibles, and the scaled trajectory will be written. This function
```
blnicho, line 619: Typo: variables

idaes/core/solvers/petsc.py
```diff
@@ -0,0 +1,688 @@
...
+        uses varaible scaling facors from the given model to unscale the
```
blnicho, line 620: Typo: variable, factors

idaes/core/solvers/petsc.py
```diff
@@ -0,0 +1,688 @@
...
+        if pth.endswith(".gz"):
+            with gzip.open(pth, "w") as fp:
+                fp.write(json.dumps(self.vecs).encode("utf-8"))
+        else:
+            with open(pth, "w") as fp:
+                json.dump(self.vecs, fp)
+
+    def from_json(self, pth):
+        """Dump the trajectory data to a json file in the form of a dictionary
+        with varaible name keys and '_time' with time vectors values.
```
blnicho, line 673: Typo: variable

blnicho (commented):
I found lots of typos and have a few implementation questions.

docs/technical_specs/core/solvers.rst
```diff
@@ -46,15 +47,194 @@ solvers and to help with basic solver testing.
...
 .. autofunction:: idaes.core.solvers.ipopt_has_linear_solver
 
+
+PETSc Utilities
+---------------
+
+IDAES provides an AMPL solver interface for the PETSc solver suite,
+`(see the PETSc website) <https://petsc.org/release/overview/>`_.  PETSc provides
+nonlinear equation (NLE) and differential algebraic equation (DAE) solvers. Both
+NLE and DAE solvers are capable of solving simulation problems with zero degrees
```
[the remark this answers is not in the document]
eslickj: I'm going to change the way the solver reports this, but basically derivatives don't count as degrees of freedom.  I'll try to think of a good way to explain that you need the same number of state variables as equations.

idaes/core/solvers/features.py
```diff
@@ -102,6 +103,219 @@ def minlp():
...
+    model.Ke = pyo.Param(initialize=34.4)
+    model.klA = pyo.Param(initialize=3.3)
+    model.Ks = pyo.Param(initialize=115.83)
+    model.pCO2 = pyo.Param(initialize=0.9)
+    # The following parameter H, is best made a parameter, but will use a
+    # variable and constraint instead to test non-time-indexed constraints
+    #model.H = pyo.Param(initialize=737)
+
+    # Problem variables ydot = dy/dt,
+    #    (dy6/dt is not explicitly in the equations, so only 5 ydots)
```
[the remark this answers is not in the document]
eslickj: Yes

idaes/core/solvers/features.py
```diff
@@ -102,6 +103,219 @@ def minlp():
...
+    model.Fin = pyo.Var(model.t, initialize=1.0)
+
+    # Non-time indexed constraint (just for testing)
+    model.H_eqn = pyo.Constraint(expr=model.H==737)
+
+    # Equations
+    @model.Constraint(model.t)
+    def eq_ydot1(b, t):
+        if t==0:
+            pyo.Constraint.Skip
```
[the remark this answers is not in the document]
eslickj: Okay.  I can leave it in for time zero.  Whether they are active, inactive, or don't exist shouldn't shouldn't matter for the PETSc setup.  If they are active, you just calculate a derivative value too for the initial condition.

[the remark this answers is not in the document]
dallan-keylogic: I understand that PETSc won't care about it, but it's occurring in a utility function to return a DAE model for solver testing. It's pretty easy to see a place down the road where someone tries to test Pyomo DAE solvers or integrators using this function.

idaes/core/solvers/features.py
```diff
@@ -102,6 +103,219 @@ def minlp():
...
+        return b.r[t, 4] == b.k[3]*b.y[t, 1]*b.y[t, 4]**2
+    @model.Constraint(model.t)
+    def eq_r5(b, t):
+        return b.r[t, 5] == b.k[4]*b.y6[t]**2*b.y[t, 2]**0.5
+    @model.Constraint(model.t)
+    def eq_Fin(b, t):
+        return b.Fin[t] == b.klA*(b.pCO2/b.H - b.y[t, 2])
+
+    # Set initial condtions and solve initial from the values of differntial
+    # variables (r and y6 well and the derivative vars too).
```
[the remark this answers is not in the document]
eslickj: I guess it was from an old version.  Just deleted.  I think I meant the initial condition solve would solve for the algebraic variable and the derivative values (must be before I decided to skip t=0)

idaes/core/solvers/petsc.py
```diff
@@ -0,0 +1,626 @@
...
+    """
+
+    def __init__(self, **kwds):
+        self._ts_vars_stub = kwds.pop("vars_stub", None)
+        if "options" in kwds and kwds["options"] is not None:
+            kwds["options"] = copy.deepcopy(kwds["options"])
+        else:
+            kwds["options"] = {}
+        kwds["options"]["--dae_solve"] = ""
+        kwds["options"]["--ts_monitor"] = ""
```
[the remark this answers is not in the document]
eslickj: No.  The above if is just to create an options dict if it doesn't exist.  This is always going to be a DAE solver so I just force those options in.

idaes/core/solvers/petsc.py
eslickj: Due to the scaling issues, I decided to remove this function.  We can do our own plots pretty easy once the trajectory is read in.

idaes/core/solvers/tests/test_solvers.py
```diff
@@ -84,6 +87,41 @@ def test_ipopt_has_ma27():
...
+@pytest.mark.skipif(not petsc.petsc_available(), reason="PETSc solver not available")
+def test_petsc_idaes_solve():
+    """
+    Make sure there is no issue with the solver class or default settings that
+    break the solver object.  Passing a bad solver option will result in failure
+    """
+    m, x = nle()
+    solver = pyo.SolverFactory("petsc_snes")
+    solver.solve(m, tee=True)
+    assert pytest.approx(x) == pyo.value(m.x) or pytest.approx(x) == pyo.value(-m.x)
```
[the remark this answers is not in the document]
eslickj: Laziness.  You're right.  I'll update it.

idaes/core/solvers/petsc.py
```diff
@@ -0,0 +1,618 @@
...
+    for var in variables:
+        if var in deriv_diff_map:
+            deriv = var
+            diffvar = deriv_diff_map[deriv]
+            m.dae_suffix[diffvar] = int(DaeVarTypes.DIFFERENTIAL)
+            m.dae_suffix[deriv] = int(DaeVarTypes.DERIVATIVE)
+            m.dae_link[diffvar] = dae_var_link_index
+            m.dae_link[deriv] = dae_var_link_index
+            dae_var_link_index += 1
+    return differential_vars
```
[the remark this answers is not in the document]
eslickj: That seems to be a mistake.  I think I fixed it, but will check again.

eslickj:
@dangunter or @lbianchi-lbl, the DMF magics tests are failing here.  Not sure if it's a package version change or something like that.

lbianchi-lbl:
@eslickj it looks like this is a breaking change in IPython introduced by 8.0.0 that apparently was released just a few hours ago: https://pypi.org/project/ipython/#history
 
image

One temporary fix to try would be to add a requirement with version constraint in `setup.py`, e.g. `ipython<8.0.0`.

idaes/core/solvers/petsc.py
eslickj: My thinking on this was if I had a lot of time point and a lot of variables, I could save time searching by assuming the times were in order then reusing the index and weights for all the variables.  I'm sure you are right.  Numpy should be a lot faster.  I'll take a closer look at this see if I can find the most efficient way to do it.

eslickj: Updated to Numpy (which is likely faster and definitely cleaner).  I still feel there is probably a better way than what I have if you share part of the computation between between all the variables, but it's probably not worth worrying about.  If slowness is a problem for some super large problem we can probably improve.

docs/technical_specs/core/solvers.rst
```diff
@@ -62,6 +62,24 @@ PETSc includes optimization solvers, but they are not currently supported by the
...
 
+DAE Terminology
+~~~~~~~~~~~~~~~
+
+For the following discussion regarding the PETSc solver interface, the following
+terminology is used.
+
+* Derivative: a time derivative
+* Differential variable: a variable that is differentiated with respect to time
+* Algebraic: a variable with no explicit time derivative appearing in the problem
```
dallan-keylogic, line 73: Derivative -> Derivative variable
Algebraic -> Algebraic variable

docs/technical_specs/core/solvers.rst
```diff
@@ -62,6 +62,24 @@ PETSc includes optimization solvers, but they are not currently supported by the
...
+* State Variables: the set of algebraic and differential variables
+* Time variable: a variable representing time
```
dallan-keylogic, line 75: Either capitalize every "Variable" or none of them.

docs/technical_specs/core/solvers.rst
```diff
@@ -74,17 +92,18 @@ Windows installation. The easiest option is to run the available precompiled
...
 written by the PETSc solver.  On Windows, some manual installation of the Python
 modules is required.  If you are using the WSL method to run PETSc, copy the
 ``petscpy`` directory from the Linux package you are using to the IDAES binary
-directory.  You can find the IDAES binary directory with the command
-``idaes bin-directory``.  The primary use for these Python modules is to read
-trajectory files saved by the TS solver.
+directory.  You can find the IDAES binary directory by running the command
+``idaes bin-directory`` in the OS command shell (e.g. Bash, Windows CMD,
+PowerShell).  The primary use for these Python modules is to read trajectory
+files saved by the TS solver.
```
dallan-keylogic, line 106: May be worth mentioning that, if you're using something like Miniconda to manage multiple environments, it has to be an environment where you installed IDAES

idaes/core/solvers/features.py
```diff
@@ -130,7 +130,8 @@ def dae():
     model.H = pyo.Param(initialize=737)
 
     # Problem variables ydot = dy/dt,
-    #    (dy6/dt is not explicitly in the equations, so only 5 ydots)
+    #    (dy6/dt is not explicitly in the equations, so only 5 ydots i.e.
+    #    y6 is an agebraic varaible and y1 to y5 are differential variables)
```
dallan-keylogic, line 134: agebraic varaible -> algebraic variable

dallan-keylogic (approved):
Just a few minor typos/style things.

dallan-keylogic (approved):

idaes/core/solvers/petsc.py
```diff
@@ -0,0 +1,680 @@
...
+            pvar = var.parent_component()
+            if (isinstance(pvar, pyodae.DerivativeVar) and not var.fixed and
+                    time in ComponentSet(pvar.get_continuousset_list())):
+                derivs.add(var)
+    deriv_diff_list = [None]*len(derivs)
+    for i, deriv in enumerate(derivs):
+        diffvar = deriv.parent_component().get_state_var()
+        idx = deriv.index()
+        deriv_diff_list[i] = (deriv, diffvar[idx])
+    return ComponentMap(deriv_diff_list)
```
Robbybp, line 309: I'd like to avoid calling `deriv.index()` here if we can. This is a linear scan over `deriv`'s indexing set, and so will be quite expensive if we are calling this for every derivative data object. See my PR into this branch, https://github.com/eslickj/idaes-pse/pull/2, where I propose an alternative implementation.

Robbybp (commented):
I think this is very close. Along the lines of the edge case where a derivative does not appear, unfixed, in an active constraint, I'd like to consider the case where a differential var does not appear, unfixed, in an active constraint. This is not necessarily a problem, as differential variables are usually fixed as initial conditions. But consider the following example:

```python
from pyomo.environ import *
from pyomo.dae import *
from idaes.core.solvers import petsc

m = ConcreteModel()

m.time = ContinuousSet(initialize=(0.0, 10.0))
m.x = Var(m.time)
m.u = Var(m.time)
m.dxdt = DerivativeVar(m.x, wrt=m.time)

def diff_eq_rule(m, t):
    return m.dxdt[t] == m.x[t]**2 - m.u[t]
m.diff_eq = Constraint(m.time, rule=diff_eq_rule)

discretizer = TransformationFactory('dae.finite_difference')
discretizer.apply_to(m,nfe=15,scheme='BACKWARD')

for t in m.time:
    m.x[t].fix(2.0*t)

m.u[0].fix(1.0)

petsc.petsc_dae_by_time_element(
    m,
    time=m.time,
    initial_constraints=[],
    initial_variables=[],
)
```
Here we fail having found one derivative var and zero differential vars. I'm not sure what the right course of action is, but would at least like the failure to be a little more graceful than it currently is.

eslickj:
@Robbybp I added a test for your example.  It seems to run without issue, as it should.  I haven't checked the solution yet though.  At some point it seems I accidentally deleted a few important line that deal with DAE problems where the differential variable doesn't appear.  If this doesn't work for you, you may have a slightly out dated version of this PR.

eslickj:
@Robbybp, okay I have two options for the fixed differential.  1.) We see the fixed differential variable, and just fix the corresponding derivative to 0.  2.) We raise an exception that says you can't fix differential variables (except the initial conditions).

Also I found that the example solves by happy accident as if the derivative were fixed.  Buried in the solver messages is an error that the jacobian matrix is missing the column for the differential variable, and it just goes on.

For now I'll take option one, but if people object I can change it.

Robbybp:
@eslickj I think I like your second option better, as a fixed differential variable corresponds to a constant derivative, but not necessarily a derivative of zero. Raising an exception gets around the question of how to calculate the "correct" derivative, which may depend on the discretization.

Robbybp:
@eslickj Remind me what happens if a differential variable doesn't appear in an active constraint. Does it get added by Petsc?

dallan-keylogic:
@Robbybp 

You've got it backwards. A fixed _derivative_ variable corresponds to a constant derivative. A fixed _differential_ variable always results in a derivative of zero. (Unless I have our naming scheme backwards).

eslickj:
> @eslickj Remind me what happens if a differential variable doesn't appear in an active constraint. Does it get added by Petsc?

It gets included in the NL file by this new pyomo option  Pyomo/pyomo#2152.  I could add it in PETSc, but adding it in Pyomo lets me read the solution back.

eslickj:
@Robbybp and @dallan-keylogic, I do like raising an error now mostly because it's easiest to implement.  A differential variable being fixed does mean the derivative is 0, but if you look at it from the perspective of the fully time discretized model, and it's fixed to a different value at each time point it means the derivative is not zero.  There is a bit of a disconnect there.  For the DAE solver, to get the expected behavior, you'd need to add a time variable and constraint instead of fixing the differential variable.

Robbybp:
A differential variable fixed to the same value at every point in time gives a time derivative of zero. A differential variable fixed to different values at every point in time specifies the values of the time derivative through the discretization equations, but not necessarily to a constant value. Hopefully this is more clear.

eslickj:
@Robbybp, okay should be good to go again.  It now raises an exception with a message that tries to help if there is a fixed differential var.  It suggests either fixing the derivative or adding a constraint instead.

idaes/core/solvers/petsc.py
```diff
@@ -294,19 +302,29 @@ def _get_derivative_differential_data_map(m, time):
         (ComponentMap): Map from derivative data objects to differential
             data objects
     """
-    derivs = ComponentSet()
+    # Get corresponding derivative and differential data objects,
+    # with no attention payed to fixed or active status.
```
dallan-keylogic, line 306: payed -> paid

idaes/core/solvers/petsc.py
```diff
@@ -279,6 +279,14 @@ def _set_dae_suffixes_from_variables(m, variables, deriv_diff_map):
...
             if not diffvar.fixed:
                 differential_vars.append(diffvar)
+            else:
+                raise RuntimeError(
+                    f"Problem cannot contain a fixed differential variable and "
+                    f"unfixed derivative. Consider either fixing the "
+                    f"corresponding derivative or adding a constraint for the "
+                    f"differential variable {diffvar} possibly using an "
+                    f"explicit time variable."
+                )
```
Robbybp, line 289: Will this catch fixed differential variables at initial conditions?

eslickj: No this only gets called after the initial conditions, so it should be ok.  Tests have fixed differential variables as i.c.s, so we should be good.

Robbybp: Okay, sounds good. I'll run this locally then this should be good.

eslickj:
I added another test that's a version @Robbybp's example that works properly using an explicit time variable.  I think it's ready to go again.

idaes/core/solvers/tests/test_petsc.py
```diff
@@ -25,6 +25,32 @@
 from idaes.core.solvers import petsc
 from idaes.core.solvers.features import dae
 
+from pyomo.environ import *
+from pyomo.dae import *
```
dallan-keylogic, line 29: Looks like you made these unnecessary (good! Namespaces want to be clean!) but forgot to remove them

dallan-keylogic (dismissed):
Remind me what happens when a derivative variable is fixed to a nonzero value?

eslickj:
> Remind me what happens when a derivative variable is fixed to a nonzero value?

It would turn the differential equation into an algebraic equation whether the derivative is 0 and any other value.

idaes/core/solvers/petsc.py
```diff
@@ -0,0 +1,699 @@
...
+                # initial constraints, don't try to solve.
+                t_block = create_subsystem_block(
+                    constraints,
+                    variables,
+                )
+                # set up the scaling factor suffix
+                _sub_problem_scaling_suffix(m, t_block)
+                with idaeslog.solver_log(solve_log, idaeslog.INFO) as slc:
+                    res = solver_snes.solve(t_block, tee=slc.tee)
+        res_list.append(res)
```
Robbybp, line 454: Should this be indented twice more? To be inside the `if len(constraints) > 0` branch?

Robbybp (approved):
I think this is good to go.

dallan-keylogic (approved):

idaes/core/solvers/petsc.py
```diff
@@ -477,6 +477,13 @@ def petsc_dae_by_time_element(
                 variables,
                 deriv_diff_map,
             )
+            # We need to check if there are derivatives in the problem before
+            # sending this to the solver.  We'll assume that if you are using
+            # this and don't have any differential equations, you are making a
+            # mistake.
+            if len(differential_vars) < 1:
+                raise RuntimeError(
+                    "No differential equations found, you do not need a DAE solver.")
```
Robbybp, line 486: This sounds good, but should this check happen earlier? E.g. at the end of `_get_derivative_differential_data_map`?

eslickj: Wait maybe it's not that hard.  I should be able to check that there is at least one derivative for each time point.  I think I can improve this.

Robbybp: What happens if you send a model with no derivatives to Petsc? Does it just solve the square system?

Robbybp: With your comment in mind, I'm fine with the check where it is, but I'll leave it up to you.

eslickj: Okay changed my mind again.  It's probably not worth the trouble of checking this earlier.

eslickj:
I added tests for a few more cases.  I have 5 tests based on @Robbybp's example.  2 work and 3 are caught with errors and raise exceptions.  I think this is ready, at least really close.

Once we get this in, I'll put out new binaries with a few very minor improvements and a couple example notebooks in the examples repo.

idaes/core/solvers/petsc.py
eslickj: > What happens if you send a model with no derivatives to Petsc? Does it just solve the square system?

It should but it crashes.  Most likely I messed up and assumed I had at least one derivative.  I’ll sort that out in the solver part eventually.  For now this check will stop it from getting there.

dallan-keylogic (approved):

idaes/core/solvers/tests/test_petsc.py
```diff
@@ -0,0 +1,541 @@
...
+        return m.dxdt[t] == m.x[t]**2 - m.u[t]
+    m.diff_eq1 = pyo.Constraint(m.time, rule=diff_eq1_rule)
+
+    discretizer = pyo.TransformationFactory('dae.finite_difference')
+    discretizer.apply_to(m,nfe=1,scheme='BACKWARD')
+
+    m.u[0].fix(1.0)
+    m.dxdt[:].fix(2)
+
+    return m
```
Robbybp, line 138: What's the difference between `rp_example5` and `rp_example3`?

Robbybp (approved):
Looks great! Thanks for all the work on this interface, John. This will be very useful.

idaes/core/solvers/tests/test_petsc.py
eslickj: You know, I don't see a difference and can't remember.  I think I must have copied it and meant to change it to test another case, and now I cannot recall what it was for.  Doing to many things at once I guess.  If the tests pass I'll merge this and make another small PR to update or remove the test if I still can't remember why it's there.

eslickj:
> Looks great! Thanks for all the work on this interface, John. This will be very useful.

Thanks!  I think now that this is squared away, we can use the same method to hook up other integrators.  We can come up with a standard for explicit and implicit-explicit methods too.  I have some ideas I may propose if I find time.

PR #558, opened 2021-10-19
Better PID controller

eslickj:
This brings in a new PID controller model.  This one is based on Jinliang's controller from the dynamic power plant work. The new PID model will work with integration-type DAE solvers.  The old PID controller got tricky to avoid creating integral and derivative error variables, but that causes some constraints to explicitly reference the previous time step.  The old PID model doesn't suffer as much from initialization trouble, but the formulation isn't as general.  For models that are very complicated, I suspect you will need some time-stepping initialization for the new PID model, either @Robbybp's tools or initialize with an integrator.  If you are curious about the time profile for the new steam tank test it's here  [pid_steam_tank_pressure.pdf](https://github.com/IDAES/idaes-pse/files/7376228/pid_steam_tank_pressure.pdf)

### Legal Acknowledgement


eslickj:
Okay, the docs are in.  I think this is ready to review.

PR #581, opened 2021-10-29
WIP: Platen Heat Exchanger

PTA0001:
## Fixes
1. Removes direct call to Cp parameters in the  unit model

- Updates the plate heat exchanger model to make use of the generic property framework

### Legal Acknowledgement


idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -426,22 +427,22 @@ def rule_cp_cold(blk, t):
                           self.PH, initialize=390, units=pyunits.K,
                           doc="Cold Temperature OUT of pass")
 
-        # ======================================================================
+        # =====================================================================
         # PERFORMANCE EQUATIONS
```
MAZamarripa, line 431: @andrewlee94 add to mass flow to property package (inkg/s)

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -426,22 +427,22 @@ def rule_cp_cold(blk, t):
...
-            return blk.cold_side.properties_in[t].flow_mol *\
-                blk.cold_side.properties_in[t].mw
+            return blk.cold_fluid.properties_in[t].flow_mol *\
+                blk.cold_fluid.properties_in[t].mw
         self.mc_in = Expression(self.flowsheet().time, rule=rule_mc_in,
                                 doc='Coldside mass flow rate [kg/s]')
 
-        # ----------------------------------------------------------------------
+        # ---------------------------------------------------------------------
         # port mass velocity[kg/m2.s]
```
MAZamarripa, line 446: in line comment with better description

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -455,42 +456,43 @@ def rule_Gpc(blk, t):
...
                                rule=rule_Pr_c,
                                doc='Coldside Prandtl number')
-        # ----------------------------------------------------------------------
+        # ---------------------------------------------------------------------
         # Film heat transfer coefficients
 
         def rule_hotside_transfer_coef(blk, t, p):
-            return (blk.hot_side.properties_in[t].thermal_cond / blk.channel_dia *
+            return (blk.hot_fluid.properties_in[t].therm_cond_phase["Liq"] /
+                    blk.channel_dia *
```
MAZamarripa, line 495: channel_diameter

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -179,86 +182,89 @@ def build(self):
...
-                             block=self.cold_side, doc='outlet Port')
-        # ======================================================================
+                             block=self.cold_fluid, doc='outlet Port')
+        # =====================================================================
         # Add performace equation method
         self._make_params()
         self._make_performance_method()
 
     def _make_params(self):
         self.P = Param(initialize=self.config.passes,
```
andrewlee94, line 255: The names for these need to be more descriptive. `P` is an especially poor choice as it invites confusion with pressure.

Also, does this actually need to be a `Param`? If it is only used for building the sets below, then I don't think it is necessary and we could just use the config argument directly.

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -179,86 +182,89 @@ def build(self):
...
-                       units=None,
+                       units=pyunits.dimensionless,
                        doc="Total number of passes for hot or cold fluid")
         self.PH = RangeSet(self.P,
                            doc="Set of hot fluid passes")
         self.PC = RangeSet(self.P,
```
andrewlee94, line 260: If these will be equal, is it necessary to make two Sets?

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -179,86 +182,89 @@ def build(self):
...
                            doc="Set of cold fluid passes(equal to PH)")
 
-        self.plate_thermal_cond = Param(mutable=True,
-                                        initialize=self.config.plate_thermal_cond,
-                                        units=pyunits.W / pyunits.m / pyunits.K,
-                                        doc="Plate thermal conductivity")
+        self.plate_thermal_cond = Param(
+            mutable=True,
+            initialize=self.config.plate_thermal_cond,
+            units=pyunits.W / pyunits.m / pyunits.K,
```
andrewlee94, line 266: It would be nice if these units were not hard coded and depended on the property package instead.

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -291,34 +297,38 @@ def _make_params(self):
         _total_active_plate_number = 2 * sum(self.config.channel_list) - 1 -\
             self.config.divider_plate_number
 
-        self.plate_area = Expression(expr=self.config.total_area /
+        self.plate_area = Expression(expr=self.config.total_area*pyunits.m**2 /
```
andrewlee94, line 300: This looks like a very bad idea. Why is the config argument itself appearing in the Expression - this looks like it should just be a unit level `Var`.

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -291,34 +297,38 @@ def _make_params(self):
...
-            _plate_pitch = self.config.plate_pact_length / _total_plate_number
+            _plate_pitch = self.config.plate_pact_length*pyunits.m / _total_plate_number
 
-            _plate_gap = _plate_pitch - self.config.plate_thickness
+            _plate_gap = _plate_pitch - self.config.plate_thickness*pyunits.m
         else:
-            _plate_gap = self.config.plate_gap
+            _plate_gap = self.config.plate_gap*pyunits.m
 
         self.plate_gap = Expression(expr=_plate_gap)
```
andrewlee94, line 314: As above - it seems like a really bad idea to build the config arguments directly into the Expressions.

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -334,83 +344,74 @@ def _make_params(self):
...
-                                      solvent_list,
-                                      rule=rule_cp_comp_hot,
-                                      doc='Component mean specific heat capacity'
-                                      ' btw inlet and outlet'
-                                      ' of hot-side temperature')
+        solvent_list = \
+            self.config.hot_side.property_package.solvent_set
+
+        def rule_mass_frac_solvent_hot(blk, t, j):
+            return blk.hot_fluid.properties_in[t].mass_frac_phase_comp['Liq', j] /\
```
andrewlee94, line 351: What exactly is this meant to be? Specifically, what is the correct value if `j` is not a solvent? Should this return `Expression.Skip` if `j` is not a solvent?

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -426,22 +427,22 @@ def rule_cp_cold(blk, t):
                           self.PH, initialize=390, units=pyunits.K,
```
andrewlee94, line 427: Hard coded units here will be a problem.

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -455,42 +456,43 @@ def rule_Gpc(blk, t):
...
                               doc='Coldside port mass velocity[kg/m2.s]')
 
-        # ----------------------------------------------------------------------
+        # ---------------------------------------------------------------------
         # Reynold & Prandtl numbers
         def rule_Re_h(blk, t, p):
             return blk.mh_in[t] * blk.channel_dia /\
                 (blk.Np[p] * blk.plate_width *
-                 blk.plate_gap * blk.hot_side.properties_in[t].visc_d)
+                 blk.plate_gap * blk.hot_fluid.properties_in[t].visc_d_phase["Liq"])
```
andrewlee94, line 464: This will probably need to have unit conversions to ensure consistency (unless you ensure that the spatial dimensions are consistent with the property package).

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -521,16 +524,16 @@ def rule_fric_c(blk, t):
                                  rule=rule_fric_c,
                                  doc='Coldside friction factor')
```
andrewlee94, line 525: Cold side

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -521,16 +524,16 @@ def rule_fric_c(blk, t):
...
                     blk.P * blk.Gph[t]**2) /\
-                (blk.hot_side.properties_in[t].dens_mass *
+                (blk.hot_fluid.properties_in[t].dens_mass *
                  blk.channel_dia) + 1.4 * blk.P * blk.Gph[t]**2 * 0.5 /\
-                blk.hot_side.properties_in[t].dens_mass + \
-                blk.hot_side.properties_in[t].dens_mass * \
-                9.81 * (blk.plate_length + blk.port_dia)
+                blk.hot_fluid.properties_in[t].dens_mass + \
+                blk.hot_fluid.properties_in[t].dens_mass* \
+                CONST.acceleration_gravity * (blk.plate_length + blk.port_dia)
```
andrewlee94, line 536: This will need unit conversions too.

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -653,32 +662,34 @@ def rule_Ebc_eq(blk, t, p):
...
                                        rule=rule_Tout_cold,
                                        doc='Cold fluid exit temperature')
 
-        # ----------------------------------------------------------------------
+        # ---------------------------------------------------------------------
         # Energy balance boundary conditions
         def rule_hot_BCIN(blk, t):
             return blk.Th_in[t, 1] == \
-                blk.hot_side.properties_in[t].temperature
+                blk.hot_fluid.properties_in[t].temperature
```
andrewlee94, line 684: Unit conversion.

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -687,29 +698,32 @@ def rule_cold_BCIN(blk, t):
 
         def rule_hot_BC(blk, t, p):
             return blk.Th_out[t, p] == blk.Th_in[t, p + 1]
-        self.hot_BC = Constraint(self.flowsheet().time, Pset,
```
andrewlee94, line 690: What is the purpose of Pset here?

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -179,86 +182,89 @@ def build(self):
...
+        self.cold_fluid = ControlVolume0DBlock(default={
             "dynamic": self.config.dynamic,
             "has_holdup": self.config.has_holdup,
             "property_package": self.config.cold_side.property_package,
-            "property_package_args": self.config.cold_side.property_package_args})
+            "property_package_args":
+                self.config.cold_side.property_package_args})
 
-        self.cold_side.add_state_blocks(has_phase_equilibrium=False)
+        self.cold_fluid.add_state_blocks(has_phase_equilibrium=False)
```
andrewlee94, line 225: Where is the energy balance for the system? Even with the NTU calculations, I would expect to see an proper energy balance for each control volume with a heat transfer term, a constraint that says Qhot = -Qcold, and then the NTU calculations used to determine the value of one of the Q terms.

andrewlee94 (changes requested):
1. This PR will need to wait until some of the other PRs have been merged. It is expected that it will depend on the SovlentReboiler PR, but I also see some other things have slipped in.
2. There are a lot of areas where unit conversions are needed- at the moment this model presumes units on everything.
3. I do not see a proper energy balance in this model.

idaes/generic_models/unit_models/plate_heat_exchanger.py
andrewlee94: Mass flow should already be a property.

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -426,22 +427,22 @@ def rule_cp_cold(blk, t):
...
+            return blk.cold_fluid.properties_in[t].flow_mol *\
+                blk.cold_fluid.properties_in[t].mw
         self.mc_in = Expression(self.flowsheet().time, rule=rule_mc_in,
                                 doc='Coldside mass flow rate [kg/s]')
 
-        # ----------------------------------------------------------------------
+        # ---------------------------------------------------------------------
         # port mass velocity[kg/m2.s]
         def rule_Gph(blk, t):
             return (4 * blk.mh_in[t] * 7) / (22 * blk.port_dia**2)
```
andrewlee94, line 448: Where do these hard coded constants come from? There should at a minimum be a comment here with an explanation or reference for the source of these.

idaes/generic_models/unit_models/plate_heat_exchanger.py
PTA0001: @andrewlee94, this might be the source of the unit consistency(applies to other similar comments), I will take the first pass at converting this but let's look at it during our 10 am meeting today.

PTA0001:
@andrewlee94 , @MAZamarripa , @anujad95 
I have updated the phe to capture the review comments. On the unit consistency, while trying to convert the unit for dynamic viscosity it gives an error that Pa/s not equal to Pa.s.  Hope to discuss this in our meeting today.(I commented the code out)

andrewlee94:
@PTA0001 There is an import error in your code somewhere that is preventing the tests from running. This really should be addressed before we review the code so that we can confirm that the tests pass.

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -284,63 +295,108 @@ def _make_params(self):
...
-        self.param_a = Var(initialize=0.3, bounds=(0.2, 0.4), units=None,
-                           doc='Nusselt parameter')
-        self.param_b = Var(initialize=0.663, bounds=(0.3, 0.7), units=None,
-                           doc='Nusselt parameter')
-        self.param_c = Var(initialize=1 / 3.0, bounds=(1e-5, 2), units=None,
-                           doc='Nusselt parameter')
-        self.param_a.fix(0.4)
-        self.param_b.fix(0.663)
-        self.param_c.fix(0.333)
+        self.nusselt_param_a = Var(initialize=0.3,
```
andrewlee94, line 369: Are these really `Vars`? These look like coefficients for a correlation, and thus I question whether these should be varied during optimization of the unit.

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -284,63 +295,108 @@ def _make_params(self):
...
+                                   bounds=(1e-5, 2),
+                                   units=pyunits.dimensionless,
+                                   doc='Nusselt parameter')
+
+        self.nusselt_param_a.fix(0.4)
+        self.nusselt_param_b.fix(0.663)
+        self.nusselt_param_c.fix(0.333)
+
+        # friction factor parameters
+        self.frict_factor_param_a = Var(initialize=0.0,
```
andrewlee94, line 387: As above.

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -429,160 +485,200 @@ def rule_cp_cold(blk, t):
...
-        def rule_mc_in(blk, t):
-            return blk.cold_fluid.properties_in[t].flow_mol *\
-                blk.cold_fluid.properties_in[t].mw
-        self.mc_in = Expression(self.flowsheet().time, rule=rule_mc_in,
-                                doc='Coldside mass flow rate [kg/s]')
 
-        # ---------------------------------------------------------------------
-        # port mass velocity[kg/m2.s]
+        # G:mass flow velocity, p :port , h:hotfluid , c :coldfluid
+        # Gph is hotfluid port mass velocity[kg/m2.s]
```
andrewlee94, line 490: Descriptive names.

idaes/generic_models/unit_models/plate_heat_exchanger.py
```diff
@@ -429,160 +485,200 @@ def rule_cp_cold(blk, t):
...
         def rule_Gph(blk, t):
-            return (4 * blk.mh_in[t] * 7) / (22 * blk.port_dia**2)
+            return (4 * blk.hot_fluid.properties_in[t].flow_mass * 7) /\
```
andrewlee94, line 492: Again, I would like to know where these hard-coded constants are coming from. A comment may suffice.

idaes/generic_models/unit_models/plate_heat_exchanger.py
andrewlee94: @PTA0001 a quick check of the property packages shows that the units were accidentally set to Pa/s instead of Pa.s (line 450 in https://github.com/IDAES/idaes-pse/blob/main/idaes/power_generation/carbon_capture/mea_solvent_system/properties/MEA_solvent.py).

andrewlee94 (changes requested):
I had a look through the changes and added a few more comments. Overall, it is coming along but I still see a number of things that need to be fixed to meet a minimum standard:

1. The hot and cold side control volumes need to have energy balances added with a heat transfer term, and there needs to be a constraint that says `Q_hot == -Q_cold`. The NTU method can then be used to define the value of `Q`.
2. Every component should have a descriptive name, not just the ones I explicitly point out.
3. There are a number of hard-coded constants that appear in the model which should be explained in comments at the least.

I also have a theoretical question about the way the model is structured. You are using the overall inlet and outlet conditions to get the `cp` values and using them across the entire unit. However, internally the model is broken down into sub-exchangers (each plate pass I believe). This would seem to imply that you really should be calculating `cp` at the inlet and outlet of each sub-exchanger; alternatively you could just apply the NTU method across the exchanger as a whole with some correlation for `eps` that accounts for the multiple passes.

andrewlee94:
Replaced by #590.

PR #582, opened 2021-10-31
Generalized Packed Column

anujad95:
## Summary/Motivation:
Introducing a generalized packed column model in IDAES 

- Generalized mass and heat transfer correlations, hydraulics correlations and initialization routine
- Alignment with generic property package framework

## Work in progress
- Updating the column test

### Legal Acknowledgement


idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -140,12 +138,18 @@ class PackedColumnData(UnitModelBlockData):
         default=250,
```
andrewlee94, line 138: Should these be config arguments, or are they just default values for design variables the user should be fixing later?

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -140,12 +138,18 @@ class PackedColumnData(UnitModelBlockData):
...
         default=0.97,
         domain=float,
         description="Void fraction of the packing",
         doc="Packing porosity or void fraction (default= 0.97 )"))
+    CONFIG.declare("packing_channel_size", ConfigValue(
+        default=0.1,
+        domain=float,
+        description="Channel size of the packing",
+        doc="Packing channel size default = 0.1 m"))
     CONFIG.declare("fix_column_pressure", ConfigValue(
```
andrewlee94, line 153: Is this the same as `has_pressure_drop`? If so, we should use the standard name.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -161,29 +165,35 @@ class PackedColumnData(UnitModelBlockData):
         domain=float,
```
andrewlee94, line 165: As above, I am not sure why this is a config argument.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -161,29 +165,35 @@ class PackedColumnData(UnitModelBlockData):
...
         description="fixed column pressure in Pa",
         doc="Fixed column operating pressure in Pa"))
+    CONFIG.declare("fix_column_pressure_drop", ConfigValue(
```
andrewlee94, line 168: Similarly, isn;t this just a design condition the user should be fixing (or providing a sub-model for)?

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -161,29 +165,35 @@ class PackedColumnData(UnitModelBlockData):
...
+If this is fixed, the column pressure profile will be calculated accordingly.
+If this is not fixed, pressure drop will be calculated based on the inbuilt correlation."""))
+    CONFIG.declare("simple_flooding_velocity_calculation", ConfigValue(
+        default=True,
+        domain=Bool,
+        description="Indicates whether the simple flooding velocity calculation is to be used",
+        doc="""Indicates whether the simple flooding velocity calculation is to be used.
+If this is true, the gas flooding velocity will be calculated based on a simplified correlation.
+If this is not true, gas flooding velocity will be calculated based on the inbuilt correlation."""))
+    CONFIG.declare("column_pressure_drop", ConfigValue(
```
andrewlee94, line 182: Doesn't this conflict with some of the options above?

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -161,29 +165,35 @@ class PackedColumnData(UnitModelBlockData):
...
         description="fixed column pressure in Pa",
         doc="Fixed column operating pressure in Pa"))
+    CONFIG.declare("fix_column_pressure_drop", ConfigValue(
+        default=True,
+        domain=Bool,
+        description="Indicates whether the column pressure drop should be fixed or not",
+        doc=""""Indicates whether the column pressure drop should be fixed or not.
+If this is fixed, the column pressure profile will be calculated accordingly.
+If this is not fixed, pressure drop will be calculated based on the inbuilt correlation."""))
+    CONFIG.declare("simple_flooding_velocity_calculation", ConfigValue(
```
andrewlee94, line 175: I am not sure what is meant by this - from reading the doc string it makes me think that this should be a general option for providing a flooding correlation (i.e. we ask the user to explicitly provide a method, of which we have a library of some standard option). We should also allow `None` as an option for cases where flooding is not an issue.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -235,7 +245,7 @@ def build(self):
         # Call UnitModel.build to build default attributes
         super(PackedColumnData, self).build()
```
andrewlee94, line 246: This can just be `super().build()`

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -235,7 +245,7 @@ def build(self):
...
```
andrewlee94, line 247: There should be some checking of general assumptions here - i.e. that the vapour property package has one phase named "Vap", the liquid properties have one phase names "Liq" and that there is at least 1 shared component between the two property packages.

See the Solvent Reboiler model for examples.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -286,40 +298,15 @@ def build(self):
                 balance_type=MomentumBalanceType.pressureTotal,
                 has_pressure_change=self.config.vapor_side.has_pressure_change)
 
-            # TO DO : remove this warning when there is support for deltaP
-            warnings.warn("""{} WARNING! WARNING!! WARNING!!!
-                  control volume class has not implemented a method
-                  for pressure drop. Constraint for deltaP must be provided if
-                  has_pressure_change is set to True""".format(self.name))
         # consistency check
         if (self.config.vapor_side.has_pressure_change and
```
andrewlee94, line 302: As mentioned above, we need to clean up the config arguments so that we don't have these possible clashes.

Normally we wouldn't have a config argument for fixing pressure drop anyway - that is an standard design variable the user needs to fix themselves or add a performance equation for.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -383,17 +370,18 @@ def _make_performance(self):
...
-        liquid_phase_list_ref = self.config.liquid_side.property_package.phase_list
-
-        # Add object reference - time
-        add_object_reference(self,
-                             "t",
-                             self.flowsheet().time)
+        equilibrium_comp = vap_comp & liq_comp
+        solvent_comp_list = \
+            self.config.liquid_side.property_package.solvent_set
+        solute_comp_list = self.config.liquid_side.property_package.solute_set
```
andrewlee94, line 379: We need to think about this more - either we need to make it a requirement that the liquids phase property package defines solvents and solutes (and that we add a check for this), or we need to generalize the model more.

My initial feeling is that we do not need this, at least at the general column level, and that we just need to make the assumption that any component that appears in both phases is in equilibrium, and any that appears in only one phase is non-condensable/non-vapourisable.

Again, see the Solvent Reboiler model for an example.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -402,13 +390,17 @@ def _make_performance(self):
 
         self.a_ref = Param(initialize=self.config.packing_specific_area,
```
andrewlee94, line 391: Needs a more descriptive name.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -402,13 +390,17 @@ def _make_performance(self):
...
                            units=pyunits.m**2 / pyunits.m**3,
-                           doc="Packing specific surface area m2/m3")
+                           doc="Packing specific surface area (m2/m3)")
 
-        self.dh_ref = Expression(expr= 4 * self.eps_ref /self.a_ref,
-                                 doc="Hydraulic diameter")
+        self.dh_ref = Expression(expr=4 * self.eps_ref / self.a_ref,
```
andrewlee94, line 395: This definitely needs a better name. At the moment it clashes with heat of reaction, etc.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -402,13 +390,17 @@ def _make_performance(self):
...
+                                 doc="Hydraulic diameter (m)")
+        
+        self.S_ref = Param(initialize=self.config.packing_channel_size,
```
andrewlee94, line 398: As above.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -402,13 +390,17 @@ def _make_performance(self):
...
+                           units=pyunits.m,
+                           doc="Packing channel size (m)")
 
-        # specific constants for volumetric mass transfer coefficients
-        # reference:  Billet and Schultes, 1999
+        # Specific constants for volumetric mass transfer coefficients
```
andrewlee94, line 402: I would leave them here for now whilst we get things working, but these should eventually be pulled out in to a sub-model that the user can optionally include.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -402,13 +390,17 @@ def _make_performance(self):
...
+        # Reference:  Billet and Schultes, 1999
         self.Cv_ref = Var(initialize=0.357,
```
andrewlee94, line 404: Naming again. This clashes with constant volume specific heat capacity.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -422,53 +414,48 @@ def _make_performance(self):
...
+        g_ref = CONST.acceleration_gravity
 
         # Unit Model Parameters/sets
         self.zi = Param(self.vapor_phase.length_domain, mutable=True,
                         doc='''Integer indexing parameter required for transfer
                              across boundaries of a given volume element''')
-        # Set the integer  indices
+        # Set the integer indices along vapor phase length domain
         for i, x in enumerate(self.vapor_phase.length_domain, 1):
             self.zi[x] = i
```
andrewlee94, line 425: Do we actually need this?

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -422,53 +414,48 @@ def _make_performance(self):
...
-
-        # fixed column pressure
-        if self.config.fix_column_pressure:
-            self.column_pressure = Param(initialize=self.config.column_pressure,
-                                         mutable=True,
-                                         units=pyunits.Pa,
-                                         doc='Fixed operating pressure of column')
-
-        # Interfacial area  parameters
+        self._homotopy_par_m = Param(
```
andrewlee94, line 428: 1. We should look at whether we actually need this, especially if we start with a simpler model with fixed mass transfer coefficients, enhancement factor and pressure drop.
2. I had some success in earlier testing of removing the energy balance homotopy step at least.
3. If it is needed, this should be an argument to initialization, not a parameter on the model.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -422,53 +414,48 @@ def _make_performance(self):
...
                                units=None,
-                               doc='''holdup parameter B''')
+                               doc='Holdup parameter B')
         self.holdup_parA.fix(24.2355)
         self.holdup_parB.fix(0.6471)
 
-
         # Unit Model Variables
         # Geometry
         self.diameter_column = Var(domain=Reals,
```
andrewlee94, line 461: These should be declared before the control volumes - notably, length is required for the control volumes so we should ensure they are consistent.

In fact, we should be creating the length domain at the unit model level and passing this to the control volumes (that is a supported option) to ensure the domains are consistent, and to make it easier to provide custom meshing.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -485,66 +472,268 @@ def _make_performance(self):
...
-                       domain=Reals,
-                       initialize=0.0,
-                       units=pyunits.mol / (pyunits.s * pyunits.m),
-                       doc='''Moles of diffusing species transfered
-                                     into liquid ''')
-        self.enhancement_factor = Var(self.t,
+        
+        # Liquid holdup model    
+        if self.config.fix_column_pressure or self.config.fix_column_pressure_drop:
+            # reference: Tsai correlation,regressed by Chinen et al. 2018
```
andrewlee94, line 490: Typo: start with capital, space after comma.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -485,66 +472,268 @@ def _make_performance(self):
...
+            return blk.eps_ref - blk.holdup_liq[t, x]
+
+        self.holdup_vap = Expression(self.flowsheet().time,
+                                     self.vapor_phase.length_domain,
+                                     rule=rule_holdup_vap,
+                                     doc='Volumetric vapor holdup [-]')
+        
+        # Column pressure profile
+        if self.config.fix_column_pressure:
+            self.column_pressure = Param(
```
andrewlee94, line 531: This should not be a Param. Users should either provide a model for delta P, or just fix the variable directly.

On that note, we should be creating a reference to one of the delta P variables before this, along with a constraint equating delta P in each phase. Then, users can just work with the unit level delta P without needing to be concerned about the control volumes.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -485,66 +472,268 @@ def _make_performance(self):
...
+            
+        self.flood_fraction = Expression(self.flowsheet().time,
+                                    self.vapor_phase.length_domain,
+                                    rule=rule_flood_fraction,
+                                    doc='Flooding fraction (expected to be below 0.8)')
+        
+        # Mass and heat transfer terms
+        
+        # Mass transfer terms
+        self.pressure_equil = Var(
```
andrewlee94, line 702: For the future, we probably want to think about this some more. In simpler models (without the film theory), this comes directly from the property packages.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -485,66 +472,268 @@ def _make_performance(self):
...
-                           domain=NonNegativeReals,
-                           initialize=0.5,
-                           units=None,
-                           doc='''Dimensionless concentration of CO2
-                                      in equilibruim with the bulk''')
-
-        # heat transfer
-        self.heat_vap = Var(self.t,
+        # Heat transfer
+        self.heat_vap = Var(self.flowsheet().time,
```
andrewlee94, line 730: I think a more descriptive name is needed. This could be confused for heat of vapourisation.

idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/column.py
```diff
@@ -485,66 +472,268 @@ def _make_performance(self):
...
                             self.vapor_phase.length_domain,
                             domain=Reals,
                             initialize=0.0,
                             units=pyunits.J / (pyunits.s * pyunits.m),
                             doc='Heat transfer rate in vapor phase')
-        self.heat_liq = Var(self.t,
+        self.heat_liq = Var(self.flowsheet().time,
```
andrewlee94, line 736: Similar to the above.

andrewlee94 (changes requested):
There is a lot to review here, so I probably missed a few things. However, some general comments:

1. There is currently a syntax error somewhere in the column model which is causing tests to fail.
2. We need to have tests for the column model.
3. I would suggest we start even simpler with the column model first to get things working. For now, I would suggest we start with the simplest possible column model to make this easier to review and debug, and to add detail later once we  have the basics in place. I.e. start with just a model of the form `transfer = K*(y-y_eq)` where `K` is a fixed constant. Once we have the foundation in place, then we can start adding details like pressure drop corrections, mass transfer models, etc.

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1497 @@
...
+        
+        # Column pressure profile
+        if not self.config.vapor_phase.has_pressure_change:
+            self.column_pressure = Var(
+                initialize=107650,
+                units=pyunits.Pa,
+                doc='Fixed operating pressure of column')
+        else:                
+            def rule_deltaP(blk, t, x):
+                return blk.vapor_phase.deltaP[t, x] == self.config.column_pressure_drop
```
PTA0001, line 428: what kind of object is `self.config.column_pressure_drop` ? if this is not a Param or Var , it will cause issues

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1497 @@
...
+                ((blk.liquid_phase.properties[t, x_liq].visc_d_phase['Liq']/mu_water)*(-0.2))*\
+                    exp(-4*(H)**(0.25)))**0.5
+                
+        self.gas_velocity_flood = Expression(self.flowsheet().time,
+                                self.vapor_phase.length_domain,
+                                rule=rule_flood_velocity,
+                                doc='Gas velocity at flooding point')
+            
+        # Flooding fraction 
+        def rule_flood_fraction(blk, t, x):
```
PTA0001, line 465: We need to apply this rule to construct the flood_fraction `Constraint` to allow for the calculation of the required vapor velocity to determine the column diameter during design. Hence `self.flood_fraction` should be a  variable. So during simulation, the `flood_fraction` is calculated and for design purposes, it is fixed as desired by the user to determine the corresponding velocity and then the diameter of the column from the admissible flows.

PTA0001 (commented):
I looked at the new hydraulics equations and made comments as required.

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+# Solutions of Sandia, LLC, Carnegie Mellon University, West Virginia University
+# Research Corporation, et al.  All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and
+# license information.
+#################################################################################
+"""
+IDAES First Generation (GEN1) General Rate-Based Packed Column.
+
+Detailed model equations can be found in the supplimentary information
```
andrewlee94, line 16: This should be removed from the general model part.

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+        doc="""Number of collocation points to use per finite element when
+discretizing length domain (default=3)"""))
+
+    CONFIG.declare("column_pressure_drop", ConfigValue(
+        default=0,
+        description="Column pressure drop per unit length in Pa/m",
+        doc="Column pressure drop per unit length in Pa/m provided as a value or expression"))
+
+    # Populate the phase side template to default values
+    _PhaseCONFIG.declare("has_pressure_change", ConfigValue(
```
andrewlee94, line 116: This should be a global level argument - this setting should be the same for both phases.

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+        Returns: None
+
+        """
+
+        # ======================================================================
+        # Custom Sets
+        vap_comp = self.config.vapor_side.property_package.component_list
+        liq_comp = self.config.liquid_side.property_package.component_list
+        equilibrium_comp = vap_comp & liq_comp
+        solvent_comp_list = \
```
andrewlee94, line 292: Do we need the solvent and solute sets in the generic model? I would think that the general model would just define mass transfer for all common components, and the user then needs to define these.

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+            self.config.liquid_side.property_package.solvent_set
+        solute_comp_list = self.config.liquid_side.property_package.solute_set
+        liq_comp_apparent = self.config.liquid_side.property_package._apparent_set
```
andrewlee94, line 295: Similarly, `_apparent_set` is not a standard attribute of property packages (it only appears if electrolytes are used).

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+        vapor_phase_list_ref = \
+            self.config.vapor_side.property_package.phase_list
+        liquid_phase_list_ref = \
+            self.config.liquid_side.property_package.phase_list
+
+        # Packing  parameters
+        self.eps_ref = Param(initialize=0.97,units=None,
```
andrewlee94, line 302: Units should be `pyunits.dimensionless` to be explicit (i.e. it is unitless, versus we were just too lazy to specify).

I would also suggest `packing_voidage` instead of `eps_ref`, as we use `eps` for a lot of purposes.

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+                                doc='Vapor superficial velocity')
+        
+        self.velocity_liq = Var(self.flowsheet().time,
+                                self.liquid_phase.length_domain,
+                                domain=NonNegativeReals,
+                                initialize=0.01,
+                                units=pyunits.m / pyunits.s,
+                                doc='Liquid superficial velocity')
+        
+        self.holdup_liq = Var(self.flowsheet().time,
```
andrewlee94, line 358: 1. I think we should call this `liquid_fraction` - it is not a `holdup` itself. This also applies to the vapour holdup below.
2. We do not need both the variable for liquid fraction and the expression for vapor fraction. Just replace the vapor fraction with `1-liquid_fraction` wherever it appears.

Also, units should be specified (`pyunits.dimensionless`).

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+            return blk.eps_ref - blk.holdup_liq[t, x]
+
+        self.holdup_vap = Expression(self.flowsheet().time,
+                                     self.vapor_phase.length_domain,
+                                     rule=rule_holdup_vap,
+                                     doc='Volumetric vapor holdup [-]')
+        
+        # Calculating gas velocity at flooding point
+        def rule_flood_velocity(blk, t, x):
+            x_liq = blk.liquid_phase.length_domain.at(blk.zi[x].value)
```
andrewlee94, line 373: This correlation should have a doc string giving a reference , and maybe a little explanation (and note the specific units used).

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+                    blk.vapor_phase.properties[t, x].vol_mol_phase['Vap']
+                    
+            H = (blk.liquid_phase.properties[t, x_liq].flow_mass_phase['Liq']/\
+                blk.vapor_phase.properties[t, x].flow_mass_phase['Vap'])*\
+                (dens_liq/dens_vap)**0.5
+                
+            mu_water = 0.001*pyunits.Pa*pyunits.s
+            
+            g_ref = CONST.acceleration_gravity
+            gref = pyunits.convert(g_ref, to_units=pyunits.ft/(pyunits.s)**2)
```
andrewlee94, line 387: Very minor comment, but I would just combine lines 386 and 387 into one.

idaes/generic_models/unit_models/column_models/column.py
andrewlee94: This is probably the best way to do this.

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+        
+        # Mass transfer terms
+        self.pressure_equil = Var(
+            self.flowsheet().time,
+            self.vapor_phase.length_domain,
+            equilibrium_comp,
+            domain=NonNegativeReals,
+            initialize=500,
+            units=pyunits.Pa,
+            doc='Equilibruim pressure of diffusing components at interface')
```
andrewlee94, line 419: I would remove "diffusing" from this comment, as the generic model does not concern itself with diffusion (that is for the more detailed/specific models to worry about).

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+            doc='Rate at which moles of diffusing species transfered into liquid')
+        
+        self.enhancement_factor = Var(self.flowsheet().time,
+                                      self.liquid_phase.length_domain,
+                                      units=None,
+                                      initialize=160,
+                                      doc='Enhancement factor')
+
+        # Heat transfer terms
+        self.heat_transfer_rate_vap = Var(self.flowsheet().time,
```
andrewlee94, line 437: Are these actually needed? The control volumes have built their internal mass transfer terms already, and they should both be equal. At most, I think we want a `Reference` to the mass transfer term in one of the phases, and maybe not even that.

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+            has_phase_equilibrium=False)
+
+        self.vapor_phase.add_material_balances(
+            balance_type=MaterialBalanceType.componentTotal,
+            has_phase_equilibrium=False,
+            has_mass_transfer=True)
+
+        self.vapor_phase.add_energy_balances(
+            balance_type=EnergyBalanceType.enthalpyTotal,
+            has_heat_transfer=True)
```
andrewlee94, line 220: The energy balances for both phases also need `has_enthalpy_transfer=True` to account for enthalpy transferred as part of the mass transfer (this is probably already lumped into the heat transfer term, but should be separated for clarity and consistency).

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+                             
+        # Set the integer indices along vapor phase length domain
+        for i, x in enumerate(self.vapor_phase.length_domain, 1):
+            self.zi[x] = i
+
+        # Unit Model Design Variables
+        # Geometry
+        self.diameter_column = Var(domain=Reals,
+                                   initialize=0.1,
+                                   units=pyunits.m,
```
andrewlee94, line 330: This should wait until we've cleaned up everything else, but the units of all these variables should be drawn from one of the property packages.

See lines 257 and 277 in https://github.com/IDAES/idaes-pse/blob/main/idaes/generic_models/unit_models/column_models/solvent_condenser.py for an example.

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+                    bk.area_column * bk.holdup_vap[t, x])
+
+            @self.Constraint(self.flowsheet().time,
+                             self.liquid_phase.length_domain,
+                             doc=control_volume_area_definition)
+            def liquid_side_area(bk, t, x):
+                return bk.liquid_phase.area[t, x] == (
+                    bk.area_column * bk.holdup_liq[t, x])
+        else:
+            self.vapor_phase.area.fix(value(self.area_column))
```
andrewlee94, line 494: I think we should just write the constraints at all times - yes, fixing the variables makes the problem a bit less complicated and removes a couple of trivial constraints, but doing so adds a structural difference between dynamic and steady-state models. We have pre-solve tools that can clean up the trivial constraints if required.

Also, why are the phase areas fixed to the column area rather than `area*phase fraction`?

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+                         self.liquid_phase.length_domain,
+                         doc='''Mechanical equilibruim: vapor-side pressure
+                                    equal liquid -side pressure''')
+        def mechanical_equil(bk, t, x):
+            return bk.liquid_phase.properties[t, x].pressure == \
+                    bk.vapor_phase.properties[t, x].pressure
+
+        # Length of control volume : vapor side and liquid side
+        @self.Constraint(doc="Vapor side length")
+        def vapor_side_length(blk):
```
andrewlee94, line 508: Add a `TODO` here saying we should add a way to define `length` at the unit model level and make `References` to that in the control volumes, and then clean up these equality constraints,

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+
+        # Equilibruim partial pressure of diffusing components at interface
+        @self.Constraint(self.flowsheet().time,
+                          self.vapor_phase.length_domain,
+                          equilibrium_comp,
+                          doc='''Equilibruim partial pressure of diffusing
+                                components at interface''')
+        def pressure_at_interface(blk, t, x, j):
+            if x == self.vapor_phase.length_domain.first():
+                return blk.pressure_equil[t, x, j] == 0.0
```
andrewlee94, line 575: Should this be `Constraint.Skip`?

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+                else:
+                    return blk.pressure_equil[t, x, j] == (
+                        lprops.vol_mol_phase['Liq'] *
+                        lprops.conc_mol_phase_comp_true['Liq',j] *
+                        lprops.pressure_sat_comp[j])
+
+        # Mass transfer of  diffusing components in vapor phase
+        def rule_mass_transfer(blk, t, x, j):
+            if x == self.vapor_phase.length_domain.first():
+                return blk.interphase_mass_transfer[t, x, j] == 0.0
```
andrewlee94, line 597: I think this could be `Constraint.Skip` (although some will argue otherwise).

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+
+        self.mass_transfer_vapor = Constraint(self.flowsheet().time,
+                                        self.vapor_phase.length_domain,
+                                        liq_comp,
+                                        rule=rule_mass_transfer,
+                                        doc="mass transfer in vapor phase")
+
+        # Liquid phase mass transfer handle
+        @self.Constraint(self.flowsheet().time,
```
andrewlee94, line 615: This all looks more complicated than necessary to me right now (I could be wrong). What I was thinking of was:

1. A constraint across all species that says `liquid_phase.mass_transfer[t, x, p, j] == vapor_phase.mass_transfer[t, x, p, j]` for all equilibrium species, `liquid_phase.mass_transfer[t, x, p, j] == 0` for liquid only species and `vapor_phase.mass_transfer[t, x, p, j] == 0` for all vapour only components.
2. A constraint across all equilibrium components that says `liquid_phase.mass_transfer[t, x, p, j] == K*A*A_inter*(y*P - P_equil)`.
3. Some constraint that defines `P_equil` which for the generic model is probably `P_equil[t, x, p, j] == liquid_phase.properties[t, x].fugactiy_phase_comp[p, j]` (but we would need to modify this for more detailed models.
4. `K` should either be a fixed value, or there should be a simple constraint that describes the actual `K` in terms of enhancement factor and `K_prime` measured by experiments.

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+        # Heat transfer coefficients
+        # Vapor-liquid heat transfer coefficient [J/m2.s.K]
+
+        self.h_v = Var(self.flowsheet().time,
+                       self.vapor_phase.length_domain,
+                       initialize=100,
+                       doc='''Vapor-liquid heat transfer coefficient''')
+
+        # Vapor-liquid heat transfer coeff modified by Ackmann factor [J/m.s.K]
+        def rule_heat_transfer_coeff_Ack(blk, t, x):
```
andrewlee94, line 656: Similarly, I think the generic model should just have `Q=U*A_inter*deltaT` - definition of `U` is something for more detailed models.

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+                     blk.interphase_mass_transfer[t, zf, j] for j in solute_comp_list) -
+                     sum(blk.liquid_phase.properties[t, x].enth_mol_phase_comp['Liq',k] *
+                     blk.interphase_mass_transfer[t, zf, k] for k in solvent_comp_list))
+
+        # Vapor phase heat transfer handle
+        @self.Constraint(self.flowsheet().time,
+                         self.vapor_phase.length_domain,
+                         doc="vapor - heat transfer handle")
+        def vapor_phase_heat_transfer_handle(blk, t, x):
+            return blk.vapor_phase.heat[t, x] == blk.heat_transfer_rate_vap[t, x]
```
andrewlee94, line 706: If this is needed in the generic model, this constraint should just be combined with the one above.

idaes/generic_models/unit_models/column_models/column.py
```diff
@@ -0,0 +1,1144 @@
...
+
+        # Liquid phase heat transfer handle
+        @self.Constraint(self.flowsheet().time,
+                         self.liquid_phase.length_domain,
+                         doc="liquid - heat transfer handle")
+        def liquid_phase_heat_transfer_handle(blk, t, x):
+            return blk.liquid_phase.heat[t, x] == -blk.heat_transfer_rate_liq[t, x]
+
+        if self.config.dynamic:
+            self.fix_initial_condition()
```
andrewlee94, line 716: This should not be part of the unit model. This is something the user should do when initializing the model, rather than done automatically during model construction.

andrewlee94:
@anujad95 I just tired checkout out this branch on my compute to confirm the tests and read through the code, and ran into a few problems:

1. The tests for the MEA column model (`idaes/power_generation/carbon_capture/mea_solvent_system/unit_models/tests/test_column.py`) all fail or produce errors on my computer.
2. Github is showing conflict between this PR and the main branch which will need to be resolved.

Finally, could you rename the generic column model to `solvent_column` rather than just `column` please - we need to be specific as to what type of column this is so that users can identify which model they want.

andrewlee94 (commented):
@anujad95 I've looked into the conflicts and it looks like it might just be confusing on GitHub's end. Try just updating your branch with the latest main branch and hopefully the conflicts will disappear.

One question I had was are we using the changes you made to Paul's column model at all ()the one in `power_generation`? As I mentioned, I am seeing test failures with it at the moment, and I am inclined to say we just leave it as it was for now unless there is a need for it (i.e. undo any changes you made).

Other than that, whilst there are a lot of thing that still need to be cleaned up, I think we can go ahead and merge this PR. We can then go about cleaning up the rest later. That said, I don't think we should put this in the November release, as it is not ready for public use yet - I think we wait until after the release (hopefully this week?) to merge it so that we have three months to get it cleaned up.

anujad95:
@andrewlee94 at this point, the simple generalized column model works, however the derived version of this model for MEA capture system will be made available separately, after detailed validation.

andrewlee94:
@anujad95 The problem is that you have made changes to the original model in `power_generation` which is likely to cause test failures when we come to running the integration tests. I am seeing test failures when I run those tests locally.

andrewlee94:
@anujad95 You can ignore the test failures here - they are occurring because we are halfway through a release at the moment. Otherwise, I think this is good to merge, but we will hold off until after the current release.

PR #588, opened 2021-11-03
Entropy fix, partial molar properties, and log mole fractions

dallan-keylogic:
## Fixes
- Entropy of mixing added to `entr_mol_phase`

## Summary/Motivation:
Several weeks ago I noticed that `entr_mol_phase` in `ceos.py` lacked an entropy of mixing term. That fix was easy, but the question of what to do with `entr_mol_phase_comp` lead me down a rabbit hole to implement real partial molar quantities. 

- `energy_internal_mol_phase_comp`, `enth_mol_phase_comp`, and `entr_mol_phase_comp` changed to be meaningful partial molar quantities instead of whatever they were before
- `vol_mol_phase_comp` added
- Refactored internals to more easily support different functions for a(T)
- Made it clear in (some) places that properties are mixing rule dependent

### Legal Acknowledgement


dallan-keylogic:
Looks like a Gibbs reactor example is failing. The problem might be that a property block is getting mole fractions of zero:

```
        m.fs.reactor.inlet.mole_frac_comp[0, "H2"].fix(0)
        m.fs.reactor.inlet.mole_frac_comp[0, "CO"].fix(0)
        m.fs.reactor.inlet.mole_frac_comp[0, "H2O"].fix(0.3)
        m.fs.reactor.inlet.mole_frac_comp[0, "CO2"].fix(0)
        m.fs.reactor.inlet.mole_frac_comp[0, "O2"].fix(0.2)
        m.fs.reactor.inlet.mole_frac_comp[0, "N2"].fix(0.05)
        m.fs.reactor.inlet.mole_frac_comp[0, "Ar"].fix(0.05)
        m.fs.reactor.inlet.mole_frac_comp[0, "CH4"].fix(0.4)
```
0 *log(0) is indeterminant, but in the context of entropy calculations it should be assigned a value of 0. Will troubleshoot this in the morning.

andrewlee94:
@dallan-keylogic There is also an issue of an implicit block replacement occurring in that test that needs to be looked into - this is quite possibly not something you did, but needs to be identified and fixed anyway,

dallan-keylogic:
It appears the Gibbs reactor was calling `gibbs_mol_phase_comp` directly , which is one of the reasons I really wanted to get partial molar quantities rather than whatever we had before. The previous code called `enth_mol_phase_comp` and `entr_mol_phase_comp` to evaluate  `gibbs_mol_phase_comp`, which was really slow after I changed those expressions. That's because those expressions have big terms involving the derivative of the fugacity coefficient. Those terms cancel out for the Gibbs energy, but (I suspect) Pyomo keeps them in their expressions, slowing things down. However, we can evaluate the Gibbs energy departure function from the fugacity coefficient directly. Doing so speeds things up massively.

In short, that test still fails, but at least it fails a lot faster. Hopefully that helps with debugging.

idaes/generic_models/properties/core/eos/ceos.py
```diff
@@ -70,10 +71,12 @@ class CubicType(Enum):
 
 
 class MixingRuleA(Enum):
+    # Rule to calculate am for cubic equations of state
     default = 0
```
andrewlee94, line 75: We probably should add another entry here with an explicit name for the default rule (van der Waal's?), and the machinery to link that up later. This would also give us the ability to properly test the machinery (even if the end result is the same).

idaes/generic_models/properties/core/eos/ceos.py
```diff
@@ -327,14 +408,48 @@ def calculate_scaling_factors(b, pobj):
...
-
+        
+        
+        setattr(param_block,cname+"_eos_options", deepcopy(CubicConfig))
+        ConfigBlock = getattr(param_block,cname+"_eos_options")
+        ConfigBlock.set_value(b.config.equation_of_state_options)
+            
+        mixing_rule_a = ConfigBlock["mixing_rule_a"]
+        mixing_rule_b = ConfigBlock["mixing_rule_b"]
+        b._mixing_rule_a = mixing_rule_a
```
andrewlee94, line 450: Minor comment, but do we need to attach the rules to the Block? They are not hard to retrieve from the config block, so it seems to be a waste of memory to do this.

idaes/generic_models/properties/core/eos/ceos.py
```diff
@@ -344,6 +459,22 @@ def build_parameters(b):
...
+        if b._cubic_type == CubicType.PR:
+            func_fw = func_fw_PR
+        elif b._cubic_type == CubicType.SRK:
+            func_fw = func_fw_SRK
+        else:
+            raise BurntToast(
+                    "{} received unrecognized cubic type. This should "
+                    "never happen, so please contact the IDAES developers "
+                    "with this bug.".format(b.name))
+        setattr(param_block, cname+"_func_fw", func_fw)
```
andrewlee94, line 472: As above, do we need to attach most of these to the Block? If it is matter of saving repeated code, then we could write a short utility unction that can retrieve these from the config block that we call when needed.

idaes/generic_models/properties/core/generic/generic_property.py
```diff
@@ -1742,7 +1785,7 @@ def calculate_scaling_factors(self):
             self.pressure, default=1, warning=True)
         sf_mf = {}
         for i, v in self.mole_frac_phase_comp.items():
-            sf_mf[i] = iscale.get_scaling_factor(v, default=1e3, warning=True)
+            sf_mf[i] = iscale.get_scaling_factor(v, default=10, warning=True)
```
andrewlee94, line 1788: I am not completely sure about this, a we use 1e3 as the default scaling factor for mole fractions everywhere else (for better or worse). Then again, users really should be providing their own scaling in a lot of cases.

andrewlee94 (commented):
A few minor comments, but the big things I think we need to do before merging are:

1. Try to validate the new entropy values against some other source - literature data would be best, but I would settle for something like CoolProp too.
2. Get someone from the NG turbine team (@eslickj?) to look this over and confirm that they are OK with the changes.

idaes/generic_models/properties/core/generic/generic_property.py
dallan-keylogic: I got this value from @eslickj when I asked him. It doesn't matter much to me either way.

idaes/generic_models/properties/core/examples/tests/test_BT_PR.py
```diff
@@ -176,9 +176,9 @@ def test_T350_P1_x5(self, m):
         assert pytest.approx(
                 value(m.fs.state[1].enth_mol_phase["Vap"]), 1e-5) == 78048.7
         assert pytest.approx(
-                value(m.fs.state[1].entr_mol_phase["Liq"]), 1e-5) == -367.558
+                value(m.fs.state[1].entr_mol_phase["Liq"]), 1e-5) == -361.794
```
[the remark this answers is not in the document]
dallan-keylogic: There are two different things that we might want to test:
1) Has our implementation changed?
2) Are the values accurate?

These sorts of checks validate (1) but not (2). There are certainly properties that we *know* give inaccurate results (see my struggles with water VLE) just from modeling using a cubic EoS.

The best bet would be to validate pure component properties against a high-fidelity EoS used by NIST or whoever. Unfortunately, this would not have been a property we could have tested against. However, we could look into creating those tests in the future, probably for hydrocarbons properties and VLE, which Cubics model really well.

If we started including experimental data, we would have to deal with the further complication of experimental error. (Not to mention it's impossible to measure entropy directly.)

If I have some extra time, I'll try to look into limiting cases for cubics. We know as pressure drops to zero they should behave like an ideal gas---we could create a battery of tests comparing them to the Ideal property package at low pressures. Even as the pressure increases some, I think we should be able to compare a second virial coefficient against some other gases.

idaes/generic_models/properties/core/eos/ceos.py
```diff
@@ -82,12 +85,25 @@ class MixingRuleB(Enum):
...
         }
 
+eps_SL = 1E-8
 
 CubicConfig = ConfigBlock()
 CubicConfig.declare("type", ConfigValue(
     domain=In(CubicType),
     description="Equation of state to use",
     doc="Enum indicating type of cubic equation of state to use."))
+CubicConfig.declare("mixing_rule_a", ConfigValue(
```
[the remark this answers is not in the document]
dallan-keylogic: On the user's end, it should be exactly the same. From the perspective of the class, the mixing rules were never made available globally, but only checked for when defining am and bm:
```
        def rule_am(m, p):
            try:
                rule = m.params.get_phase(p).config.equation_of_state_options[
                    "mixing_rule_a"]
            except (KeyError, TypeError):
                rule = MixingRuleA.default

            a = getattr(m, cname+"_a")
            if rule == MixingRuleA.default:
                return rule_am_default(m, cname, a, p)
            else:
                raise ConfigurationError(
                    "{} Unrecognized option for Equation of State "
                    "mixing_rule_a: {}. Must be an instance of MixingRuleA "
                    "Enum.".format(m.name, rule))
        b.add_component(cname+'_am',
                        Expression(b.phase_list, rule=rule_am))
```
However, I wanted to make the dependency of the fugacity coefficient on the mixing rules explicit and the bubble and dew property methods didn't even check for mixing rules, so attaching it to the parameter block made sense. I'm not completely consistent, because I also attach it to the phase object (perhaps in anticipation of permitting different mixing rules in different phases).

idaes/generic_models/properties/core/eos/ceos.py
andrewlee94: I am not sure what you mean there. Mixing rules are phase specific (they get set in the "equation_of_state_options" part of the config). My point is that the data is already stored on the model in the config block, so we should just retrieve the information from there when we need it, rather than storing it in multiple places.

dallan-keylogic: Mixing rules are set in a phase-specific way, but, as it stands, if one phase using PR has already created `am`, `bm`, etc., the second phase using PR just skips their creation entirely and uses what the first phase created. Also, I think the way we index components like `a` by phase would also cause issues if we wanted to use a cubic EoS for the vapor phase but an activity coefficient model for the liquid phase with a nonvolatile component. The nonvolatile component doesn't have Tc, Pc, or omega defined, so at present the user would have to assign them dummy values so the vapor phase could create the (unused) `am[Liq]`.

idaes/generic_models/properties/core/eos/ceos.py
```diff
@@ -82,12 +85,25 @@ class MixingRuleB(Enum):
         CubicType.SRK: {'u': 1, 'w': 0, 'omegaA': 0.42748, 'coeff_b': 0.08664}
         }
 
+eps_SL = 1E-8
```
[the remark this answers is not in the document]
dallan-keylogic: For now I just left a comment.

idaes/generic_models/properties/core/examples/tests/test_BT_PR.py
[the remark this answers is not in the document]
dallan-keylogic: Added a new test featuring comparisons to CoolProp's implementation of cubic EoS  and values of partial molar quantities calculated using a cubic EoS from a paper. I'd say that most of our methods now have at least some independent validation, but there was a bit of mismatch between the partial molar enthalpies calculated using our equations and those included in that paper. CoolProp doesn't provide partial molar quantities, so it cannot serve as a tiebreaker.

eslickj (approved):
Looks good I think.  I looked at the properties changes and I don't see anything wrong, although I may not be familiar enough with that part to really understand.  The turbine changes look good.

andrewlee94 (approved):
This all looks good to me. Well done Doug.

eslickj (approved):

andrewlee94 (approved):

PR #590, opened 2021-11-05
Generic heat exchanger using NTU method and MEA case example

andrewlee94:
## Summary/Motivation:
This PR takes the MEA specific plate heat exchanger model and turns it into a generic NTU-base heat exchanger model with a derived MEA specific case study/example.

- Create new HeatExchangerNTU model
- Refactor MEA plate heat exchanger to derive from HeatExchangerNTU

### Legal Acknowledgement


docs/technical_specs/model_libraries/power_generation/carbon_capture/mea_solvent_system/unit_models/phe.rst
```diff
@@ -98,15 +98,15 @@ post-combustion carbon capture using MEA solvent.
 PHE Class
 ----------
 
-.. module:: idaes.power_generation.carbon_capture.mea_solvent_system.unit_models.phe
```
eslickj, line 101: Should add the NTU heat exchanger page too.

eslickj (approved):
As far as I can tell it looks good.  There are a couple things you may consider.  

* I think you need a doc page for the general NTU heat exchanger.
* You may want to consider providing the Cratio expression since it is commonly used to in the effectiveness calculation.
* There are some standard effectiveness factor calculations that can be derived.  It may be nice to have a standard way to add an effectiveness factor calculation.  You know I'm a fan of callbacks.  That would allow you to work it into the initialization and make it clear how the user should add it.  I know a user could just add a constraint which is probably fine too.

eslickj (approved):

MAZamarripa (approved):
lgtm

PR #595, opened 2021-11-08
Update Sphinx theme to address unordered list spacing issues

lbianchi-lbl:
## Summary/Motivation:

The list of collaborating institutions on the IDAES docs index page should be rendered as a bulleted list in HTML, but currently isn't:

image

This PR fixes this:

image

### Legal Acknowledgement


ksbeattie:
Wow, removing the entire file, seemed to work!

lbianchi-lbl:
> Wow, removing the entire file, seemed to work!

Yeah, it looks like there was a `theme.css` that appears to be taken from a previous version of `sphinx-rtd-theme` and copied to IDAES's own `docs/static` directory that (my guess) was probably interfering in some way with the corresponding file from the `pip install`ed version of the theme.

One hint pointing in this direction is that, in the HTML of the docs built without removing the extra CSS file, `theme.css` is included twice:

image

This could also potentially explain the different behavior between local builds and RTD since the mechanism for copying and including CSS files might differ between the two.

From a very cursory glance, removing `theme.css` seems to address the issue without breaking anything else, but we should probably check to make sure that we're not missing the reason why the file was there in the first place (https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence). Maybe @dangunter can shed some light on this?

FiannaOBrien (approved):
Great work sussing it out!

ksbeattie (approved):
LGTM

PR #599, opened 2021-11-16
Fixing some bugs in alamopy wrapper

andrewlee94:
## Summary/Motivation:
@MAZamarripa identified a couple of bugs in the Alamopy wrapper.

- Corrects domain validators for mono- and multi-power config arguments to allow for any real number instead of just integers.
- Fixes bug with `evaluate_surrogate` method where calling the method a second time would result in an exception due to `fcn` being accessed before assignment.
- Added tests for the above cases.

### Legal Acknowledgement

### DIFF
--- a/idaes/surrogate/alamopy.py
+++ b/idaes/surrogate/alamopy.py
@@ -19,7 +19,7 @@ import numpy as np
 import pandas as pd
 import json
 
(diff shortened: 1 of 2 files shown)


carldlaird (approved):
Looks good to me. sorry for the bug.

eslickj (approved):

PR #604, opened 2021-11-18
Updates to Surrogate Visualization Module

bpaul4:
## Summary/Motivation:
Modifications to existing surrogate visualization utilities

- Reorganize plotting tools into main 'surrogate' and private 'non-surrogate' plotting methods
- Clean up formatting, and integrate PDF print functionality into main code for more efficient memory usage
- Minor changes in response to #599 

### Legal Acknowledgement


idaes/surrogate/plotting/sm_plotter.py
```diff
@@ -30,22 +30,30 @@
...
+    _scatter2D(xdata=input_data.values,
+               zdata=output_data.values,
+               zfit=output_surrogate.values,
+               xlabels=surrogate.input_labels(),
+               zlabels=surrogate.output_labels(),
+               filename=filename)
+
+
+def _scatter2D(xdata, zdata, zfit, xlabels=None, zlabels=None,
+               show=True, filename=None):
```
andrewlee94, line 46: Comment for discussion - the private method has a `show` argument, but the public method does not. Should we include the `show` argument in the public method as well to allow for cases where the user would prefer to just save the plots to file before looking at them?

idaes/surrogate/plotting/sm_plotter.py
```diff
@@ -60,64 +68,59 @@ def scatter2D(xdata, zdata, xtest, zfit, xlabels=None, zlabels=None, dz=None,
         zlabels = []
         for j in range(numouts):
             zlabels.append('z' + str(j+1))
-    if clabel is None:
-        clabel = 'Confidence Intervals'
 
-    pointidx = -1  # dummy index for each plot produced
-    fig, ax = [], []
+    if filename is not None:
+        if filename[-4:] != '.pdf':  # adding extension if not present
```
andrewlee94, line 73: We might want to be careful here. This will just append `.pdf` to the filename if the last four characters are not `.pdf`. However, if the user specified an extension that was not `.pdf` we would get something like `foo.bar.pdf` as the filename which might cause some problems (depending on OS).

idaes/surrogate/plotting/sm_plotter.py
```diff
@@ -60,64 +68,59 @@ def scatter2D(xdata, zdata, xtest, zfit, xlabels=None, zlabels=None, dz=None,
...
-                ax[pointidx].scatter(xdata[:, i], dz[:, j], c='grey', marker='s',
-                                     label='Data')
-                ax[pointidx].scatter(xdata[:, i], dzfit[:, j], c='b',
-                                     marker='.', label='Model')
-                ax[pointidx].set_xlabel(xlabels[i])
-                ax[pointidx].set_ylabel('d[' + str(zlabels[j]) + ']/d[' +
-                                        xlabels[i] + ']')
-                ax[pointidx].set_title('2D Derivative Plot')
-                ax[pointidx].legend()
+            fig = plt.figure()
```
andrewlee94, line 81: I am not completely familiar with the plotting tools, but this looks to me that this keeps all the plots open which will result in the memory issues @carldlaird was discussing.

idaes/surrogate/plotting/sm_plotter.py
```diff
@@ -182,55 +195,62 @@ def parity(zdata, zfit, zlabels=None, clo=None, chi=None,
...
+        error = np.divide(error, np.maximum(output_data, 1.0))
+    _residual(xdata=input_data.values, error=error.values,
+              xlabels=surrogate.input_labels(),
+              elabels=surrogate.output_labels(),
+              filename=filename)
 
 
-def residual(xdata, e, xlabels=None, elabels=None, show=True, PDF=False,
-             filename='results_residual.pdf'):
+def _residual(xdata, error, xlabels=None, elabels=None, show=True,
```
andrewlee94, line 239: Instead of "error", would "residual" be a better name? When I see "error", I tend to think exception first.

andrewlee94 (commented):
A couple of comments for thoguht/discussion.

idaes/surrogate/plotting/sm_plotter.py
bpaul4: Testing some specific cases, saving to '.xps' and no extension produce files that show the plot images with no issues (the second one is strange to me), and even saving to '.txt' produces a text file successfully. I initially wanted to ensure the PdfPages() call would print to .pdf, although we could just let the user enter any name including the extension they want.

idaes/surrogate/plotting/sm_plotter.py
bpaul4: The new form of the loop keeps the file open, but overwrites the figure/axis on each iteration rather than saving a list of figure objects as before. If we close the file after saving each figure, PdfPages() will overwrite the previous file and plot. There is another package (PyPDF) that can open files and add new pages, but it would require an additional dependency since it is not part of any already installed package.

idaes/surrogate/plotting/sm_plotter.py
andrewlee94: In many ways, the extension itself doesn't matter - it merely tells the OS what programme to use to open the file by default. The more important issue however is how the OS would handle a case where there are two periods in the file name. I think a lot of OS's can handle that these days (not that I try), but it used to be a problem and should probably be avoided. There is also a further issue that the user might not be aware that we append `.pdf` to the file name if it is not present, and might deliberately use the name `foo.bar` and expect the resulting file to have that name.

For sanity, I would often take the most restrictive approach and instead raise and `Exception` if the filename does not end in `.pdf`. That way there is no ambiguity.

bpaul4: I did try '[name].bar.pdf' and it opened normally on my Windows machine, however I agree with you about enforcing '.pdf' in the filename.

idaes/surrogate/plotting/sm_plotter.py
bpaul4: We could....I'll leave the default to 'True' since most users would want the plots to print to their IDE viewer.

idaes/surrogate/plotting/sm_plotter.py
bpaul4: I agree, as long as it's not an issue with the argument sharing a name with the method.

bpaul4:
@andrewlee94 I've addressed your comments, and changing how the figures are saved should resolve @carldlaird's memory issue as well. The test failure is due to the lack of patch coverage for the new code, and the project coverage test passes.

idaes/surrogate/plotting/sm_plotter.py
```diff
@@ -60,64 +69,59 @@ def scatter2D(xdata, zdata, xtest, zfit, xlabels=None, zlabels=None, dz=None,
...
         for j in range(numouts):
             zlabels.append('z' + str(j+1))
-    if clabel is None:
-        clabel = 'Confidence Intervals'
 
-    pointidx = -1  # dummy index for each plot produced
-    fig, ax = [], []
+    if filename is not None:
+        if filename[-4:] != '.pdf':  # checking if extension is present
+            raise Exception('Filename does not end in .pdf, please amend '
```
carldlaird, line 75: This is fine for now, but we should be using the "os" package to get the extension instead of relying on it being the last 4 characters.

idaes/surrogate/plotting/sm_plotter.py
carldlaird: This is exactly what I wanted. Thanks, @bpaul4

carldlaird (approved):
This looks reasonable to me.

andrewlee94 (approved):

PR #607, opened 2021-11-18
Configure Codecov to improve when report updates are created

lbianchi-lbl:
## Summary/Motivation:

This PR should take care of temporary inconsistencies in the reports that Codecov generates when a PR is updated, resulting in e.g. unrelated files being flagged as having coverage changes because of coverage data being submitted to Codecov at different times (because tests take longer to run on Windows).

### Legal Acknowledgement

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -11,3 +11,10 @@ coverage:
       default:
         target: auto
         threshold: 1%
+codecov:
+  require_ci_to_pass: false
+  notify:
+    wait_for_ci: false
+    # after_n_builds should match the number of CI jobs configured to upload reports to Codecov
+    # in IDAES there are 2 such jobs (1 Windows + 1 Linux for a single sample Python version)
+    after_n_builds: 2


lbianchi-lbl:
So it looks like there are still some files that are flagged as having coverage changes even though they're not affected by the PR:

- `idaes/ver.py`: this is a known issue (I don't remember off the top of my head why this happens but IIRC it had something to do with git)
- `idaes/surrogate/pysmo/kriging.py`: the lines associated with the coverage changes are here https://app.codecov.io/gh/IDAES/idaes-pse/compare/607/changes#D2L362
  - As I understand, it looks like there's a variation in which branch is taken (i.e. which `opt_resultsN.fun` is greater) with respect to the base branch
  - Does anyone have any insight on why this happens? More specifically, is this completely stochastic (random), or there are external circumstances that can affect the results of the methods used there?

ksbeattie (approved):
LGTM

Issue #597, opened 2021-11-15
Degeneracy Hunter SVD analysis


dallan-keylogic:
Using Degeneracy Hunter to debug a flowsheet that isn't converging, I'm running into problems with `scipy.sparse.linalg.svds` in computing the smallest singular values of the Jacobian. Apparently, `ARPACK` [has problems finding small eigenvalues](https://docs.scipy.org/doc/scipy/reference/tutorial/arpack.html) without shifting the spectrum, and `svds` makes no attempt to shift the spectrum. We probably need to create a special implementation of `svds` for our purposes here.

A minor, tangentially related issue is that `check_rank_equality_constraints` only computes the SVD once, then returns the same SVD for future calls. This would cause problems if a user, for example, wanted to check the rank at the initialization point, then check the rank again at the point IPOPT converges to,.

ksbeattie:
@dallan-keylogic, let me know if you disagree with the priority and target release.

Robbybp:
I've had success using `solver="lobpcg"` in `svds`.

adowling2:
@dallan-keylogic How large is the model? Are there specific solver options we should add in Degeneracy Hunter?

dallan-keylogic:
@adowling2 It's actually only around an 800x800 square problem. Using a normal SVD works fine. Using `solver="lobpcg"` didn't help in my case. We need to form the matrix X'X then call `eigsh` in shift-invert mode.

I'm going to create a small PR replacing `svds` with `svd`, as well as some other necessary changes (get the scaled Jacobian rather than the raw Jacobian and make sure pretty printing constraints isn't on by default so users don't get bombarded with 20 pages of thermo garbage). We can homebrew our own `svds` using the right option in the medium term.

adowling2:
@dallan-keylogic Good plan. My suggestion is to add `svd_option` as an input argument to degeneracy hunter. We can then support `dense`, `sparse`, etc.

adowling2:
@shermanjasonaf has another moderate size system (7698 x 15484) where this is causing an issue. @dallan-keylogic where does this stand? @shermanjasonaf (with help from me) may be willing to test out changes in a draft PR.

dallan-keylogic:
I spent a morning trying a few things without success, then just resigned myself to using the dense SVD. However, I finally have some time to work on this again, so hopefully I can find a good working method.

dallan-keylogic:
[...]

PR #610, opened 2021-11-24
DegeneracyHunter scaled Jacobian and SVD

dallan-keylogic:
## Fixes
-Use dense SVD rather than sparse SVD
-Use scaled Jacobian rather than raw Jacobian to perform rank analysis

## Summary/Motivation:
Mitigate #597 by switching from Scipy's unreliable (for small singular values) sparse SVD to a dense SVD. Also, ensure that a scaled Jacobian is being used to detect nearly degenerate constraints.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/model_diagnostics.py
+++ b/idaes/core/util/model_diagnostics.py
@@ -637,6 +637,3 @@ class DegeneracyHunter():
             nothing
         '''
         print(v,"\t\t",v.lb,"\t",v.value,"\t",v.ub)
-        
-def _special_svds():
-    pass
--- a/idaes/core/util/model_diagnostics.py
+++ b/idaes/core/util/model_diagnostics.py
@@ -21,11 +21,12 @@ import pyomo.environ as pyo
 from pyomo.core.expr.visitor import identify_variables
 from pyomo.contrib.pynumero.interfaces.pyomo_nlp import PyomoNLP
 import numpy as np
+from scipy.linalg  import svd
 from scipy.sparse.linalg import svds
 from scipy.sparse import issparse, find
 
 from idaes.core.util.model_statistics import large_residuals_set, variables_near_bounds_set
-
+import idaes.core.util.scaling as iscale
 from pyomo.opt import SolverStatus, TerminationCondition
 
 import matplotlib.pyplot as plt
@@ -61,11 +62,8 @@ class DegeneracyHunter():
             # setup pynumero interface
             self.nlp = PyomoNLP(self.block)
 
-            # calculate Jacobian of equality constraints in COO sparse matrix format
-            jac_eq = self.nlp.evaluate_jacobian_eq()
-
-            # save the Jacobian
-            self.jac_eq = jac_eq
+            # Get the scaled Jacobian of equality constraints
+            self.jac_eq = iscale.get_jacobian(self.block)[0]
         
             # Create a list of equality constraint names            
             self.eq_con_list = PyomoNLP.get_pyomo_equality_constraints(self.nlp)
@@ -112,7 +110,7 @@ class DegeneracyHunter():
         self.min_nonzero_nu = 1E-5
         
         
-    def check_residuals(self, tol=1e-5, print_level=2, sort=True):
+    def check_residuals(self, tol=1e-5, print_level=1, sort=True):
         """
         Method to return a ComponentSet of all Constraint components with a
         residual greater than a given threshold which appear in a model.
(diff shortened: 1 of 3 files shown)


idaes/core/util/model_diagnostics.py
```diff
@@ -508,23 +507,31 @@ def svd_analysis(self, n_smallest_sv=10):
...
             # (U or V may be smaller in economy mode)
             # Thus we really only care about U
-            u, s, v = svds(self.jac_eq, k = n_sv, which='SM')
-        
+            #u, s, v = svds(self.jac_eq, k = n_sv, which='SM')#, solver='lobpcg')
+            
+            u, s, vT = svd(self.jac_eq.todense())
+            u = u[:,-1-n_smallest_sv:]
+            s = s[-1-n_smallest_sv:]
+            vT = vT[-1-n_smallest_sv:,:]
```
Robbybp, line 530: I would prefer to have an option for sparse vs dense SVD, rather than hard-coding one or the other. I'd also prefer the default to be sparse to avoid the memory error that `jac_eq.todense()` will cause for large enough matrices.

idaes/core/util/model_diagnostics.py
```diff
@@ -628,4 +635,7 @@ def print_variable_bounds(v):
         Return:
             nothing
         '''
-        print(v,"\t\t",v.lb,"\t",v.value,"\t",v.ub)
\ No newline at end of file
+        print(v,"\t\t",v.lb,"\t",v.value,"\t",v.ub)
+        
+def _special_svds():
+    pass
```
Robbybp, line 641: Should this be removed?

idaes/core/util/scaling.py
```diff
@@ -504,6 +505,8 @@ def constraint_autoscale_large_jac(
             scaled too much.
         no_scale: just calculate the Jacobian and scaled Jacobian, don't scale
             anything
+        equality_constraints_only: Only include the equality constraints in the
+        Jacobian calculated and scaled
```
Robbybp, line 509: Minor formatting: four more spaces in second line of docstring.

idaes/core/util/scaling.py
```diff
@@ -564,11 +573,14 @@ def get_jacobian(m, scaled=True):
     Args:
         m: model to get Jacobian from
         scaled: if True return scaled Jacobian, else get unscaled
+        equality_constraints_only: Only include equality constraints in the
+        Jacobian calculated and scaled
```
Robbybp, line 577: Same formatting comment as above.

Robbybp (commented):
Looks good. Main comment is I'd like to retain `svds` as the default, and we can use `svd` with a `sparse=False` argument or something.

idaes/core/util/model_diagnostics.py
dallan-keylogic: Given my experience with `svds`, I'm rather skeptical. Do we have a use case where 1) The Jacobian is too large to reasonably factor using `svd` and 2) `svds` actually converges? Maybe some of the control problems you've been looking at---it's possible their block structure works better with `svds` than a steady-state problem for a medium sized flowsheet.

Robbybp: Not off the top of my head - I usually try to only call SVD on small matrices. That said, the dimension above which `todense` fails due to memory is relatively small on my (and I suspect many people's) workstation. The following code fails for me, for instance:
```python
import scipy.sparse as sps
iden = sps.identity(int(5e4))
iden = iden.todense()
```
`svds` is definitely less reliable than I'd like, but we should at least be able to try it for moderately large matrices, rather than running out of memory. I'd be fine with a try/except block that tries `svd` and runs `svds` only if it gets a memory error.

idaes/core/util/model_diagnostics.py
```diff
@@ -507,20 +506,29 @@ def svd_analysis(self, n_smallest_sv=10):
         '''
         
         if self.n_eq > 1:
-        
+
```
Robbybp, line 509: NFC: whitespace

idaes/core/util/tests/test_model_diagnostics.py
```diff
@@ -242,8 +242,18 @@ def test_problem2_with_degenerate_constraint():
...
-    # TODO: Add MILP solver to idaes get-extensions and add more tests
\ No newline at end of file
+    # Test DH with SVD
+    # Is there any way to test whether the sparse SVD is actually used?
+    # Trying to do an SVD of a 50000x50000 identity matrix would work, but
+    # would produce an extremely slow test failure
+    dh3 = DegeneracyHunter(m2)
+    n_rank_deficient = dh3.check_rank_equality_constraints(dense=False)
+    assert n_rank_deficient == 1
+    # TODO: Add MILP solver to idaes get-extensions and add more tests
```
Robbybp, line 259: I don't see good way to test that sparse SVD is actually used. Can you add the above assertions (e.g. `dh3.u.shape == (2,2)`) to the `dense=False` test? And maybe test the actual values that come from the SVD? Although that might be too sensitive?

Robbybp (changes requested):
Minor change request on the test. After that I'm fine with it.

andrewlee94:
@dallan-keylogic What is the status on this PR? We are starting to make changes to move towards IDAES 2.0 now, so there are going to be significant changes in the repo, which means long lasting PRs are going to run into a lot of conflicts.

dallan-keylogic:
@andrewlee94 I have a suite of degeneracy-hunting and scaling functions that I really need to create PRs for. I'm not going to have time to make those PRs until the beginning of April. At that point, I could make it a priority to merge them.

dallan-keylogic:
Subsumed this stale PR into #799

PR #619, opened 2021-12-06
Fixing some inter-dependency between component and unit tests

andrewlee94:
## Summary/Motivation:
There were a few cases where the outcome of some `component` tests relied on effects of `unit` tests. This resulted in failures of these tests if the `component` tests were run in isolation. This slipped through testing as we almost never run `component` tests in isolation (at least not as part of the CI).

### Legal Acknowledgement


jsiirola (approved):
This looks reasonable

blnicho (approved):

PR #622, opened 2021-12-09
Fixing some Pyomo warnings

andrewlee94:
## Summary/Motivation:
The most recent version of Pyomo has begun emitting warnings when the value of a `Var` is set outside its bounds. Some IDAES code was triggering these warnings, so this PR addresses the cause of those.

- When initializing mole fractions in the generic property package, ensure that the value is never set less than the lower bound.
- Correct bounds in one example code snippet where the value was being set outside the bounds.

### Legal Acknowledgement

### DIFF
--- a/idaes/generic_models/properties/core/generic/generic_property.py
+++ b/idaes/generic_models/properties/core/generic/generic_property.py
@@ -1263,9 +1263,17 @@ class _GenericStateBlock(StateBlock):
                             blk[k].true_to_appr_species[p, j])
                     # Need to calculate all flows before doing mole fractions
                     for p, j in blk[k].params.apparent_phase_component_set:
-                        calculate_variable_from_constraint(
-                            blk[k].mole_frac_phase_comp_apparent[p, j],
-                            blk[k].appr_mole_frac_constraint[p, j])
+                        x = value(
+                            blk[k].flow_mol_phase_comp_apparent[p, j] /
+                            sum(blk[k].flow_mol_phase_comp_apparent[p, jj]
+                                for jj in blk[k].params.apparent_species_set))
+                        lb = blk[k].mole_frac_phase_comp_apparent[p, j]._lb
+                        if lb is not None and x <= lb:
+                            blk[k].mole_frac_phase_comp_apparent[
+                                p, j].set_value(lb)
+                        else:
+                            blk[k].mole_frac_phase_comp_apparent[
+                                p, j].set_value(x)
                 elif blk[k].params.config.state_components == \
                         StateIndex.apparent:
                     # First calculate initial values for true species flows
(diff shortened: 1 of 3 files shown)


eslickj (dismissed):

idaes/generic_models/properties/core/generic/generic_property.py
```diff
@@ -1263,9 +1263,17 @@ def initialize(blk, state_args=None, state_vars_fixed=False,
...
-                            blk[k].mole_frac_phase_comp_apparent[p, j],
-                            blk[k].appr_mole_frac_constraint[p, j])
+                        x = value(
+                            blk[k].flow_mol_phase_comp_apparent[p, j] /
+                            sum(blk[k].flow_mol_phase_comp_apparent[p, jj]
+                                for jj in blk[k].params.apparent_species_set))
+                        lb = blk[k].mole_frac_phase_comp_apparent[p, j]._lb
+                        if lb is not None and x <= lb:
+                            blk[k].mole_frac_phase_comp_apparent[
+                                p, j].set_value(lb)
```
jsiirola, line 1273: Why are you only snapping to the lower bound?  Should you not also check and snap to the upper bound?

idaes/generic_models/properties/core/generic/generic_property.py
```diff
@@ -1263,9 +1263,17 @@ def initialize(blk, state_args=None, state_vars_fixed=False,
...
                     # Need to calculate all flows before doing mole fractions
                     for p, j in blk[k].params.apparent_phase_component_set:
-                        calculate_variable_from_constraint(
-                            blk[k].mole_frac_phase_comp_apparent[p, j],
-                            blk[k].appr_mole_frac_constraint[p, j])
+                        x = value(
+                            blk[k].flow_mol_phase_comp_apparent[p, j] /
+                            sum(blk[k].flow_mol_phase_comp_apparent[p, jj]
+                                for jj in blk[k].params.apparent_species_set))
+                        lb = blk[k].mole_frac_phase_comp_apparent[p, j]._lb
```
jsiirola, line 1270: Why are you directly accessing the `_lb` attribute?  The public `lb` attribute returns a float or `None` (the `_lb` attribute could be an expression...)

idaes/generic_models/properties/core/generic/generic_property.py
```diff
@@ -1275,9 +1283,15 @@ def initialize(blk, state_args=None, state_vars_fixed=False,
...
-                        calculate_variable_from_constraint(
-                            blk[k].mole_frac_phase_comp_true[p, j],
-                            blk[k].true_mole_frac_constraint[p, j])
+                        x = value(
+                            blk[k].flow_mol_phase_comp_true[p, j] /
+                            sum(blk[k].flow_mol_phase_comp_true[p, jj]
+                                for jj in blk[k].params.true_species_set))
+                        lb = blk[k].mole_frac_phase_comp_true[p, j]._lb
+                        if lb is not None and x <= lb:
+                            blk[k].mole_frac_phase_comp_true[p, j].set_value(lb)
```
jsiirola, line 1292: Same questions as above...

idaes/generic_models/properties/core/generic/generic_property.py
andrewlee94: The upper bound _should_ not be an issue as there is some slack there (the bound is slightly greater than 1), so there should be no way the calculation returns a value greater than 1. Given this, I think it would actually be better not to have the check - that way if something ever does occur that causes a value greater than 1, we will start seeing warnings to tell us that something is off.

jsiirola (dismissed):

eslickj (approved):

jsiirola (approved):

eslickj (approved):

PR #626, opened 2021-12-13
Fix two bugs in model serializer.

eslickj:
This fixes two bugs in the model serializer.  First @carldlaird noticed that the from_json() function was passing the component object to the filter function for attributes when reading component data objects.  It should have been passing the component data object.  I also added an option to the canned value StoreSpec object that lets you only read values of unfixed variables based on whether the model variable in fixed.

The second issue was with sets like FiniteSetOf not writing right.  This seems to come up sometimes, but not always with References attached to the model.  May also have something to do with DAE transformations not sure.  Anyway now when the serializer encounters FiniteSetOf and similar it won't fail.

I added test for both of these fixes and they seem to work.

### Legal Acknowledgement


idaes/core/util/model_serializer.py
```diff
@@ -701,8 +729,7 @@ def _read_component_data(sd, o, wts, lookup=None, suffixes=None):
     except AttributeError:
         item_keys = [None]
     for key in item_keys:
-        if key is None and isinstance(o, ComponentData) \
-            and not isinstance(o, Component):
+        if key is None and isinstance(o, ComponentData):
             el = o
         else:
             el = o[key]
```
jsiirola, line 735: Technically, you shouldn't need this test: `o[None] == o` for Scalar components.  (same comment applies to line 518)

jsiirola (approved):

idaes/core/util/model_serializer.py
eslickj: Yeah, I shouldn't need to check this.  The thing is I sort of messed this up, if a component doesn't have keys, I pretend that it does and put the data under the None index for some sort of consistency.  I newly discovered that FiniteSetOf is a component but apparently doesn't have keys.  I guess I should probably just skip if there are no keys.  If it's a FiniteSetOf or the like, I don't have any need to write it.  I guess this was just me not being able to decide how to handle this.

carldlaird (approved):

PR #627, opened 2021-12-13
Adding InitializationErrors to core UnitModels

andrewlee94:
## Summary/Motivation:
Adds core to raise InitializationErrors if solvers fail to converge during initialization to all core unit models

-
-

### Legal Acknowledgement

### DIFF
--- a/idaes/core/unit_model.py
+++ b/idaes/core/unit_model.py
@@ -14,7 +14,7 @@
 Base class for unit models
 """
 
-from pyomo.environ import Reference
+from pyomo.environ import Reference, SolverStatus, TerminationCondition
 from pyomo.network import Port
 from pyomo.common.config import ConfigValue
 
@@ -28,7 +28,8 @@ from .control_volume_base import (ControlVolumeBlockData,
 from idaes.core.util.exceptions import (BurntToast,
                                         ConfigurationError,
                                         PropertyPackageError,
-                                        BalanceTypeNotSupportedError)
+                                        BalanceTypeNotSupportedError,
+                                        InitializationError)
 from idaes.core.util.tables import create_stream_table_dataframe
 import idaes.core.util.unit_costing
 import idaes.logger as idaeslog
@@ -655,6 +656,14 @@ Must be True if dynamic = True,
         # ---------------------------------------------------------------------
         # Release Inlet state
         blk.control_volume.release_state(flags, outlvl)
+
+        if (results is not None and (
+                results.solver.termination_condition !=
+                TerminationCondition.optimal or
+                results.solver.status != SolverStatus.ok)):
+            raise InitializationError(
+                f"{blk.name} failed to initialize successfully. Please check "
+                f"the output logs for more information.")
 
         init_log.info('Initialization Complete: {}'
                       .format(idaeslog.condition(results)))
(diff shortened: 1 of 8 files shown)


eslickj (dismissed):

idaes/core/unit_model.py
```diff
@@ -656,5 +657,13 @@ def initialize(blk, state_args=None, outlvl=idaeslog.NOTSET,
         # Release Inlet state
         blk.control_volume.release_state(flags, outlvl)
 
+        if (results is not None and (
+                results.solver.termination_condition !=
+                TerminationCondition.optimal or
+                results.solver.status != SolverStatus.ok)):
```
jsiirola, line 663: Instead of repeating this everywhere, why not just:
```python
from pyomo.environ import check_optimal_termination
# ...
if not check_optimal_termination(results):
   raise InitializationError(#...
```

idaes/core/unit_model.py
```diff
@@ -656,5 +657,13 @@ def initialize(blk, state_args=None, outlvl=idaeslog.NOTSET,
         # Release Inlet state
         blk.control_volume.release_state(flags, outlvl)
 
+        if (results is not None and (
```
jsiirola, line 660: Question: under what conditions would results be `None`?

andrewlee94: Looks like this got accidentally included here by copying from property package (where there are cases where results is `None`).

idaes/core/unit_model.py
andrewlee94: Mostly because I copied code that I already had that did it this way.

andrewlee94:
@jsiirola @eslickj It is going to prove to be challenging to come up with tests that cover these new exceptions, as it means we have to construct initialization cases that fail for each model. Obviously, this is case specific so it needs a bit of effort.

Seeing as this is a safety error, do we feel it is necessary to test this or would we be OK with letting this through (even though test coverage drops slightly)? I am wondering if it is really worth the effort.

idaes/generic_models/unit_models/pressure_changer.py
```diff
@@ -778,21 +787,25 @@ def init_adiabatic(blk, state_args, outlvl, solver, optarg):
...
             res = opt.solve(blk, tee=slc.tee)
-        init_log.info_high("Initialization Step 3 {}."
+        init_log.info_high("Initialization Step 2 {}."
                            .format(idaeslog.condition(res)))
 
         # ---------------------------------------------------------------------
         # Release Inlet state
         blk.control_volume.release_state(flags, outlvl)
+
+        if (res is not None and (
```
jsiirola, line 801: Can res ever be None here?

idaes/generic_models/unit_models/pressure_changer.py
```diff
@@ -989,11 +1002,21 @@ def tmp_rule(b, t):
...
                 res = opt.solve(blk, tee=slc.tee)
-            init_log.info_high(f"Initialization Step 5 {idaeslog.condition(res)}.")
+            init_log.info_high(
+                f"Initialization Step 5 {idaeslog.condition(res)}.")
 
         # ---------------------------------------------------------------------
         # Release Inlet state
         blk.control_volume.release_state(flags, outlvl)
+
+        if (res is not None and (
```
jsiirola, line 1012: Can res ever be None here?

jsiirola (approved):
This is OK.  Some minor questions that can either be addressed here, or else in a later PR where we switch to using `check_optimal_termination()`

jsiirola (approved):

eslickj (dismissed):

eslickj (approved):

jsiirola (approved):

PR #631, opened 2021-12-16
Temporarily skip failing integration test to address #493

### LINKED ISSUE #493: get_examples() fails when latest available version is a release candidate (causes integration tests failures)
- This happens in the narrow time period between when the first release candidate for a release series and the first final (non-candidate) versions are created
- Causes failures in the integration tests (e.g. #481) that are independent on the PR's changes, and can therefore be ignored as a temporary workaround until the root cause is addressed

lbianchi-lbl:
- Move `pytest.mark.skip` from entire test module to individual failing test

### Legal Acknowledgement

### DIFF
--- a/idaes/commands/tests/test_commands.py
+++ b/idaes/commands/tests/test_commands.py
@@ -26,7 +26,6 @@ import uuid
 # third-party
 from click.testing import CliRunner
 import pytest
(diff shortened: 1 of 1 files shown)


ksbeattie (approved):
LGTM

PR #634, opened 2021-12-17
FB0D and methane iron OC properties updates

bpaul4:
## Fixes
- Add unit consistency checks to FB0D and methane iron OC test modules
- Updates to FB0D model addressing unit inconsistency issues and derived units
- Updated units and scaling in methane iron OC properties

## Summary/Motivation:
- Updating FB0D and associated files to new IDAES units standards

-
-

### Legal Acknowledgement

### DIFF
--- a/idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
+++ b/idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
@@ -51,7 +51,7 @@ from idaes.core.util.model_statistics import (
     degrees_of_freedom,
     number_unfixed_variables_in_activated_equalities)
 import idaes.logger as idaeslog
-from idaes.core.util import get_solver
+from idaes.core.util import get_solver, scaling as iscale
 
 # Some more information about this module
 __author__ = "Chinedu Okoli"
@@ -123,30 +123,30 @@ class PhysicalParameterData(PhysicalParameterBlock):
         # H_comp = H - H(298.15) = A*T + B*T^2/2 + C*T^3/3 +
         # D*T^4/4 - E/T + F - H where T = Temp (K)/1000 and H_comp = (kJ/mol)
         cp_param_dict = {
-                        ('CH4', 1): -0.7030290,
-                        ('CH4', 2): 108.4773000,
-                        ('CH4', 3): -42.5215700,
-                        ('CH4', 4): 5.8627880,
-                        ('CH4', 5): 0.6785650,
-                        ('CH4', 6): -76.8437600,
-                        ('CH4', 7): 158.7163000,
-                        ('CH4', 8): -74.8731000,
-                        ('CO2', 1): 24.9973500,
-                        ('CO2', 2): 55.1869600,
-                        ('CO2', 3): -33.6913700,
-                        ('CO2', 4): 7.9483870,
-                        ('CO2', 5): -0.1366380,
-                        ('CO2', 6): -403.6075000,
-                        ('CO2', 7): 228.2431000,
-                        ('CO2', 8): -393.5224000,
-                        ('H2O', 1): 30.0920000,
-                        ('H2O', 2): 6.8325140,
-                        ('H2O', 3): 6.7934350,
-                        ('H2O', 4): -2.5344800,
-                        ('H2O', 5): 0.0821390,
-                        ('H2O', 6): -250.8810000,
-                        ('H2O', 7): 223.3967000,
-                        ('H2O', 8): -241.8264000
+                        ('CH4', 1): -0.7030290*pyunits.kJ/pyunits.mol/pyunits.K,
+                        ('CH4', 2): 108.4773000*pyunits.kJ/pyunits.mol/pyunits.K**2,
+                        ('CH4', 3): -42.5215700*pyunits.kJ/pyunits.mol/pyunits.K**3,
+                        ('CH4', 4): 5.8627880*pyunits.kJ/pyunits.mol/pyunits.K**4,
+                        ('CH4', 5): 0.6785650*pyunits.kJ/pyunits.mol*pyunits.K,
+                        ('CH4', 6): -76.8437600*pyunits.kJ/pyunits.mol,
+                        ('CH4', 7): 158.7163000*pyunits.kJ/pyunits.mol/pyunits.K,
+                        ('CH4', 8): -74.8731000*pyunits.kJ/pyunits.mol,
+                        ('CO2', 1): 24.9973500*pyunits.kJ/pyunits.mol/pyunits.K,
+                        ('CO2', 2): 55.1869600*pyunits.kJ/pyunits.mol/pyunits.K**2,
+                        ('CO2', 3): -33.6913700*pyunits.kJ/pyunits.mol/pyunits.K**3,
+                        ('CO2', 4): 7.9483870*pyunits.kJ/pyunits.mol/pyunits.K**4,
+                        ('CO2', 5): -0.1366380*pyunits.kJ/pyunits.mol*pyunits.K,
+                        ('CO2', 6): -403.6075000*pyunits.kJ/pyunits.mol,
+                        ('CO2', 7): 228.2431000*pyunits.kJ/pyunits.mol/pyunits.K,
+                        ('CO2', 8): -393.5224000*pyunits.kJ/pyunits.mol,
+                        ('H2O', 1): 30.0920000*pyunits.kJ/pyunits.mol/pyunits.K,
+                        ('H2O', 2): 6.8325140*pyunits.kJ/pyunits.mol/pyunits.K**2,
+                        ('H2O', 3): 6.7934350*pyunits.kJ/pyunits.mol/pyunits.K**3,
+                        ('H2O', 4): -2.5344800*pyunits.kJ/pyunits.mol/pyunits.K**4,
+                        ('H2O', 5): 0.0821390*pyunits.kJ/pyunits.mol*pyunits.K,
+                        ('H2O', 6): -250.8810000*pyunits.kJ/pyunits.mol,
+                        ('H2O', 7): 223.3967000*pyunits.kJ/pyunits.mol/pyunits.K,
+                        ('H2O', 8): -241.8264000*pyunits.kJ/pyunits.mol
                         }
         self.cp_param = Param(self.component_list,
                               range(1, 10),
@@ -440,8 +440,8 @@ class GasPhaseStateBlockData(StateBlockData):
         # Sum mole fractions if not inlet block
         if self.config.defined_state is False:
             def sum_component_eqn(b):
-                return 1e2 == 1e2 * sum(b.mole_frac_comp[j]
-                                        for j in b._params.component_list)
+                return 1 == sum(b.mole_frac_comp[j]
+                                for j in b._params.component_list)
             self.sum_component_eqn = Constraint(rule=sum_component_eqn)
 
     def _mw(self):
@@ -472,8 +472,11 @@ class GasPhaseStateBlockData(StateBlockData):
                             units=pyunits.mol/pyunits.m**3)
 
         def ideal_gas(b):
-            return (b.dens_mol*b._params.gas_const*b.temperature*1e-2 ==
-                    b.pressure)
+            return (b.dens_mol * pyunits.convert(
+                b._params.gas_const,
+                to_units=pyunits.J/pyunits.mol/pyunits.K) *
+                    b.temperature ==
+                    pyunits.convert(b.pressure, to_units=pyunits.Pa))
         try:
             # Try to build constraint
             self.ideal_gas = Constraint(rule=ideal_gas)
@@ -532,20 +535,21 @@ class GasPhaseStateBlockData(StateBlockData):
 
         def visc_d_comp(i):
             return self._params.visc_d_param[i, 1] * \
-                    (self.temperature**self._params.visc_d_param[i, 2]) \
-                    / ((1 + (self._params.visc_d_param[i, 3]/self.temperature))
+                    ((self.temperature/pyunits.K) **
+                     self._params.visc_d_param[i, 2]) \
+                    / ((1 + (self._params.visc_d_param[i, 3] /
+                             (self.temperature/pyunits.K)))
                         + (self._params.visc_d_param[i, 4] /
-                           (self.temperature**2)))
+                           ((self.temperature/pyunits.K)**2))) * \
+                    pyunits.kg/pyunits.m/pyunits.s
 
         def visc_d_constraint(b):
-            return 1e6*b.visc_d == 1e6*sum(b.mole_frac_comp[i]*visc_d_comp(i)
-                                           / (sum(b.mole_frac_comp[j]
-                                                  * (b._params.mw_comp[j] /
-                                                     b._params.mw_comp[i])**0.5
-                                                  for j in
-                                                  b._params.component_list))
-                                           for i in
-                                           b._params.component_list)
+            return b.visc_d == sum(b.mole_frac_comp[i]*visc_d_comp(i)
+                                   / (sum(b.mole_frac_comp[j]
+                                          * (b._params.mw_comp[j] /
+                                             b._params.mw_comp[i])**0.5
+                                          for j in b._params.component_list))
+                                   for i in b._params.component_list)
         try:
             # Try to build constraint
             self.visc_d_constraint = Constraint(rule=visc_d_constraint)
@@ -565,12 +569,17 @@ class GasPhaseStateBlockData(StateBlockData):
                                   units=pyunits.cm**2/pyunits.s)
 
         def D_bin(i, j):
-            # 1e3 used to multiply MW to convert from kg/mol to kg/kmol
             return ((1.43e-3*(self.temperature**1.75) *
-                     ((1e3 * self._params.mw_comp[i] +
-                       1e3 * self._params.mw_comp[j])
-                     / (2 * (1e3 * self._params.mw_comp[i]) *
-                        (1e3*self._params.mw_comp[j])))**0.5)
+                     ((pyunits.convert(self._params.mw_comp[i],
+                                       to_units=pyunits.kg/pyunits.kmol) +
+                       pyunits.convert(self._params.mw_comp[j],
+                                       to_units=pyunits.kg/pyunits.kmol))
+                     / (2 *
+                        (pyunits.convert(self._params.mw_comp[i],
+                                         to_units=pyunits.kg/pyunits.kmol)) *
+                        (pyunits.convert(self._params.mw_comp[j],
+                                         to_units=pyunits.kg/pyunits.kmol))
+                        ))**0.5)
                     / ((self.pressure)
                         * ((self._params.diff_vol_param[i]**(1/3))
                             + (self._params.diff_vol_param[j]**(1/3)))**2))
@@ -599,12 +608,14 @@ class GasPhaseStateBlockData(StateBlockData):
                               units=pyunits.kJ/pyunits.m/pyunits.K/pyunits.s)
 
         def therm_cond_comp(i):
-            return self._params.therm_cond_param[i, 1] \
-                    * (self.temperature**self._params.therm_cond_param[i, 2]) \
+            return (self._params.therm_cond_param[i, 1]
+                    * ((self.temperature/pyunits.K) **
+                       self._params.therm_cond_param[i, 2])
                     / ((1 + (self._params.therm_cond_param[i, 3] /
-                             self.temperature))
+                             (self.temperature/pyunits.K)))
                         + (self._params.therm_cond_param[i, 4] /
-                       (self.temperature**2)))
+                       ((self.temperature/pyunits.K)**2)))) * \
+                pyunits.J/pyunits.m/pyunits.K/pyunits.s  # units from source
 
         def A_bin(i, j):
             return (1 + ((therm_cond_comp(j)/therm_cond_comp(i))**0.5)
@@ -614,8 +625,7 @@ class GasPhaseStateBlockData(StateBlockData):
                              self._params.mw_comp[i])))**0.5
 
         def therm_cond_constraint(b):
-            # The 1e-3 term is used as a conversion factor to a kJ basis
-            return 1e6*b.therm_cond == 1e6*(1e-3) * \
+            return b.therm_cond == pyunits.convert(
                                        sum(b.mole_frac_comp[i]
                                            * therm_cond_comp(i)
                                            / (sum(b.mole_frac_comp[j] *
@@ -623,7 +633,9 @@ class GasPhaseStateBlockData(StateBlockData):
                                                   for j in
                                                   b._params.component_list))
                                            for i in
-                                           b._params.component_list)
+                                           b._params.component_list),
+                                       to_units=pyunits.kJ/pyunits.m /
+                                       pyunits.K/pyunits.s)
         try:
             # Try to build constraint
             self.therm_cond_constraint = Constraint(rule=therm_cond_constraint)
@@ -643,12 +655,19 @@ class GasPhaseStateBlockData(StateBlockData):
                                units=pyunits.kJ/pyunits.mol/pyunits.K)
 
         def pure_component_cp_mol(b, j):
+            # for simplicity, coefficients using K/1000 to different powers
+            # are used as if in K to different powers (cp_param_dict), and
+            # temperatures are converted from K/1000 to K below. The equation
+            # gives heat capacity in J/mol-K, so we multiply by 1e-3 to get kJ
+            # since the coefficients are defined in kJ above
+            # SHOULDN'T HAVE TO ADD UNITS MANUALLY, SHOULD COME FROM CP_PARAM
             return b.cp_mol_comp[j] == 1e-3*(
                         b._params.cp_param[j, 1] +
-                        b._params.cp_param[j, 2]*(b.temperature*1e-3) +
-                        b._params.cp_param[j, 3]*(b.temperature*1e-3)**2 +
-                        b._params.cp_param[j, 4]*(b.temperature*1e-3)**3 +
-                        b._params.cp_param[j, 5]/((b.temperature*1e-3)**2))
+                        b._params.cp_param[j, 2]*(b.temperature*1e-3/pyunits.K) +
+                        b._params.cp_param[j, 3]*(b.temperature*1e-3/pyunits.K)**2 +
+                        b._params.cp_param[j, 4]*(b.temperature*1e-3/pyunits.K)**3 +
+                        b._params.cp_param[j, 5]/((b.temperature*1e-3/pyunits.K)**2)) * \
+                pyunits.kJ/pyunits.mol/pyunits.K
         try:
             # Try to build constraint
             self.cp_shomate_eqn = Constraint(self._params.component_list,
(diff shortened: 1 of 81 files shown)


bpaul4:
Tests may fail due to recent unrelated updates to the test_commands script. All gas_solid_contactors tests pass when run locally.

Robbybp (commented):
Very happy to see these changes implemented. Have some clarifying questions.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
```diff
@@ -472,8 +472,11 @@ def _dens_mol(self):
                             units=pyunits.mol/pyunits.m**3)
 
         def ideal_gas(b):
-            return (b.dens_mol*b._params.gas_const*b.temperature*1e-2 ==
-                    b.pressure)
+            return (b.dens_mol * pyunits.convert(
+                b._params.gas_const,
```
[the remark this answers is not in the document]
bpaul4: I see that the oxidation gas thermo package had this updated in #416, I can update this here as well.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
```diff
@@ -532,20 +535,21 @@ def _visc_d(self):
 
         def visc_d_comp(i):
             return self._params.visc_d_param[i, 1] * \
-                    (self.temperature**self._params.visc_d_param[i, 2]) \
-                    / ((1 + (self._params.visc_d_param[i, 3]/self.temperature))
+                    ((self.temperature/pyunits.K) **
+                     self._params.visc_d_param[i, 2]) \
+                    / ((1 + (self._params.visc_d_param[i, 3] /
+                             (self.temperature/pyunits.K)))
```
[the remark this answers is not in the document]
bpaul4: Yes, the 1/Kelvin units are to address unit inconsistency issues.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/solid_phase_thermo.py
```diff
@@ -585,3 +596,10 @@ def default_material_balance_type(blk):
 
     def default_energy_balance_type(blk):
         return EnergyBalanceType.enthalpyTotal
+
+    def calculate_scaling_factors(self):
+        super().calculate_scaling_factors()
+
+        if hasattr(self, "sum_component_eqn"):
+            for t, v in self.sum_component_eqn.items():
+                iscale.set_scaling_factor(v, 1e2)
```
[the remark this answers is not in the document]
bpaul4: Some of the component mole fractions may become very small, either through inlet specifications (see the MB flowsheet example where the inlet water fraction is 1E-5) or if methane approaches full conversion by the reactor outlet. The scaling helps address convergence issues; @chineduobiora may be able to expand upon the reasoning behind the scaling factor.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -217,13 +217,16 @@ def solids_material_holdup_constraints(b, t, j):
                     self.solids_material_holdup,
                     wrt=self.flowsheet().config.time,
                     doc="Solids material accumulation",
-                    units=units_meta_solid('mass')/units_meta_solid('time'))
+                    units=units_meta_solid('mass'))
+        # should be mass/time, to get around DAE limitations var is
+        # divided by time units when used
```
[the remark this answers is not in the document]
bpaul4: Pyomo does not currently support units for ContinuousSet objects, and @andrewlee94 has opened an issue requesting this feature. Since we cannot specify mass/time units here but the numerical calculations are still correct, the accumulation variables are manually corrected for units in the accumulation constraints.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/solid_phase_thermo.py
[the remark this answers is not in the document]
andrewlee94: Assuming this is a sum of mole fractions, you should try to get the currently defined scaling factor first, with a fall back to a default if one has not been defined.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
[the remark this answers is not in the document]
andrewlee94: I think you need to be a bit more careful here. The first reference to temperature needs to have the units removed as it is raised to a variable, non-integer power. However, in the later terms the parameters should have units which cancel with T - e.g. parameter 3 should have units of `1/K` which will cancel with those of `temperature` as required. Similar for parameter 4 (units should be 1/K**2).

Note that this required you to create separate `Vars` for each parameter (rather than a single indexed parameter) as all elements of an indexed `Var` must have the same units. However, this is what we are doing everywhere else in IDAES, and is probably the better way to do this.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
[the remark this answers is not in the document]
andrewlee94: @bpaul4 I am not sure that this is correct. In the 1D control volume, we assign the correct units to the `DerivativeVar` (see line 424 in CV1D). I am not sure how we made this work, but it does.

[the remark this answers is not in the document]
bpaul4: It appears line 268 in the 0D control volume does something similar, although Pyomo threw an InconsistentUnits error when I attempted to set expected units on the derivative variable itself. An internal constraint defined the discretization as accumulation = Δ(mass)/Δ(time) with units [m/s] = [m]/[dimensionless since ContinuousSets are not passed with units].

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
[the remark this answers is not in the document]
bpaul4: I'm reviewing property packages with Parameters for each coefficient, indexed by component, for proper syntax. Would you recommend modifying the gas/solid Cp and gas thermo_cond parameters in this way as well?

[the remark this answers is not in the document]
andrewlee94: It needs to happen at some point. However, if the PR starts getting too big you might want to try breaking it up into stages.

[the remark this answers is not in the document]
bpaul4: This PR is focusing only on the FixedBed0D model and related updates  to the methane iron OC properties. There will be other PRs for the MovingBed and BubblingFluidizedBed models, and I could leave the Param updates until then to simplify this PR.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/solid_phase_thermo.py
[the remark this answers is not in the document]
Robbybp: I have no real problem with this scaling factor, but my intuition is that adding a scaling factor of 1e2 to a constraint actually makes it _harder_ to converge. I.e., we need to converge our mole fraction variables to a tighter tolerance to satisfy this constraint to an absolute tolerance of 1e-8. This factor also increases the constraint's Jacobian entries from 1 to 100, which we may or may not want? 

Again no real problem here, I'm just being difficult because I'm trying to arrive at a more systematic understanding of scaling myself and am always curious about hard-coded scaling factors.

idaes/gas_solid_contactors/unit_models/fixed_bed_0D.py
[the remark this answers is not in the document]
Robbybp: Okay, I think I see. The problem is that the "Delta t" constants in the time discretization equation don't have units. It's not a problem for the space discretization equations because our length domains are normalized (dimensionless).

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
```diff
@@ -123,30 +123,30 @@ def build(self):
...
+                        ('CO2', 2): 55.1869600*pyunits.kJ/pyunits.mol/pyunits.K**2,
+                        ('CO2', 3): -33.6913700*pyunits.kJ/pyunits.mol/pyunits.K**3,
+                        ('CO2', 4): 7.9483870*pyunits.kJ/pyunits.mol/pyunits.K**4,
+                        ('CO2', 5): -0.1366380*pyunits.kJ/pyunits.mol*pyunits.K,
+                        ('CO2', 6): -403.6075000*pyunits.kJ/pyunits.mol,
+                        ('CO2', 7): 228.2431000*pyunits.kJ/pyunits.mol/pyunits.K,
+                        ('CO2', 8): -393.5224000*pyunits.kJ/pyunits.mol,
+                        ('H2O', 1): 30.0920000*pyunits.kJ/pyunits.mol/pyunits.K,
+                        ('H2O', 2): 6.8325140*pyunits.kJ/pyunits.mol/pyunits.K**2,
+                        ('H2O', 3): 6.7934350*pyunits.kJ/pyunits.mol/pyunits.K**3,
```
chineduobiora: The example gas proppack we went through during our last meeting has an updated code for the cp param. If you get a chance, please take a look at it? I think it is preferable to the setup shown here as the parameters are grouped by units. The same idea is applicable to the cp parameters in the solids propack, and to the parameters for viscosity, thermal conductivity, and diffusivity in the gas proppack.

bpaul4: @chineduobiora the most updated version from commit "Replacing dimensionless parameter dictionary with dimensional parameters" has the updated setup - the coefficients are extracted to 8 parameters, each indexed by the component list.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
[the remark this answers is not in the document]
chineduobiora: I think that's the approach to take. @bpaul4 the gas proppack file we discussed at our last meeting has the cp param, and enthalpy and cp_mol variables setup in this manner. Please take a look when you get the opportunity.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
```diff
@@ -162,11 +206,34 @@ def build(self):
...
-                                  mutable=True,
-                                  initialize=visc_d_param_dict,
-                                  doc="Dynamic viscosity constants")
+        self.visc_d_param_1 = Param(self.component_list,
+                                    mutable=True,
+                                    initialize={k: v for (k, j), v in
+                                                visc_d_param_dict.items()
+                                                if j == 1},
+                                    doc="Dynamic viscosity constants",
+                                    units=pyunits.kg/pyunits.m/pyunits.s)
```
chineduobiora, line 215: This unit is kg * m**-1 * s**-1 * K**-visc_d_param_2

chineduobiora: The equation is in this form: k=(a*(T^b))/(1+c/T +d/(T^2 ))

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
```diff
@@ -176,11 +243,34 @@ def build(self):
...
-                                      mutable=True,
-                                      initialize=therm_cond_param_dict,
-                                      doc="Thermal conductivity constants")
+        self.therm_cond_param_1 = Param(self.component_list,
+                                        mutable=True,
+                                        initialize={k: v for (k, j), v in
+                                                    therm_cond_param_dict.items()
+                                                    if j == 1},
+                                        doc="Dynamic viscosity constants",
+                                        units=pyunits.J/pyunits.m/pyunits.K/pyunits.s)
```
chineduobiora, line 252: This unit is J * m**-1 * s**-1 * K**-[1 + therm_cond_param_2]

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
bpaul4: Yes, this form is used for thermal conductivity and viscosity. If the units of one of those variables is [x], then the parameters should have the units {a: [x], b:[], c: [K], d: [K^2]}, correct? In the expressions the first T is divided by [K] before the exponent is applied.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
```diff
@@ -531,21 +624,21 @@ def _visc_d(self):
                           units=pyunits.kg/pyunits.m/pyunits.s)
 
         def visc_d_comp(i):
-            return self._params.visc_d_param[i, 1] * \
-                    (self.temperature**self._params.visc_d_param[i, 2]) \
-                    / ((1 + (self._params.visc_d_param[i, 3]/self.temperature))
-                        + (self._params.visc_d_param[i, 4] /
-                           (self.temperature**2)))
+            return (self._params.visc_d_param_1[i] *
+                    ((self.temperature/pyunits.K) **  # need /K to cancel units
```
chineduobiora, line 628: Division by pyunits.K will no longer be necessary when the units for visc_d_param_1 are updated per prior comment.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
```diff
@@ -595,16 +693,17 @@ def _therm_cond(self):
...
-                              units=pyunits.kJ/pyunits.m/pyunits.K/pyunits.s)
+                              doc="Thermal conductivity of gas [J/m.K.s]",
+                              units=pyunits.J/pyunits.m/pyunits.K/pyunits.s)
 
         def therm_cond_comp(i):
-            return self._params.therm_cond_param[i, 1] \
-                    * (self.temperature**self._params.therm_cond_param[i, 2]) \
-                    / ((1 + (self._params.therm_cond_param[i, 3] /
+            return (self._params.therm_cond_param_1[i] *
+                    ((self.temperature/pyunits.K) **  # need /K to cancel units
```
chineduobiora, line 701: Division by pyunits.K will no longer be necessary when the units for therm_cond_param_1 are updated per prior comment.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/solid_phase_thermo.py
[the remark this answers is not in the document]
chineduobiora: 1e2 was originally hardcoded here because my experience is to have the mole fraction variables scaled by at least a factor of 100. For the generic application I would lean towards @andrewlee94's suggestion to scale this based on a default scaling factor (using the default scaling for the mole fraction variables is a good choice). However, to @Robbybp's point the "right" scaling factor also tends to be problem specific, and a user should be able to adjust scaling factors from the flowsheet level as needed.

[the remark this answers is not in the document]
chineduobiora: @bpaul4 the gas proppack we discussed at our last meeting has examples of how default scaling can be set for individual variables (see the end of the PhysicalParameterData Class), and how they can be applied to scale constraints.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
```diff
@@ -789,3 +889,18 @@ def default_material_balance_type(blk):
...
         return EnergyBalanceType.enthalpyTotal
+
+    def calculate_scaling_factors(self):
+        super().calculate_scaling_factors()
+
+        if hasattr(self, "sum_component_eqn"):
+            for t, v in self.sum_component_eqn.items():
+                iscale.set_scaling_factor(v, 1e2)
+
+        if hasattr(self._k_rxn, "visc_d_constraint"):
```
chineduobiora, line 900: What does _k_rxn represent?

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
```diff
@@ -789,3 +889,18 @@ def default_material_balance_type(blk):
...
+            for t, v in self.visc_d_constraint.items():
+                iscale.set_scaling_factor(v, 1e6)
+
+        if hasattr(self._k_rxn, "therm_cond_constraint"):
```
chineduobiora, line 904: What does _k_rxn represent?

bpaul4: That is a typing error, I will remove the extra attributes on those lines.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
chineduobiora: "In the expressions the first T is divided by [K] before the exponent is applied." That works, but it might get confusing for someone who compares the code to the actual equation.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
```diff
@@ -789,3 +933,21 @@ def default_material_balance_type(blk):
...
+
+    def calculate_scaling_factors(self):
+        super().calculate_scaling_factors()
+
+        if self.is_property_constructed("sum_component_eqn"):
+            for t, v in self.sum_component_eqn.items():
+                iscale.constraint_scaling_transform(
+                    v,
+                    iscale.get_scaling_factor(self.mole_frac_comp[t]),
+                    overwrite=False)
```
chineduobiora, line 945: This is only one equation (`1 = sum(x_i)`), so the `for loop` isn't required. The component index should also be specified i.e. choose one of the components from the list.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
```diff
@@ -789,3 +933,21 @@ def default_material_balance_type(blk):
...
+
+        if hasattr(self, "visc_d_constraint"):
+            for t, v in self.visc_d_constraint.items():
+                iscale.set_scaling_factor(v, 1e6)
+
+        if hasattr(self, "therm_cond_constraint"):
+            for t, v in self.therm_cond_constraint.items():
+                iscale.set_scaling_factor(v, 1e6)
```
chineduobiora, line 953: Update these in the same way as the scaling for the sum_component_eqn constraint i.e. define default scaling for visc_d and therm_cond like you did for mole_frac_comp, and then use them to set scaling factors for the constraints.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/hetero_reactions.py
```diff
@@ -166,7 +167,9 @@ def build(self):
                           domain=Reals,
                           initialize=8e-4,
                           doc='Pre-exponential factor'
-                          '[mol^(1-N_reaction)m^(3*N_reaction -2)/s]')
+                          '[mol^(1-N_reaction)m^(3*N_reaction -2)/s]',
```
chineduobiora, line 170: For consistency purposes, please change all `N_reaction` occurences to `rxn_order`.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/hetero_reactions.py
```diff
@@ -398,7 +403,7 @@ def _OC_conv(self):
                            doc='Fraction of metal oxide converted')
 
         def OC_conv_eqn(b):
-            return 1e6 * b.OC_conv * \
+            return b.OC_conv * \
```
chineduobiora, line 406: `OC_conv` and `OC_conv_temp` have dimensionless units.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/solid_phase_thermo.py
```diff
@@ -48,7 +48,7 @@
     degrees_of_freedom,
     number_unfixed_variables_in_activated_equalities)
 import idaes.logger as idaeslog
-from idaes.core.util import get_solver
+from idaes.core.util import get_solver, scaling as iscale
```
chineduobiora, line 52: The suggested changes to the gas property package also apply to the solid property package.

chineduobiora (commented):
Happy New Year @bpaul4. The changes so far look good and I think we are close to completion of this PR, but I have included a couple more minor comments to be addressed.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
bpaul4: @chineduobiora Would you recommend updating the reaction module scaling in this manner as well?

chineduobiora: Yes, that will be good. However, unlike the physical properties blocks you can't set scaling factors for variables using the set_default_scaling method in the reaction properties block (it doesn't seem to be supported). So instead, I suggest using the set_scaling_factor method to set the scaling factors for variables under the calculate_scaling_factors method, and then you can set the scaling for the related constraints as before. See the example below:

```
    def calculate_scaling_factors(self):
        super().calculate_scaling_factors()

        def _set_default_factor(v, s):
            for i in v:
                if iscale.get_scaling_factor(v[i]) is None:
                    iscale.set_scaling_factor(v[i], s)
        _set_default_factor(self.reaction_rate, 1)
        _set_default_factor(self.rate_constant, 1e2)
```

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/solid_phase_thermo.py
```diff
@@ -666,8 +669,7 @@ def calculate_scaling_factors(self):
...
 
         if self.is_property_constructed("sum_component_eqn"):
-            for t, v in self.sum_component_eqn.items():
-                iscale.constraint_scaling_transform(
-                    v,
-                    iscale.get_scaling_factor(self.mass_frac_comp[t]),
-                    overwrite=False)
+            iscale.constraint_scaling_transform(
+                self.sum_component_eqn,
+                iscale.get_scaling_factor(self.mole_frac_comp['CH4']),
```
chineduobiora, line 674: Select a component from the solids component list. CH4 is only in the gas property package.

chineduobiora (dismissed):
Looks good to me.

andrewlee94 (approved):
I don't see anything wrong, and I'll trust @chineduobiora for the details.

andrewlee94:
There are some issues that showed up in pyling that need to be fixed - in lines 968 and 973 of the gas phase thermo properties you are missing an argument to  `iscale.constraint_scaling_transform` (the constraint to be transformed). This also suggests that you are not testing the scaling methods completely - you should have a test the runs the scaling method with all properties present and make sure that the correct scaling factor is set for each.

The other failing test is a known issue separate from this PR. It will be fixed as part of #647.

chineduobiora:
> There are some issues that showed up in pyling that need to be fixed - in lines 968 and 973 of the gas phase thermo properties you are missing an argument to `iscale.constraint_scaling_transform` (the constraint to be transformed). This also suggests that you are not testing the scaling methods completely - you should have a test the runs the scaling method with all properties present and make sure that the correct scaling factor is set for each.
> 
> The other failing test is a known issue separate from this PR. It will be fixed as part of #647.

Mistake was from me. I've made the update. Yes, we probably should have a test for the scaling method.

chineduobiora (approved):

andrewlee94 (approved):
Looks good.

Robbybp:
I have a fair amount of research code that depends on these property packages, and some of it is failing when I use this branch, so I'd like to figure that out before this PR is merged. Most of it is probably that I need to adjust ranges on some of my tests, so hopefully it won't take too long to sort through.

Robbybp:
Okay, most of the errors I encountered I could fix by changing a tolerance or something, but I have one dynamic optimization problem that solves with the old version of the property packages (with hard-coded scaling factors on sum_component_eqn, OC_conv_eqn, etc. constraints), but not the new version (without the hard-coded scaling factors). I'm not sure this is a compelling reason to re-introduce the factors, however, unless others report problems as well. Hopefully this is not too difficult for me to work around. Another option is that we could manually scale constraints with a unit conversion rather than hard-coded constants. Let me know your thoughts:
1. Has anybody done an convergence reliability studies on the old/new forms of the model?
2. Should we keep the factors on these constraints so as to not potentially break users' optimization problems?
3. If yes, should we replace the constant factors with unit conversions?

chineduobiora:
@ My inclination is to minimize the number of variable and constraint scaling factors that are introduced to the model. I think it is best that a user updates the scaling of the variables and constraints as they deem fit, because the appropriate scale factors will vary by application (even for the same proppack). At the flowsheet level you can also update the scaling factors of individual constraints and variables to factors of your choice. A user can also update a local copy of their proppack with the scale factor of interest for a variable or constraint.

chineduobiora:
@Robbybp, for the other questions, see below:

1. Has anybody done any convergence reliability studies on the old/new forms of the model?
     To the best of my knowledge we haven’t done this. We will consider this when updating the MB and BFB to have units.
2. Should we keep the factors on these constraints so as to not potentially break users' optimization problems?
     I’ll prefer not to keep the factors on the constraints. Users can update the factors in their local copies of the proppack, or at the flowsheet level as their application dicatates. The scaling module supports two methods (`extreme_jacobian_entries(m)`, and `badly_scaled_var_generator(m)`) that can allow a user investigate their model and update scaling factors as needed. If you aren’t already doing this, you’ll need to apply the scaling transform using `calculate_scaling_factors(m)` at the flowsheet level.
3. If yes, should we replace the constant factors with unit conversions?
    I recommend updating the constant factors with unit conversions (using the scaling tool).

andrewlee94:
@Robbybp @chineduobiora Are we ready to merge this PR (assuming the tests pass)?

chineduobiora:
@andrewlee94 I am happy for it to be merged. @Robbybp what are your thoughts? Are my responses to your comments on convergence reliability studies, and scaling sufficient?

Robbybp:
@chineduobiora @andrewlee94 I'd prefer to not alter default algebraic form of models unless we have a good reason for doing so. A good reason might be "make dens_mass a variable rather than an expression so that it can be fixed, as it is a common degree of freedom." Another good reason would be "change a hard-coded factor on a constraint to make convergence more reliable for a suite of cases we are interested in." I'm not sure "make constraint expressions simpler" is a good enough reason to change the form of a model, given that doing this may break users' optimization problems.

Here I am being somewhat selfish -- I have an NMPC problem that I'm solving for a presentation that breaks when we change these constraint factors. I'm sure I can find a way to work around this change, but I imagine it would be a much more frustrating experience for a user if they update versions of IDAES and all of a sudden one of their optimization problems stops converging.

chineduobiora:
@Robbybp The default algebraic forms of the models we currently have are not being changed in these PRs. The only changes being made are to add units (including using pyconvert to ensure that units are consistent in equations), and to remove hard-coded factors from constraints. The variable and constraint scaling transformations are optional and only work on the model when called. Users can update the scaling transformations as they deem fit for their particular problem (i.e. scale a particular constraint as needed to improve convergence). Also, we haven't applied these changes to the MB and BFB yet so there might be some inconsistencies at the moment if you using the proppacks in this PR with those models. The MB is next on @bpaul4 list, before the BFB.

Does this address your comment? If not we can discuss further at the CLC meeting on Thursday.

Robbybp:
> and to remove hard-coded factors from constraints

This is what I mean by "changing the algebraic form." This can affect convergence, so I'd rather only do this if we have a good enough reason. Making the expressions simpler may be a good enough reason, but I'd like to bring this up at a core dev meeting to see what others think. On one hand, it seems very restrictive to say "we will never change a constraint expression ever," but on the other, it would be nice if model-solver performance/convergence changed as little as possible (ideally not at all) from one IDAES version to the next.

PR #635, opened 2021-12-17
Address test failures from #633

### LINKED ISSUE #633: Test failures in test_commands::test_conf_set after #628
## Symptoms

- As part of #628, a module-level `pytest.mark.skip` marker was added to `idaes/commands/tests/test_commands.py` because of the known issue with the intermittently failing integration tests described in #493
- However, this seems to have shadowed a real test failure happening in `test_conf_set` of the same file, that emerged once #631 converted the `skip` marker from module-level to targeting the known problematic integration test

## Diagnosis

- The test failure is hard to debug because it causes the pytest process to exit abruptly, without running other tests or displaying any post-test-failure information
- Additionally, the test passes if the `-s` flag is used, which disables the default pytest behavior of capturing stdout/err, instead displaying it to the terminal
- After some research, one possible cause seems to be a Tensorflow bug (see e.g. https://github.com/tensorflow/tensorflow/issues/29842#issuecomment-506841955): a side-effect of `import tensorflow` causes conflict with the root logger, which in turn is presumably what's causing the test failure in our test suite

## Remedies

- Short-term: disabling the sets of arguments (2 out of 3) that cause `test_conf_set` to fail
- After the release:
  - Investigate the issue to validate the hypothesized cause described above
  - If confirmed, figure out a way to work around this Tensorflow long-standing issue

lbianchi-lbl:
-
-

### Legal Acknowledgement

### DIFF
--- a/idaes/commands/tests/test_commands.py
+++ b/idaes/commands/tests/test_commands.py
@@ -663,8 +663,9 @@ def test_conf_set(runner):
         if os.path.exists(fname):
             os.remove(fname)
 
-    _tst(["--global", "--file", fname, "--file_as_global"])
-    _tst(["--local", "--file", fname, "--file_as_local"])
+    # lbianchi-lbl: commenting these line out temporarily because of #633
+    # _tst(["--global", "--file", fname, "--file_as_global"])
+    # _tst(["--local", "--file", fname, "--file_as_local"])
     _tst(["--file", fname])
 
 
--- a/idaes/commands/tests/test_commands.py
+++ b/idaes/commands/tests/test_commands.py
@@ -20,6 +20,7 @@ import os
 from pathlib import Path
 from shutil import rmtree
 import subprocess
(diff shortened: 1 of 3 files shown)


lbianchi-lbl:
Reviewers: feel free to take a look and leave a review. Since this might result in GitHub invalidating previous checks, I might go ahead and scary-red-button-merge this since all checks are currently passing: 
image

jsiirola (approved):
This looks reasonable.  Does it make any sense to move other things from requirements-dev.txt into appropriage targets in `extras_require` as well?

ksbeattie (approved):

lbianchi-lbl:
> This looks reasonable. Does it make any sense to move other things from requirements-dev.txt into appropriage targets in `extras_require` as well?

Eventually (i.e. after the release), yes, as part of the planned general dependency cleanup.

PR #636, opened 2021-12-17
Making tensorflow optional in test_keras_surrogate

carldlaird:
## Fixes
- using pyomo attempt_import to make 'tensorflow.keras' an optional import and skipping all of test_keras_surrogate file to skip tests if tensorflow.keras is not available
- removing tensorflow from the list of required dependencies

-
-

### Legal Acknowledgement

### DIFF
--- a/idaes/surrogate/keras_surrogate.py
+++ b/idaes/surrogate/keras_surrogate.py
@@ -16,7 +16,13 @@ Interface for importing Keras models into IDAES
 import numpy as np
 import json
 import os.path
-import tensorflow.keras as keras
+
+try:
+    import tensorflow.keras as keras
+    keras_available = True
+except:
+    keras_available = False
+    
 from enum import Enum
 import pandas as pd
 from idaes.surrogate.base.surrogate_base import SurrogateBase
--- a/idaes/surrogate/keras_surrogate.py
+++ b/idaes/surrogate/keras_surrogate.py
@@ -17,12 +17,9 @@ import numpy as np
 import json
 import os.path
 
-try:
-    import tensorflow.keras as keras
-    keras_available = True
-except:
-    keras_available = False
-    
+from pyomo.common.dependencies import attempt_import
+keras, keras_available = attempt_import('tensorflow.keras')
+
 from enum import Enum
 import pandas as pd
 from idaes.surrogate.base.surrogate_base import SurrogateBase
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,6 @@ kwargs = dict(
         "numpy",
         "networkx",
         "omlt==0.3", # fix the version for now as package evolves
-        "tensorflow",
         "pandas",
         "pint",
         "psutil",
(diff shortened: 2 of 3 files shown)


idaes/surrogate/keras_surrogate.py
```diff
@@ -16,7 +16,13 @@
 import numpy as np
 import json
 import os.path
-import tensorflow.keras as keras
+
+try:
+    import tensorflow.keras as keras
+    keras_available = True
+except:
+    keras_available = False
```
jsiirola, line 24: You aren't testing that keras is available before using it, which means that a user that does not have tensorflow installed will just get a `NameError: name `keras` is not defined`.  I would recommend either adding a test for keras_available (and an appropriate exception), or just use Pyomo's `attempt_import()` which will generate the exception automatically when `keras` is used:
```
from pyomo.common.dependencies import attempt_import
keras, keras_available = attempt_import('tensorflow.keras`)
```

lbianchi-lbl (approved):
There was a failing pylint check, but using `attempt_import()` might have taken care of that.

setup.py
```diff
@@ -65,7 +65,6 @@ def rglob(path, glob):
         "numpy",
         "networkx",
         "omlt==0.3", # fix the version for now as package evolves
-        "tensorflow",
```
jsiirola, line 68: @lbianchi-lbl: I am all for culling required dependencies, but as we do that, should we start an "optional" target with something like:
```
extras_require = {
    'optional': [
        'tensorflow',  # idaes.surrogates.keras_surrogate
    ]
}
```

lbianchi-lbl: Yeah, I think it's a good idea. I thought we could deal with the optional dependencies all at once, but this looks like a good place to start.

idaes/surrogate/keras_surrogate.py
carldlaird: I made this change.

setup.py
lbianchi-lbl: @carldlaird @jsiirola I can do this as part of #633. I propose we get this merged in as soon as all the tests pass.

idaes/surrogate/keras_surrogate.py
```diff
@@ -16,7 +16,10 @@
 import numpy as np
 import json
 import os.path
-import tensorflow.keras as keras
+
+from pyomo.common.dependencies import attempt_import
+keras, keras_available = attempt_import('tensorflow.keras')
```
jsiirola, line 21: One more quick change: by default, the error message emitted by `attempt_import` is:
```
The tensorflow.keras module (an optional Pyomo dependency) failed to import: ModuleNotFoundError: No module named 'tensorflow'
```
If we don't want this to blame Pyomo, then we can specify our own message:
```
keras, keras_available = attempt_import(
    'tensorflow.keras', 'The tensorflow.keras module (an optional IDAES dependency) failed to import'
)

jsiirola (approved):

PR #639, opened 2021-12-21
Links connect to unit models horizontally with a gap

elbashandy:
## Fixes
Adds the multi connections with the gap I/O horizontal vertices that the router links have to go through per unit model. 

## Summary/Motivation:
This way the flowsheet will look cleaner and would make more sense

### Legal Acknowledgement


dangunter:
@elbashandy ping me when you've added tests to address codecov

dangunter (approved):
Looks good. Let's get this puppy merged! @ksbeattie your turn

ksbeattie (approved):
This is an improvement, but I did notice that they start out not horizontal and if you move them they correct, but you can get them to be non-horizontal again if you move the units around in just the "wrong" way.

Also, hitting refresh completely messes it all up.  But I think that was already the case.

So, I'm approving this as an improvement, but there's still things to improve on here.

PR #641, opened 2021-12-22
Clean up of Packed Column Models

### LINKED ISSUE #244: MEA model TODO list

andrewlee94:
## Summary/Motivation:
This PR cleans up the current solvent column models. The biggest change is dividing  the existing model into two parts - a simple generic column model and a more rigorous version for MEA capture.

Additionally, a number of things have been cleaned up such as units of measurement, scaling and removing some assumptions/requirements on the property package.

- Remove old column model in `power_generation/carbon_capture`
- Add a new `mea_solvent_column` model in `power_generation/carbon_capture` which derived from the more general model in `generic_models/unit_models/column_models`
- Move most heat and mass transfer correlations from the generic model to the new `mea_solvent_column`
- Standardized handling of units of measurement and addressed some lingering unit consistency issues
- Removed requirement for liquids properties to define solvent and solute components
- Added additional tests to verify values of key variable for all test cases

### Legal Acknowledgement


andrewlee94:
@anujad95 @aostace01 Would you be able to test this PR with the MEA flowsheet/UQ example? I do not have access to those models so I wasn't able to test it against it. Note that you will need to to switch to using the new `MEAColumn` model in `power_plant/carbon_capture/mea_solvent_systems/unit_models/mea_solvent_column.py` instead of the one in `generic_models/unit_models` (I broke the model into two parts).

Additionally, a few things have changed in this PR. A number of config arguments were removed, and `vapor_side` became `vapor_phase` for consistency (same for the liquid). You will also see some small changes in the model results due to me trying to standardize some things - those changes were fairly minor for the test cases I can't guarantee that. You can hopefully see how to use the new version via the tests.

aostace01:
I do not have access to those models either. @anujad95, can you please point me to them/share them with me?

ksbeattie:
@anujad95, @aostace01, @MAZamarripa is there any progress on this?

andrewlee94:
Closing this as things have moved to far since it was created.

Issue #644, opened 2022-01-03
Reoccurring Pytest Failure on Ubuntu Python 3.8/3.9


bpaul4:
Recently the test called by idaes/surrogate/pysmo/tests/test_utils.py::TestNumpyEvaluator::test_eval_numpy has been failing due to a seemingly minor precision error (see GitHub Actions pytest output below). The fix may be simply updating the called Numpy version, or related to Ubuntu-specific processes as this error does not occur for Windows either locally or through GitHub Actions.

Ubuntu w/ Python 3.8
```
______________________ TestNumpyEvaluator.test_eval_numpy ______________________

self = <idaes.surrogate.pysmo.tests.test_utils.TestNumpyEvaluator object at 0x7f335d3e28b0>

    @pytest.mark.unit
    def test_eval_numpy(self):
        m = ConcreteModel()
        m.p = Param([1,2], mutable=True)
        m.x = Var()
    
        data = np.array([[0,-1,2],
                         [.1,.2,.3],
                         [4,5,6]])
        cMap = ComponentMap()
        cMap[m.p[1]] = data[0]
        cMap[m.p[2]] = data[1]
        cMap[m.x] = data[2]
    
        npe = NumpyEvaluator(cMap)
    
        result = npe.walk_expression(sin(m.x))
/home/runner/work/idaes-pse/idaes-pse/idaes/apps/matopt/opt/pyomo_modeling.py:1355: SyntaxWarning: "is not" with a literal. Did you mean "!="?
>       assert result[0] == sin(4)
  if PosZic is not 0:
E       assert -0.7568024953079284 == -0.7568024953079282
/home/runner/work/idaes-pse/idaes-pse/idaes/apps/matopt/opt/pyomo_modeling.py:1357: SyntaxWarning: "is not" with a literal. Did you mean "!="?
E        +  where -0.7568024953079282 = sin(4)
  if NegZic is not 0:

/home/runner/work/idaes-pse/idaes-pse/idaes/apps/matopt/opt/pyomo_modeling.py:1471: SyntaxWarning: "is not" with a literal. Did you mean "!="?
/home/runner/work/idaes-pse/idaes-pse/idaes/surrogate/pysmo/tests/test_utils.py:55: AssertionError
  if PosZic is not 0:

/home/runner/work/idaes-pse/idaes-pse/idaes/apps/matopt/opt/pyomo_modeling.py:1473: SyntaxWarning: "is not" with a literal. Did you mean "!="?
[...]

PR #647, opened 2022-01-05
Fixing bug in generic properties initialization

andrewlee94:
## Summary/Motivation:
There was a bug introduced in a recent PR that caused KeyErrors during initialization of generic property models due to sparse phase-component sets.

- Add checks for sparse phase-component sets in the new code.
- Add a check for potential division-by-zero errors in cases with spare phase-component sets.

### Legal Acknowledgement


lbianchi-lbl (dismissed):
I don't know enough to comment on the code itself but I've tested this PR's branch in the context of watertap-org/watertap#287 and the tests that were failing are now passing.

There is a failing check in the IDAES test suite that I think is likely unrelated to these changes (see #644). Not sure how we want to go about addressing that, but using `pytest.approx()` should fix the error if we don't plan to investigate that issue more thoroughly.

lbianchi-lbl (approved):
Looks good!

eslickj (approved):

PR #650, opened 2022-01-07
Update to the derived_units method in property_meta.

chineduobiora:
## Summary/Motivation:
This update allows non-base default units such as energy, pressure defined in property packages to override unit declarations in the _create_derived_units method.

-
-

### Legal Acknowledgement

### DIFF
--- a/idaes/core/property_meta.py
+++ b/idaes/core/property_meta.py
@@ -153,6 +153,8 @@ class PropertyClassMetadata(object):
         # which has not defined units.
         if self._derived_units is None:
             self._create_derived_units()
+            # Update derived_units dict with default_units dict
+            self._derived_units.update(self.default_units)
         return self._derived_units
 
     @property


andrewlee94:
This is going to need a bit of thought and testing before we merge it, especially to think of potential edge cases. I already have one case where this could/would be problematic, which is the energy requirement for a pump: `E = deltaP*Q`. The equation assumes E, P and Q are also defined using consistent units, so if we allow for users to override specific quantities then we would need to have a unit conversion here (which is the whole reason we went with the restrictive set of units).

Robbybp:
I'm in favor of a change along these lines. By coincidence, I started prototyping something like this recently as well. See my pull request into this branch, https://github.com/chineduobiora/idaes-pse/pull/1, if you are curious.

Along the lines of what @andrewlee94 is saying, my initial implementation treats the calculation of units as a DAG, so that if a user specifies energy, energy_mass and energy_mole will be updated as well. However, the control volumes assume throughout that (a) specified units are consistent with each other and (b) expressions returned by the property package are consistent with these. This means that allowing this would probably require a potential unit conversion in each equation written by the control volumes, which could get to be expensive. Like I said, I think I'm in favor of allowing this, but can see the case against it.

eslickj:
I’d really like to avoid filling constraints with unit conversions, but I don’t mind giving users options if they know what they are doing.  Of course,  it’s hard to have a firm grip on what’s going on behind the scenes to set up all the constraints.

It’d be nice to allow any units, but it seems like a massive effort.

Robbybp:
> It’d be nice to allow any units, but it seems like a massive effort. Not to mention computation effort especially for dynamic problems.

Unless I'm missing something (which is likely), I don't think this would be a huge programming effort, but the extra computational effort could be huge. If we could somehow calculate and cache conversion factors, then re-apply them for every point in time, space, or whatever other indexing sets without an additional calculation, this might be manageable.

eslickj:
> > It’d be nice to allow any units, but it seems like a massive effort. Not to mention computation effort especially for dynamic problems.
> 
> Unless I'm missing something (which is likely), I don't think this would be a huge programming effort, but the extra computational effort could be huge. If we could somehow calculate and cache conversion factors, then re-apply them for every point in time, space, or whatever other indexing sets without an additional calculation, this might be manageable.

Well, it wouldn't be hard, but wouldn't we need to add convert calls to pretty much all constraints to ensure the units are consistent, and make sure everyone does it in their models going forward?  We'd also need to test (maybe multiple) inconsistent unit sets to make sure that there are converts statements to cover everything.  Maybe I'm think it's more complicated than it is, but to me it sounds really painful.  We'd also need to consider the unit conversion factors in  scaling, since they also act as scaling factors.  That may depend on what you decide the base units are. The choice of what the consistent set of units are could also vary from constraint to constraint.  Potentially it's just a big mess.

I guess potentially a unit model could write expressions for quantities in a consistent set of units then the unit models could use those expressions instead of variables.  If the variables are indexed maybe we could easily find the conversion factor for one element and apply it to all.  That would probably take care of the computational effort concern.

I may be wrong, but my feeling was the best thing to do is use consistent unit sets.  If the scaling is bad, we can scale the variables and constraints.  If the units are just weird to work with, we can convert inputs and output maybe by setting some display units.

I'm not super confident that I'm right here, so I can definitely be convinced otherwise.  Supporting arbitrary units would be really cool.

Robbybp:
I think I see more of why this would be so complicated. Adding a bunch of conversion statements to control volumes wouldn't be that bad, but testing them for every combination of inconsistent units would be tedious. Also, most unit models would be incompatible with property packages that don't use self-consistent units. I'm not sure this is a huge problem (unit models can choose what types of property packages to support), but will almost certainly cause some confusion/headache. Some way to check whether a property package defines "inherently consistent units" would be necessary.

I still think this change is worth thinking about if we can eliminate the overhead.

If we don't support assigning arbitrary units, I'd like to somehow support changing units "in-place", something along the lines of the Pyomo scaling transformation, but supporting units. Basically I'd like to be able to somehow arrive at a Pyomo model that uses units that aren't inherently consistent, e.g. km for length, m^2 for area, hours for time, kJ for energy, and bar for pressure, and (for now) I'm not a huge fan of needing to assign potentially solver-specific scaling factors to accomplish something like this.

andrewlee94:
@eslickj @Robbybp This is why we went with the restrictive requirement of only supporting internally consistent units. As soon as you allow for non-consistent units, even in a sub-part of the system, you need to start adding conversion factors to every constraint, and we know from experience that unit conversions can quickly become expensive when using deep expression trees. Whilst a lot of people would like to be able to use units that result in better scaling, I will note that this is effectively applying a simple order-of-magnitude scaling factor that is generally know in advance. I.e. if I think kJ is better scaled than J, then I know that any energy constraint just needs to be scaled by 1e-3. So, units-based scaling should not be solver specific.

We should bring @carldlaird and @jsiirola into this discussion to see if they have any ideas on how this could be achieved without a significant performance hit.

Robbybp:
@andrewlee94 Agreed. And I miscommunicated above. I meant that the implementation of scaling is solver-specific (e.g. how they get assigned, communicated, and used), not that the factors themselves are.

chineduobiora:
What I think could be useful is having block level default units declarations and maintaining unit consistency in each block based on those declared units i.e. property-package and unit model. I think this is a similar idea is similar to what @eslickj was alluding to with unit models using expressions for quantities of units. This will allow us to scale each block independently of each other. 
 We can reduce convert statements by ensuring that variables originating from a block should be enforced to be in the default units of that block. If in that block a call to a variable in a lower-level block (i.e. call from unit model to property block) exists in a constraint, then that variable should have a pyconvert to the units in the block (upper-level block).
 If consistency is maintained/enforced when writing each block this will reduce the amount of pyconvert statements needed.
 I also agree that caching units and applying across index sets (in space and time) will reduce the computational burden of using convert statements in constraints.

chineduobiora:
Regarding edge cases, I think that the edge cases in thermo-physical properties are easy to capture and update with pyconvert, while maintaining unit consistency in the unit models will reduce the number of other edge cases originating from them.

andrewlee94:
@chineduobiora The issue with units keeping consistency between block - within a Block a user is free to do what they want and take on the responsibility for ensuring unit consistency. The problem comes (as always) with dealing with consistency between block, as then the developer of a Block has no control over the other blocks around it. The only way to deal with this is to have a `convert` statement on any term that originates outside the current block, and with IDAES models that means basically every term as (almost) all terms in unit models depend on something from a property block.

Further, if we start letting each unit model decide on its own preferred units, then we need to add convert statements to all the `Arcs` in the model which Pyomo does not support (and probably does not want to support).

andrewlee94:
@Robbybp In the end, unit consistency always falls upon the model developer, but we have taken the effort to ensure that all core unit models are unit-independent. The more general answer is that a unit checking framework is only valuable if you apply it everywhere; as soon as units in one expression are inconsistent, the whole model is inconsistent and then there is no point to having units anymore.

andrewlee94:
So, to summarize the discussion we had on this at the Dev Team meeting:

1. Allow custom units in models, we would need to add `convert` statements to basically every term in every constraint in the core IDAES models. Whilst individually not hard, this is a large volume of work for which we currently do not have the resources.
2. Additionally, adding all these `convert` statements adds a very significant computational overhead in Pyomo at the moment. We have a propose solution to cut this down by adding `unit_hints` to named Expressions which we would trust for unit conversion and only check during consistency assertion, but this feature needs to be implemented in Pyomo, and then we would need to add these `unit_hints` in the IDAES code.

So, in short, we have a path forward, but we lack the resources to implement it right now, and it is probably not a short path. For now, I will close this PR and move the conversation to an issue for the future.

PR #638, opened 2021-12-20
Grid integration (double loop) modeling extension

xiangao1:
## Changes proposed in this PR:
- generalize `Bidder`, `Tracker`, `DoubleLoopCoordinator` classes.
- Simple Rankine cycle example
- Documentation

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #652, opened 2022-01-11
Add optional requirement and simple test case for Prescient 

lbianchi-lbl:
- Add `gridx-prescient` as an optional dependency in `setup.py`
- Add simple test case with built-in input CSV files (so that no network calls are required to run the tests)

## TODO before undrafting

- [x] Update URL for requirement in `setup.py` once grid-parity-exchange/Prescient#128 is merged in
- [x] Update options to reduce simulation runtime in `test_prescient.py` (or update marker to `pytest.mark.integration`)

## Reviewers

- @bknueven (for some reason I wasn't able to add you through the usual menu when creating the PR)

### Legal Acknowledgement


idaes/tests/prescient/test_prescient.py
```diff
@@ -0,0 +1,81 @@
...
+    @pytest.fixture
+    def prescient_options(self, data_path: Path, output_dir: Path) -> PrescientOptions:
+        return {
+            "data_path": str(data_path),
+            "input_format":"rts-gmlc",
+            "simulate_out_of_sample":True,
+            "run_sced_with_persistent_forecast_errors":True,
+            "output_directory": str(output_dir),
+            "start_date":"07-10-2020",
+            "num_days":7,
```
bknueven, line 48:
```suggestion
            "num_days":2,
```

idaes/tests/prescient/test_prescient.py
```diff
@@ -0,0 +1,81 @@
...
+            "sced_horizon":1,
+            "ruc_mipgap":0.01,
+            "reserve_factor":0.1,
+            "deterministic_ruc_solver":"cbc",
+            "deterministic_ruc_solver_options":{"feas":"off", "DivingF":"on",},
+            "sced_solver":"cbc",
+            "sced_frequency_minutes":5,
```
bknueven, line 55:
```suggestion
            "sced_frequency_minutes":60,
```

idaes/tests/prescient/test_prescient.py
```diff
@@ -0,0 +1,81 @@
...
+            "ruc_horizon":36,
+            "compute_market_settlements":True,
+            "monitor_all_contingencies":True,
```
bknueven, line 58:
```suggestion
            "monitor_all_contingencies":False,
```

bknueven (commented):
The suggested option changes below should get you a much faster regression test.

lbianchi-lbl:
@bknueven thanks for the updated options! The runtime is ~10-15 s on the GHA runners, which I think is perfect for a component test.

Do you have any suggestion for a basic validation of the results once the simulation is complete? The current implementation just asserts that at least two files have been created in the output directory, which is not a particularly stringent check...

bknueven:
Prescient's own regression tests are available here: https://github.com/grid-parity-exchange/Prescient/blob/main/tests/simulator_tests/test_sim_rts_mod.py

That said, something simpler might just be to check for the existence of `overall_simulation_output.csv` in the solution directory, which is only created and written to as one of the final steps of Prescient execution.

If you want to check the values in any of the files, you should update the `ruc_mipgap` option to `0.0`. Also, it's probably a bad idea to use strict file equality as we occasionally add columns to these csv files if we find something else worth reporting.

lbianchi-lbl:
@bknueven Is there a planned timeline for when a new version of Prescient including these changes will be released on PyPI? From the IDAES side, I think this is ready to go, but I imagine it would be preferable to use a PyPI release (rather than `main`) to specify the requirement.

bknueven:
We could cut the Prescient release when convenient. We were hoping to get one more feature in, but we also don't want to hold up the IDAES release **or** have IDAES released against a specific Prescient commit.

lbianchi-lbl:
@bknueven Thanks for the update. My question for purely for keeping track of this PR's status: personally I don't know how urgent it would be to have this merged into IDAES (and therefore how urgently we'd need to have the Prescient PyPI release). 

@adowling2, do you have any thoughts on the required or desired timeline for enabling Prescient in IDAES?

setup.py
```diff
@@ -108,9 +108,10 @@ def dmf_data_files(root: str = DMF_DATA_ROOT) -> List[Tuple[str, List[str]]]:
         "prerelease": DEPENDENCIES_FOR_PRERELEASE_VERSION,
         "optional": [
             "tensorflow",  # idaes.surrogate.keras_surrogate
+            "gridx-prescient==2.1",  # idaes.tests.prescient
```
bknueven, line 111: I'm not sure it matters, but we're fairly careful not to make breaking changes to Prescient at this point.
```suggestion
            "gridx-prescient>=2.1",  # idaes.tests.prescient
```

bknueven (approved):
This seems reasonable to me

setup.py
lbianchi-lbl: Good point, I think it makes sense.

bknueven:
This should be merged before #638 so #638 can include Prescient integration tests.

ksbeattie (approved):
LGTM

.github/workflows/core.yml
```diff
@@ -56,6 +56,7 @@ jobs:
           - python-version: '3.8'
             # only generate coverage report for a single python version in the matrix
             # to avoid overloading Codecov
+            # FIXME: reinstate coverage report before merging PR
```
bknueven, line 59: Should this be removed now?

lbianchi-lbl: Thanks for catching that, I've removed it in 559b788.

bknueven (approved):

ksbeattie (approved):
LGTM

PR #655, opened 2022-01-11
Stream table brushing

elbashandy:
## Summary/Motivation:
Potential user showed interest in marking links when they hover over streams in the stream table and vice versa.

-
-

### Legal Acknowledgement


ksbeattie:
Two minor suggestions:
1. Add the ability to turn this feature on/off.  Maybe via the "View" menu.  Not sure if the default should be on as it might be disorienting if the user hasn't enabled it.
2. When the table column is highlighted and text within each cell moves a little bit.  Anyway to prevent that movement?

dangunter (approved):
Looks good

ksbeattie (dismissed):
LGTM

dangunter (approved):
Looks good

PR #662, opened 2022-01-12
Pin IPython version to address test failures (#661)

### LINKED ISSUE #661: Test failures in DMF test_magics after IPython 8.0.0 release
As originally brought up by @eslickj in #552, it looks like there is a breaking change in IPython introduced by 8.0.0 that apparently was released just a few hours ago: https://pypi.org/project/ipython/#history
 
image

One temporary fix to try would be to add a requirement with version constraint in `setup.py`, e.g. `ipython<8.0.0`.

lbianchi-lbl:
- Add requirement `ipython<8.0.0`

### Legal Acknowledgement

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,6 +77,8 @@ kwargs = dict(
         "sympy",
         "tinydb",
         "rbfopt",
+        # lbianchi-lbl: see https://github.com/IDAES/idaes-pse/issues/661
+        "ipython<8.0.0",
     ],
     entry_points={
         "console_scripts": [


blnicho (approved):
I'm fine with this as a quick temporary fix but I think we should open an issue (or keep #661 open) to make our code compatible with ipython 8.0.0

lbianchi-lbl:
@blnicho As far as I understand from the error message, the error occurs in `_display_markdown()` which is an internal IPython function; if this is the case, the problem is not in the IDAES code and therefore we have more limited options in addressing the problem directly.

eslickj (approved):

PR #555, opened 2021-10-15
FTPx initialization

dallan-keylogic:
## Summary/Motivation:
I was having problems initializing flash calculations in which a noncondensable phase was present. It turned out that no attempt was made to calculate the vapor fraction or component phase splits when noncondensable or nonvaporizable phases were present in the `FTPx.py` initialization. 

## Changes proposed in this PR:
I derived a method (closely related to ideal bubble and dew point calculations) to calculate the vapor fraction of an ideal mixture obeying Raoult's law through a 1D root finding problem. Mathematically, one can prove that the root finding algorithm converges so long as the system pressure is between the ideal dew and bubble pressures, all the saturation pressures, Henry's law constants, and mole fractions are nonnegative, and the mole fractions add up to 1.  I added a failsafe for if the system pressure does not fall into that range, and exceptions if "significantly" nonphysical values for component parameters are calculated, as well as a failsafe if ill-conditioning or "insignificant" unphysical values cause a convergence failure.

A new IDAES exception, the `UserModelError` is added to handle cases in which unphysical values computed from user-supplied methods prevent execution from continuing.

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #665, opened 2022-01-17
Bring some NGCC branch items

eslickj:
This brings in changes to a few power generation library files from the NGCC branch.  With these, I should be able to write an updated NGCC flowsheet for the examples repo.  These are mostly minor updates to allow more formulation options.  Some of these were reformatted with black, so there looks like more change than there really is.

### Legal Acknowledgement

### DIFF
--- a/idaes/power_generation/properties/flue_gas_ideal.py
+++ b/idaes/power_generation/properties/flue_gas_ideal.py
@@ -18,9 +18,20 @@ Main assumptions:
     - components in flue gas: O2, N2, NO, CO2, H2O, SO2
 """
 # Import Pyomo libraries
-from pyomo.environ import (Constraint, Param, PositiveReals, Reals, value, log,
-                           sqrt, Var, Expression, units as pyunits)
+from pyomo.environ import (
+    Constraint,
+    Param,
+    PositiveReals,
+    Reals,
+    value,
+    log,
+    sqrt,
+    Var,
+    Expression,
+    units as pyunits,
+)
 from pyomo.opt import SolverFactory
+from pyomo.common.config import ConfigValue
 
 # Import IDAES cores
 from idaes.core import (
@@ -29,12 +40,13 @@ from idaes.core import (
     StateBlockData,
     StateBlock,
     Component,
-    VaporPhase
+    VaporPhase,
 )
 from idaes.core.util.model_statistics import (
-    degrees_of_freedom, number_activated_constraints)
-from idaes.core import (
-    MaterialBalanceType, EnergyBalanceType, MaterialFlowBasis)
+    degrees_of_freedom,
+    number_activated_constraints,
+)
+from idaes.core import MaterialBalanceType, EnergyBalanceType, MaterialFlowBasis
 from idaes.core.util.initialization import fix_state_vars, revert_state_vars
 from idaes.core.util import constants, get_solver
 import idaes.core.util.scaling as iscale
@@ -48,7 +60,7 @@ __author__ = "Boiler Subsystem Team  J. Ma, M. Zamarripa, T. Burgard"
 __version__ = "3"
 
 # Set up logger
-_log = idaeslog.getLogger('idaes.unit_model.properties')
+_log = idaeslog.getLogger("idaes.unit_model.properties")
 
 
 @declare_process_block_class("FlueGasParameterBlock")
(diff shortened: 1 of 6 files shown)


idaes/power_generation/properties/flue_gas_ideal.py
```diff
@@ -366,34 +501,39 @@ def initialize(
...
+            }
 
         if state_vars_fixed is False:
             flags = fix_state_vars(self, state_args)
         # Check when the state vars are fixed already result in dof 0
         for b in self.values():
             if degrees_of_freedom(b) != 0:
-                raise Exception(f"{self.name} initializtion error: State vars "
-                                "fixed but degrees of freedom not equal to 0")
+                raise Exception(
```
andrewlee94, line 528: An `InitializationError` would be better here (rather than a base `Exception`).

idaes/power_generation/properties/flue_gas_ideal.py
```diff
@@ -419,11 +560,10 @@ def initialize(
...
                 res = "skipped"
             init_log.info_high(
-                "Initialization Step 2 {}.".format(idaeslog.condition(res)))
+                "Initialization Step 2 {}.".format(idaeslog.condition(res))
+            )
 
-        init_log.info(
-            'Initialisation Complete, {}.'.format(
-                idaeslog.condition(res)))
+        init_log.info("Initialisation Complete, {}.".format(idaeslog.condition(res)))
```
andrewlee94, line 566: We should also add an `InitializationError` here if the final solve was not optimal:

```
if not check_optimal_termination(res):
    raise InitializationError()
```

idaes/power_generation/unit_models/boiler_heat_exchanger.py
```diff
@@ -77,11 +79,92 @@ class DeltaTMethod(Enum):
     coCurrent = 1
 
 
+def delta_temperature_lmtd_callback(b):
```
andrewlee94, line 82: I didn't do a detailed comparison, but are these the same the ones in the base HX module? If so, shouldn't we just import them from there?

idaes/power_generation/unit_models/boiler_heat_exchanger.py
```diff
@@ -300,57 +446,65 @@ def _set_geometry(self):
...
 
         # Thickness of tube
-        self.tube_thickness = Var(initialize=0.005,
-                                  within=PositiveReals,
-                                  doc='Tube thickness - m',
-                                  units=pyunits.m)
+        self.tube_thickness = Var(
+            initialize=0.005,
+            within=PositiveReals,
+            doc="Tube thickness - m",
```
andrewlee94, line 487: Minor comment, but I would avoid listing units in the doc strings now that we have formal definition of units in the component. It is effectively duplicating the information, and adds a point where things can become inconsistent.

idaes/power_generation/properties/natural_gas_PR.py
```diff
@@ -82,6 +90,18 @@
...
 
+_phase_dicts_ideal = {
+    "Vap": {
+        "type": VaporPhase,
+        "equation_of_state": Ideal,
+    },
+    "Liq": {
+        "type": LiquidPhase,
+        "equation_of_state": Cubic,
+        "equation_of_state_options": {"type": CubicType.PR}
```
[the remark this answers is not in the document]
eslickj: I was, but it didn't really work right.  I'll take it out.

idaes/power_generation/unit_models/boiler_heat_exchanger.py
eslickj: Yeah.  I think they are I'll double check and do that.

idaes/power_generation/properties/natural_gas_PR.py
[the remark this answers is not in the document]
eslickj: I didn't really look at it too close.  I just didn't get answers that made sense.  I was trying to separate hydrogen and water.   I asked @dallan-keylogic to look into it, and he probably knows a lot more than I do at this point.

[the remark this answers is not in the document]
dallan-keylogic: @andrewlee94 @eslickj The problem I found with the separation of hydrogen and water was that there was no initialization for flash calculations if there was a single phase component. That's fixed in #555 , but is suspended in limbo now because Andrew wanted tests for Henry's Law initialization, and I didn't have the time to write them last month. I don't have the time to write them now either, but hopefully it will happen sometime soon.

idaes/power_generation/unit_models/boiler_heat_exchanger.py
eslickj: Ok, this exists in the base heat exchanger, but it's named different, so I'll leave it.  Ideally this would inherit the base heatexchanger, but I guess I won't push that.  It's probably too late to go back.

eslickj: Okay.  I figured out why I rewrote the call backs here.  This model has different variable and expression names than the generic heat exchanger.

andrewlee94: Should we standardize the names between the two models then? To someone who knows nothing about the details of the model, it sounds like that would be a good idea.

eslickj: I feel like that ship has sailed.  We should inherit the standard heat exchanger, but there are a lot of legacy models (not automatically tested and not in the repo) that probably depend on us not changing this in a way that would break backward compatibility.

andrewlee94: I am not sure how I feel about that. whilst I don't want to break other peoples models, we really should try to enforce some standards on naming, otherwise we are just going to end up proliferating models which exist only because someone used a different set of names. I will also note that whilst we guarantee that if there is a test we won't break it, the reverse is not true.

Is there a way we could set up some `References` or some deprecation path so that we could standardize this?

eslickj: Probably.  I'll take a look.

eslickj: Ok, I added the references and used the callbacks from the generic heat exchanger.

idaes/power_generation/properties/natural_gas_PR.py
```diff
@@ -431,7 +447,15 @@ def get_prop(components=None, phases="Vap"):
     if isinstance(phases, str):
         phases = [phases]
     for k in phases:
-        configuration["phases"][k] = copy.deepcopy(_phase_dicts[k])
+        if eos==EosType.PR:
+            configuration["phases"][k] = copy.deepcopy(_phase_dicts_pr[k])
+        elif eos==EosType.IDEAL:
+            if k == "Liq":
+                raise ConfigurationError(
+                    "This paraemeter set does not support Ideal EOS with liquid")
```
dallan-keylogic, line 455: paraemeter -> parameter

idaes/power_generation/properties/flue_gas_ideal.py
```diff
@@ -61,191 +75,301 @@ class FlueGasParameterData(PhysicalParameterBlock):
...
+                    "CO2": 244.0,
+                    "H2O": 572.4,
+                    "SO2": 252.0,
+                }.items()
+                if k in self.component_list
+            },
             doc="characteristic energy of interaction between pair of "
-                "molecules, K = Boltzmann constant in Kelvin",
-            units=pyunits.K)
+            "molecules, K = Boltzmann constant in Kelvin",
```
dallan-keylogic, line 370: Maybe "reduced by the Boltzmann constant to put in units of Kelvin"?

idaes/power_generation/unit_models/tests/test_boiler_heat_exchanger.py
```diff
@@ -176,3 +180,75 @@ def test_units():
...
+@pytest.mark.unit
+def test_config_uwt():
+    tc(delta_temperature_underwood_tune_callback)
+
+@pytest.mark.skipif(not iapws95.iapws95_available(),
+                    reason="IAPWS not available")
+@pytest.mark.skipif(solver is None, reason="Solver not available")
+@pytest.mark.component
+def test_boiler_hx_am():
+    # arethmetic mean is pretty far off
```
dallan-keylogic, line 207: arethmetic -> arithmetic

dallan-keylogic (dismissed):

idaes/power_generation/properties/flue_gas_ideal.py
eslickj: I didn't write this, so it took a bit to find the correct description.

dallan-keylogic (dismissed):

idaes/power_generation/unit_models/boiler_heat_exchanger.py
```diff
@@ -163,21 +134,27 @@ class BoilerHeatExchangerData(UnitModelBlockData):
     """
 
     CONFIG = ConfigBlock()
```
andrewlee94, line 136: Minor comment, but why not use `CONFIG = UnitModelBlockData.CONFIG()` (or just rely on inheritance directly)? As far as I can tell, you are just re-declaring the contents of the base classes Config block anyway (dynamic and has_holdup).

idaes/power_generation/unit_models/boiler_heat_exchanger.py
```diff
@@ -720,6 +699,9 @@ def _make_performance(self):
             units=pyunits.K,
         )
 
+        self.delta_temperature_in = Reference(self.deltaT_1)
```
andrewlee94, line 702: This is something for the IDAES 2.0 discussion, but we really should standardize names across all the HX models (including the names for sides and Ports).

andrewlee94 (approved):
Only one minor comment at this point.

idaes/power_generation/unit_models/boiler_heat_exchanger.py
eslickj: Will be fixed in a PR coming in a day or two I promise.

dallan-keylogic (approved):

eslickj:
@andrewlee94 and @dallan-keylogic I had to make a minor change to the NGFC flowsheet to set some initial mole fractions to 1e-11 instead of 0 because there was a failing integration test.  Could you guys re-approve?

andrewlee94 (approved):
Looks good to me.

PR #668, opened 2022-01-20
Change Slice Tags to Reference

eslickj:
In the ModeTag class, when you tag a slice convert it to a Reference instead of storing the slice.  That makes the tag value easier to deal with since the indexing is preserved better.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/tags.py
+++ b/idaes/core/util/tags.py
@@ -63,7 +63,10 @@ class ModelTag:
         """
         super().__init__()
         self._format = format_string  # format string for printing expression
-        self._expression = expr  # tag expression (can be unnamed)
+        if isinstance(expr, IndexedComponent_slice):
+            self._expression = pyo.Reference(expr)  # tag expression (can be unnamed)
+        else:
+            self._expression = expr
         self._doc = doc  # documentation for a tag
         self._display_units = display_units  # unit to display value in
         self._cache_validation_value = {}  # value when converted value stored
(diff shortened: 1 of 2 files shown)


andrewlee94 (approved):
I think this looks OK (although I wonder a bit about some of the flow on effects like type checking and backwards compatibility). I do have to ask about tests for this however - notably the fact hat you could remove the `is_slice` property without having to change the tests implies that there aren't enough at the moment.

eslickj:
@andrewlee94, I had a lot of tests for this, but it looks like you were right.  I overlooked slices.  I added a test.  It's not likely to have broken anything.  It would just change the way the indexing looks for slices, and I'm almost certainly the only person to ever use this.

andrewlee94:
@eslickj My biggest concern there was the fact that the `is_slice` property returned a tuple, which was no longer going to happen. Thus, anything expecting the tuple (if any such thing existed) would likely fail.

eslickj:
@andrewlee94, is_slice was used internally.  Someone could have used it but hopefully didn't.  Since slices lost their indexing before this, the main use I had for them was plotting, which still works with the reference.  An example is like a distillation column, where you want to plot mole fraction of a component and the variable is indexed by time, position and composition.  You can get a tag just indexed by position.  You probably don't care if it's a slice or Reference since you'll just iterate through.

dallan-keylogic (approved):

PR #675, opened 2022-01-25
Update 0D boiler heatexchanger to inherit from generic heat exchaner

eslickj:
This updates the OD boiler heat exchanger model to inherit the generic heat exchanger model.  This is to make the models more standard.  This shouldn't break anything, but it could slightly affect the scaling.  You'll get some depreciation warnings if you use some old config options.  We can move away from the deprecated interface over time.  This changes surprisingly little about the model.

### Legal Acknowledgement

### DIFF
--- a/idaes/generic_models/unit_models/heat_exchanger.py
+++ b/idaes/generic_models/unit_models/heat_exchanger.py
@@ -21,17 +21,14 @@ from enum import Enum
 # Import Pyomo libraries
 from pyomo.environ import (
     Var,
-    Param,
-    Expression,
     log,
     Reference,
     PositiveReals,
-    SolverFactory,
     ExternalFunction,
     Block,
     units as pyunits,
-    NonNegativeReals,
-    value,
+    SolverStatus,
+    TerminationCondition
 )
 from pyomo.common.config import ConfigBlock, ConfigValue, In
 
@@ -52,7 +49,7 @@ from idaes.generic_models.unit_models.heater import (
 import idaes.core.util.unit_costing as costing
 from idaes.core.util.misc import add_object_reference
 from idaes.core.util import get_solver, scaling as iscale
-from idaes.core.util.exceptions import ConfigurationError
+from idaes.core.util.exceptions import ConfigurationError, InitializationError
 
 _log = idaeslog.getLogger(__name__)
 
@@ -236,7 +233,7 @@ def delta_temperature_underwood_callback(b):
     """
     dT1 = b.delta_temperature_in
     dT2 = b.delta_temperature_out
-    temp_units = pyunits.get_units(dT1[dT1.index_set().first()]) #handle Refernce
+    temp_units = pyunits.get_units(dT1[dT1.index_set().first()])
 
     # external function that ruturns the real root, for the cuberoot of negitive
     # numbers, so it will return without error for positive and negitive dT.
@@ -614,6 +611,14 @@ class HeatExchangerData(UnitModelBlockData):
         if hasattr(self, "costing"):
             self.costing.activate()
             costing.initialize(self.costing)
+
+        if (res is not None and (
+                res.solver.termination_condition !=
+                TerminationCondition.optimal or
+                res.solver.status != SolverStatus.ok)):
+            raise InitializationError(
+                f"{self.name} failed to initialize successfully. Please check "
+                f"the output logs for more information.")
 
     def _get_performance_contents(self, time_point=0):
         var_dict = {
--- a/idaes/generic_models/unit_models/heat_exchanger.py
+++ b/idaes/generic_models/unit_models/heat_exchanger.py
@@ -353,7 +353,7 @@ class HeatExchangerData(UnitModelBlockData):
             if not hasattr(self, "side_1"):
                 add_object_reference(self, "side_1", cold_side)
             if not hasattr(self, "side_2"):
-                add_object_reference(self, "side_2", hot_side)    
+                add_object_reference(self, "side_2", hot_side)
         if not hasattr(self, "hot_side"):
             add_object_reference(self, "hot_side", hot_side)
         if not hasattr(self, "cold_side"):
@@ -425,17 +425,29 @@ class HeatExchangerData(UnitModelBlockData):
             block=cold_side,
             doc="Cold side outlet")
 
+        if not config.side_1_is_hot:
+            i1, i2 = i2, i1
+            o1, o2 = o2, o1
+
         # Using Andrew's function for now.  I want these port names for backward
         # compatablity, but I don't want them to appear if you iterate throught
         # components and add_object_reference hides them from Pyomo.
         if not hasattr(self, "inlet_1"):
             add_object_reference(self, "inlet_1", i1)
+        if not hasattr(self, "side_1_inlet"):
+            add_object_reference(self, "side_1_inlet", i1)
         if not hasattr(self, "inlet_2"):
             add_object_reference(self, "inlet_2", i2)
+        if not hasattr(self, "side_2_inlet"):
+            add_object_reference(self, "side_2_inlet", i2)
         if not hasattr(self, "outlet_1"):
             add_object_reference(self, "outlet_1", o1)
+        if not hasattr(self, "side_1_outlet"):
+            add_object_reference(self, "side_1_outlet", o1)
         if not hasattr(self, "outlet_2"):
             add_object_reference(self, "outlet_2", o2)
+        if not hasattr(self, "side_2_outlet"):
+            add_object_reference(self, "side_2_outlet", o2)
 
         if not hasattr(self, "hot_inlet"):
             add_object_reference(self, "hot_inlet", i1)
--- a/idaes/generic_models/unit_models/heat_exchanger.py
+++ b/idaes/generic_models/unit_models/heat_exchanger.py
@@ -21,14 +21,17 @@ from enum import Enum
 # Import Pyomo libraries
 from pyomo.environ import (
     Var,
+    Param,
+    Expression,
     log,
     Reference,
     PositiveReals,
+    SolverFactory,
     ExternalFunction,
     Block,
     units as pyunits,
-    SolverStatus,
-    TerminationCondition
+    NonNegativeReals,
+    value,
 )
 from pyomo.common.config import ConfigBlock, ConfigValue, In
 
@@ -49,7 +52,7 @@ from idaes.generic_models.unit_models.heater import (
 import idaes.core.util.unit_costing as costing
 from idaes.core.util.misc import add_object_reference
 from idaes.core.util import get_solver, scaling as iscale
-from idaes.core.util.exceptions import ConfigurationError, InitializationError
+from idaes.core.util.exceptions import ConfigurationError
 
 _log = idaeslog.getLogger(__name__)
 
@@ -233,7 +236,7 @@ def delta_temperature_underwood_callback(b):
     """
     dT1 = b.delta_temperature_in
     dT2 = b.delta_temperature_out
-    temp_units = pyunits.get_units(dT1)
+    temp_units = pyunits.get_units(dT1[dT1.index_set().first()]) #handle Refernce
 
     # external function that ruturns the real root, for the cuberoot of negitive
     # numbers, so it will return without error for positive and negitive dT.
@@ -611,14 +614,6 @@ class HeatExchangerData(UnitModelBlockData):
         if hasattr(self, "costing"):
             self.costing.activate()
             costing.initialize(self.costing)
-
-        if (res is not None and (
-                res.solver.termination_condition !=
-                TerminationCondition.optimal or
-                res.solver.status != SolverStatus.ok)):
-            raise InitializationError(
-                f"{self.name} failed to initialize successfully. Please check "
-                f"the output logs for more information.")
 
     def _get_performance_contents(self, time_point=0):
         var_dict = {
--- a/idaes/generic_models/unit_models/heat_exchanger.py
+++ b/idaes/generic_models/unit_models/heat_exchanger.py
@@ -21,17 +21,14 @@ from enum import Enum
 # Import Pyomo libraries
 from pyomo.environ import (
     Var,
-    Param,
-    Expression,
     log,
     Reference,
     PositiveReals,
-    SolverFactory,
     ExternalFunction,
     Block,
     units as pyunits,
-    NonNegativeReals,
-    value,
+    SolverStatus,
+    TerminationCondition
 )
 from pyomo.common.config import ConfigBlock, ConfigValue, In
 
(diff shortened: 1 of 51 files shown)


eslickj:
@dangunter or @elbashandy, I made an update to the heat exchange boiler and it broke a ui test for model serialization.  I suspect it just serializes a bit differently now.  How can I update the test?

idaes/power_generation/unit_models/boiler_heat_exchanger.py
```diff
@@ -523,9 +490,11 @@ def area_flow_shell(b):
             return b.tube_length * (b.pitch_y - b.do_tube) * b.tube_ncol
 
         # Total heat transfer area based on outside diameter
-        @self.Expression(doc="Total heat transfer " "area based on tube outside diamer")
-        def area_heat_transfer(b):
-            return c.pi * b.do_tube * b.tube_length * b.tube_ncol * b.tube_nrow
+        @self.Constraint(doc="Total heat transfer " "area based on tube outside diamer")
```
andrewlee94, line 493: Extra "" in middle of string, typo: diameter.

idaes/generic_models/unit_models/heat_exchanger.py
```diff
@@ -287,10 +295,16 @@ def _process_config(self):
             # compatible with the tube and shell notation
             setattr(config, config.cold_side_name, config.cold_side_config)
 
-        if config.cold_side_name in ["hot_side", "side_1"]:
-            raise ConfigurationError("Cold side name cannot be in ['hot_side', 'side_1'].")
-        if config.hot_side_name in ["cold_side", "side_2"]:
-            raise ConfigurationError("Hot side name cannot be in ['cold_side', 'side_2'].")
+        if config.side_1_is_hot:
```
andrewlee94, line 298: This seems like additional complexity and a potential point of confusion for users in the future - I assume there is a reason we need to know the hot side explicitly at some point for the more complex models?

My first (perhaps naive) reaction is that we should either force the issue and just make one side hot_side (and drop the generic side_1, side_2 names) if if actually matters, or just not assign any actual meaning to the names at all (i.e. you can call it hot side if you want, but we won't treat it any differently). Alternatively, we could use a standard (internal at least) that e.g. side_1 will always be the hot side.

idaes/generic_models/unit_models/heat_exchanger.py
```diff
@@ -330,10 +344,16 @@ def build(self):
         # compatability with older models.  Using add_object_reference keeps
         # these from showing up when you iterate through pyomo compoents in a
         # model, so only the user specified control volume names are "seen"
-        if not hasattr(self, "side_1"):
-            add_object_reference(self, "side_1", hot_side)
-        if not hasattr(self, "side_2"):
-            add_object_reference(self, "side_2", cold_side)
+        if config.side_1_is_hot:
```
andrewlee94, line 347: I wonder if we should have a deprecation warning here too (to try an push things towards using the standard names). I.e. let the user provide aliases, but also force them to have the standard names in the background. (and log a deprecation warning if they don't).

idaes/power_generation/unit_models/boiler_heat_exchanger.py
```diff
@@ -110,9 +123,9 @@ def delta_temperature_underwood_tune_callback(b):
     This uses a cube root function that works with negative numbers returning
```
andrewlee94, line 123: I know I've said this before, but do we really need this option? The author who came up with this originally had a misconception about the whole Underwood method, and their claimed "improvements" were really no different from the original method (basically they overfitted the parameters using limited data and claimed it was better).

Notably, c1 ~= 1/3 and c2 ~=2 which were the original parameters from Underwood.

In fact, that makes me suggest an alternative: if we really think we want this, then we should just take the Underwood method and make the parameters Vars with the default values and let users "tune" them if they want.

andrewlee94 (commented):
A few comments on things I saw in a first pass.

idaes/generic_models/unit_models/heat_exchanger.py
eslickj: Yeah, this is a bit of a mess.  It generally doesn't matter which side is hot or cold.  But designating one side as hot helps since it may matter sometimes and it give you a point of reference if say the heat transfer is negative.  I don't think it hurts for the user to specify which side they think is hot.  The problem with side_1 and side_2 is that, at least, I can't keep them straight.  Of course, you can always name the sides anything you want.

The complex business with the side_1 and side_2 is for backward compatibility.  Also the boiler heat exchanger was different than the standard heat exchanger so I had to add another option about whether side_1 was cold or hot.

Anyway by default the sides are named "shell" and "tube".  It's a mess so as not to break any old models.  I guess it's an IDAES 2.0 cleanup action.

idaes/generic_models/unit_models/heat_exchanger.py
eslickj: I don't know.  I like the idea, but then the boiler heat exchanger would raise deprecation warning that the user couldn't fix.

idaes/power_generation/unit_models/boiler_heat_exchanger.py
eslickj: This was the original LMTD calculation in the boiler model.  I actually originally changed the boiler heat exchanger so I could pick something else.  The reason it is still here is that I promised I wouldn't break anything, so it has to remain the default option.

andrewlee94: I think the boiler team really need to rethink that then, as they should not be using that correlation.

Maybe a deprecation path then? We update the Underwood form to allow adjustment of parameters, and then this callback can use it, update the parameters and log a deprecation warning?

eslickj: Ok.  I can add a depreciation warning to this and add a tunable Underwood with adjustable parameters as variables.  This PR is going to unleash a flood of depreciation warnings where-ever this model is used.  I can't update the plain Underwood approximation with adjustable parameters.  Right now it doesn't care which side is hot and which side is cold, but once you adjust the exponent the hot side needs to be hot or you'll get evaluation errors.

andrewlee94: If the Underwood correlation requires us to know the hot side and cold side, then that beings to imply we should just bite the bullet and force one side to be hot in the base model (of course that risks breaking even more things). However, this is an important enough case that if the hot side matters, maybe we should drop the generalization at the base level.

andrewlee94: However, this is beginning to sound like another IDAES 2.0 thing.

eslickj: Well I looked back and I do have variables for the parameters.  I agree it shouldn't be the default, but I can't figure out how to turn that into a meaningful deprecation warning.  I guess we can put it on the IDAES 2.0 list to change the default.  The only thing I can think of is to produce a warning message that says we'll change the default behavior in IDAES 2.0, but I don't know if I should do that.  If we switch to a standard LMTD calculation it shouldn't break things, but it will change model results a bit (more or less depending on the delta Ts).

andrewlee94: I would do the following (if possible):

1. Keep the current method for tune Underwood, but make it call the base Underwood method to set up the Vars and expression/constraint.
2. Then update the parameters as part of the tune Underwood method this should keep the result the same)
3. Log a deprecation warning that the tuned Underwood method will go away and that users should use the base Underwood method instead.

If that is not feasible, then I would just add the deprecation warning to the tuned Underwood method, but do nothing else. Just say that the method is going away, and that you should instead use the base method and adjust parameters.

eslickj: Ok.  I'll add the deprecation warning.  The issue with using the base Underwood method is that it uses a cubroot function that returns the real root, so it doesn't matter if delta T is positive or negative.  It's not really possible to change the parameters there.

docs/technical_specs/model_libraries/power_generation/unit_models/boiler_heat_exchanger.rst
```diff
@@ -6,19 +6,19 @@ BoilerHeatExchanger
 
 .. currentmodule:: idaes.power_generation.unit_models.boiler_heat_exchanger
 
-The BoilerHeatExchanger model can be used to represent boiler heat exchangers in 
-sub-critical and super critical power plant flowsheets (i.e. econmizer, primary superheater, secondary superheater, finishing superheater, reheater, etc.).  
-The model consists of a shell and tube crossflow heat exchanger, in which the shell is used as the gas side and the tube is used as the water or steam side. 
+The BoilerHeatExchanger model can be used to represent boiler heat exchangers in
+sub-critical and super critical power plant flowsheets (i.e. econmizer, primary superheater, secondary superheater, finishing superheater, reheater, etc.).
```
andrewlee94, line 10: Typo: econmizer

andrewlee94 (changes requested):
A found a few typos (including one in the docs), so they should be fixed before merging.

I also have one more general question - do we have a timeline planned for full deprecation? We have a lot of warnings, and generally there should be a plan for when they will be fully deprecated and removed.

Finally, it would be good if we could get the coverage a bit higher - unfortunately CodeCov is refusing to let me look at the coverage report, so I can't suggest any areas for improvement.

idaes/power_generation/unit_models/tests/test_boiler_heat_exchanger.py
```diff
@@ -46,86 +57,92 @@
...
+            "side_1_water_phase": "Liq",
+            "has_radiation": True,
+        }
+    )
 
     # Check unit config arguments
     # There are 8 to 10 arguments since you can add a side 1 and 2 config by
     # side_1, side_2, or whatever the user named them
-    assert len(m.fs.unit.config) >= 12 and len(m.fs.unit.config) <= 16
+    # assert len(m.fs.unit.config) >= 12 and len(m.fs.unit.config) <= 16
```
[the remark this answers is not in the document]
eslickj: Yeah.  I'll do that.

eslickj:
@andrewlee94, I'm working on more tests.  I don't really have a plan for removing the deprecated stuff.  Depending on how soon we think these versions will happen I'm thinking 2.0 or 3.0.  I want people to have a very long time to respond to the warnings.

andrewlee94:
@eslickj I just had a look at the CodeCov report (it finally let me view it), and it looks like the new deprecation warnings in the boiler heat exchanger are the problem. For the sake of the power generation team, we should probably add tests to make sure the deprecation paths work (and whilst we are at it, we should confirm we see a deprecation warning). Hopefully those tests are not hard to write, and they should be fairly fast as we don't need to solve the model for those.

eslickj:
@andrewlee94, will do.  It's easy.

idaes/power_generation/unit_models/boiler_heat_exchanger.py
```diff
@@ -96,9 +99,23 @@ class TubeArrangement(Enum):
     staggered = 1
 
 
-class DeltaTMethod(Enum):
-    counterCurrent = 0
-    coCurrent = 1
+class _DeprecateDeltaTMethod(EnumMeta):
+    # This is used to log a deprication warning if someone uses DeltaTMethod
```
andrewlee94, line 103: V. minor comment: Typo: deprecation.

idaes/power_generation/unit_models/boiler_heat_exchanger.py
```diff
@@ -110,9 +127,14 @@ def delta_temperature_underwood_tune_callback(b):
     This uses a cube root function that works with negative numbers returning
     the real negative root. This should always evaluate successfully.
     """
-    temp_units = pyunits.get_units(b.deltaT_1)
-    dT1 = b.deltaT_1
-    dT2 = b.deltaT_2
+    _log.warning(
+        "DEPRECATED: delta_temperature_underwood_tune_callback will be "
+        "removed. Use another standard delta_temperature callback or "
+        "provide a custom callback.  This will be removed in IDAES 3.0."
```
andrewlee94, line 133: IDAES 3.0 seems a bit far in the future (seeing as we haven't got a concrete date for IDAES 2.0 yet). Would IDAES 2.1 be OK? (still vague, but sooner).

idaes/power_generation/unit_models/boiler_heat_exchanger.py
```diff
@@ -128,47 +150,15 @@ def delta_temperature(b, t):
...
-            doc="""Indicates whether holdup terms should be constructed or not.
-Must be True if dynamic = True,
-**default** - False.
-**Valid values:** {
-**True** - construct holdup terms,
-**False** - do not construct holdup terms}""",
-        ),
-    )
+class BoilerHeatExchangerData(HeatExchangerData):
+    CONFIG = HeatExchangerData.CONFIG(implicit=True)
```
andrewlee94, line 154: Do we need implicit declaration here? We generally avoid that except for very specific circumstances (placeholders for other model's configs).

idaes/power_generation/unit_models/boiler_heat_exchanger.py
```diff
@@ -313,13 +304,87 @@ class BoilerHeatExchangerData(UnitModelBlockData):
...
-        "delta_temperature_callback",
-        ConfigValue(
-            default=delta_temperature_underwood_tune_callback,
-            description="Callback for for temperature difference calculations",
-        ),
-    )
+
+    def _process_config(self):
+        """Deal with old style config arguments by converting them to be
+        consitent with the generic heat exchanger.  Log a deprication warning
```
andrewlee94, line 310: Typo: deprecation.

andrewlee94 (commented):
I have one question regarding implicit declaration of config arguments, but otherwise looks good. I found a couple of very minor typos and one comment on the release for when the deprecated features will be removed (although I know we discussed this somewhere else).

idaes/power_generation/unit_models/boiler_heat_exchanger.py
eslickj: I'm not sure when 2 will be, but I figure it has to be a major release, and I want to leave a lot of time before I break anything.  I'm open to other numbers though this will mostly affect @MAZamarripa and Jinliang.  I'll happily put whatever in there.

idaes/power_generation/unit_models/boiler_heat_exchanger.py
eslickj: Yes.  The heat exchanger lets you put the configuration for each side in under whatever you named it.  There are sub config blocks for each side.

andrewlee94 (dismissed):

idaes/generic_models/unit_models/heat_exchanger.py
```diff
@@ -121,7 +121,15 @@ def _make_heat_exchanger_config(config):
...
         ),
     )
-
+    config.declare(
+        "side_1_is_hot",
+        ConfigValue(
+            default=True,
+            domain=bool,
+            doc="""If True, side_1 is an alias for hot_side. Otherwise, side_1
+is alias an alias for cold_side""",
```
dallan-keylogic, line 130: "alias an alias" => "an alias"

dallan-keylogic (approved):

eslickj:
Looks like I broke an example.  I think it is probably due to a change in the model serialization and reading back initialization.  I don't think it will require any changes here, but I may have a chicken and egg problem fixing the example.

PR #679, opened 2022-01-31
Updating CV1D reaction variables to use reaction package units

bpaul4:
## Fixes
When running the spatially discretized CLC unit models with updated units, unit inconsistency issues arise when the reaction package units do not match the units of the associated phase (e.g. solid phase for CLC models). This update ensures reaction generation and extent variables use the correct units defined in the reaction package.

- Updated CV1D reaction variables to use reaction package units
-

### Legal Acknowledgement


andrewlee94 (changes requested):
I do not think this will work as is - it actually introduces more problems rathe than fixing them. The units of measurement of the material balances are based on those of the thermo property package in the control volume, and there is meant to be a check that ensures that the reaction package units are consistent with these. This might be getting skipped however with these heterogeneous reaction packages.

The correct fix is to enforce unit consistency between the reaction and thermo packages.

andrewlee94:
@bpaul You can't just mix mass and mole bases like that without a conversion somewhere, and that appears to be missing right now.  Simply put - you are either changing the wrong part of the control volume code, or trying to do something that just won't work.

The control volume does have some code to try to handle these conversions, but they depend on the packages correctly identifying the flow basis. You should look at the basis conversion in the actual material balances for this.

chineduobiora:
@andrewlee94 The issue I think, is that as things are currently setup reaction_rate_generation in the control volume is defined as the stoichiometric number of the reaction [units of mol_component/mol_extent] multiplied by the rate_reaction_extent. To ensure units consistency, the rate_reaction_extent has to have numerator units of mol_extent. If needed, the _rxn_rate_conv term (introduces molecular weight I think) in the mass balance constraint should then convert from a mole basis to a mass basis.

When we define the units of reaction_rate_generation and reaction_rate based on the units of the associated state block we get the units inconsistency issue.

@bpaul4's fix aims to overcome this, though it should be updated to only apply to units of rate_reaction_extent and reaction_rate_generation.

andrewlee94:
@chineduobiora @bpaul4 OK, I think I am beginning to understand the problem, and I think I see the actual problem. I was slightly confused by the fact you looked up the reaction package units metadata however.

Looking more closely however, it looks like you are using the units from the thermo package in all places that matter - the only place you are actually checking the reaction units is in the check for `if rxn_units('length') is not None:`. So, my main question is why you need to check the reaction package units here? For one, there should be a check elsewhere that ensures the reaction package units are consistent with the thermo package. Secondly, units should never be `None` at this point - I think IDAES has fully deprecated support for skipping unit declarations.

bpaul4:
@andrewlee94 Lines 434 and 438 should be calling `rxn_units('flow_mole')` and `rxn_units('length')`, I will correct that.

@chineduobiora, is there a specific reason to use the reaction package units for rate reaction generation & extent variables but not equilibrium reaction generation & extent variables? Currently I updated the variable units for both reaction types.

chineduobiora:
@bpaul4 Maybe the changes should also apply to the equilibrium reactions (and maybe inherent reactions reactions), but I don't have a good grasp of what the reaction proppacks for those reactions will look like. @andrewlee94 might have a better idea of if the changes are also applicable to those reaction types.

andrewlee94:
@chineduobiora Equilibrium and inherent reactions will need similar updates.

idaes/core/control_volume1d.py
```diff
@@ -426,19 +426,15 @@ def _add_material_balance_common(self,
...
-                else:  # reaction package has no unit container objects
+                if (self.reactions[self.flowsheet().time.first(),
+                                   self.length_domain.first()]
+                        .get_reaction_rate_basis() == MaterialFlowBasis.molar):
+                    rxn_flow_l_units = units('flow_mole')/units('length')
+                elif (self.reactions[self.flowsheet().time.first(),
+                                     self.length_domain.first()]
+                      .get_reaction_rate_basis() == MaterialFlowBasis.mass):
+                    rxn_flow_l_units = units('flow_mass')/units('length')
+                else:  # reaction basis not defined
```
andrewlee94, line 437: Minor comment, this case which should raise an `Exception` of some type, as if the basis is "other" then we don't know what to do and should not try guessing (and `None` is not the right answer).

This test might already exist somewhere else, but it is probably worth adding here anyway - I would probably go with a `PropertyPackageError`.

andrewlee94 (changes requested):
Looking good, but there are two things I think that still need to be done.

1. Most significantly, equivalent changes are probably also needed in the CV0D code, so we should do that now for consistency and completeness.
2. I had one comment about cases where the reaction basis is "other" - we should not try to handle that in any way and just raise an Exception.

idaes/core/control_volume1d.py
bpaul4: Regarding your second point, what is your opinion of adding a similar `Exception` to the property package "other basis" case on line 400?

andrewlee94: It should not be in the property package itself (otherwise there would be no point to having the option) - that option is there to allow for custom cases, but precludes the use of control volumes (hence the check should be in the control volumes).

bpaul4: To clarify, I'm asking if we should add a `PropertyPackageError` to the control volume if the flow basis is not found to be Molar or Mass.

andrewlee94: Now that I look at that, I am not sure - we are basically saying we will accept "other" as a basis and assume you know what you are doing. So, we should do the same here then, and not raise the exception, just set units to `None`.

andrewlee94 (dismissed):
Looks good, although we really should add some tests for this too (but that takes a bit more work to create a case for testing so I won't insist).

EDIT: It looks like the CodeCov patch diff is too low, so you will need to add some new tests then. I tried looking at the CodeCov report, and as usual it is saying I am unauthorized to view it so I can't be sure about what lines you need to cover. However, what I think you will need to do is add new tests for each type of control volume, each with all 3 types of reactions (rate, equilibrium and inherent). You will need to do one test with each reaction basis, and just check that the units of the generation and extent terms are correct.

andrewlee94:
@bpaul4 To give you somewhere to start with the new tests, look at line 309 in https://github.com/IDAES/idaes-pse/blob/main/idaes/core/tests/test_control_volume_0d.py.

This is doing the first part of what you need to do for rate based reactions at least. This shows how you can use the `ReactionParameterTestBlock` to test different reaction bases easily - all you need to do for rate based reactions is to then add the material balances and check the units o the generation term. This should also work for equilibrium reactions, but see below.

Inherent reactions are actually going to be harder - and notably we might need to think this through more. Notably, inherent and equilibrium reactions actually don't have a flow basis like rate based reactions. They just force a relationship between specific species in the system, and the reaction extent and generation are solved implicitly. Basically - the generation and extent term should probably always be in molar basis (regardless of anything else), as we need to apply stoichiometry to the reactions.

chineduobiora (approved):
Looks good to me.

andrewlee94 (approved):

andrewlee94:
@bpaul4 It looks like the CodeCov diff is still too low. See https://app.codecov.io/gh/IDAES/idaes-pse/compare/679/diff for more information. Notably, the two columns on the left show what is covered and missed (the first column is the original, the second column is the current PR). Notably those lines in dark red are the ones that are missed.

Now, for this particular case, we see that the only lines that are missed are those where reaction units are `None`. Whilst we could write additional tests for these, I don't think we actually need to care (they are no longer supported and are just there for backwards compatibility).

So, assuming all the other tests pass and nobody complains, I suggest an administrator merge for this.

PR #680, opened 2022-01-31
Correcting units for CLC methane reduction properties

bpaul4:
## Fixes
When the base unit defaults (mass, time, length, temperature, amount) use standard SI units, any non-standard defaults are overwritten by the `create_derived_units()` method. This updates the CLC methane reduction property and reaction packages to ensure consistent energy and pressure units.

## Summary/Motivation:
- The CLC methane reduction modules require a unit checking scheme internal to property and reaction packages for non-standard pressure and energy default units, and this update originally added this scheme to the gas, solid and reaction packages. However, to maintain consistency with IDAES standards, for now the property packages and any relevant flowsheets will be reverted to use only units derived from the five base standard quantities (m, kg, s, K, mol).

- Update energy units from kJ to J and pressure units from bar to Pa where required in methane reduction files
- Minor text comments regarding molecular weight and heat of reaction unit requirements

### Legal Acknowledgement

### DIFF
--- a/idaes/dmf/tests/test_util.py
+++ b/idaes/dmf/tests/test_util.py
@@ -26,6 +26,7 @@ from typing import Union
 import pytest
 
 #
+from pyomo.environ import value
 from idaes.dmf import util, resource
 from .util import init_logging
 
@@ -101,7 +102,7 @@ def test_datetime_timestamp():
     ts = time.time()
     dt = datetime.datetime.fromtimestamp(ts)
     ts1 = util.datetime_timestamp(dt)
-    assert pytest.approx(ts, ts1, 0.000001)
+    assert pytest.approx(ts, 0.000001) == ts1
 
 
 @pytest.mark.unit
--- a/idaes/generic_models/properties/core/eos/tests/test_ceos_PR.py
+++ b/idaes/generic_models/properties/core/eos/tests/test_ceos_PR.py
@@ -156,6 +156,33 @@ def m():
     # Set a distinct value for _teq so it can be distinguished from temperature
     m.props[1]._teq[("Vap", "Liq")].value = 100
 
+
+
+    m.props[1].energy_internal_mol_phase_comp = Var(
+        m.params.phase_list, m.params.component_list, initialize=1)
+    m.props[1].enth_mol_phase_comp = Var(m.params.phase_list,
+                                         m.params.component_list)
+    m.props[1].vol_mol_phase_comp = Var(m.params.phase_list,
+                                         m.params.component_list)
+    m.props[1].entr_mol_phase_comp = Var(m.params.phase_list,
+                                         m.params.component_list,
+                                         initialize=1)
+    
+    m.props[1].enth_mol_phase = Var(m.params.phase_list)
+    m.props[1].entr_mol_phase = Var(m.params.phase_list)
+    
+    m.props[1].dens_mol_phase = Var(m.params.phase_list)
+    m.props[1].mw_phase = Var(m.params.phase_list)
+    
+
+    for j in m.params.component_list:
+        m.params.config.include_enthalpy_of_formation = False
+        m.params.get_component(j).config.enth_mol_liq_comp = dummy_call
+        m.params.get_component(j).config.enth_mol_ig_comp = dummy_call
+
+        m.params.get_component(j).config.entr_mol_liq_comp = dummy_call
+        m.params.get_component(j).config.entr_mol_ig_comp = dummy_call
+    
     return m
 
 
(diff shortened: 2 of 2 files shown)


andrewlee94 (changes requested):
@bpaul4 This is not going to work - the basis of IDAES is that units of measurement must be internally consistent, so we can't go adding different units into our models. As soon as you do this, you need to add unit conversion statements to **every** term in every expression and constraint in the model, otherwise you are going to run into unit inconsistencies eventually.

Whilst it is restrictive, with the way things are now we have to enforce units derived from the base 5 quantities only.

chineduobiora:
@andrewlee94  I also had my initial doubts when @bpaul4 originally suggested the changes. However, upon taking time to look through things, I realized that the changes are local only to these proppacks and **have no impact on the rest of the codebase** including no code changes to property_meta. Note also that the pressure and energy updates are only implemented if default units of pressure and energy are defined in define_metadata, and things work as normal if those default units aren’t present. I have also reviewed the MB, BFB and FixedBed0D models and confirm that their units have been defined such that units consistency is maintained without the need to add any specific unit conversions to support the proposed changes.

These changes will allow a user of the CLC models to specify units of pressure and energy that work best for their application. From my experience this minimizes the effort to figure out an appropriate scaling factor for the pressure and energy related variables. 

We tout the ability to use IDAES and make local custom changes for niche applications. I think this is one of those cases. We also always say that the burden of units consistency is on the user, these changes don’t add or subtract to the amount of that burden.

andrewlee94:
@chineduobiora Whilst I wouldn't mind so much if this was in project specific models outside the main repository, this is in `idaes-pse` and thus it needs to follow the standards set out for the rest of the code base. For now, that includes the requirement that all units of measurement be derived from the base 5 quantities.

I have opened issue #652 to discuss the possibility of allowing more flexible units, but until that is implemented I am afraid that I will have to say no to this. You can do this with other models that sit outside the main repo, but anything that is committed here must follow the standards to maintain consistency with the rest of the codebase.

chineduobiora:
Sounds good. @bpaul4 let's revert to working with the base units only.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
```diff
@@ -91,22 +92,25 @@ def build(self):
...
+        # Std. heat of formation of comp. - units = J/(mol comp) - ref: NIST
+        enth_mol_form_comp_dict = {'CH4': -74.8731E3, 'CO2': -393.5224E3,
+                                   'H2O': -241.8264E3}
         self.enth_mol_form_comp = Param(
                 self.component_list,
                 mutable=False,
                 initialize=enth_mol_form_comp_dict,
-                doc="Component molar heats of formation [kJ/mol]",
-                units=pyunits.kJ/pyunits.mol)
+                doc="Component molar heats of formation [J/mol]",
```
andrewlee94, line 112: Minor comment, but I generally avoid putting units in doc strings now - it is effectively duplicating the information in the units and adds the possibility of the two getting out of synch.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
```diff
@@ -307,31 +311,29 @@ def define_metadata(cls, obj):
                 'temperature': {'method': None, 'units': 'K'},
```
andrewlee94, line 311: Units are no longer needed here, as they are recorded in the property itself.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/hetero_reactions.py
```diff
@@ -116,12 +116,16 @@ def build(self):
...
+        dh_rxn_dict = {"R1": 136.5843E3}
+        # Heat of Reaction should be defined in default units
+        # (default energy units)/(default amount units)
+        # per the define_meta.add_default_units method in the
+        # relevant phase - in this case, the solid phase properties
         self.dh_rxn = Param(self.rate_reaction_idx,
                             initialize=dh_rxn_dict,
-                            doc="Heat of reaction [kJ/mol]",
-                            units=pyunits.kJ/pyunits.mol)
+                            doc="Heat of reaction [J/mol]",
```
andrewlee94, line 127: As above r.e. units in doc strings.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/hetero_reactions.py
```diff
@@ -116,12 +116,16 @@ def build(self):
                                             ("R1", "Sol", "Fe3O4"): 8,
                                             ("R1", "Sol", "Al2O3"): 0}
 
-        # Standard Heat of Reaction - kJ/mol_rxn
-        dh_rxn_dict = {"R1": 136.5843}
+        # Standard Heat of Reaction - J/mol_rxn
+        dh_rxn_dict = {"R1": 136.5843E3}
```
andrewlee94, line 120: A couple of minor comments:

1. Do we really need to declare this separately from its `Param` declaration below?
2. Moe for information that an actual change, but you can add units here too for clarity (i.e. `136.6*pyunits.J/pyunits.mol`). Pyomo will then automatically convert this value to those of the parameter you are assigning them to if required. This is more for use at the flowsheet level, but still applies here.

andrewlee94 (approved):
Only a couple of minor comments, but nothing that absolutely needs to be changed.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
bpaul4: In the future, if we change these files back to custom units I will remove the doc strings. I left them for now since we are enforcing standard base units everywhere.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/hetero_reactions.py
bpaul4: We don't need to define it this way, but it is consistent with how the Parameters are defined in the gas and solid property packages and is flexible for adding more reaction values if needed.

chineduobiora (approved):
The changes to the property packages look good to me. I'll leave comments relating to the MB and BFB for their respective PRs.

PR #681, opened 2022-01-31
Moving Bed Updates Scaling and Units

bpaul4:
## Fixes
This adds updated scaling and units to the CLC moving bed model. The changes here are dependent upon #679 and #680 , and unit inconsistency test failures will resolve once these two updates are merged.

- New scaling and units definitions for the moving bed model
-

### Legal Acknowledgement

### DIFF
--- a/idaes/gas_solid_contactors/unit_models/moving_bed.py
+++ b/idaes/gas_solid_contactors/unit_models/moving_bed.py
@@ -1,15 +1,15 @@
-###############################################################################
+#################################################################################
 # The Institute for the Design of Advanced Energy Systems Integrated Platform
 # Framework (IDAES IP) was produced under the DOE Institute for the
 # Design of Advanced Energy Systems (IDAES), and is copyright (c) 2018-2021
 # by the software owners: The Regents of the University of California, through
 # Lawrence Berkeley National Laboratory,  National Technology & Engineering
-# Solutions of Sandia, LLC, Carnegie Mellon University,
-# West Virginia University Research Corporation, et al.  All rights reserved.
+# Solutions of Sandia, LLC, Carnegie Mellon University, West Virginia University
+# Research Corporation, et al.  All rights reserved.
 #
 # Please see the files COPYRIGHT.md and LICENSE.md for full copyright and
 # license information.
-###############################################################################
+#################################################################################
 """
 IDAES Moving Bed Model.
 
@@ -38,7 +38,7 @@ import matplotlib.pyplot as plt
 # Import Pyomo libraries
 from pyomo.environ import (Var, Param, Reals, value,
                            TransformationFactory, Constraint,
-                           check_optimal_termination)
+                           TerminationCondition, units as pyunits)
 from pyomo.common.config import ConfigBlock, ConfigValue, In, Bool
 from pyomo.util.calc_var_value import calculate_variable_from_constraint
 from pyomo.dae import ContinuousSet
@@ -60,7 +60,7 @@ from idaes.core.control_volume1d import DistributedVars
 from idaes.core.util.constants import Constants as constants
 from idaes.core.util.math import smooth_abs
 import idaes.logger as idaeslog
-from idaes.core.util import get_solver
+from idaes.core.util import get_solver, scaling as iscale
 
 __author__ = "Chinedu Okoli", "Anca Ostace"
 
@@ -246,7 +246,7 @@ see reaction package for documentation.}"""))
             None
         """
         # Call UnitModel.build to build default attributes
-        super(MBRData, self).build()
+        super().build()
 
         # Consistency check for transformation method and transformation scheme
         if (self.config.transformation_method == "dae.finite_difference" and
@@ -320,6 +320,13 @@ see reaction package for documentation.}"""))
         else:
             has_heat_of_reaction_solid_phase = False
 
+        # local aliases used to shorten object names
+        solid_phase = self.config.solid_phase_config
+
+        # Get units meta data from property packages
+        units_meta_solid = \
+            solid_phase.property_package.get_metadata().get_derived_units
+
         # Create a unit model length domain
         self.length_domain = ContinuousSet(
                                 bounds=(0.0, 1.0),
@@ -327,7 +334,8 @@ see reaction package for documentation.}"""))
                                 doc="Normalized length domain")
 
         self.bed_height = Var(domain=Reals, initialize=1,
-                              doc='Bed length [m]')
+                              doc='Bed length',
+                              units=units_meta_solid('length'))
 
     # =========================================================================
         """ Build Control volume 1D for gas phase and
@@ -508,55 +516,70 @@ see reaction package for documentation.}"""))
         gas_phase = self.config.gas_phase_config
         solid_phase = self.config.solid_phase_config
 
+        # Get units meta data from property packages
+        units_meta_gas = \
+            gas_phase.property_package.get_metadata().get_derived_units
+        units_meta_solid = \
+            solid_phase.property_package.get_metadata().get_derived_units
+
         # Declare Mutable Parameters
         self.eps = Param(mutable=True,
                          default=1e-8,
-                         doc='Smoothing Factor for Smooth IF Statements')
+                         doc='Smoothing Factor for Smooth IF Statements',
+                         units=pyunits.dimensionless)
 
         # Unit Model variables
-        self.bed_diameter = Var(domain=Reals,
-                                initialize=1,
-                                doc='Reactor diameter [m]')
+        self.bed_diameter = Var(initialize=1,
+                                doc='Reactor diameter',
+                                units=units_meta_solid('length'))
         self.bed_area = Var(domain=Reals,
                             initialize=1,
-                            doc='Reactor cross-sectional area [m2]')
+                            doc='Reactor cross-sectional area',
+                            units=units_meta_solid('area'))
 
         # Phase specific variables
         self.velocity_superficial_gas = Var(
                 self.flowsheet().time,
                 self.length_domain,
                 domain=Reals, initialize=0.05,
-                doc='Gas superficial velocity [m/s]')
+                doc='Gas superficial velocity',
+                units=units_meta_gas('velocity'))
         self.velocity_superficial_solid = Var(
                 self.flowsheet().time,
                 domain=Reals, initialize=0.005,
-                doc='Solid superficial velocity [m/s]')
+                doc='Solid superficial velocity',
+                units=units_meta_solid('velocity'))
 
         # Dimensionless numbers, mass and heat transfer coefficients
         self.Re_particle = Var(self.flowsheet().time,
                                self.length_domain,
                                domain=Reals, initialize=1.0,
-                               doc='Particle Reynolds number [-]')
+                               doc='Particle Reynolds number',
+                               units=pyunits.dimensionless)
 
         self.Pr = Var(self.flowsheet().time,
                       self.length_domain,
                       domain=Reals, initialize=1.0,
-                      doc='Prandtl number of gas in bed [-]')
+                      doc='Prandtl number of gas in bed',
+                      units=pyunits.dimensionless)
 
         self.Nu_particle = Var(self.flowsheet().time,
                                self.length_domain,
                                domain=Reals, initialize=1.0,
-                               doc='Particle Nusselt number [-]')
+                               doc='Particle Nusselt number',
+                               units=pyunits.dimensionless)
         self.gas_solid_htc = Var(self.flowsheet().time,
                                  self.length_domain,
                                  domain=Reals, initialize=1.0,
-                                 doc='Gas-solid heat transfer coefficient'
-                                 '[kJ/(m2Ks)]')
+                                 doc='Gas-solid heat transfer coefficient',
+                                 units=units_meta_gas(
+                                     'heat_transfer_coefficient'))
 
         # Fixed variables (these are parameters that can be estimated)
         self.bed_voidage = Var(domain=Reals,
                                initialize=0.4,
-                               doc="Bed voidage [-]")
+                               doc="Bed voidage",
+                               units=pyunits.dimensionless)
         self.bed_voidage.fix()
 
     # =========================================================================
@@ -577,7 +600,9 @@ see reaction package for documentation.}"""))
                          doc="Gas side area")
         def gas_phase_area(b, t, x):
             return (b.gas_phase.area[t, x] ==
-                    b.bed_area*b.bed_voidage)
+                    pyunits.convert(b.bed_area,
+                                    to_units=units_meta_gas('area')) *
+                    b.bed_voidage)
 
         @self.Constraint(self.flowsheet().time,
                          self.length_domain,
@@ -594,7 +619,9 @@ see reaction package for documentation.}"""))
                          self.length_domain,
                          doc="Gas superficial velocity")
         def gas_super_vel(b, t, x):
-            return (b.velocity_superficial_gas[t, x] * b.bed_area *
+            return (b.velocity_superficial_gas[t, x] *
+                    pyunits.convert(b.bed_area,
+                                    to_units=units_meta_gas('area')) *
                     b.gas_phase.properties[t, x].dens_mol ==
                     b.gas_phase.properties[t, x].flow_mol)
 
@@ -620,9 +647,15 @@ see reaction package for documentation.}"""))
                              doc="Gas side pressure drop calculation -"
                                  "simplified pressure drop")
             def gas_phase_config_pressure_drop(b, t, x):
-                return b.gas_phase.deltaP[t, x]*1e5 == -0.2*(
+                #  add 1/pyunits.s to account for missing parameter units
+                return pyunits.convert(
+                    b.gas_phase.deltaP[t, x],
+                    to_units=units_meta_solid('pressure') /
+                    units_meta_solid('length')) == -0.2/pyunits.s*(
                         b.velocity_superficial_gas[t, x] *
-                        (b.solid_phase.properties[t, x].dens_mass_particle -
+                        (pyunits.convert(
+                            b.solid_phase.properties[t, x].dens_mass_particle,
+                            to_units=units_meta_solid('density_mass')) -
                          b.gas_phase.properties[t, x].dens_mass))
         elif (self.config.has_pressure_change and
               self.config.pressure_drop_type == "ergun_correlation"):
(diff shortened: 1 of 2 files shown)


idaes/gas_solid_contactors/unit_models/moving_bed.py
```diff
@@ -38,7 +38,7 @@
 # Import Pyomo libraries
 from pyomo.environ import (Var, Param, Reals, value,
                            TransformationFactory, Constraint,
-                           check_optimal_termination)
+                           TerminationCondition, units as pyunits)
```
andrewlee94, line 41: @bpaul4 This change needs to be reverted-  `check_optimal_termination` is the correct way to check solver convergence as it is less solver specific.

idaes/gas_solid_contactors/unit_models/moving_bed.py
```diff
@@ -508,55 +516,70 @@ def _make_performance(self):
...
+            gas_phase.property_package.get_metadata().get_derived_units
+        units_meta_solid = \
+            solid_phase.property_package.get_metadata().get_derived_units
+
         # Declare Mutable Parameters
         self.eps = Param(mutable=True,
                          default=1e-8,
-                         doc='Smoothing Factor for Smooth IF Statements')
+                         doc='Smoothing Factor for Smooth IF Statements',
+                         units=pyunits.dimensionless)
```
andrewlee94, line 529: Be careful here - epsilon needs to have the same units as the quantity it is scaling.

idaes/gas_solid_contactors/unit_models/moving_bed.py
```diff
@@ -620,9 +647,15 @@ def solid_super_vel(b, t, x):
                              doc="Gas side pressure drop calculation -"
                                  "simplified pressure drop")
             def gas_phase_config_pressure_drop(b, t, x):
-                return b.gas_phase.deltaP[t, x]*1e5 == -0.2*(
+                #  add 1/pyunits.s to account for missing parameter units
+                return pyunits.convert(
+                    b.gas_phase.deltaP[t, x],
+                    to_units=units_meta_solid('pressure') /
+                    units_meta_solid('length')) == -0.2/pyunits.s*(
```
andrewlee94, line 654: What is the `-0.2` in this case? Is it a unitted constant (i.e. it has units of `1/s`?

idaes/gas_solid_contactors/unit_models/moving_bed.py
```diff
@@ -632,20 +665,33 @@ def gas_phase_config_pressure_drop(b, t, x):
...
-                return (1e2*-b.gas_phase.deltaP[t, x]*1e5 ==
-                        1e2*(
-                        150*(1 - b.bed_voidage) ** 2 *
+                #  add 1/pyunits.s to account for missing parameter units
+                return (-pyunits.convert(
+                    b.gas_phase.deltaP[t, x],
+                    to_units=units_meta_solid('pressure') /
+                    units_meta_solid('length')) ==
+                        (
+                        150/pyunits.s*(1 - b.bed_voidage) ** 2 *
```
andrewlee94, line 674: As above, where does this equation come from?

idaes/gas_solid_contactors/unit_models/moving_bed.py
```diff
@@ -632,20 +665,33 @@ def gas_phase_config_pressure_drop(b, t, x):
                              doc="Gas side pressure drop calculation -"
                                  "ergun equation")
```
andrewlee94, line 666: Typo: Ergun

idaes/gas_solid_contactors/unit_models/moving_bed.py
```diff
@@ -1177,15 +1246,39 @@ def initialize(blk, gas_phase_state_args=None, solid_phase_state_args=None,
...
                             )
             else:
-                _log.warning('{} Initialisation Step 4 Failed.'
+                _log.warning('{} Initialization Step 4 Failed.'
                              .format(blk.name))
 
+    def calculate_scaling_factors(self):
+        super().calculate_scaling_factors()
+
+        if hasattr(self, "gas_phase_config_pressure_drop"):
```
andrewlee94, line 1262: At a glance, this is not a lot of constraints being scaled. Make sure you scale every constraint added at the unit model level here.

idaes/gas_solid_contactors/unit_models/tests/test_MB.py
```diff
@@ -18,11 +18,13 @@
 
 import pytest
 
-from pyomo.environ import (check_optimal_termination,
-                           ConcreteModel,
+from pyomo.environ import (ConcreteModel,
+                           TerminationCondition,
```
andrewlee94, line 22: As mentioned earlier, you should use `check_optimal_termination`.

idaes/gas_solid_contactors/unit_models/tests/test_MB.py
```diff
@@ -228,7 +230,9 @@ def test_solve(self, iron_oc):
         results = solver.solve(iron_oc)
 
         # Check for optimal solution
-        assert check_optimal_termination(results)
```
andrewlee94, line 231: Revert this change,

idaes/gas_solid_contactors/unit_models/tests/test_MB.py
```diff
@@ -247,6 +251,10 @@ def test_solution(self, iron_oc):
                 iron_oc.fs.unit.gas_inlet.pressure[0] -
                 iron_oc.fs.unit.gas_outlet.pressure[0]) >= 0
```
andrewlee94, line 252: This check looks a little odd - it isn't actually checking for an expected result, just that pressure decreased over the column,

idaes/gas_solid_contactors/unit_models/tests/test_MB.py
```diff
@@ -421,7 +429,9 @@ def test_solve(self, iron_oc):
         results = solver.solve(iron_oc)
 
         # Check for optimal solution
-        assert check_optimal_termination(results)
```
andrewlee94, line 424: As above.

andrewlee94 (changes requested):
A few comments.

idaes/gas_solid_contactors/unit_models/moving_bed.py
bpaul4: Thank you for checking this. This parameter is only used in one place, to smooth the Reynold's number on line 798 and the units match the dimensionless number (which is explicitly given dimensionless units on line 558).

idaes/gas_solid_contactors/unit_models/moving_bed.py
bpaul4: Yes, this is a unitted constant to satisfy unit consistency.

idaes/gas_solid_contactors/unit_models/moving_bed.py
bpaul4: Similar to above.

idaes/gas_solid_contactors/unit_models/moving_bed.py
andrewlee94: I would change the comment to note that 0.2/s is a unitted costant, and proably put the actual parameter in parenthesis for clarity: `(0.2/pyunits.s)`

idaes/gas_solid_contactors/unit_models/moving_bed.py
bpaul4: I originally added constraints under this method which previously had hard-coded factors, keeping @chineduobiora 's scaling factors. Would you recommend explicitly setting remaining constraint scaling factors to 1E0?

andrewlee94: I would recommend all other constraint have methods that determine a good scaling factor - you can specify a default scaling factor of 1 in the absence of more information, but always look for something better.

idaes/gas_solid_contactors/unit_models/tests/test_MB.py
bpaul4: I believe it may be redundant since the outlet pressure is asserted to be a lower value than the fixed inlet pressure. @chineduobiora do you believe we should keep this check?

andrewlee94: We should make this a check for an explicit value.

chineduobiora: @bpaul4 Yes, we should keep it, but update it to check for an explicit value (i.e. the outlet pressure) like @andrewlee94 mentioned.

idaes/gas_solid_contactors/unit_models/moving_bed.py
chineduobiora: My recommendation is to scale the equation based on the scale factors of the variables in the equation i.e.
`f: 	a*b = c*d.`

The constraint, f, can be scaled either with the LHS or the RHS i.e.
`sf1 = iscale.get_scaling_factor(a)`
`sf2 = iscale.get_scaling_factor(b)`
`iscale.constraint_scaling_transform(c, sf1 * sf2, overwrite=False)`

If the variables a and b do not have scale factors then they can be set from the equations they are derived from i.e.
`a = e/f`

`sf1 = iscale.get_scaling_factor(e)`
`sf2 = iscale.get_scaling_factor(f)`
`iscale.set_scaling_factor(v, sf1 * sf2)`

or a factor can be set if they are not derived from any equation i.e.
`iscale.set_scaling_factor(v, 1e1)`

In this way the scale factors for the constraints will end up being mostly derived from the scale factors that have been set in the property packages, and the factors of other geometric variables like the diameter and length of the reactor.

Reach out to me if it’s unclear and we can schedule a meeting to go through an example I have for the BFB.

idaes/gas_solid_contactors/unit_models/tests/test_MB.py
```diff
@@ -307,6 +314,88 @@ def test_conservation(self, iron_oc):
         ebal_tol = ebal_gas + ebal_solid - e_reaction
         assert abs(ebal_tol) <= 1e-2
 
+    @pytest.mark.component
```
chineduobiora, line 317: We will also need a test to check that the model converges with the default scale factors.

chineduobiora (changes requested):
I suggest updating the constraint scaling transformations to be dependent on the scale factors of the variables in the equation. If the variables in the equation don't have scale factors, my recommendation is to set the scale factors of those variables based on the equations they are derived from. In the case where they are not dependent on any other equations, i.e. bed diameter, then we can set a default value (or use the inverse of its fixed value to approximate a scale). In this way the scale factors for the constraints will end up being mostly derived from the scale factors that have been set in the property packages, and the factors of other geometric variables like the diameter and length of the reactor.

I also suggest that we have a test to ensure that the model converges when we use the default scale factors.

bpaul4:
@chineduobiora @andrewlee94 I have updated the tests to cover scaling and initialization updates, and codecov is passing now. When you have a moment, please provide your feedback on the scaling method and tests (I am updating the BFB model in parallel).

chineduobiora (approved):
I made a few changes, most of them are scaling related. The major ones are:
1.     Updated scaling factors in the property package and unit model to ensure that the scaling was robust and the resulting scaled model was well conditioned. 2. Updated tests to include testing for collocation transformation scheme as this is one of the popular transformation schemes used 3. Updated flowsheet example to ensure it converges for both scaled and unscaled simulations.

Everything else looks good. Hopefully I didn't break anything!

idaes/gas_solid_contactors/flowsheets/ss_MB_methane_combustion.py
```diff
@@ -112,14 +112,26 @@ def main():
             'Fe3O4': blk.solid_inlet.mass_frac_comp[0, 'Fe3O4'].value,
             'Al2O3': blk.solid_inlet.mass_frac_comp[0, 'Al2O3'].value}}
 
+    print()
```
andrewlee94, line 115: Should use the IDAES logger instead of print statmenets.

idaes/gas_solid_contactors/flowsheets/ss_MB_methane_combustion.py
```diff
@@ -112,14 +112,26 @@ def main():
...
                        gas_phase_state_args=gas_phase_state_args,
                        solid_phase_state_args=solid_phase_state_args)
 
     t_initialize = time.time()  # Initialization time
 
+    print()
+    print("Solve the model")
     # Create a solver
     solver = get_solver()
+    solver.options = {'tol': 1e-5}
```
andrewlee94, line 134: You need to be careful when doing this as this actually changes the global solver settings (i.e. it can effect other models running after this, such as during testing).

andrewlee94 (commented):
A couple of quick comments.

idaes/gas_solid_contactors/unit_models/moving_bed.py
```diff
@@ -620,33 +652,57 @@ def solid_super_vel(b, t, x):
...
             @self.Constraint(self.flowsheet().time,
                              self.length_domain,
                              doc="Gas side pressure drop calculation -"
-                                 "ergun equation")
+                                 "Ergun equation")
             def gas_phase_config_pressure_drop(b, t, x):
-                return (1e2*-b.gas_phase.deltaP[t, x] ==
-                        1e2*(
-                        150*(1 - b.bed_voidage) ** 2 *
+                #  150/s is a unitted constant in the correlation
```
chineduobiora, line 673: This line can be deleted. The 150 parameter is dimensionless

idaes/gas_solid_contactors/flowsheets/ss_MB_methane_combustion.py
chineduobiora: This is in the example flowsheet not in the unit model.

chineduobiora:
Test failure is from the BFB tests. Since we already have a PR for the BFB should we resolve only the MB comments/tests and then resolve the test failure when we handle the BFB, or resolve things in the PR and risk having conflicts when we try merging the BFB PR after its review? Is there another alternative to pursue?

idaes/gas_solid_contactors/flowsheets/ss_MB_methane_combustion.py
andrewlee94: There is a context manager somewhere to do this - the quickest way is probably to ask @eslickj.

andrewlee94:
@chineduobiora We cannot merge this with a failing test, so something has to be done to fix it.

idaes/gas_solid_contactors/unit_models/moving_bed.py
chineduobiora: Removed.

idaes/gas_solid_contactors/flowsheets/ss_MB_methane_combustion.py
eslickj: I actually think this is okay.  The line in question only sets options on that specific solver object.

chineduobiora: Thanks @eslickj

idaes/gas_solid_contactors/unit_models/tests/test_BFB.py
```diff
@@ -451,19 +453,19 @@ def test_solve(self, iron_oc):
     def test_solution(self, iron_oc):
         assert (pytest.approx(0.44, abs=1e-2) ==
                 iron_oc.fs.unit.velocity_superficial_gas[0, 0].value)
-        assert (pytest.approx(1.05, abs=1e-2) ==
+        assert (pytest.approx(1.05, abs=1e-1) ==
```
andrewlee94, line 456: And absolute tolerance of 1e-1 leaves a lot of room for error. If these are things you are actually solving for, then you should be checking to slight larger than solver tolerance (e.g. `rel=1e-5`).

If these are fixed inputs, then the value should be exact in most cases (or at least to machine precision).

chineduobiora: We will make the updates in the BFB PR. The changes here were just to make sure that the BFB tests were not failing and holding up the PR.

andrewlee94: OK, these were the changes you were referring to then.

andrewlee94 (approved):
Assuming @chineduobiora is happy with the changes I think this is close to ready to merge. However, the test coverage appears to be low, so I would ask that you try to improve that a bit first.

idaes/gas_solid_contactors/unit_models/tests/test_BFB.py
chineduobiora: Not this exactly. The model initialization and solve tests for the BFB were failing. Since we had a PR to specifically address the BFB I didn't want to spend too much time on it.

chineduobiora (dismissed):
Looks good to me.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
```diff
@@ -962,17 +962,17 @@ def entropy_correlation(b):
...
             return (
                 (self.entr_mol + r_gas * log(p/1e5)) * b.flow_mol ==
                 sum(b.flow_mol*b.mole_frac_comp[j] * (
-                    self.params.cp_mol_ig_comp_coeff_A[j]*log(t) +
-                    self.params.cp_mol_ig_comp_coeff_B[j]*t +
-                    self.params.cp_mol_ig_comp_coeff_C[j]*t**2 / 2 +
-                    self.params.cp_mol_ig_comp_coeff_D[j]*t**3 / 3 -
-                    self.params.cp_mol_ig_comp_coeff_E[j]/t**2 / 2 +
-                    self.params.cp_mol_ig_comp_coeff_G[j] +
+                    self.params.cp_param_1[j]*log(t) +
```
andrewlee94, line 970: This looks like there would be a unit consistency issue. From what I can tell, `t` has units of `kK`, which means that the `log()` term here will have a unit consistency failure.

1. In these cases, you just need to do `log(t/pyunits.kK)`.
2. The fact that we don't see a test failure implies that we do not have a test for unit consistency. Is this an oversight, or are we not quite ready for unit consistency tests overall?

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
```diff
@@ -312,7 +312,7 @@ def build(self):
         self.set_default_scaling("cp_mass", 1e-6)
         self.set_default_scaling("entr_mol", 1e-2)
         self.set_default_scaling("entr_mol_phase", 1e-2)
-        self.set_default_scaling("dens_mol", 1)
+        # self.set_default_scaling("dens_mol", 1)  # defined later
```
andrewlee94, line 315: Do we need this commented line?

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/tests/test_CLC_gas_prop.py
```diff
@@ -66,13 +66,145 @@ def test_build_inlet_state_block(gas_prop):
...
+    assert isinstance(gas_prop.fs.unit.entr_mol, Var)
 
 
 @pytest.mark.unit
 def test_setInputs_state_block(gas_prop):
     assert degrees_of_freedom(gas_prop.fs.unit) == 0
 
 
+@pytest.fixture(scope="class")
+def gas_prop_unscaled(gas_prop):
```
andrewlee94, line 79: Rather than copy the model, I would suggest you just rebuild the model for each case. It does not cost that much to build a model, and copying the model runs the risk of test failures due to something going wrong during copying.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/tests/test_CLC_solid_prop.py
```diff
@@ -71,6 +70,102 @@ def test_setInputs_state_block(solid_prop):
     assert degrees_of_freedom(solid_prop.fs.unit) == 0
 
 
+@pytest.fixture(scope="class")
+def solid_prop_unscaled(solid_prop):
```
andrewlee94, line 74: As above.

andrewlee94 (changes requested):

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/tests/test_CLC_gas_prop.py
bpaul4: Okay, I will update the fixtures. Would you recommend rebuilding the model rather than copying for the MB and BFB tests as well?

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
bpaul4: We don't need this since we define the scaling in the scaling factor method, I'll remove this line.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
chineduobiora: @bpaul4 similar to the challenges we had with the scaling tests, it is possible that the property constraints have not been constructed therefore the units consistency check might not be testing the constraints. If that's the case, then the property constraints will have to be constructed like was done in the scaling tests before calling the assert_units_consistent method.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/gas_phase_thermo.py
chineduobiora: @bpaul4 that line can be deleted.

idaes/gas_solid_contactors/properties/methane_iron_OC_reduction/tests/test_CLC_solid_prop.py
```diff
@@ -203,4 +216,25 @@ def test_solution(solid_prop):
 
 @pytest.mark.component
 def test_units_consistent(solid_prop):
+
+    # Construct property methods to build the constraints
+
+    assert hasattr(solid_prop.fs.unit, "flow_mass")
+    assert hasattr(solid_prop.fs.unit, "particle_porosity")
+    assert hasattr(solid_prop.fs.unit, "temperature")
+    assert hasattr(solid_prop.fs.unit, "mass_frac_comp")
```
chineduobiora, line 225: These can be removed as they are state variables, so there are not property methods associated for them.

chineduobiora (approved):
Looks good. All my comments have been addressed. I think this is now ready to be merged.

andrewlee94 (approved):

PR #683, opened 2022-01-31
Make IDAES compatible with Click 8

lbianchi-lbl:
## Summary/Motivation:
- Click 8 is a dependency of Black, which has recently released its first stable release
- As many of IDAES's dependencies depend on Black, this effectively results in IDAES having to support Click 8 as well
- The Click requirement for IDAES was fixed to `<8` because of an unspecified error occurring (at the time) with Click 8

### Legal Acknowledgement

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ kwargs = dict(
         # idaes core / dmf
         "backports.shutil_get_terminal_size",
         "bunch",
-        "click<=7.1.2", # problems with 8.x
+        "click", # problems with 8.x
         "colorama",
         "flask",  # for ui/fsvis
         "flask-cors",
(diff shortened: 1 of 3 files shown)


andrewlee94 (dismissed):
LGTM

setup.py
```diff
@@ -49,7 +49,7 @@ def rglob(path, glob):
         # idaes core / dmf
         "backports.shutil_get_terminal_size",
         "bunch",
-        "click<=7.1.2", # problems with 8.x
+        "click", # problems with 8.x
```
andrewlee94, line 52: If this works, we should remove the comment.

andrewlee94 (approved):
LGTM

eslickj (approved):

andrewlee94 (approved):

PR #685, opened 2022-02-01
Power plant costing update

AlexNoring:
## Summary/Motivation:
This brings in the updates made to the power plant costing library from the NGCC branch.

- The addition of functions to build constraints for fixed and variable O&M costs
- The inclusion of custom dictionaries within the power_plant_costing.py file to make it easier to add new cost accounts
- The custom dictionaries have been used to add an account for CO2 capture equipment at a capture rate above 90%

### Legal Acknowledgement


andrewlee94:
At the risk of creating havoc, should we hold off on this until the new costing framework proposal has been settled and implemented? Otherwise, we are going to need to rework this once that is done.

AlexNoring:
@eslickj what are your thoughts on that? Could you push the NGCC model without costing until the new costing framework is settled?

eslickj:
I think we should go ahead with this, but if you guys want to hold off, I can finish the NGCC model release without costing.  @jghouse88 and @MAZamarripa, are you guys using costing or just the NGCC model?  At this point we can't delay too long.   I think I have to get the NGCC out by the end of March and at the rate things happen I won't make it.

andrewlee94:
I will note that the new costing proposal is moving along fairly quickly, and we might have a prototype ready in the next few weeks.

eslickj:
Ok, let's put a pin in this.  We'll see where we are in March.  If a new interface is ready, we'll use that.  If not we'll go with this.

andrewlee94:
@eslickj I have no objection, but we do need to be aware that this will need to be updated before IDAES 2.0 as the current costing framework will be deprecated. That said, the changes should not be that significant (based on converting the general costing package).

eslickj (approved):

MAZamarripa (approved):
lgtm

PR #689, opened 2022-02-03
 Full Docs Diataxis Re-Org

FiannaOBrien:
## Summary/Motivation:

Full Docs Diataxis Re-Org: Restructure docs to Diataxis framework by moving all docs pages into a diataxis group.

* Move all pages into "Tutorials", "How-To Guides", "Explanations", "Reference Guides"
* Create new index pages for Diataxis groups
* Remove unnecessary index pages for removed directories
* Update all internal page references to point to new locations or to similar pages (see change above)

### Legal Acknowledgement


lbianchi-lbl:
@dangunter and I should look at the test(s) that check the Sphinx build to make sure that they're being exercised correctly (broken references/links, etc).

andrewlee94 (commented):
There are too many files to go through by hand and check them all, but the ones I've looked over so far look fine and a quick check of the build for this PR on the read-the-docs also looked good.

One thing I did notice in passing was that we have a couple of links in our readme.md that pint to different parts of the docs (notably there are links at the end of the Getting Started and Contributing sections which point to specific pages). I did a quick search, and it looks like the readme.md file was not updated, so we should update those links (as I am pretty certain they are broken now).

docs/index.rst
```diff
@@ -14,52 +14,48 @@ of the full range of advanced fossil energy systems, including chemical looping
...
-    technical_specs/index
-    related_packages/index
-    faq
-    license
-    copyright
+
+    tutorials/index
+    how_to_guides/index
+    explanations/index
+    reference_guides/index
```
Robbybp, line 57: To clarify a comment I made during a recent meeting, I'm wondering if "About" or "About IDAES" might be a better label here than "Explanations." I think "Explanations" makes sense to somebody already familiar with the Diataxis framework, but I'm not sure its meaning is intuitive to a layperson in the same way that "Turorials," "How-to guides," and "Reference guides" are.

dangunter (approved):
LGTM

ksbeattie (approved):
Damn the re-org, full speed ahead!

FiannaOBrien:
> There are too many files to go through by hand and check them all, but the ones I've looked over so far look fine and a quick check of the build for this PR on the read-the-docs also looked good.
> 
> One thing I did notice in passing was that we have a couple of links in our readme.md that pint to different parts of the docs (notably there are links at the end of the Getting Started and Contributing sections which point to specific pages). I did a quick search, and it looks like the readme.md file was not updated, so we should update those links (as I am pretty certain they are broken now).

@andrewlee94 Good catch! Thank you!
After discussing it with the team, we're going to update the readme.md after the changes have been released, so we can verify the branches are accurate (since they're outside the sphinx reference space) & that we can keep the links pointing to stable rather than latest.

PR #693, opened 2022-02-04
Insert value() method to resolve Boolean assertion error

bpaul4:
## Fixes
- Adds value() method to resolve failing assertion tests

-
-

### Legal Acknowledgement

### DIFF
--- a/idaes/dmf/tests/test_util.py
+++ b/idaes/dmf/tests/test_util.py
@@ -26,6 +26,7 @@ from typing import Union
 import pytest
 
 #
+from pyomo.environ import value
 from idaes.dmf import util, resource
 from .util import init_logging
 
@@ -101,7 +102,7 @@ def test_datetime_timestamp():
     ts = time.time()
     dt = datetime.datetime.fromtimestamp(ts)
     ts1 = util.datetime_timestamp(dt)
-    assert pytest.approx(ts, ts1, 0.000001)
+    assert pytest.approx(value(ts), value(ts1), 0.000001)
 
 
 @pytest.mark.unit
--- a/idaes/dmf/tests/test_util.py
+++ b/idaes/dmf/tests/test_util.py
@@ -102,7 +102,7 @@ def test_datetime_timestamp():
     ts = time.time()
     dt = datetime.datetime.fromtimestamp(ts)
     ts1 = util.datetime_timestamp(dt)
-    assert pytest.approx(value(ts), value(ts1), 0.000001)
+    assert pytest.approx(ts, 0.000001) == ts1
 
 
 @pytest.mark.unit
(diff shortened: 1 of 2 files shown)


andrewlee94 (dismissed):

dallan-keylogic (dismissed):

andrewlee94:
@bpaul4 It looks like there is a second issue here as well. It looks like these two cases are using `pytest.approx(a == b)` or `pytest.approx(a, b)`.

These should be replaced with `a == pytest.approx(b)` (with `value()` where appropriate).

bpaul4:
The first test is straightforward and passes:

```
@pytest.mark.unit
def test_datetime_timestamp():
    ts = time.time()
    dt = datetime.datetime.fromtimestamp(ts)
    ts1 = util.datetime_timestamp(dt)
    assert value(ts1) == pytest.approx(value(ts), 0.000001)
```

The second test fails (is this the correct way to update this?):

```
for i in m.params.component_list:
        for j in m.params.component_list:
            assert value(m.props[1].PR_daij_dT[i,j]) == \
                                 pytest.approx(value((1-m.params.PR_kappa[i, j])
                                 * (m.props[1].PR_fw[j] 
                                 * sqrt(m.props[1].PR_a[i]
                                     *  m.params.get_component(j).temperature_crit
                                     / m.params.get_component(j).pressure_crit)
                                  + m.props[1].PR_fw[i]
                                  * sqrt(m.props[1].PR_a[j]
                                      * m.params.get_component(i).temperature_crit
                                      / m.params.get_component(i).pressure_crit))))
```

`FAILED test_ceos_PR.py::test_common - AssertionError: assert -0.005903679876332056 == 0.03637534575014918 ± 3.6e-08`

andrewlee94:
@bpaul4 When using `pytest.approx` you should provide a tolerance as well for IDAES I generally use `rel=1e-5` (i.e. 1 order of magnitude greater than solve tolerance). You should do this for both tests (the first one passes with the default tolerance, but I generally prefer to be explicit).

bpaul4:
@andrewlee94 the second test passes when the tolerance is set to `rel=1E1` and this seems a bit large to me. Should I go ahead and push this change?

bpaul4:
Yes. In the failure message above, the values seem very different from each other and now I'm suspecting something changed in the test module itself. We are comparing the same two quantities as the original test before the Boolean error occurred.

andrewlee94:
If there is a difference in the values, then this problem goes deeper (and you will need to work out why).

bpaul4:
The syntax seems correct and other test checks pass in the module using the same variables, so I believe this may be due to the PR_dai_dT equations inside the ceos module. @dallan-keylogic , would you have time to take a closer look at these equations and the chain rule expressions?

andrewlee94:
@bpaul4 @dallan-keylogic As a note, it is not he expression that is the problem, it is a question of how it is being initialized/solved and why it is not getting the expected answer.

dallan-keylogic (approved):
I figured out what the problem was. I had incompletely cut and paste some stuff from Radhakrishna's test, and I suppose my incorrect call to `approx`  hid the issue before now. At first I thought it was being caused by side effects to the fixture because I was looking at the wrong test, so I also moved all code that modified the fixture `m` into the code defining that fixture.

lbianchi-lbl (approved):
I don't have enough domain knowledge to comment on the specific model and/or expression, but the changes look OK to me and they make pytest happy.

PR #699, opened 2022-02-08
Adding a toggle to hide the stream names and show them

elbashandy:
## Summary/Motivation:
Adds Toggle to Streams' names in the flowsheet canvas

Screen Shot 2022-02-09 at 11 34 08 AM
Screen Shot 2022-02-09 at 11 34 15 AM

-
-

### Legal Acknowledgement

### DIFF
--- a/idaes/ui/fsvis/static/css/main.css
+++ b/idaes/ui/fsvis/static/css/main.css
@@ -54,7 +54,7 @@ hr {
   position: absolute;
   top: 20px;
   right: 40px;
-  width: 350px;
+  width: 500px;
   height: 40px;
   z-index: 1;
 }
--- a/idaes/ui/fsvis/static/js/toolbar.js
+++ b/idaes/ui/fsvis/static/js/toolbar.js
@@ -155,6 +155,32 @@ export class Toolbar {
             };
         });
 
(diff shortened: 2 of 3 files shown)


sspoon:
Functionality wise, I think this is great! I wonder if, since we have some thoughts about hiding the streams themselves, if we should label this toggle "Stream Names". I know it's not as clean as just Streams. We might as a separate task do some reorg of that menu, especially once we are able to show and hide it.

dangunter:
@sspoon @elbashandy What about putting Show Stream Names and Show Unit Labels in the View menu instead, so the only toggle left standing would be "Grid"? I guess that could go too though for old times' sake I like it as toggle

sspoon:
I think that would be ok. In general, all the things we put in the menu will be less easily discovered, vs controls directly on screen. But our list of toggles is getting rather large and potentially growing ... putting them in the menu could be a decent compromise.

ksbeattie (approved):
The functionality of this looks good to me (without any opinion on the title of the button or moving that into a menu item as discussed by @dangunter & @sspoon)

dangunter (approved):
LGTM

PR #700, opened 2022-02-08
Adding a toggle feature to turn off the stream brushing

elbashandy:
## Summary/Motivation:
Adds the capability to turn off the stream brushing/highlighting feature

Screen Shot 2022-02-09 at 11 33 04 AM

-
-

### Legal Acknowledgement

### DIFF
--- a/idaes/ui/fsvis/static/css/bootstrap/bootstrap.css
+++ b/idaes/ui/fsvis/static/css/bootstrap/bootstrap.css
@@ -3207,7 +3207,7 @@ input[type="button"].btn-block {
   z-index: 1000;
   display: none;
   float: left;
-  min-width: 10rem;
+  min-width: 12rem;
   padding: 0.5rem 0;
   margin: 0.125rem 0 0;
   font-size: 1rem;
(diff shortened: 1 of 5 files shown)


dangunter:
Nitpick:
I would have the menu items spelled like this:
`Stream table`
`Stream highights`

dangunter:
Actually thinking about it more, the 3rd item is different. Can you do a divider in the list? e.g.

**View**
  ✔️ Flowsheet
  ✔️ Stream table
   . . . . . . . . . . . . . . . . . . . .
   ✔️ Highlight streams

sspoon:
So I like Dan's suggestion, but let's make the whole thing verb forward. Also, Dan and I discussed this, and we agreed that title case is more common for menus (although Dan think it's wrong). So, let's go with:

**View** 
✔️ Show Flowsheet 
✔️ Show Stream Table 
. . . . . . . . . . . . . . . . . . . . 
✔️ Highlight Streams

ksbeattie:
I like @sspoon suggestion.  Other than that I'd approve this PR.

But wait, there's more....  try this:
- Uncheck all 3 in the View menu
- Hit refresh
- re-check all 3 in the View menu

The flowsheet is kinda gone, but comes back if you hit the zoom-to-fit button
And the Stream Table is all squished up.

(I know, I know - you _love_ me!)  I don't think this needs to be fixed in this PR, but this should probably be moved to a new issue or fixed in a new PR.

dangunter (approved):
LGTM

PR #704, opened 2022-02-10
Switch cubic EOS extrernal functions to generic cubic root, and move some expression back to pyomo 

### LINKED ISSUE #701: Cubic function root finder and Cubic EoS
I want to take the generic Cubic EoS stuff out of the cubic root calculator function.  This will make the external function simpler and the whole thing more flexible if we want to add a new type of cubic EoS.

Right now how it works is the external function take eos type (Peng-Robinson, SRK, or Van der Waals) and A and B.  Then based on the EoS type, A and B and calculates the roots.  What I want to do is just provide a function that returns the answers to  0 = z^3 + bz^2 + cz +d given b, c, and d.  User's won't know the difference and it's probably a (mostly) a simple find and replace to update the IDAES code.  I feel like nothing should be in an external C function without good reason, and this chunk of the cubic EoS doesn't really have a good reason for being there.

Any thoughts?  I can probably handle this update, if everyone is okay with it.

eslickj:
## Fixes #701 

This makes the change from issue #701.  I found two cubic equations of state, and they had some common stuff so I tried to centralize a bit.  This changes nothing from a user perspective, but it gives us a little more cubic formulation flexibility.   I'm going to rework the external functions to get the junk out, but I'll do that later. 

@andrewlee94, I probably need a little input on your style preferences, since you own this stuff I think.  

### Legal Acknowledgement


andrewlee94 (approved):
Looks good. I do wonder if we need to have some separate tests for ceos_common and some basic documentation though.

agarciadiego (approved):
LGTM

eslickj:
@andrewlee94, I'll write some tests shortly along with the second part of this.  I'm going rework the external functions.  At least for now they are tested at a higher level.  I'll add docs then too.  It's way to low level most people to care about, so it should be okay for a week or so.

PR #706, opened 2022-02-11
B-Truncated Abbort Virial Equation of State

PTA0001:
This adds thermodynamically consistent methods for computing gas properties
using the B-truncated volume explicit virial equations of state with validity
in the low to medium pressure range.

- 
-

### Legal Acknowledgement

### DIFF
--- a/idaes/generic_models/properties/core/eos/virial.py
+++ b/idaes/generic_models/properties/core/eos/virial.py
@@ -16,17 +16,17 @@ Methods for virial equations of state (veos) for gases
 Currently supports B-truncated volume explicit virial equation of state with
 validity at low to moderate pressures where the compresiibility factor is
 approximately a linear function of  pressure. It is most accurate for
- non-polar species.
+non-polar species.
 
-Refernces
+References
 [1] J. M Prausnitz, R, N Lichtenthaler, and E. G de Azvedo,
-    Molecular thermodynamics of Fluid-phase Equilibruim 3rd ed.
-    Prentice-Hall, Nj, 1998
+Molecular thermodynamics of Fluid-phase Equilibruim 3rd ed.
+Prentice-Hall, Nj, 1998
 [2] Any standard chemical engineering thermodynamic textbook
 
 Author: Akula Paul
-
 """
+
 
 from copy import deepcopy
 from pyomo.environ import (Expression,
@@ -58,8 +58,9 @@ VirialConfig.declare("use_pseudocritical_rules", ConfigValue(
 
 
 class entr_mol_ig_comp_G_H_ref(entr_mol_ig_comp):
-    # This updates the build_parameter method to avoid setting S_ref
-    # So that S_sef is calculated from H_ref and G_ref.
+    """This updates the build_parameter method to avoid setting S_ref
+    So that S_sef is calculated from H_ref and G_ref.
+    """
     @staticmethod
     def build_parameters(cobj):
         if not hasattr(cobj, "cp_mol_ig_comp_coeff_A"):
@@ -74,11 +75,25 @@ class entr_mol_ig_comp_G_H_ref(entr_mol_ig_comp):
 
 class Virial(EoSBase):
 
+    """Methods for virial equations of state (veos) for gases"""
     @staticmethod
     def common(b, pobj):
+        """Common objects
 
+        Args:
+            b : Current block
+            pobj (block): Phase object
+        """
         # Add expressions for pseudocritical parameters
         def func_p_omega(m):
+            """Pseudocritical accentric factor
+
+            Args:
+                m (block): Phase object
+
+            Returns:
+                Expression: Pseudocritical parameter
+            """
             return sum(m.mole_frac_phase_comp['Vap', i] *
                        m.params.get_component(i).omega
                        for i in m.components_in_phase('Vap'))
@@ -88,6 +103,14 @@ class Virial(EoSBase):
                                    doc='Pseudocritical omega for gas mixture'))
 
         def func_p_Tc(m):
+            """Pseudocritical temperature
+
+            Args:
+               m (block): Phase object
+
+            Returns:
+                Expression: Pseudocritical parameter
+            """
             return sum(m.mole_frac_phase_comp['Vap', i] *
                        m.params.get_component(i).temperature_crit
                        for i in m.components_in_phase('Vap'))
@@ -97,6 +120,14 @@ class Virial(EoSBase):
                                    doc='Pseudocritical temperature for gas mixture'))
 
         def func_p_Pc(m):
+            """Pseudocritical Pressure
+
+            Args:
+               m (block): Phase object
+
+            Returns:
+                Expression: Pseudocritical parameter
+            """
             return sum(m.mole_frac_phase_comp['Vap', i] *
                        m.params.get_component(i).pressure_crit
                        for i in m.components_in_phase('Vap'))
@@ -115,6 +146,16 @@ class Virial(EoSBase):
 
         # Add combining rules proposed by Prausnitz et al. [1]
         def rule_omega_ij(m, i, j):
+            """Accentric factor i-j molecular pair
+
+            Args:
+               m (block): Phase object
+               i : Component
+               j : Component
+
+            Returns:
+                Expression: Interaction parameter
+            """
             return 0.5 * (m.params.get_component(i).omega +
                           m.params.get_component(j).omega)
         b.add_component('omega_ij',
@@ -124,6 +165,16 @@ class Virial(EoSBase):
                                    doc='Omega combining rule for i-j molecular pair'))
 
         def rule_Zc_ij(m, i, j):
+            """Critical compressibility factor i-j molecular pair
+
+            Args:
+               m (block): Phase object
+               i : Component
+               j : Component
+
+            Returns:
+                Expression: Interaction parameter
+            """
             return 0.5 * (m.params.get_component(i).compress_fact_crit +
                           m.params.get_component(j).compress_fact_crit)
         b.add_component('Zc_ij',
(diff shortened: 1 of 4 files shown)


idaes/generic_models/properties/core/eos/virial.py
```diff
@@ -0,0 +1,797 @@
...
+    composition and critical properties of pure components:
+**default** - True.
+**Valid values:** {
+**True** - use pseudocritical_rule: mixing rule based on composition and
+critical properties of pure components.
+**False** - use mixing rule based on composition and corresponding
+pure component properties.}'''))
+
+
+class enth_mol_ig_comp():
```
andrewlee94, line 60: If there is a bug in a separate module, you should be fixing it there, not recreating it. Please open a separate PR for this.

idaes/generic_models/properties/core/eos/virial.py
```diff
@@ -0,0 +1,797 @@
...
+
+        b.add_component('pseudo_Tc',
+                        Expression(rule=func_p_Tc,
+                                   doc='Pseudocritical temperature for gas mixture'))
+
+        def func_p_Pc(m):
+            return rule_pseudocritical_pressure(m)
+
+        b.add_component('pseudo_Pc',
+                        Expression(rule=func_p_Pc,
```
andrewlee94, line 134: As above.

idaes/generic_models/properties/core/eos/virial.py
```diff
@@ -0,0 +1,797 @@
...
+    @staticmethod
+    def calculate_scaling_factors(b, pobj):
+        pass
+
+    @staticmethod
+    def build_parameters(b):
+        param_block = b.parent_block()
+        # virial eos supports only vapor phase
+        if not b.is_vapor_phase():
+            raise PropertyNotSupportedError(f"{b.parent_block().name} "
```
andrewlee94, line 216: Why use the parent block here, and not the current block?

idaes/generic_models/properties/core/eos/virial.py
```diff
@@ -0,0 +1,797 @@
...
+                    deepcopy(VirialConfig))
+            ConfigBlock = getattr(b.parent_block(), "virial_eos_options")
+            ConfigBlock.set_value(b.config.equation_of_state_options)
+
+        # Get critical volume of components
+        units = b.parent_block().get_metadata().derived_units
+        for comp in b.parent_block().component_list:
+            cobj = b.parent_block().get_component(comp)
+            if not hasattr(cobj, 'volume_crit'):
+                setattr(cobj, "volume_crit",
```
andrewlee94, line 243: This is not the correct way to do this. `volume_crit` should be created by the component if the data is available - i.e this should already exist, or the user forgot to provide the data. In the second case, we should be raising an Exception, rather than trying to create the Var (as the data obviously doesn't exist). If the code does not yet exist to automatically build `volume_crit`, then that should be done elsewhere.

idaes/generic_models/properties/core/eos/virial.py
```diff
@@ -0,0 +1,797 @@
...
+                    within=NonNegativeReals,
+                    initialize=0.0,
+                    doc='Binary interaction parameters',
+                    units=pyunits.dimensionless))
+
+        # Since standard property changes of formation are readily available for
+        # Enthalpy(H_ref) and Gibbs free energy (G_ref),
+        # This builds G_ref and set its value from config block, and  then
+        # computes S_ref = (H_ref-G_ref)/T_ref
+        for comp in b.parent_block().component_list:
```
andrewlee94, line 277: Again, this is not the right place for this. The `component` modules are responsible for doing this, not the EoS.

idaes/generic_models/properties/core/eos/virial.py
```diff
@@ -0,0 +1,797 @@
...
+                cobj.entr_mol_form_vap_comp_ref.set_value(value(S_ref))
+            else:
+                cobj.entr_mol_form_vap_comp_ref.set_value(0.0)
+                cobj.gibbs_mol_form_vap_comp_ref.set_value(0.0)
+
+    # -------------------------------------------------------------------------
+    # METHODS FOR PURE SPECIES
+
+    @staticmethod
+    def compress_fact_vap_comp_pure(b, comp):
```
andrewlee94, line 300: How are these methods intended to be used? They are not part of the build-on-demand system, so they will not be called that way, so this suggests that other methods below will be calling them directly. Some explanation of this would be useful here.

Further, why not just build these as part of the `common` method above?

andrewlee94 (changes requested):
@PTA0001 I am not sure when we will have time to review this thoroughly, but I had a quick look over and had a number of comments on implementation to start with.

You are also going to need to add some documentation for this as well as part of this PR.

andrewlee94:
@PTA0001 You also have some test failures you will need to address.

idaes/generic_models/properties/core/eos/virial.py
```diff
@@ -0,0 +1,797 @@
...
+            (cobj.cp_mol_ig_comp_coeff_D / 4) * (T**4 - Tref**4) +
+            (cobj.cp_mol_ig_comp_coeff_C / 3) * (T**3 - Tref**3) +
+            (cobj.cp_mol_ig_comp_coeff_B / 2) * (T**2 - Tref**2) +
+            cobj.cp_mol_ig_comp_coeff_A * (T - Tref), units["energy_mole"]) +
+            cobj.enth_mol_form_vap_comp_ref)
+
+        return h
+
+
+class entr_mol_ig_comp_G_H_ref(entr_mol_ig_comp):
```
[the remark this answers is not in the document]
PTA0001: @andrewlee94  Yes, I agree we need a discussion on this especially for clarity between the standard entropy reference (absolute entropy) that users may want to set. For consistency with chemical reactions and data on heat formation and Gibbs energy of formation, this will be required.  This is critical for the framework.

idaes/generic_models/properties/core/eos/virial.py
PTA0001: @ Andrew as a first pass, these are built here but yes, some changes will be required in the component module, on-demand property system and I will suggest we resolved these in a meeting.

idaes/generic_models/properties/core/eos/virial.py
```diff
@@ -0,0 +1,797 @@
...
+                           value, NonNegativeReals,
+                           Var, units as pyunits)
+from pyomo.common.config import ConfigBlock, ConfigValue, Bool
+
+from idaes.core.util.exceptions import PropertyNotSupportedError
+from idaes.generic_models.properties.core.generic.utility import (
+    get_method, get_component_object as gcobj)
+from idaes.generic_models.properties.core.eos.eos_base import EoSBase
+from idaes.core.util.misc import set_param_from_config
+from idaes.generic_models.properties.core.pure.RPP4 import entr_mol_ig_comp
```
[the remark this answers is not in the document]
PTA0001: @ Andrew, this only fixes the bug in the imported class as a first pass(I wanted to be sure it not just happening on my system), the EOS doesn't depend on the specific pure property.

idaes/generic_models/properties/core/eos/virial.py
```diff
@@ -0,0 +1,797 @@
...
+                                            f"{b}. Virial equation of"
+                                            "state supports only Vap phase.")
+
+        # get user specified eos options and save on parameter block
+        try:
+            use_pseudocritical_rules = b.config.equation_of_state_options[
+                "use_pseudocritical_rules"]
+            b._use_pseudocritical_rules = use_pseudocritical_rules
+            setattr(b.parent_block(), "virial_eos_options",
+                    deepcopy(VirialConfig))
```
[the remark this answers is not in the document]
PTA0001: @ Andrew, we would need a discussion on how this matters and the fix. I just replicated the same procedure in the Cubic equation of state.

idaes/generic_models/properties/core/eos/virial.py
```diff
@@ -0,0 +1,797 @@
...
+                            units=units["volume"] / pyunits.mol))
+                set_param_from_config(cobj, param="volume_crit")
+
+        # kappa binary interaction parameters for i-j molecular pair
+        # for  i=j(like molecules), kappa = 0
+        # for i not equal to  j, kappa is a small positive number.
+        # set to zero if data is unavailable
+
+        try:
+            kappa_data = param_block.config.parameter_data["kappa"]
```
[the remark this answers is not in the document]
PTA0001: Awesome

idaes/generic_models/properties/core/eos/virial.py
PTA0001: lets discuss about these

andrewlee94:
@PTA0001 I think we might need to rethink how you are going about this. Adding to the modular property framework is very complicated and takes a lot of work, and right now the IDAES team has very limited bandwidth for reviewing this. There is a good reason why nothing like this has been implemented previously in the modular property framework, and it is likely that there would need to be infrastructure changes to make all of this work.

It might be far easier (and faster) in the short term to do this as a stand-alone (i.e. custom) property package which implements everything directly and contribute it to `idaes.models_extra.column_models.properties` instead. This would allow you the freedom to implement the equations in the exact form you want without needing to worry about the structures (and limitations) forced upon you be the existing modular framework. This would also give us a working version of the model that we can test against when building the modular version in future. This would also allow you to do a more specific version for MEA only at this stage rather than having to take the extra steps to make it completely generic if necessary.

This would also separate the implementation of this model from the necessary infrastructure changes for the modular properties (i.e. you can get a working model up and running without needing to worry about or wait for the infrastructure changes).

ksbeattie:
@PTA0001 should we close this now?

ksbeattie:
@PTA0001, I'm going to close this as it is now quite old.  You can re-open it if you have time to get back to it.

PR #711, opened 2022-02-14
Adding Methanol Synthesis Flowsheets, Properties and Documentation

bpaul4:
## Summary/Motivation:
- Adding two methanol synthesis flowsheets (with and without recycle) and related properties to the IDAES repository

- Methanol synthesis flowsheets with importable methods, and relevant documentation
- Thermophysical properties for ideal VLE syngas (H2, CH4, CO, CO2) and H2O system
- Properties for kinetic and equilibrium reaction forms

### Legal Acknowledgement


bpaul4:
I am writing a test module to resolve the codecov issue. Please let me know if you have any recommendations regarding the added code or documentation.

MAZamarripa (approved):
lgtm

idaes/generic_models/properties/examples/methanol_water_ideal.py
```diff
@@ -0,0 +1,221 @@
...
+# Research Corporation, et al.  All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and
+# license information.
+#################################################################################
+"""
+Phase equilibrium package for methanol synthesis using ideal VLE.
+Author: Brandon Paul
+"""
+# Import Python libraries
```
andrewlee94, line 17: There needs to be some sort of test for this property package if it is going to go in core. It should solve the model for at least one test case and check the values against some source (preferably literature).

idaes/generic_models/properties/examples/methanol_water_ideal.py
```diff
@@ -0,0 +1,221 @@
...
+import logging
+
+# Import Pyomo units
+from pyomo.environ import units as pyunits
+
+# Import IDAES cores
+from idaes.core import LiquidPhase, VaporPhase, Component
+from idaes.core.phases import PhaseType as PT
```
andrewlee94, line 25: This should be imported from `idaes.core` in the line above.

As it stands this will have deprecation warnings due to the reorganization of core.

idaes/generic_models/properties/examples/methanol_reactions.py
```diff
@@ -0,0 +1,76 @@
...
+# Syngas conversion to methanol: CO + 2H2 => CH3OH (assumed sole reaction here)
+
+
+config_dict = {
+    "base_units": {"time": pyunits.s,
+                   "length": pyunits.m,
+                   "mass": pyunits.kg,
+                   "amount": pyunits.mol,
+                   "temperature": pyunits.K},
+#    "equilibrium_reactions": {
```
andrewlee94, line 47: Commented code.

idaes/generic_models/properties/examples/methanol_reactions.py
```diff
@@ -0,0 +1,76 @@
...
+               "concentration_form": ConcentrationForm.molarity,
+               "rate_constant": arrhenius,
+               "rate_form": power_law_rate,
+               "parameter_data": {
+                    "reaction_order": {("Vap", "CO"): 1,
+                                       ("Vap", "H2"): 2},
+                    "arrhenius_const": (1, pyunits.m**6 / pyunits.mol**2
+                                        / pyunits.s),
+                    "dh_rxn_ref": (-90640, pyunits.J/pyunits.mol),
+                    "energy_activation": (0, pyunits.J/pyunits.mol)}}}}
```
andrewlee94, line 76: As for the thermo properties, this needs some sort of test.

idaes/generic_models/flowsheets/tests/test_methanol_flowsheet.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+@pytest.mark.unit
+def test_set_inputs(model):
+    set_inputs(model)
+
+    assert degrees_of_freedom(model) == 0
+
+
+@pytest.mark.unit
```
andrewlee94, line 88: This should be a `component` test, if not an `integration` test, as it requires a solver.

idaes/generic_models/flowsheets/tests/test_methanol_flowsheet.py
```diff
@@ -0,0 +1,193 @@
...
+    assert model.fs.F101.liq_outlet.flow_mol[0].expr.value == pytest.approx(
+        142.57, 1e-3)
+
+
+@pytest.mark.integration
+def test_unit_consistency(model):
+    assert_units_consistent(model)
+
+
+@pytest.mark.unit
```
andrewlee94, line 111: As above this should be a `component ` or `integration` test.

idaes/generic_models/flowsheets/tests/test_methanol_flowsheet.py
```diff
@@ -0,0 +1,193 @@
...
+        237.6005, abs=1e-2)
+    assert value(model.fs.R101.heat_duty[0])/1E6 == pytest.approx(
+        -45.2192, abs=1e-2)
+    assert value(model.fs.T101.work_isentropic[0])/1E6 == pytest.approx(
+        -0.9593, abs=1e-2)
+    assert value(model.fs.F101.recovery*100) == pytest.approx(
+        60.0047, abs=1e-2)
+
+
+@pytest.mark.unit
```
andrewlee94, line 130: This is definitely an `integration` test.

idaes/generic_models/flowsheets/tests/test_methanol_flowsheet.py
```diff
@@ -0,0 +1,193 @@
...
+    assert degrees_of_freedom(model) == 0
+
+
+@pytest.mark.unit
+def test_initialize_flowsheet(model):
+    initialize_flowsheet(model)
+
+    assert degrees_of_freedom(model) == 0
+
+    assert model.fs.M101.outlet.flow_mol[0].value == pytest.approx(954, 1e-3)
```
andrewlee94, line 94: Do you have a source for these values, or are they just the values you got when you ran the model? If you have a source, it would be good to report it here.

idaes/generic_models/flowsheets/tests/test_methanol_flowsheet.py
```diff
@@ -0,0 +1,193 @@
...
+    assert model.fs.C101.outlet.flow_mol[0].value == pytest.approx(954, 1e-3)
+    assert model.fs.H101.outlet.flow_mol[0].value == pytest.approx(954, 1e-3)
+    assert model.fs.R101.outlet.flow_mol[0].value == pytest.approx(478.8, 1e-3)
+    assert model.fs.T101.outlet.flow_mol[0].value == pytest.approx(478.8, 1e-3)
+    assert model.fs.H102.outlet.flow_mol[0].value == pytest.approx(478.8, 1e-3)
+    assert model.fs.F101.vap_outlet.flow_mol[0].expr.value == pytest.approx(
+        336.23, 1e-3)
+    assert model.fs.F101.liq_outlet.flow_mol[0].expr.value == pytest.approx(
+        142.57, 1e-3)
```
andrewlee94, line 103: A check for conservation of mass and energy would be good too. In fact, I don't see any checks for temperature and pressure above at all.

idaes/generic_models/flowsheets/methanol_flowsheet.py
```diff
@@ -0,0 +1,396 @@
...
+    m.fs.T101.initialize()
+    propagate_state(arc=m.fs.s06)  # turbine to cooler
+
+    m.fs.H102.initialize()
+    propagate_state(arc=m.fs.s07)  # cooler to flash
+
+    m.fs.F101.initialize()
+
+
+def add_costing(m):
```
andrewlee94, line 245: This is going to need to be updated before November when the old costing framework is deprecated.

andrewlee94 (changes requested):
From a quick look, I see a few things that need to be considered and addressed.

First, is `generic_models` the correct place for this, or would it be better to put this in the new `models_extra` folder? I don't mind it going in `generic_models`, but it would need to be tested a lot more thoroughly for that to happen.

Additionally, there are a few things here that are going to need to be updated for IDAES 2.0, and I feel it would be better to address these before this is merged if at all possible (especially if this is going into `generic_models`).

bpaul4:
It seems the end of each flowsheet test module has a second-to-last test asserting that extent equals some expected nonzero value, and a last test that fails because extent equals zero. I'm still investigating 1) why extent changes values between the two tests, and 2) why I don't see these tests failing locally.

andrewlee94:
@bpaul4 As a more general comment on this, I have to ask why the `report` function in the main code has an `assert` statement in it. An `assert` like that makes sense in a test, but not so much in the actual code, as it means that if it ever gets triggered the user will see an `AssertionError` with no guidance on what it means or how to fix it.

As for why you are seeing the error - the problem is that you are only solving the model as an `integration` tests. That means that when we run the normal tests in Github Actions, these tests are being skipped and the reporting test is being called on a model which has not been solved (hence the 0 extents). In short, we only run tests with the `integration` flag once the PR is approved (and they are run separately), so you need to consider how this interreacts with other tests.

Short answer, I think you need to do three things:
1. Flag the test of `report` as an integration test too if you expect it to show real results.
2. Remove the `assert` from the `report` method, as it does not belong there.
3. Add an assertion in the test code that checks the output from the `report` method. See line 255 in  https://github.com/IDAES/idaes-pse/blob/main/idaes/models/unit_models/tests/test_cstr.py

bpaul4:
I added the assert statement to check why the error was occurring, as I did not see any test failures locally and thought it may have been a numpy or other error by the CI client (which has occurred before) - I'll remove it in the next update. I appreciate your insights, and will make the necessary changes in the code.

idaes/generic_models/flowsheets/methanol_flowsheet.py
```diff
@@ -11,386 +11,12 @@
 # at the URL "https://github.com/IDAES/idaes-pse".
 ##############################################################################
 """
-Task: IDAES Support for ARPE-E Differentiate
-Scenario: Methanol Synthesis From Syngas
-Author: B. Paul and M. Zamarripa
+Deprecation path for renamed module.
```
andrewlee94, line 14: This file is not needed - the methanol flowsheet is a new addition to the repo so it does not need a deprecation path.

idaes/generic_models/flowsheets/methanol_flowsheet_w_recycle.py
```diff
@@ -11,471 +11,12 @@
 # at the URL "https://github.com/IDAES/idaes-pse".
 ##############################################################################
 """
-Task: IDAES Support for ARPE-E Differentiate
-Scenario: Methanol Synthesis From Syngas
-Author: B. Paul and M. Zamarripa
+Deprecation path for renamed module.
```
andrewlee94, line 14: As before, deprecation paths are only needed for things that were in the repo before this PR.

idaes/models/flowsheets/methanol_flowsheet.py
```diff
@@ -0,0 +1,736 @@
...
+    # product recovery
+    m.fs.F101 = Flash(
+        default={"property_package": m.fs.thermo_params_VLE,
+                 "has_heat_transfer": True,
+                 "has_pressure_change": True})
+
+    # Build the flowsheet
+    print('Unit degrees of freedom')
+    for unit in ('M101', 'C101', 'H101', 'R101', 'T101', 'H102', 'F101'):
+        if unit == 'M101':
```
andrewlee94, line 167: This needs some comments to explain what is going on and what the expected output is. Notably, you need to explain where the `spec` value come from.

idaes/models/flowsheets/methanol_flowsheet.py
```diff
@@ -0,0 +1,736 @@
...
+
+    # Add unit and stream specifications
+    print('Total DOF: ', degrees_of_freedom(m))
+    return m
+
+
+def set_inputs(m):
+
+    #  feed streams, post WGS
+    m.fs.M101.H2_WGS.flow_mol[0].fix(637.2)  # mol/s, relative to 177 kmol/h
```
andrewlee94, line 202: You should include units in all these `fix` methods.

E.g. `m.fs.M101.H2_WGS.flow_mol[0].fix(177*pyo.units.kmol/pyo.units.hour)`

Also, as mentioned elsewhere, these feed conditions should be set by `Feed` blocks rather than through the `Port` of the `Mixer`.

idaes/models/flowsheets/methanol_flowsheet.py
```diff
@@ -0,0 +1,736 @@
...
+    m.fs.T101.initialize()
+    propagate_state(arc=m.fs.s06)  # turbine to cooler
+
+    m.fs.H102.initialize()
+    propagate_state(arc=m.fs.s07)  # cooler to flash
+
+    m.fs.F101.initialize()
+
+
+def add_costing(m):
```
andrewlee94, line 519: As I noted in the previous review, this is using the old style costing methods. If this is going into `idaes.models`, then I will need to insist that this is updated to the new style before it can be merged.

I would strongly encourage you to make the change anyway, as otherwise you will just need to update this in the next few months when we fully deprecate the old-style approach.

idaes/models/flowsheets/methanol_flowsheet.py
```diff
@@ -0,0 +1,736 @@
...
+                iscale.set_scaling_factor(getattr(
+                    getattr(m.fs.thermo_params_vapor, comp), attr), 1)
+        m.fs.thermo_params_vapor.set_default_scaling("flow_mol_phase_comp",
+                                                     1, index=comp)
+
+    m.fs.reaction_params = GenericReactionParameterBlock(
+        default={"property_package": m.fs.thermo_params_vapor,
+                 **reaction_props.config_dict})
+
+    # mixing feed streams
```
andrewlee94, line 117: If this is going to be a proper example, then you need to use `Feed` blocks to to define the inlets and connect those to the mixer.

idaes/models/flowsheets/tests/test_methanol_flowsheet.py
```diff
@@ -0,0 +1,673 @@
...
+            if hasattr(unit, "pressure_equality"):
+                for t, c in unit.pressure_equality.items():
+                    assert pytest.approx(1e-5, rel=1e-5) == \
+                        (iscale.
+                         get_constraint_transform_applied_scaling_factor(c))
+
+
+@pytest.mark.component
+def test_initialize_flowsheet(model):
+    solver = get_solver()
```
andrewlee94, line 384: Minor comment, but you can pass the `optarg` dict here to set the options at the same time as creating the solver object.

idaes/models/flowsheets/tests/test_methanol_flowsheet.py
```diff
@@ -0,0 +1,673 @@
...
+    optarg = {'tol': 1e-6,
+              'max_iter': 5000}
```
andrewlee94, line 386: 5000 iterations is excessive - you should set this to a reasonable value. This is especially important if something goes wrong later so that the tests fail quickly.

idaes/models/flowsheets/tests/test_methanol_flowsheet.py
```diff
@@ -0,0 +1,673 @@
...
+    solver.options = optarg
+    initialize_flowsheet(model)
+
+    assert degrees_of_freedom(model) == 0
+
+    assert model.fs.M101.outlet.flow_mol[0].value == pytest.approx(954, 1e-3)
```
andrewlee94, line 392: This is a rather high tolerance for checking the result (0.1%) - we should be able to check to a far higher tolerance that this (I would aim for `rel=1e-5`). It is probably also best to explicitly set this to a relative tolerance as well (i.e. do `rel=`) in case something changes in `pytest` in the future.

idaes/models/flowsheets/tests/test_methanol_flowsheet.py
```diff
@@ -0,0 +1,673 @@
...
+        -51.3636, abs=1e-2)
+    assert value(model.fs.T101.work_isentropic[0])/1E6 == pytest.approx(
+        -1.9905, abs=1e-2)
+    assert value(model.fs.F101.recovery*100) == pytest.approx(
+        95.5433, abs=1e-2)
+    assert value(model.fs.objective)/1E6 == pytest.approx(
+        81.0476, abs=1e-2)
+
+
+@pytest.mark.unit
```
andrewlee94, line 581: This cannot be marked as a `unit` test as it depends on output from previous `component` tests. This means that if someone were to un only the `unit` tests this test would fail as the model would not yet have been solved.

At a higher level, how long do these tests take to run? As this is a flowsheet example, I think most of these tests would be better classified as integration tests.

idaes/models/flowsheets/tests/test_methanol_flowsheet.py
```diff
@@ -0,0 +1,673 @@
...
+def test_report(model):
+    stream = StringIO()
```
andrewlee94, line 583: I know I said to do this last time, but we have come to realize that this can be a bit too fragile. However, with the way the `report` method is currently written I don't think there are many alternatives.

idaes/models/flowsheets/tests/test_methanol_flowsheet.py
```diff
@@ -0,0 +1,673 @@
...
+    sys.stdout = stream
```
andrewlee94, line 584: This looks like a very bad idea (even if you revert the change later). You should rewrite the `report` method to take  a stream as an option argument which will allow you (and users) to redirect the output to wherever you want.

idaes/models/properties/examples/methanol_ideal_VLE.py
```diff
@@ -22,17 +22,18 @@
```
andrewlee94, line 22: As part of the broader question of where this example belongs, I wonder if it would be better to keep the property modules with the flowsheet so that it is easier for users to find them,

idaes/models/properties/examples/tests/test_methanol_ideal_VLE.py
```diff
@@ -0,0 +1,298 @@
...
+def _as_quantity(x):
+    unit = pyunits.get_units(x)
+    if unit is None:
+        unit = pyunits.dimensionless
+    return value(x) * unit._get_pint_unit()
+
+
+# Test for configuration dictionaries with parameters from Properties of Gases
+# and liquids 4th edition
+class TestParamBlock(object):
```
andrewlee94, line 66: This test is not required - all of this gets covered by the tests in the modular properties code.

idaes/models/properties/examples/tests/test_methanol_ideal_VLE.py
```diff
@@ -0,0 +1,298 @@
...
+        # Fix state
+        model.props[1].flow_mol.fix(415.44)
+        model.props[1].enth_mol.fix(-1.3643e5)
+        model.props[1].pressure.fix(5.1e6)
+        model.props[1].mole_frac_comp["CH4"].fix(2.2964e-6)
+        model.props[1].mole_frac_comp["CO"].fix(0.11438)
+        model.props[1].mole_frac_comp["H2"].fix(0.23743)
+        model.props[1].mole_frac_comp["CH3OH"].fix(0.64818)
+
+        assert degrees_of_freedom(model.props[1]) == 0
```
andrewlee94, line 186: I would not put the `assert` in the `fixture`, as this will cause all associated tests to fail. Instead, there should be a separate test for degrees of freedom.

idaes/models/properties/examples/tests/test_methanol_ideal_VLE.py
```diff
@@ -0,0 +1,298 @@
...
+
+        return model
+
+    @pytest.mark.unit
```
andrewlee94, line 190: I don't think this test is necessary.

idaes/models/properties/examples/tests/test_methanol_ideal_VLE.py
```diff
@@ -0,0 +1,298 @@
...
+        assert model.props[1].mole_frac_comp["CH4"].value == 2.2964e-6
+        assert model.props[1].mole_frac_comp["CO"].value == 0.11438
+        assert model.props[1].mole_frac_comp["H2"].value == 0.23743
+        assert model.props[1].mole_frac_comp["CH3OH"].value == 0.64818
+
+    @pytest.mark.integration
+    def test_unit_consistency(self, model):
+        assert_units_consistent(model)
+
+    @pytest.mark.unit
```
andrewlee94, line 219: This test is not required here,

idaes/models/properties/examples/tests/test_methanol_ideal_VLE.py
```diff
@@ -0,0 +1,298 @@
...
+    def test_define_state_vars(self, model):
+        sv = model.props[1].define_state_vars()
+
+        assert len(sv) == 4
+        for i in sv:
+            assert i in ["flow_mol", "mole_frac_comp", "enth_mol", "pressure"]
+
+    @pytest.mark.unit
```
andrewlee94, line 227: This test is not required here,

idaes/models/properties/examples/tests/test_methanol_ideal_VLE.py
```diff
@@ -0,0 +1,298 @@
...
+        # Check phase equilibrium results
+        assert model.props[1].mole_frac_phase_comp[
+            "Vap", "CH3OH"
+        ].value == pytest.approx(0.64818, abs=1e-4)
+        assert model.props[1].mole_frac_phase_comp[
+            "Liq", "CH3OH"
+        ].value == pytest.approx(1.00000, abs=1e-4)
+
+    @pytest.mark.unit
+    def test_report(self, model):
```
andrewlee94, line 297: This test is not required here,

idaes/models/properties/examples/tests/test_methanol_ideal_VLE.py
```diff
@@ -0,0 +1,298 @@
...
+
+        assert len(fin_act_consts) == len(orig_act_consts)
+        assert len(fin_fixed_vars) == len(orig_fixed_vars)
+
+        for c in fin_act_consts:
+            assert c in orig_act_consts
+        for v in fin_fixed_vars:
+            assert v in orig_fixed_vars
+
+    @pytest.mark.integration
```
andrewlee94, line 279: This should be a component test.

idaes/models/properties/examples/tests/test_methanol_reactions.py
```diff
@@ -0,0 +1,187 @@
...
+                assert v == rate_config[k[0]].stoichiometry[k[1], k[2]]
+            else:
+                assert v == 0
+
+        assert isinstance(model.rxn_params.reaction_R1, Block)
+
+        assert_units_consistent(model)
+
+
+class TestStateBlock(object):
```
andrewlee94, line 79: Similar to the thermo properties, I think most of these tests can be dropped - all you need to test here is that you get the expected result.

andrewlee94 (changes requested):
I had a number of inline comments, but there are a few general things that need to be considered.

1. I am not sure that `idaes-pse` is the correct place for this. It is primarily an example, which would suggest that it belongs in `examples-pse` instead (note that not everything in `examples=pse` needs to be a notebook). I have added this to the agenda for the next IDAES dev team call to discuss.
2. If we do decide to keep this in `idaes-pse`, then I am not sure if `idaes.models` is the correct place for it either. Notably, if it is to go in `idaes.models` then it needs to adhere to all the core standards, which means that the costing needs to be updated to the new standard first.

andrewlee94:
@bpaul4 We discussed where this model belongs in the IDAES dev team call just now, and the consensus is that example flowsheets like this should generally go in the examples repository. I think it also makes the most sense to keep the property packages with the example for things like this too (so that they are easily found and examined by the user), so I would suggest you move all of this into you other PR to `examples-pse`.

bpaul4:
@andrewlee94 okay. I will close this pull request and update the example accordingly.

PR #712, opened 2022-02-15
Solvent column corrections

anujad95:
## Fixes
- Unit of convective heat flow corrected
- Convective and total inter phase heat transfer equations corrected

## Summary/Motivation:
Corrections have been made to the heat transfer equations in the solvent column model. This is expected to improve robustness.

- Using average liquid and vapor phase temperatures in the convective heat transfer equation
- Using the appropriate sign conventions and phase properties for calculating the total rate of heat transfer

### Legal Acknowledgement

### DIFF
--- a/idaes/generic_models/unit_models/column_models/solvent_column.py
+++ b/idaes/generic_models/unit_models/column_models/solvent_column.py
@@ -312,17 +312,17 @@ documentation for supported schemes,
         self.diameter_column = Var(domain=Reals,
                                    initialize=0.1,
                                    units=pyunits.m,
-                                   doc='Column diameter')
+                                   doc='Column diameter (m)')
         
         self.area_column = Var(domain=Reals,
                                initialize=0.5,
                                units=pyunits.m**2,
-                               doc='Column cross-sectional area')
+                               doc='Column cross-sectional area (m2)')
         
         self.length_column = Var(domain=Reals,
                                  initialize=4.9,
                                  units=pyunits.m,
-                                 doc='Column length')
+                                 doc='Column length (m)')
 
         # Hydrodynamics
         self.velocity_vap = Var(self.flowsheet().time,
@@ -330,14 +330,14 @@ documentation for supported schemes,
                                 domain=NonNegativeReals,
                                 initialize=2,
                                 units=pyunits.m / pyunits.s,
-                                doc='Vapor superficial velocity')
+                                doc='Vapor superficial velocity (m/s)')
         
         self.velocity_liq = Var(self.flowsheet().time,
                                 self.liquid_phase.length_domain,
                                 domain=NonNegativeReals,
                                 initialize=0.01,
                                 units=pyunits.m / pyunits.s,
-                                doc='Liquid superficial velocity')
+                                doc='Liquid superficial velocity (m/s)')
         
         self.holdup_liq = Var(self.flowsheet().time,
                               self.liquid_phase.length_domain,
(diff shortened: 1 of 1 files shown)


idaes/generic_models/unit_models/column_models/solvent_column.py
```diff
@@ -312,32 +312,32 @@ def _make_performance(self):
         self.diameter_column = Var(domain=Reals,
                                    initialize=0.1,
                                    units=pyunits.m,
-                                   doc='Column diameter')
+                                   doc='Column diameter (m)')
```
andrewlee94, line 315: I generally suggest not including the units in the doc string, as it presumes a specific set of units and effectively duplicates the information stored in units.  Thus, it becomes possible for the doc string to become inconsistent with the actual units if one s changed but not the other,

idaes/generic_models/unit_models/column_models/solvent_column.py
```diff
@@ -622,38 +624,37 @@ def rule_heat_transfer_coeff_Ack(blk, t, x):
...
+        # Convective interphase heat transfer [J/s.m]
         @self.Constraint(self.flowsheet().time,
                          self.vapor_phase.length_domain,
-                         doc="heat transfer - vapor side ")
-        def vapor_phase_volumetric_heat_flux(blk, t, x):
+                         doc="Convective interphase heat transfer")
+        def vapor_phase_convective_heat_flow(blk, t, x):
             if x == self.vapor_phase.length_domain.first():
-                return blk.heat_flux_vap[t, x] == 0
+                return blk.heat_flow_vap[t, x] == 0
```
andrewlee94, line 633: This should be `Constraint.Skip`

idaes/generic_models/unit_models/column_models/solvent_column.py
```diff
@@ -622,38 +624,37 @@ def rule_heat_transfer_coeff_Ack(blk, t, x):
...
             else:
                 zb = self.vapor_phase.length_domain.at(value(self.zi[x]) - 1)
-                return blk.heat_flux_vap[t, x] == blk.h_v_Ack[t, x] * \
-                    (blk.liquid_phase.properties[t, zb].temperature -
-                      blk.vapor_phase.properties[t, x].temperature)
+                return blk.heat_flow_vap[t, x] == blk.h_v_Ack[t, x] * \
```
andrewlee94, line 636: Why did you update this to use the average temperature between the two points? I believe this model is written explicitly using a finite element approach assuming that the state in each cell is equal to the value of the finite element. I.e. I believe the original form was correct.

andrewlee94 (changes requested):
Mostly looks good, but I had a few comments. The most significant is asking why you changed the way the convective heat transfer is calculated.

idaes/generic_models/unit_models/column_models/solvent_column.py
anujad95: Units have been removed from all doc strings, as suggested.

idaes/generic_models/unit_models/column_models/solvent_column.py
anujad95: I agree, this change has been made.

idaes/generic_models/unit_models/column_models/solvent_column.py
anujad95: I agree with this as well. I have changed it to the original equation.

andrewlee94 (approved):
Looks OK to me.

PR #713, opened 2022-02-16
Prevent references from being descaled multiple times

dallan-keylogic:
## Fixes

Previously, if a variable with multiple references had a scaling factor, it would get descaled once for each reference. This PR ensures they are descaled only once.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/solvers/petsc.py
+++ b/idaes/core/solvers/petsc.py
@@ -655,9 +655,11 @@ class PetscTrajectory(object):
         Returns:
             None
         """
+        # Variables might show up more than once because of References
+        already_scaled = set()
         for var in m.component_data_objects():
             vname = str(var)
-            if vname in self.vecs:
+            if vname in self.vecs and vname not in already_scaled:
                 s = None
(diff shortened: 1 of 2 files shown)


eslickj (approved):
Looks good.  Thanks!

idaes/core/solvers/petsc.py
```diff
@@ -666,6 +668,7 @@ def _unscale(self, m):
                 if s is not None:
                     for i, x in enumerate(self.vecs[vname]):
                         self.vecs[vname][i] = x/s
+                already_scaled.add(vname)
```
Robbybp, line 671: Since `vname` is being used already, using it in this set isn't a huge deal, but I would rather we didn't use `str(var)` altogether. Name generation can be very time consuming for variables with large indexing sets. Using `id(var)` would be much faster.

Robbybp (approved):
Looks good. I didn't notice we were using `str(var)` in my previous review, but that has nothing to do with this fix.

Issue #673, opened 2022-01-24
Add wall temperature and heat accumulation terms to the 0D heat exchanger


rustygentile:
# Summary 
Hello IDAES team. I have a unit model that, I think, would make a nice contribution as a pull request. A draft of it can be found [here](https://github.com/CATER-UCF/sco2-air-cooler/blob/master/models/heat_exchanger_lumped_capacitance.py). Our use case was for a transient simulation of an sCO2 power cycle, though there's nothing really sCO2 or power plant specific here. This could be used to model any heat exchanger.

# Questions 
* Would this be better suited as a new unit model in `/generic_models` or in `/power_generation` ? Another option could be to add my new code directly to `/idaes/generic_models/unit_models/heat_exchanger.py`
* I'd also like to include an example flowsheet. Should this be added to one of the `/flowsheets` folders here or to the `examples-pse` repository? (See also #577). If we opt for the latter, how would you prefer to keep the changes in sync? Maybe submit the PR's to feature branches instead of main?

Any other early feedback or considerations would be welcome too, of course.

rustygentile:
Another note: it seems that, as an external user, I don't have permission to assign this to myself as the [docs](https://idaes-pse.readthedocs.io/en/stable/advanced_user_guide/developer/devsw.html#create-an-issue-on-github) suggest.

eslickj:
This looks great.  I didn't look at the details really close, but I have some notes on style.  To accommodate the possibility of indexed models and to have your constraints built automatically, you can inherit HeatExchangerData and implement the build method.

Here's an example (see one of the IDAES model for a concrete example):

```python
...
@declare_process_block_class("HeatExchangerLumpedCapacitance", doc="...")
class HeatExchangerLumpedCapacitanceData(HeatExchangerData):
    CONFIG = HeatExchangerData.CONFIG(implicit=True)
    ...Additional config options if any...
    def build(self):
        super().build()
        self.add_dynamic_variables()
        self.add_dynamic_variable_constraints()
        ...

 
    ...special initialization methods? if yes implement initialize...

    ...special scaling factor calculations? if yes implement calculate_scaling_factors...
```

eslickj:
[...]

PR #714, opened 2022-02-17
Add wall temperature and heat accumulation terms to the 0D heat exchanger

rustygentile:
## Fixes
See: #673 

## Summary/Motivation:

Adding a new unit model that extends the 0D HeatExchanger by including wall temperature and heat holdup terms in the overall energy balance. This will improve transient simulations.

- Adding `HeatExchangerLumpedCapacitance` unit model class
- Adding unit tests for the new model
- Adding a readthedocs page that includes an example with sample code

### Legal Acknowledgement


idaes/generic_models/unit_models/tests/test_heat_exchanger_lc.py
```diff
@@ -0,0 +1,515 @@
...
+        m.fs.unit.ua_hot_side[:].fix(100)
+
+        return m
+
+    @pytest.mark.unit
+    def test_units(self, model):
+        # Note: using the discretizer makes the units of measure on the time
+        # derivative term inconsistent...
+        with pytest.raises(InconsistentUnitsError):
+            assert_units_consistent(model)
```
andrewlee94, line 319: Rather than expecting an error here, I think the correct approach would be to `xfail` the test (i.e. tell pytest that for now it is expected to fail). The difference is rather semantic, but in this case to me an `xfail` tends to indicate this is wrong and needs to be fixed but we will accept it for now, whilst the `pytest.rasies` tends to imply an expected behaviour we are testing for.

docs/reference_guides/model_libraries/generic/unit_models/heat_exchanger_lc.rst
```diff
@@ -0,0 +1,227 @@
+Heat Exchanger (Lumped Capacitance)
+===================================
```
andrewlee94, line 2: Feel free to add a line here taking credit for the contribution. Not only do we want to acknowledge external input, but it might also encourage others to contribute as well.

andrewlee94 (commented):
From an initial read through, this looks really good. My only comments were a couple of minor things.

idaes/generic_models/unit_models/heat_exchanger_lc.py
```diff
@@ -0,0 +1,203 @@
...
+from pyomo.dae import DerivativeVar
+from idaes.core import declare_process_block_class
+from .heat_exchanger import HeatExchangerData
+
+
+@declare_process_block_class("HeatExchangerLumpedCapacitance",
+                             doc="0D heat exchanger for transient simulations")
+class HeatExchangerLumpedCapacitanceData(HeatExchangerData):
+
+    CONFIG = HeatExchangerData.CONFIG(implicit=True)
```
eslickj, line 29: Unless you want to add configuration options, I don't think you need this line since CONFIG should already exist as part of the inherited HeatExchangerData.

idaes/generic_models/unit_models/heat_exchanger_lc.py
```diff
@@ -0,0 +1,203 @@
...
+                return b.hot_side.heat[t] + b.cold_side.heat[t] == \
+                       -b.dT_wall_dt[t] * b.heat_capacity_wall
+
+            self.activate_dynamic_heat_eq()
+            self.dT_wall_dt[self.flowsheet().config.time.first()].fix(0)
+            self.dT_wall_dt[:].value = 0
+
+    def initialize(self, *args, **kwargs):
+
+        if self.flowsheet().config.dynamic:
```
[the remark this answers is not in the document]
rustygentile: The reason for this is that I want to use control volumes with `dynamic=False`. And the way the base class is written, the control volumes inherit their `dynamic` flags from the unit model config. In `heat_exchanger.py` we have:

```
        hot_side = _make_heater_control_volume(
            self,
            config.hot_side_name,
            config.hot_side_config,
            dynamic=config.dynamic,
            has_holdup=config.has_holdup,
        )
```

What you're suggesting would call for a config like so:

```
        m.fs.unit = HeatExchangerLumpedCapacitance(default={
            "shell": {"property_package": m.fs.properties,
                      "dynamic": False},
            "tube": {"property_package": m.fs.properties,
                     "dynamic": False},
            "dynamic": True})
```

And then, for instance, in `heat_exchanger.py`:

```
        try:
            hot_dynamic = hot_side_config.dynamic
        except KeyError:
            hot_dynamic = config.dynamic
        
        hot_side = _make_heater_control_volume(
            self,
            config.hot_side_name,
            config.hot_side_config,
            dynamic=hot_dynamic,
            has_holdup=config.has_holdup,
        )
```

That would give us three different levels where the `dynamic` flag could be configured: flowsheet -> unit model -> control volume. And in that case, we should probably do the same for `has_holdup`. Actually, the new logic probably belongs in `_process_config()`.

I'd be happy to try this, but I thought it best not to make any changes to `heat_exchanger.py` in case someone else is also working on it. Let me know which you prefer, to leave it as-is, or to make the changes to both unit models and have `dynamic` be configurable at each level.

idaes/generic_models/unit_models/heat_exchanger_lc.py
rustygentile: Removed with my last commit. Thanks for the feedback!

idaes/generic_models/unit_models/heat_exchanger_lc.py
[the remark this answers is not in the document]
andrewlee94: First, I think we need to work out exactly what you want to allow for, as I am not quite sure what is happening. Is it that you want to allow separate the decision about holdup dynamics from that of the wall energy balance then - i.e. allow the wall energy balance to be dynamic even if the holdups are steady-state?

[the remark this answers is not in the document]
rustygentile: > Is it that you want to allow separate the decision about holdup dynamics from that of the wall energy balance then - i.e. allow the wall energy balance to be dynamic even if the holdups are steady-state?

Yes exactly!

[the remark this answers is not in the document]
andrewlee94: OK, my first thought then is that we should allow the user to choose wall dynamics separately from the material balance -  I.e. have a separate configuration option for the wall dynamics that the user can set, and then have some logic on the default setting (much like how the material dynamics are done).

We might want to think about any combinations that do not make sense however - I am not sure that there are any however (at most it would be setting everything to steady-state, in which case this devolves into the basic heat exchanger, and that will still work).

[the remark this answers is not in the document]
rustygentile: > OK, my first thought then is that we should allow the user to choose wall dynamics separately from the material balance - I.e. have a separate configuration option for the wall dynamics that the user can set, and then have some logic on the default setting (much like how the material dynamics are done).

This is a nice solution. If there are no other concerns or objections, I'll go ahead and implement it. Do you have a preference for the new variable name? I was thinking either `dynamic_wall` or `dynamic_heat_balance`. Also, are there any other similar unit models? Because it might be good to standardize or reuse this variable name if possible.

> We might want to think about any combinations that do not make sense however - I am not sure that there are any however (at most it would be setting everything to steady-state, in which case this devolves into the basic heat exchanger, and that will still work).

Agreed. That did occur to me. My initial thought was that there wasn't a use case for having everything be steady-state, since one could just use the base heat exchanger model. But with this model, you at least get the wall temperature in addition. So that might still be useful to have. Especially if you're troubleshooting a flowsheet or model that's difficult to solve, it could be helpful to see this temperature at steady-state.

Also, if you're in "troubleshooting mode" I think it's better to change something with a flag rather than swapping out the entire unit model. Since using a different unit model introduces more unknowns. So I think it's important to make sure this model still works with steady-state flowsheets.

The same argument could be made for the other config combinations. The user might want to make sure their model solves with dynamic control volumes before activating the wall dynamics. And vice versa too.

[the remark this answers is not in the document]
andrewlee94: `dynamic_heat_balance` is the best of anything I can think of so far.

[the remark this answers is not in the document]
rustygentile: > `dynamic_heat_balance` is the best of anything I can think of so far.

Ok done! I also added a test to make sure each combination of dynamic / dynamic_heat_balance can be built. And some xfail-ing tests for invalid configurations.

idaes/generic_models/unit_models/tests/test_heat_exchanger_lc.py
```diff
@@ -381,6 +395,55 @@ def test_solve(self, model):
...
+    @pytest.mark.unit
+    @pytest.mark.xfail(reason="static flowsheet with dynamic heat balance",
+                       raises=DAE_Error)
+    def test_static_flowsheet(self, static_flowsheet_model):
+        m = static_flowsheet_model
+        m.fs.unit = HeatExchangerLumpedCapacitance(default={
+            "shell": {"property_package": m.fs.properties},
+            "tube": {"property_package": m.fs.properties},
+            "dynamic": False,
+            "dynamic_heat_balance": True})
```
andrewlee94, line 407: For these cases, we generally prefer the following to an `xfail`:

```
with pytest.raises(DAE_Error, match="error message"):
    m.fs.unit = HeatExchangerLumpedCapacitance(default={
            "shell": {"property_package": m.fs.properties},
            "tube": {"property_package": m.fs.properties},
            "dynamic": False,
            "dynamic_heat_balance": True})
```

The reason being that this is actually a successfully test of an expected failure, so this way the test shows up as passed, plus it also checks that the error message is what was expected.

idaes/generic_models/unit_models/heat_exchanger_lc.py
```diff
@@ -26,6 +27,22 @@
                              doc="0D heat exchanger for transient simulations")
 class HeatExchangerLumpedCapacitanceData(HeatExchangerData):
 
+    CONFIG = HeatExchangerData.CONFIG(implicit=True)
+
+    CONFIG.declare(
+        "dynamic_heat_balance",
+        ConfigValue(
+            default=True,
```
andrewlee94, line 35: Two comments here:

1. Do we want to allow this to take a default value based on the flowsheet, in the same way the `dynamic` flag does? In this case, we should make the default `useDefault` and the domain `DefaultBool` (see https://github.com/IDAES/idaes-pse/blob/main/idaes/core/unit_model.py).
2. I think it would be helpful to add validation of this config argument in the code (and it would be necessary if we do suggestion 1). E.g. we should check for `dynamic_heat_balance=True` in flowsheets where `dynamic=False` and raise an exception of our own to inform users what happened. Otherwise, they will eventually see a `DAE_Error`, but that is likely not going to help them understand what they did wrong and how to fix it.

andrewlee94 (commented):
Looking good. I just had a couple of suggestions on how to improve the usability of the new config argument and on a way to replace some of the `xfails` with tests that will check the error messages.

idaes/generic_models/unit_models/tests/test_heat_exchanger_lc.py
rustygentile: That makes sense. I didn't do this for the InconsistentUnitsError though, since that's not something we really intended. And this doesn't seem to be caused by IDAES. So I'll try and setup a simple example and submit an issue to the Pyomo folks.

idaes/generic_models/unit_models/heat_exchanger_lc.py
rustygentile: Agreed. `useDefault` makes sense here. Regarding point 2., I found this: https://github.com/IDAES/idaes-pse/blob/main/idaes/core/util/exceptions.py, which seems more appropriate than `DAE_Error`

idaes/generic_models/unit_models/tests/test_heat_exchanger_lc.py
andrewlee94: @rustygentile Pyomo is already aware of the issue with units: https://github.com/Pyomo/pyomo/issues/1790. However, more people commenting about this being a problem might help motivate a fix.

andrewlee94 (approved):
Looks good to me.

PR #715, opened 2022-02-18
Add SOEC design point model

eslickj:
Adds a simple design point SOEC model.  There are two purposes for this.  1) to have a model that solves easily with no parameters for the flowsheet design phase and 2) serve as the basis for a thermodynamically consistent surrogate.

The model lets you just specify water utilization and outlet temperature.  These should be readily matched by a more detailed model, and the detailed model can be used for off-design cases.  

### Legal Acknowledgement


idaes/power_generation/unit_models/soec_design.py
```diff
@@ -0,0 +1,621 @@
...
+from idaes.power_generation.properties.natural_gas_PR import (
+    get_prop, get_rxn, EosType)
+from idaes.core.util.exceptions import ConfigurationError, InitializationError
+import idaes.logger as idaeslog
+
+
+@declare_process_block_class(
+    "SoecDesign", doc="Simple SOEC model for process design.")
+class SoecDesignData(UnitModelBlockData):
+    """Simple 0D SOEC model. This model can be used to develop a flowsheet at the
```
andrewlee94, line 43: I have been asking my self if "design" is the best name for this, but I am not sure of a better alternative. Looking at the core libraries for inspiration, I would have said this would be just `SOEC`, and then the more detailed models would derive from this (as appropriate) and specify more details in their names,

idaes/power_generation/unit_models/soec_design.py
```diff
@@ -0,0 +1,621 @@
...
+
+    def build(self):
+        """Construct a unit model data block"""
+        super().build()
+
+        # Do a little validation.
+        # The hydrogen side needs water and hydrogen
+        if set(self.config.hydrogen_side_package.component_list) != {"H2", "H2O"}:
+            raise ConfigurationError(
+                "SOEC hydrogen side must contain exactly H2 and H2O"
```
andrewlee94, line 116: Just an observation, but this is a very specific set of components, so I would almost suggest we call this `H2SOEC`. It is also a prime candidate for the new "devel" folder due to being fairly specific.

andrewlee94 (commented):
A couple of observations, but I don't know enough about SOECs to be the first reviewer here.

idaes/power_generation/unit_models/soec_design.py
dallan-keylogic: For a real 0D surrogate, we'd want to have some equation that touches on the voltage-current relationship, and at present this model doesn't have that.

idaes/power_generation/unit_models/soec_design.py
eslickj: I'm not sure.  It's a specific unit for some power generation like task, but hopefully it can be finished.  I don't want it to be in perpetual development.  Does devel mean in development and not fully supported, or is it anything that's not core and totally generic?

idaes/power_generation/unit_models/soec_design.py
eslickj: @dallan-keylogic, that's right.  You'd have to add that on.  You could inherit this.  But that's what I mean about the 0D name.  I don't want to imply that is more than just a simple design point calculation.

idaes/power_generation/unit_models/soec_design.py
andrewlee94: That is still up for debate (i.e. we might need two separate folders).

eslickj:
I'm going to fix a scaling issue or two in generic units as I'm testing this more.  Best wait to merge this until after the Feb release to be safe.

idaes/power_generation/unit_models/soec_design.py
```diff
@@ -0,0 +1,686 @@
...
+        )
+        self.o2_seperator = um.Separator(
+            default={
+                "property_package": self.electrolysis_prop_params,
+                "split_basis": um.SplittingType.componentFlow,
+                "outlet_list": ["h2_strm", "o2_strm"],
+            }
+        )
+        self.h2_outlet_translator = um.Translator(
+            doc="Translate hydrogen side properties to electrolysis properties",
```
dallan-keylogic, line 185: Incorrect doc

idaes/power_generation/unit_models/soec_design.py
```diff
@@ -0,0 +1,686 @@
...
+            default={
+                "inlet_property_package": self.electrolysis_prop_params,
+                "outlet_property_package": self.config.hydrogen_side_package,
+                "outlet_property_package_args": self.config.hydrogen_side_package_args,
+                "outlet_state_defined": False
+            }
+        )
+        self.o2_translator = um.Translator(
+            doc="Translate hydrogen side properties to electrolysis properties",
```
dallan-keylogic, line 194: Incorrect doc

idaes/power_generation/unit_models/soec_design.py
```diff
@@ -0,0 +1,686 @@
...
+            source=self.electrolysis_reactor.outlet,
+            destination=self.o2_seperator.inlet,
+        )
+        self.strm3 = Arc(
+            doc="Oxygen seperation to hydrogen side outlet translator",
+            source=self.o2_seperator.h2_strm,
+            destination=self.h2_outlet_translator.inlet,
+        )
+        self.strm4 = Arc(
+            doc="Oxygen seperation to O2 to sweep translator",
```
dallan-keylogic, line 236: Awk phrasing

idaes/power_generation/unit_models/soec_design.py
```diff
@@ -0,0 +1,686 @@
...
+            return (
+                b.electrolysis_reactor.control_volume.properties_out[t].flow_mol *
+                b.electrolysis_reactor.control_volume.properties_out[
+                    t
+                ].mole_frac_comp["H2O"] ==
+                b.electrolysis_reactor.control_volume.properties_in[t].flow_mol *
+                b.electrolysis_reactor.control_volume.properties_in[
+                    t
+                ].mole_frac_comp["H2O"] *
+                (1.0 - b.water_utilization[t])
```
dallan-keylogic, line 300: Minor, but why not just use `flow_mol_comp` here?

dallan-keylogic (approved):

idaes/power_generation/unit_models/soec_design.py
```diff
@@ -0,0 +1,686 @@
...
+    CONFIG.declare(
+        "oxygen_side_package_args",
+        ConfigBlock(
+            implicit=True,
+            description="Property package arguments for the oxygen side.",
+            doc="Property package arguments for the oxygen side."
+        )
+    )
+    CONFIG.declare(
+        "hydrogen_side_package_args",
```
andrewlee94, line 80: As above.

andrewlee94 (dismissed):
I didn't look too much at the actual model form but I had a few minor comments, nothing critical however.

idaes/power_generation/unit_models/soec_design.py
eslickj: No reason.  Good call that's a bit cleaner.

eslickj:
@andrewlee94 and @dallan-keylogic, I addressed all the comments by making the requested changes.  Should be good to go.

andrewlee94 (approved):

dallan-keylogic (approved):

PR #724, opened 2022-02-23
IDAES v2.0: Initialization changes

andrewlee94:
## Summary/Motivation:
This PR updates unit model initialization to have a central, common `initialize` method which handles general steps for initialization and calls out to unit-specific `initialize_unit` methods to handle the actual initialization of each unit model. Unit models should thus implement an `initialize_unit` method instead of overloading `initialize` from now on. The core unit models have been updated to reflect this (with the exception of `SkeletonUnit` which is an oddity), and a deprecation warning put in place.

- Rename `UnitModelBlockData` `initialize` method to `initialize_unit`.
- Add new common `initialize` method to `UnitModelBlockData`, and include basic hooks for handling costing methods (including both existing methods and upcoming improvements).
- Add a check in `UnitModelBlockData` for overloading `initialize` which logs a deprecation warning,
- Update all core unit models with custom `initialize` methods by renaming to `initialize_unit` (and fixing some internal calls to `super().initialize`.
- Remove all references to `costing` in `initialize_unit` methods, as this is now handled in the core `initialize` method.
- Fix some tests in Boiler Heat Exchanger that were looking for expected warnings in logger.

### Legal Acknowledgement


eslickj:
Why are we doing this?

andrewlee94:
@eslickj That's right - you aren't part of the costing discussion. The main driver is that we want costing blocks to live on the associated unit model, which means that the unit model `initialize` method needs to be aware of this and what to do with it (deactivate costing, initialize the unit, re-activate costing and then initialize costing). 

Going by our current code, we cannot rely on the unit model developer to handle this, so we want to have a single core method that will handle costing (and any other future extensions), and leave the unit model developer to only worry about the initialization of the unit model itself. Hence, breaking initialization into two parts - the core `initialize` method and the unit-specific `initialize_unit` part.

Further ,we think that this gives us a lot more capability in going forwards for initialization, by giving us a place where we can introduce new common functionality without needing to change all the unit models.

eslickj:
@andrewlee94 okay.  Seems good.  I looked it over and it looks good.  Just one more question.  Have you thought about parallels with other types of process blocks?  Does this mean flowsheets property models etc. still overload ``initialize``?  Seems like things would be easier to learn and use if everything is the same as much as possible.  Should there be a similar system for other types of blocks?

This doesn't really matter to me.  It's just a thought.

eslickj (approved):

andrewlee94:
@eslickj We hadn't really thought of other cases yet, but we could do the same there as needed. However, at this point I only know of a user case for unit models.

idaes/core/unit_model.py
```diff
@@ -92,6 +92,20 @@ def build(self):
         """
         super(UnitModelBlockData, self).build()
 
+        # Add a placeholder for referecnes to costing blocks
+        self._costing_block_ref = None
+
+        # Check for overloading of initialize method
+        # TODO: Remove in IDAES v2.0
+        if type(self).initialize != UnitModelBlockData.initialize:
```
jsiirola, line 100: Performance note: we re-test this for *every* instance of *every* unit model.  This could add up.  I am not sure if I know of a better approach, though.

idaes/core/unit_model.py
```diff
@@ -590,8 +604,60 @@ def _get_stream_table_contents(self, time_point=0):
...
+        """
+        # Get any arguments for costing if provided
+        cost_args = kwargs.pop("costing_args", {})
+
+        # Get the costing block if present
+        # TODO: Clean up in IDAES v2.0
+        if hasattr(blk, "costing"):
+            costing = blk.costing
+        else:
+            costing = blk._costing_block_ref
```
jsiirola, line 635: Shouldn't this be the other way around?  That is, the required name `_costing_ref` should be the definitive pointer to the costing block, and fall back on the old-style `costing` if it's not there?

idaes/generic_models/unit_models/skeleton_model.py
```diff
@@ -117,6 +117,7 @@ def add_ports(self, name, member_dict, doc=None):
             p.add(member_dict[k], name=k)
 
 
+    # TODO : Work out how to make this work iwth new UnitModel initialization
```
jsiirola, line 120: Typo: with

jsiirola (commented):
A couple typos and one logical question, but otherwise looks good.

idaes/core/unit_model.py
eslickj: There aren't usually a lot of unit models.  I doubt it would have any significant impact.

eslickj (dismissed):

jsiirola (approved):

idaes/core/unit_model.py
```diff
@@ -590,8 +604,60 @@ def _get_stream_table_contents(self, time_point=0):
...
+        modeling components attached to them, such as costing blocks. First,
+        any attached components are deactivated, after which the
+        initialize_unit method is called. Following this, the attached
+        components are reactivated and theri initialize method called.
+
+        Currently, this method supports the following attached components:
+            * UnitModelCostingBlocks
+
+        Args:
+            costing_args - dict oarguments to be passed to costing block
```
Robbybp, line 622: typo: arguments

andrewlee94: I'll fix this in a subsequent PR.

eslickj (approved):

idaes/core/unit_model.py
```diff
@@ -92,6 +92,20 @@ def build(self):
         """
         super(UnitModelBlockData, self).build()
 
+        # Add a placeholder for initialization order
+        self._initialization_order = []
+
+        # Check for overloading of initialize method
+        # TODO: Remove in IDAES v2.0
+        if type(self).initialize is not UnitModelBlockData.initialize:
```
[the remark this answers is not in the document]
andrewlee94: I think @jsiirola was referring to the fact that this test would be triggered every time an instance of a model is created.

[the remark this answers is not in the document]
eslickj: Yeah, I guess I was thinking making the test faster would help.  Still probably not a significant issue, but we could check this in ```declare_process_block_class()``` if you want to get serious and check once.  None of it is probably worth the effort.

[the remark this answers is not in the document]
andrewlee94: Especially seeing as this is just backwards compatibility until IDAES v2.

eslickj (approved):

jsiirola (approved):

Issue #537, opened 2021-09-30
Revise workflow for running test suite with user installation


lbianchi-lbl:
- Having (a subset of) the IDAES test suite runnable when IDAES is installed in user mode (without cloning the repo) is valuable, as it provides an effective "smoke test" to verify that the installation was successful
- The relevant command (`pytest --pyargs idaes`) is already mentioned in the documentation; however, some tests require extra dependencies (including `pytest` itself) that are considered development dependencies and described accordingly in the "Advanced User Guide"
- This might be misleading if users try to run the test suite after installing IDAES without cloning the repo, as in e.g. #530 
- My suggestion would be to work on both fronts (test suite configuration, and documentation) so that the complete workflow works and is documented properly

Issue #530, opened 2021-09-29
Installation - pytest fails


REI-Gardner:
I am on a windows machine and I have installed idaes-pse for the first time. However, when I do the pytest command as specified in the online idaes documentation I get the error message shown below.

I have opened the tutorials and when I do: 
`run "notebook_test_script.py"`
I get that everything passes. 

Is it likely that I will run into any future because of the problem causing the error message from the pytest? Is there a simple fix to resolve the pytest fail?

![idaes error](https://user-images.githubusercontent.com/90857416/135337210-cfda51ce-ba84-4a23-b757-acb6f32c13fc.PNG)

lbianchi-lbl:
Hi @REI-Gardner, the specific error reported in this issue can be fixed by installing the missing `addheader` package by running `pip install addheader` in the same `idaes` Conda environment.

More in general, `pytest` and `addheader` are both considered development dependencies for IDAES. The development dependencies can be installed after cloning this repository from GitHub following the "Installation with GitHub" section of the [Advanced User Installation](https://idaes-pse.readthedocs.io/en/stable/advanced_user_guide/advanced_install/index.html) guide in the IDAES documentation. In summary:

```sh
git clone https://github.com/IDAES/idaes-pse.git && cd idaes-pse
conda activate idaes
pip install -r requirements-dev.txt
pytest
```

lbianchi-lbl:
@REI-Gardner I've just opened #537 so that we can improve this workflow. Let me know if the steps described in my previous comment didn't work, otherwise we can probably close this issue and keep track of this there.

PR #738, opened 2022-03-02
Ensure that pytest is the only strict requirement for running test suite for all installation modes

### LINKED ISSUE #740: Failing tests with conda install of IDAES
A Watertap user just came to me with an issue about failing tests with a new install of IDAES when using a conda install.

error_image

If I were to take a guess, it is because `omlt` does not have a conda package available, and thus does not get included in the conda install of IDAES. However, `omlt` is required by some of the tests, and thus the entire test run fails when installing via conda unless the user goes and gets `omlt` manually.

Based on this, I think we need to either:

1. Fix the conda install so that `omlt` is included (which might require an `omlt` conda package).
2. Add conditional imports on `omlt` in the code to avoid the test failures if `omlt` is not present.

lbianchi-lbl:
## Fixes #740, #537, #530 (and possibly many others)

- Test failures caused by `addheader/omlt not available` errors

## Summary/Motivation:

- Being able to run (a reasonable subset of) the IDAES test suite regardless of how it was installed is very valuable to validate installation
- We want this process to be as reliable and simple as possible
- Therefore, for simplicity, we should implement a policy that the test suite should run without errors having only `pytest` as a strict requirement
- In other words, any test that requires extra dependencies should be seamlessly skipped when that dependency is not available

- Add `pytest.importorskip()` to skip `test_headers` module if `addheader` package is not available
- Move `omlt` to `optional` dependencies similarly to `tensorflow` in `setup.py` to match Conda package
- Add optional import logic to `keras_surrogate` and `test_keras_surrogate` to account for `omlt` being optional
- Update documentation and CI check for site-packages/non-editable/"user mode" installation to use `pip install pytest` as only extra step

### Legal Acknowledgement

### DIFF
--- a/idaes/surrogate/tests/test_keras_surrogate.py
+++ b/idaes/surrogate/tests/test_keras_surrogate.py
@@ -15,6 +15,9 @@ Tests for KerasSurrogate
 """
 import pytest
 
+pytest.importorskip("tensorflow.keras", reason="tensorflow.keras not available")
+pytest.importorskip("omlt", reason="omlt not available")
+
 import os.path
 import pandas as pd
 from pyomo.common.fileutils import this_file_dir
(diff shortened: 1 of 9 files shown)


idaes/surrogate/tests/test_keras_surrogate.py
```diff
@@ -15,19 +15,20 @@
 """
 import pytest
 
+pytest.importorskip("tensorflow.keras", reason="tensorflow.keras not available")
+pytest.importorskip("omlt", reason="omlt not available")
+
```
carldlaird, line 20: I am surprised that this works with Pyomo's attempt_import stuff. Is there some magic going on under the hood?

carldlaird (approved):
This looks reasonable to me. Thanks for this. I am happy merging once the tests are complete.

PR #741, opened 2022-03-04
New IDAES costing framework

andrewlee94:
## Summary/Motivation:
This PR introduces the core code for the new IDAES costing framework and updates the old general costing package to the new format. It also adds deprecation warnings to all the old costing methods.

- Deprecate old costing framework
- Add new `idaes/generic_models/costing` folder
- Adds new `costing_base.py` module with code code and tests.
- Adds new `SSLW.py` module with updated general costing package and tests.

### Legal Acknowledgement


andrewlee94:
@MAZamarripa You might want to check the changes made to the power plant costing - they are minor but I had to assign a year to the currency in the code to get the tests to play nice. You should double check the year I assigned (2018) to make sure there are no unexpected side effects.

andrewlee94:
@MAZamarripa Also note that I changed the CE index values for 2018 and 2019 based on values I found online. These were notably lower than the values we had in the original costing module, so there will be some changes in results if reported in these years.

idaes/generic_models/costing/costing_base.py
```diff
@@ -0,0 +1,598 @@
...
+        # Clean up references from fsb and unit_model
+        try:
+            fcb._registered_unit_costing.remove(self)
+        except ValueError:
+            # Assume this is OK
+            pass
+        try:
+            self.unit_model._initialization_order.remove(self)
+        except ValueError:
+            # Assume this is OK
```
jsiirola, line 597: Is this a good assumption?  I think we might want to not suppress this exception (this kind of thing has burned me in the past).

idaes/generic_models/costing/SSLW.py
```diff
@@ -0,0 +1,1493 @@
...
+from idaes.core.util.misc import StrEnum
+
+from idaes.generic_models.costing.costing_base import (
+    FlowsheetCostingBlockData, register_idaes_currency_units)
+
+# Some more information about this module
+__author__ = "Miguel Zamarripa, Andrew Lee"
+
+
+class HXType(StrEnum):
```
jsiirola, line 58: General comment on these Enums: I would (somewhat strongly) advocate the symbol and the string be the same.

idaes/generic_models/costing/SSLW.py
```diff
@@ -0,0 +1,1493 @@
...
+            HXType.Utube: {1: 11.3852, 2: 0.9186, 3: 0.09790},
+            HXType.kettle_vap: {1: 12.2052, 2: 0.8709, 3: 0.09005}}
+        alpha = alpha_dict[hx_type]
+
+        # Convert area to square feet
+        area_unit = (pyo.units.convert(blk.unit_model.area,
+                                       to_units=pyo.units.ft**2) /
+                     blk.number_of_units)
+
+        def hx_cost_rule(blk):
```
jsiirola, line 321: Why use decorators for the Expression, but not here for the Constraint?

jsiirola (commented):
This looks pretty good.  A couple comments (mostly that you can either use or ignore).

idaes/generic_models/costing/SSLW.py
andrewlee94: I was somewhat conflicted between backwards compatibility and more standard names here, and probably got stuck in the middle. We probably should just go all the way and make the symbol and string be the same.

I assume you would advocate they be the same down to case as well (i.e. `pumpType` should render as `"pumpType"` and not `"pump_type"`)?

idaes/core/unit_model.py
```diff
@@ -727,3 +727,19 @@ def initialize_build(blk, state_args=None, outlvl=idaeslog.NOTSET,
...
         return None
+
+    def del_component(self, name_or_object):
+        """
+        Delete a component from this block. Need to introduce code to handle
+        un-registering costing blocks
+        """
+        obj = self.component(name_or_object)
+
+        # TODO: See if Pyomo can give us a call-back to do this
```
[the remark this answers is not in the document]
andrewlee94: https://github.com/Pyomo/pyomo/issues/2337

idaes/generic_models/costing/costing_base.py
```diff
@@ -0,0 +1,598 @@
...
+                    "equal to or greater than zero to avoid negative costs.")
+        else:
+            # Get bounds from expression
+            ebounds = compute_bounds_on_expr(flow_expr)
+            if ebounds[0] is None or ebounds[0] < 0:
+                _log.warning(
+                    "flow_expr is an expression with a lower bound of less "
+                    "than zero. Costing requires that all flows have a lower "
+                    "bound equal to or greater than zero to avoid negative "
+                    "costs.")
```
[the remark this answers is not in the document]
andrewlee94: I tired this, but I got the following error:

`AttributeError: 'IndexedVar' object has no attribute 'is_potentially_variable'`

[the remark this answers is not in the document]
jsiirola: OK - you need to catch things in a slightly different order:
```python
if type(flow_expr) in pyo.native_types:
    # this is a constant?!?
    pass
elif flow_expr.is_indexed():
    raise TypeError(
        f"{flow_expr.name} is an indexed component. Flow costing only "
        "supports unindexed components.")
elif not flow_expr.is_potentially_variable():
    # this is a constant?!?
    pass
#...
```

idaes/generic_models/costing/SSLW.py
jsiirola: Yeah - I think I would make them the same (characters and case) -- unless there was a really good reason to make them different (`20ft` comes to mind).

idaes/generic_models/costing/costing_base.py
andrewlee94: I think I fixed this.

idaes/generic_models/costing/SSLW.py
andrewlee94: That is how the original costing code was written, and I was trying to keep my effort to a minimum.

andrewlee94:
@jsiirola I think I've addressed all your comments, so this should be ready for re-review.

jsiirola (approved):
This looks good!

eslickj (approved):

PR #750, opened 2022-03-15
PySMO: Fix pandas deprecation warnings 

OOAmusat:
## Fixes
Pandas deprecation warnings for `pd.Series()` and `series.append()` in PySMO

## Summary/Motivation:
New changes in Pandas this year led to a number of deprecation warnings popping up in PySMO. This PR fixes these minor issues.

- Specify `pd.Series()` dtype as `float64`
- Replace pandas append with `pd.concat()`

### Legal Acknowledgement


jsiirola (approved):

idaes/surrogate/pysmo/polynomial_regression.py
```diff
@@ -949,20 +949,18 @@ def results_generation(self, beta, order):
...
         This function prints the results of the fitting to the screen.
         """
-        results_df = pd.Series()
+        results_df = pd.Series(dtype='float64')
         counter = 1
         print("\n------------------------------------------------------------")
         print("The final coefficients of the regression terms are: \n")
         print("k               |", beta[0, 0])
-        results_df = results_df.append(pd.Series({"k": beta[0, 0]}))
+        results_df = pd.concat([results_df, pd.Series({"k": beta[0, 0]})])
```
lbianchi-lbl, line 957: I would recommend being more explicit and specifying the `axis` kwarg to `pd.concat()`.

idaes/surrogate/pysmo/polynomial_regression.py
```diff
@@ -949,20 +949,18 @@ def results_generation(self, beta, order):
         """
         This function prints the results of the fitting to the screen.
         """
-        results_df = pd.Series()
+        results_df = pd.Series(dtype='float64')
```
lbianchi-lbl, line 952: This is more of a general observation, but do we need `results_df` to be a DataFrame while we collect the results? From what I understand in the function, it should be possible to use a simple list to accumulate each item as a `pd.Series` (or even as the underlying `dict` literal), and then convert the list of `pd.Series`/`dict` to a DataFrame at the end:

```py
result_items = []
for i in range(...):
    col_name = f"..."
    result_items.append({col_name: beta[i, 0]})
return pd.DataFrame(result_items)
```

Note that I might have misunderstood the expected shape of the dataframe (I can never picture what `pd.concat()` does for all possible inputs, see the comment below), in which case I'd recommend to add a comment or a docstring that shows what's the expected format of the dataframe (e.g. just call the function manually and copy `df.head(5)` in the docstring)

lbianchi-lbl (commented):
I've left a couple of general comments. However, if this should be merged sooner rather than later and we're confident that the output of this function is sufficiently tested, feel free to dismiss the suggestion and I'll approve as-is.

idaes/surrogate/pysmo/polynomial_regression.py
OOAmusat: Done

lbianchi-lbl (approved):

OOAmusat:
@jsiirola , minor changes dismissed your review. Could you please re-approve? Thanks.

blnicho (approved):

PR #756, opened 2022-03-16
Bubbling Fluidized Bed Updates Scaling and Units

bpaul4:
## Fixes
This adds updated scaling and units to the CLC bubbling fluidized bed model.

- New scaling and units definitions for the bubbling fluidized bed model
- Will add scaling tests pending updates to #681 to address any gaps in code coverage
- Minor documentation fix

### Legal Acknowledgement


idaes/gas_solid_contactors/unit_models/bubbling_fluidized_bed.py
```diff
@@ -2087,9 +2249,79 @@ def initialize(blk, gas_phase_state_args=None, solid_phase_state_args=None,
                         idaeslog.condition(results))
                         )
         else:
-            init_log.warning('{} Initialisation Step 6 Failed.'
+            init_log.warning('{} Initialization Step 6 Failed.'
                              .format(blk.name))
```
andrewlee94, line 2253: We have a new standard where initialization routines should raise an `InitializationError` (new error type) if the final step fails. This is to allow other models/flowsheets to be aware when a unit model (or property model, etc.) fails to initialize properly (which used to pass quietly in the background).

idaes/gas_solid_contactors/unit_models/bubbling_fluidized_bed.py
```diff
@@ -1518,7 +1652,7 @@ def initialize(blk, gas_phase_state_args=None, solid_phase_state_args=None,
                         property package(s) to provide an initial state for
```
andrewlee94, line 1652: Moving towards IDAES 2.0, we are renaming this method to `initialize_build`. There is a general purpose `initialize` method in the parent class which will automatically call `initialize_build`, so nothing changes for the end user. However, this allows us to automate handling of things that get added on to the unit model (such as costing blocks).

andrewlee94 (commented):
I can't speak to the details (not enough time to go through in detail and too long since I worked on this model), but I did have a couple of comments on new IDAES standards we should include in this PR if possible (they should be easy changes).

My only other comment is that I was surprised to see changes in the model results for something that apparently only changes scaling and units definitions. Is this due to changes in the units for certain variable, or did something more significant change?

idaes/gas_solid_contactors/unit_models/bubbling_fluidized_bed.py
bpaul4: Are there any potential compatibility issues with renaming this now? I know @chineduobiora would like to reference these IDAES models in external flowsheets, so although tedious it may be better to open a future PR to update all unit models, flowsheets and tests together.

bpaul4: @chineduobiora I can make the changes Andrew suggests in the MB model now since we have an open PR to push commits to. For the FB0D model, I believe it would be best to open a new PR at some point to add these updates along with scaling/tests for consistency with the other CLC models.

chineduobiora: @bpaul4 sounds good

idaes/gas_solid_contactors/unit_models/bubbling_fluidized_bed.py
```diff
@@ -650,93 +688,106 @@ def _make_vars_params(self):
...
         self.Kd = Var(domain=Reals,
                       initialize=1,
-                      doc='Bulk Gas Permeation Coefficient [m/s]')
+                      doc='Bulk Gas Permeation Coefficient',
+                      units=units_meta_gas('velocity'))
         self.Kd.fix()
         self.deltaP_orifice = Var(domain=Reals,
-                                  initialize=3.4E5,
-                                  doc='Pressure Drop Across Orifice [Pa]')
+                                  initialize=3.4E4,
```
chineduobiora, line 788: This should be 3.4E3 Pa

idaes/gas_solid_contactors/unit_models/tests/test_BFB.py
```diff
@@ -243,28 +244,31 @@ def test_solve(self, iron_oc):
...
         assert (pytest.approx(0.307, abs=1e-2) ==
+                iron_oc.fs.unit.delta[0, 0].value)
+        assert (pytest.approx(0.384, abs=1e-2) ==
                 iron_oc.fs.unit.delta[0, 1].value)
-        assert (pytest.approx(124169.5118, abs=1e-2) ==
+        assert (pytest.approx(100671.070, abs=1e-2) ==
+                iron_oc.fs.unit.gas_outlet.pressure[0].value)
+        assert (pytest.approx(85328.9301, abs=1e-2) ==
+                iron_oc.fs.unit.gas_inlet.pressure[0].value -
                 iron_oc.fs.unit.gas_outlet.pressure[0].value)
```
chineduobiora, line 267: These tests results (and potentially some other tests results) will change when you update the orifice pressure drop to 3.4E3 Pa.

idaes/gas_solid_contactors/unit_models/bubbling_fluidized_bed.py
```diff
@@ -2087,9 +2249,79 @@ def initialize(blk, gas_phase_state_args=None, solid_phase_state_args=None,
                         idaeslog.condition(results))
                         )
         else:
-            init_log.warning('{} Initialisation Step 6 Failed.'
+            init_log.warning('{} Initialization Step 6 Failed.'
                              .format(blk.name))
 
+    def calculate_scaling_factors(self, b, x):
```
chineduobiora, line 2255: I will include comments for the scaling when it is updated.

chineduobiora (changes requested):

idaes/gas_solid_contactors/unit_models/bubbling_fluidized_bed.py
chineduobiora: > I can't speak to the details (not enough time to go through in detail and too long since I worked on this model), but I did have a couple of comments on new IDAES standards we should include in this PR if possible (they should be easy changes).
> 
> My only other comment is that I was surprised to see changes in the model results for something that apparently only changes scaling and units definitions. Is this due to changes in the units for certain variable, or did something more significant change?

I think the most significant changes to the model results are because of the changes to the value of the pressure drop across the orifice. The correct value to use 3.4E3 Pa.

andrewlee94:
@bpaul4 @chineduobiora Any idea on the timeline for getting this merged?

I am starting to do a reorganization of the repository, and all the gas-solids contactor models are slated to be moved in a new `idaes/models_extra` folder as part of this in the near future. However, this has the potential to create a lot of conflicts in this PR (although it might not), so I am hoping we can get this merged quickly so that I can do a clean move for you.

chineduobiora:
> @bpaul4 @chineduobiora Any idea on the timeline for getting this merged?
> 
> I am starting to do a reorganization of the repository, and all the gas-solids contactor models are slated to be moved in a new `idaes/models_extra` folder as part of this in the near future. However, this has the potential to create a lot of conflicts in this PR (although it might not), so I am hoping we can get this merged quickly so that I can do a clean move for you.

Our target was the end of this month. I think we are really close. Updating the scaling method to use better scaling defaults for constraints is the last major thing left.

bpaul4:
> > @bpaul4 @chineduobiora Any idea on the timeline for getting this merged?
> > I am starting to do a reorganization of the repository, and all the gas-solids contactor models are slated to be moved in a new `idaes/models_extra` folder as part of this in the near future. However, this has the potential to create a lot of conflicts in this PR (although it might not), so I am hoping we can get this merged quickly so that I can do a clean move for you.
> 
> Our target was the end of this month. I think we are really close. Updating the scaling method to use better scaling defaults for constraints is the last major thing left.

Thank you @andrewlee94 for letting us know. The updates to the MB and BFB are close to completion, and updating the FB0D model for your initialization updates can wait until after the reorganization if needed.

bpaul4:
@andrewlee94 @chineduobiora merging our changes on the MB side and temporary BFB fixes was surprisingly clean, and I updated scaling and solution checks as needed. I reviewed the entire gas_solid_contactors directory comparing versions before/after merging from main and confirmed that the commits from other pull requests successfully merged into this branch.

idaes/gas_solid_contactors/unit_models/bubbling_fluidized_bed.py
```diff
@@ -1103,13 +1236,16 @@ def bubble_cloud_bulk_heat_trans(b, t, x):
...
+                conc_diff_units = (units_meta_gas('amount'))
                 return (
                     b.Hgbulk[t, x] * b.bubble_diameter[t, x] ==
-                    6 * b.Kd * b.delta[t, x] * b.bed_area *
+                    6 * b.Kd * b.delta[t, x] *
+                    pyunits.convert(b.bed_area,
+                                    to_units=units_meta_gas('area')) *
                     (b.gas_emulsion.properties[t, x].enth_mol *
-                     smooth_max(conc_diff, 0, b.eps) +
+                     smooth_max(value(conc_diff), 0, b.eps)*conc_diff_units +
```
andrewlee94, line 1246: This will not work - `value` returns a single value using the current value of the expression, so this effectively becomes a constant rather than an expression.

In this case, both `0` and `eps` should have the same units as `conc_diff` and you should just do `smooth_max(conc_diff, 0, b.eps)`

idaes/gas_solid_contactors/unit_models/bubbling_fluidized_bed.py
```diff
@@ -1165,20 +1301,26 @@ def bubble_heat_transfer(b, t, x):
                              self.length_domain,
                              doc="Gas Emulsion - Heat Transfer")
             def gas_emulsion_heat_transfer(b, t, x):
-                return (b.gas_emulsion.heat[t, x] ==
+                return (b.gas_emulsion.heat[t, x]*pyunits.m**3 ==
```
andrewlee94, line 1304: 1. Why are you multiplying by `m**3` here? This implies that units are not consistent in this constraint.
2. Hardcoding `pyunits.m` here presumes that length is defined in metres - you should always go via the units meta data to get units.

idaes/gas_solid_contactors/unit_models/bubbling_fluidized_bed.py
```diff
@@ -1021,12 +1125,16 @@ def bubble_cloud_mass_trans_coeff(b, t, x, j):
...
+            conc_diff_units = (units_meta_gas('amount') /
+                               units_meta_gas('volume'))
             return (b.Kgbulk_c[t, x, j] * b.bubble_diameter[t, x] ==
-                    6 * b.Kd * b.delta[t, x] * b.bed_area *
+                    6 * b.Kd * b.delta[t, x] *
+                    pyunits.convert(b.bed_area,
+                                    to_units=units_meta_gas('area')) *
                     (b.gas_emulsion.properties[t, x].mole_frac_comp[j] *
-                     smooth_max(conc_diff, 0, b.eps) +
+                     smooth_max(value(conc_diff), 0, b.eps)*conc_diff_units +
```
andrewlee94, line 1135: AS other comment about why this won't work.

idaes/gas_solid_contactors/unit_models/bubbling_fluidized_bed.py
```diff
@@ -1073,11 +1191,20 @@ def bubble_cloud_heat_trans_coeff(b, t, x):
...
-                    1e6 * 0.03 * b.gas_emulsion.properties[t, x].therm_cond *
-                    ((b._reform_var_5[t, x]**2 + b.eps)**0.5) ** 1.3)
+                    b.htc_conv[t, x] *
+                    ((pyunits.mol/pyunits.kg)**1.3) *
+                    pyunits.convert(b.solid_emulsion.properties[t, x]
+                                    ._params.particle_dia,
+                                    to_units=units_meta_gas('length')) ==
+                    0.03 *
+                    b.gas_emulsion.properties[t, x].therm_cond *
+                    ((b._reform_var_5[t, x]**2 + b.eps*reform_var_5_units**2
```
andrewlee94, line 1206: 1. Why does `eps` appear here?
2. Note that once `eps` gets units added you will need different `eps` objects for each place it is used.

idaes/gas_solid_contactors/unit_models/bubbling_fluidized_bed.py
```diff
@@ -1142,19 +1278,19 @@ def gas_emulsion_mass_transfer(b, t, x, j):
...
+                        gas_emulsion_hetero_rxn_term(b, t, x, j)))
 
         if self.config.energy_balance_type != EnergyBalanceType.none:
             # Bubble - heat transfer
             @self.Constraint(self.flowsheet().time,
                              self.length_domain,
                              doc="Bubble - Heat Transfer")
             def bubble_heat_transfer(b, t, x):
-                return (b.bubble.heat[t, x] == b.Hgbulk[t, x] -
+                return (b.bubble.heat[t, x] * pyunits.m**3 == b.Hgbulk[t, x] -
```
andrewlee94, line 1293: Same as comment below.

idaes/gas_solid_contactors/unit_models/bubbling_fluidized_bed.py
```diff
@@ -1832,14 +2009,14 @@ def initialize(blk, gas_phase_state_args=None, solid_phase_state_args=None,
         init_log.info('Initialize Mass Balances')
         init_log.info_high('initialize mass balances with no reactions')
         with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
-            results = opt.solve(blk, tee=slc.tee)
+            results = opt.solve(blk, tee=True)
```
andrewlee94, line 2012: You should not be hardcoding `tee` here - use `slc.tee`.

idaes/gas_solid_contactors/unit_models/bubbling_fluidized_bed.py
```diff
@@ -1893,11 +2070,11 @@ def initialize(blk, gas_phase_state_args=None, solid_phase_state_args=None,
                             Constraint, descend_into=False):
                         c.activate()
 
-            blk.bubble.reactions.initialize(outlvl=outlvl,
+            blk.bubble.reactions.initialize(outlvl=idaeslog.DEBUG,
```
andrewlee94, line 2073: Do not hard code the output level - use the `outlvl` argument provided when initialization is called.

idaes/gas_solid_contactors/unit_models/bubbling_fluidized_bed.py
```diff
@@ -1947,22 +2124,22 @@ def initialize(blk, gas_phase_state_args=None, solid_phase_state_args=None,
...
+            blk.solid_emulsion.reactions.initialize(outlvl=idaeslog.DEBUG,
                                                     optarg=optarg,
                                                     solver=solver)
 
         if (gas_phase.reaction_package is not None or
                 solid_phase.reaction_package is not None):
             init_log.info_high('initialize mass balances with reactions')
             with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
-                results = opt.solve(blk, tee=slc.tee)
+                results = opt.solve(blk, tee=True)
```
andrewlee94, line 2135: AS above.

idaes/gas_solid_contactors/unit_models/tests/test_BFB.py
```diff
@@ -217,11 +220,269 @@ def test_build(self, iron_oc):
     def test_dof(self, iron_oc):
         assert degrees_of_freedom(iron_oc) == 0
 
+    @pytest.fixture(scope="class")
+    def iron_oc_unscaled(self, iron_oc):
+        import copy
+        m = copy.deepcopy(iron_oc)
```
andrewlee94, line 226: As I commented in the previous PR, I would suggest you just rebuild the model rather than copying.

andrewlee94 (changes requested):
I found a couple of things that will not work as intended and a few places where something was hardcoded that probably should not be.

Another general topic (that goes a little beyond this PR), but how is the default value for the `eps` parameters set? I suggest having a set of module level constants (defined just after imports) which provide the default values which you then use when initializing the `eps` Params. This makes it easier to change the default values later if required - you just need to go to the module level constants at the top of the file rather than searching through the code to find where each `eps` Param is actually declared. Seeing as I think you are about to see an increase in the number of `eps` Params, this would probably be a good idea to implement now.

idaes/gas_solid_contactors/unit_models/bubbling_fluidized_bed.py
bpaul4: I updated how the `eps` Param objects are defined to ensure unit consistency when we call the smoothing factors.

idaes/gas_solid_contactors/unit_models/bubbling_fluidized_bed.py
bpaul4: In regards to this and other related comments below, I re-checked and corrected units for the heat and mass transfer variables. This removed some of the floating unit multipliers, and I attached remaining ones to constants wrapped in `pyunit.convert` statements to use the metadata units.

idaes/gas_solid_contactors/unit_models/tests/test_BFB.py
```diff
@@ -507,23 +543,23 @@ def test_solve(self, iron_oc):
     def test_solution(self, iron_oc):
         assert (pytest.approx(0.140, abs=1e-2) ==
                 iron_oc.fs.unit.velocity_superficial_gas[0, 0].value)
-        assert (pytest.approx(1.349, abs=1e-2) ==
+        assert (pytest.approx(1.028, abs=1e-2) ==
```
andrewlee94, line 546: Are these changes expected?

idaes/gas_solid_contactors/unit_models/bubbling_fluidized_bed.py
```diff
@@ -560,10 +560,16 @@ def _make_vars_params(self):
             solid_phase.property_package.get_metadata().get_derived_units
 
         # Declare Mutable Parameters
-        self.eps = Param(mutable=True,
-                         default=1e-8,
-                         doc='Smoothing Factor for Smooth IF Statements',
-                         units=pyunits.dimensionless)
+        self.eps_bulk = Param(mutable=True,
+                              default=1e-8,
```
andrewlee94, line 564: I strongly suggest you create two new module level constants to define these values at the beginning of the module:

```
EPD_BULK = 1e-8
EPS_CONV = 1e-8
```

Then, use these constants to initialize the `Params` when you create them (i.e. `default=EPS_BULK`).

idaes/gas_solid_contactors/unit_models/tests/test_BFB.py
andrewlee94: Also, these tolerances are very loose - we should be checking to close to solver tolerance in most cases (~`rel=1e-5`).

bpaul4: I believe the new values are okay. In the last couple commits, I noticed that some scaling factors changed in magnitude. This may be due to the unit fixes that occurred earlier, although I'm not sure why the tests didn't fail then. The only change I made to the model equations was to correct a small error in how `reform_var_5` is smoothed on line 1228 in the most current file version.

chineduobiora: I've confirmed that the new values of the variables are correct. I cross checked against an older version of the model and they match. The changes are as a result of updating the orifice pressure drop from 3E5 Pa to 3E3 Pa.

andrewlee94 (approved):
As long as @chineduobiora is happy and approves, I think we can merge this.

chineduobiora (approved):
I updated some of the constraint scaling transformations, and added a test for the orthogonal collocation transformation method. Everything looks good and ready to merge. @bpaul4, thanks for all the work in getting this completed.

PR #761, opened 2022-03-22
Remove outdated gas turbine model

eslickj:
This removes the gas turbine model.  I don't think this should affect anyone.  I don't think it works with the NGCC branch anymore and the new NGCC model is here https://github.com/IDAES/examples-pse/pull/90.  I would deprecate this, but I don't think there is anyway a person could really be using this.

### Legal Acknowledgement


andrewlee94 (approved):
Looks good to me.

dallan-keylogic (approved):

Issue #757, opened 2022-03-17
Improvements for IDAES interface to PETSc


dallan-keylogic:
After some use of the PETSc integrator to initialize a Pyomo model discretized in time, I've run into a few situations where additional tools within IDAES could be helpful.

- [X]  Because PETSc doesn't return values of derivatives, `DerivativeVar`s are not initialized. It would be handy to have a function to automatically initialize them using, e.g., a backwards difference. --- Added in #774.

- [ ] Often we would like a Pyomo model to have a fine mesh, while only a coarse mesh is needed to run PETSc (because it integrates between time points anyway). Furthermore, if we use PETSc with a fine mesh, there is a huge amount of overhead in calling PETSc at every time point, PETSc determining an appropriate step size, then the results getting loaded back into the Pyomo model after PETSc is done time stepping. There are currently tools in IDAES to interpolate PETSc trajectories, but it would be nice to be able to automatically initialize a fine mesh using PETSc trajectories from a coarser mesh.

- [x] It would be nice to have a tool that automatically concatenated PETSc trajectories between different time points. Right now it saves a single file for the gap between every two time points.

eslickj:
@Robbybp, I think this would be useful.  We have the thing to do the PETSc trajectory interpolation.  We mainly need to map the scalar variables from the problem PETSc solves back to time indexed ones with I guess would be in another copy of the model with finer discretization.

blnicho:
> I've implemented a tool to interpolate a Pyomo DAE model onto a model with a more fine discretization grid. (Works for PD(A)Es too!) I use this to do discretization stability testing of my models. Would this be useful? What I've done is operating on Pyomo models, so it's completely agnostic of any Petsc trajectory.

@Robbybp is your tool available somewhere?

eslickj:
@blnicho, thanks that would probably be helpful.  I'm going to get to this in the next could days.

Robbybp:
The code I'm talking about ("tool" might be a little generous) is here:
https://github.com/IDAES/publications/tree/main/parker_dycops2022/parker_dycops2022/common/serialize
See approximately lines 300-330 of this file for an example of use:
https://github.com/IDAES/publications/blob/main/parker_dycops2022/parker_dycops2022/run_stability_analysis.py

Robbybp:
[...]

PR #774, opened 2022-03-29
Add utility to calculate time derivative vars.

eslickj:
## Fixes

This adds a utility function to calculate time derivative variables after solving with PETSc. This is part of #757.

### Legal Acknowledgement


idaes/core/solvers/petsc.py
```diff
@@ -519,6 +520,30 @@ def petsc_dae_by_time_element(
...
+            if time in ComponentSet(var.get_continuousset_list()):
+                parent = var.parent_block()
+                name = var.local_name + "_disc_eq"
+                disc_eq = getattr(parent, name)
+                for i, v in var.items():
+                    try:
+                        if disc_eq[i].active:
+                            calculate_variable_from_constraint(v, disc_eq[i])
+                    except KeyError:
+                        pass # discretization equation may not exist at first time
```
andrewlee94, line 544: Should we check that this is the first time point in this case, and raise an error if it is not?

andrewlee94 (commented):
Mostly looks good, but I have one comment on more rigorous checking of the case where the is no discretization equation.

idaes/core/solvers/petsc.py
eslickj: That should be covered by the KeyError exception.  The first time point shouldn't have a discretization equation.

viiibhav (approved):
Looks good to me

idaes/core/solvers/petsc.py
eslickj: Oh yeah, now I understand.  I'm not sure I have a clear way to know which ones should be missing with all the options.  I guess the ones that are missing should be missing.  The DAE transformation writes them so there probably not a lot of room for users to just do something weird.

eslickj:
I put the tests in an integration tests, so codecov didn't catch them.  I added tests to unit test too.

blnicho (approved):

viiibhav (approved):
LGTM

idaes/core/solvers/petsc.py
Robbybp: You could check this without try/except by just checking `i in disc_eq`

PR #780, opened 2022-04-01
Add enthalpy mixing constraint to pipeline node models

Robbybp:
## Fixes
Properly specifying degrees of freedom in pipeline networks, in particular with multiple inlets to a node, was very awkward due to some constant-temperature constraints being enforced. This PR replaces these constraints with an enthalpy mixing constraint. Some tests had to be updated, but I believe this makes a lot more sense overall.

### Legal Acknowledgement


michaelbynum (approved):
Looks good!

idaes/models_extra/gas_distribution/unit_models/node.py
```diff
@@ -363,6 +363,31 @@ def component_mixing_rule(b, t, j):
...
+        time = self.flowsheet().time
+
+        def get_enthalpy_rate(state):
+            return state.temperature * state.flow_mol * state.cp_mol
+
+        def enthalpy_mixing_rule(b, t):
+            supplies = [self.supplies[i].state[t] for i in self.supply_set]
+            inlets = [self.inlets[i].state[t] for i in self.inlet_set]
+            return (
+                sum(get_enthalpy_rate(supply) for supply in supplies)
```
andrewlee94, line 382: Looking at this, is `get_enthalpy_rate()` even required? This is what the `get_enthalpy_flow_terms` method in the property package is for.

idaes/models_extra/gas_distribution/unit_models/node.py
```diff
@@ -363,6 +363,31 @@ def component_mixing_rule(b, t, j):
...
+        """
+        Adds and equation that calculates temperature of the node from
+        temperatures, heat capacities, and flow rates of the supplies
+        and inlet pipelines.
+
+        """
+        time = self.flowsheet().time
+
+        def get_enthalpy_rate(state):
+            return state.temperature * state.flow_mol * state.cp_mol
```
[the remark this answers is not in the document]
Robbybp: Right now the unit model has a check to make sure the property package has only a single phase, which is called `"Vap"`. Generalizing the models to handle multiple phases is future work. I will move the calculations (including the Tref formulation) into the property package, however.

idaes/models_extra/gas_distribution/unit_models/node.py
Robbybp: It should not be required. I just haven't implemented `get_enthalpy_flow_terms` in this property package. Will update.

idaes/models_extra/gas_distribution/unit_models/node.py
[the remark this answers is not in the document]
Robbybp: @andrewlee94 For my future reference when we extend these models to support multiple phases, would a "total enthalpy" mixing equation or a "phase enthalpy" mixing equation make more sense for a system like this.

Robbybp:
@andrewlee94 @michaelbynum I just moved the enthalpy flow calculation to the property package. Can you give this another look?

idaes/models_extra/gas_distribution/unit_models/node.py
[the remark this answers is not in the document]
andrewlee94: I would generally say go with a total balance, unless you specifically need to consider each phase separately (e.g. different velocities for each phase).

idaes/models_extra/gas_distribution/properties/tests/test_natural_gas.py
```diff
@@ -135,6 +135,20 @@ def test_build(self):
             pyo.units.kmol / pyo.units.hr,
         )
 
+        enth_flow = m.state.get_enthalpy_flow_terms("Vap")
+        pred_enth_flow = pyo.value(
+            (293.15 - 298.15) * m.state.cp_mol * m.state.flow_mol
+        )
+        self.assertAlmostEqual(
```
andrewlee94, line 142: Generally comment, but we generally use `assert pytest.approx() == value`  for these cases rather than `unittest`.

andrewlee94 (approved):
Looks good to me.

idaes/models_extra/gas_distribution/properties/tests/test_natural_gas.py
Robbybp: Muscle memory I guess. I like `pyomo.common.unittest` because I have access to `assertStructuredAlmostEqual`. Don't think I used that anywhere here though.

Issue #745, opened 2022-03-10
Notify users when attempting to install IDAES extensions on incompatible hardware


lbianchi-lbl:
## Symptoms

- We've been trying to set up a developer environment on a macOS laptop with an Apple Silicon (ARM-based) processor
- Since macOS is not supported by IDAES, we're using an Ubuntu 20.04 virtual machine through Parallels
- The IDAES installation procedure seems to proceed without errors, including `idaes get-extensions`
- However, as soon as any of the IDAES executables are accessed, an error occurs (see watertap-org/watertap#398), e.g. `OSError: [Errno 8] Exec format error: '/home/xb-ubt20/.idaes/bin/ipopt'`

## Diagnosis

- The cause for the errors is most likely the fact that the ARM-based version of the Ubuntu guest OS is being installed (as opposed to the x86-64 version, which we know is compatible with the IDAES extensions; this is something we're trying to work out independently)

## Solutions

- Since we know that hardware architectures other than `x86-64` are not supported by the IDAES extensions, we could add a pre-check in `idaes get-extensions` so that users are made aware of this issue at an earlier stage, e.g. (untested naive example):
  - Get hardware architecture from e.g. `uname -a`
  - Check that it is in a list of supported architectures such as `x86-64`
  - If not, raise an error `Error: hardware architecture not supported: aarch64`

eslickj:
@jsiirola, I'm using Pyomo's FileDownloader class.  I did a quick look at the code and I don't immediately see a way to get the architecture.  Is there a way?  If not should we add that in Pyomo.  Once we have that, I can check it, and add that info to the package names.

lbianchi-lbl:
@eslickj if it can be useful, I've just found out that there's a `platform.machine()` function in Python from the `platform` built-in module that outputs `x86_64` on my laptop.

eslickj:
@lbianchi-lbl, I discovered it's really easy to use docker to build the aarach64 binaries.  I guess it would double the builds and tests but it's pretty trivial to do aside from the time.  If you think it's worth doing I can do it.

andrewlee94:
Completed in #788

PR #788, opened 2022-04-06
Add machine name to binary packages.

eslickj:
## Fixes issue #745

This adds machine names to the binary builds.  It should show an error if no build is available and facilitate future non-x86_64 builds.  There are a lot of moving pieces here.  This should work, but I may need to follow up with another PR or binary release to clean up a few things.  As it is, since the package names changed, I can't properly test all the binaries until this is merged.  At least Windows and Ubuntu 18.04 are tested with the regular tests.

-
-

### Legal Acknowledgement


eslickj:
I added "distro" as a requirement here.  That will help with easily identifying Linux distros for the binary downloads.

idaes/commands/extensions.py
```diff
@@ -233,11 +223,21 @@ def _write_hash(fp, pack, plat):
...
                 _write_hash(f, pack, plat)
+
+@cb.command(name="bin-platform", help="Show the compatible binary build.")
+def bin_platform():
+    fd, arch = idaes.util.download_bin._get_file_downloader(False, None)
+    try:
+        platform = idaes.util.download_bin._get_platform(fd, "auto", arch)
+    except idaes.util.download_bin.UnsupportedPlatformError:
+        click.echo("No compatible binaries found.")
+    else:
```
andrewlee94, line 242: Probably not that important, but does this really need to be in an `else` clause? Wouldn't it also work if this was in the `try` part?

idaes/config.py
```diff
@@ -19,67 +19,62 @@
...
-    "rhel8": "Red Hat Enterprise Linux 8",
-    "ubuntu1804": "Ubuntu 18.04",
-    "ubuntu1810": "Ubuntu 18.10",
-    "ubuntu1904": "Ubuntu 19.04",
-    "ubuntu1910": "Ubuntu 19.10",
-    "ubuntu2004": "Ubuntu 20.04",
-    "ubuntu2010": "Ubuntu 20.10",
-    "ubuntu2104": "Ubuntu 21.04",
+# Machine map
+binary_machine_map = {
```
andrewlee94, line 50: Again not important, but looking at this I wonder if it could be made more compact and robust by using Pythons string capitalization methods.

andrewlee94 (dismissed):
I am not really an expert on this, but I didn't see anything that jumped out at me as wrong. I did note two minor things, but they wouldn't affect the function of the code.

One other comment related to the IDAES v2.0 reorganization is where should the `download-bin.py` file live in the new repo structure?

idaes/config.py
eslickj: I thought about that.  I think I caught the reasonable possibilities.  This lets me just use a dictionary without worrying about calling a function (not that it's really a big deal).

andrewlee94 (approved):

jsiirola (dismissed):
I think this looks good.

eslickj:
@andrewlee94 and @jsiirola I made a few minor changes to get the test coverage up.  I took out some useless or confusing stuff and went through the docs to make sure they were up to date

docs/tutorials/getting_started/index.rst
```diff
@@ -71,20 +71,21 @@ Linux
...
 
-   Ubuntu 18.04 and 19.10 and distributions based on them::
+       sudo apt-get install libgfortran3 libgomp1 liblapack3 libblas3
+
+   Ubuntu 18.04, Ubuntu 20.04, and Debian 10::
 
       sudo apt-get install libgfortran4 libgomp1 liblapack3 libblas3
 
-   Ubuntu 20.04 and distributions based on it ::
+   Ubuntu 22.05 and Debian 11::
```
jsiirola, line 82: Shouldn't this be `22.04`?

eslickj: Yes.

PR #791, opened 2022-04-06
add pytest importorskip for test that requires prescient

xiangao1:
## Summary/Motivation:
Given Prescient is an optional dependency, we need to skip the tests which require Prescient.

- Add pytest.importorskip() for test that requires prescient.
-

### Legal Acknowledgement

### DIFF
--- a/idaes/apps/grid_integration/tests/test_coordinator.py
+++ b/idaes/apps/grid_integration/tests/test_coordinator.py
@@ -1,4 +1,9 @@
 import pytest
+
+pytest.importorskip(
+    "prescient", reason="prescient is a requirement to run grid integration tests"
+)
+
 import pyomo.environ as pyo
 from idaes.apps.grid_integration.bidder import Bidder
 from idaes.apps.grid_integration.tracker import Tracker


xiangao1:
Hi @lbianchi-lbl! Though you suggested to add pytest.importorskip to all 3 test modules I have for double loop, I only added it to the one that import Prescient. Is this enough? Thanks!

lbianchi-lbl:
> Hi @lbianchi-lbl! Though you suggested to add pytest.importorskip to all 3 test modules I have for double loop, I only added it to the one that import Prescient. Is this enough? Thanks!

@xiangao1 If that's the only way `prescient` is imported, then it should be OK. Without knowing the structure of the `grid_integration` package, I had assumed that `prescient` could be imported indirectly (i.e. importing a module that in turn imports `prescient`) in addition to directly. If that's not the case, then it should be OK.

More in general, as part of the `Integration` CI workflow, we have checks that fail if optional dependencies are not handled correctly; as long as those pass, it's OK.

xiangao1:
> > Hi @lbianchi-lbl! Though you suggested to add pytest.importorskip to all 3 test modules I have for double loop, I only added it to the one that import Prescient. Is this enough? Thanks!
> 
> @xiangao1 If that's the only way `prescient` is imported, then it should be OK. Without knowing the structure of the `grid_integration` package, I had assumed that `prescient` could be imported indirectly (i.e. importing a module that in turn imports `prescient`) in addition to directly. If that's not the case, then it should be OK.
> 
> More in general, as part of the `Integration` CI workflow, we have checks that fail if optional dependencies are not handled correctly; as long as those pass, it's OK.
> 
> **EDIT**: For reference, the CI check is "Integration / Run tests from site-packages (non-editable) installation"

@lbianchi-lbl Thanks! I will merge this now then.

lbianchi-lbl:
@xiangao1 It looks like there is at least one indirect import of `prescient` that causes that check to fail:

```log
_________ ERROR collecting apps/grid_integration/tests/test_bidder.py __________
ImportError while importing test module '/usr/share/miniconda/envs/idaes-env/lib/python3.8/site-packages/idaes/apps/grid_integration/tests/test_bidder.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/share/miniconda/envs/idaes-env/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
/usr/share/miniconda/envs/idaes-env/lib/python3.8/site-packages/idaes/apps/grid_integration/tests/test_bidder.py:3: in <module>
    from idaes.apps.grid_integration.bidder import Bidder
/usr/share/miniconda/envs/idaes-env/lib/python3.8/site-packages/idaes/apps/grid_integration/__init__.py:3: in <module>
    from .coordinator import DoubleLoopCoordinator
/usr/share/miniconda/envs/idaes-env/lib/python3.8/site-packages/idaes/apps/grid_integration/coordinator.py:1: in <module>
    import prescient.plugins as pplugins
E   ModuleNotFoundError: No module named 'prescient'
```

Issue #794, opened 2022-04-07
First time point in dynamic models


dallan-keylogic:
Right now, some IDAES models obtain an initial condition for a model with dynamics by adding logic constraining the first time point to be at steady state. See, for example, [the PID controller model](https://github.com/IDAES/idaes-pse/blob/main/idaes/models/control/controller.py):
```
@self.Constraint(time_set, doc="Error calculated by derivative of integral")
            def error_from_integral_eqn(b, t):
                if t == time_0:
                    if self.config.calculate_initial_integral:
                        if self.config.type == ControllerType.PI:
                            return (
                                self.integral_of_error[t]
                                == (
                                    b.manipulated_var[t]
                                    - b.mv_ref[t]
                                    - b.gain_p[t] * b.error[t]
                                )
                                / b.gain_i[t]
                            )
                        return (
                            self.integral_of_error[t]
                            == (
                                b.manipulated_var[t]
                                - b.mv_ref[t]
                                - b.gain_p[t] * b.error[t]
                                - b.gain_d[t] * b.derivative_of_error[t]
                            )
                            / b.gain_i[t]
                        )
                    return pyo.Constraint.Skip
                return b.error[t] == b.integral_of_error_dot[t]
```

However, [in the documentation for Pyomo DAE simulators](https://pyomo.readthedocs.io/en/stable/modeling_extensions/dae.html), we find:
> Notice that the initial conditions are set by fixing the values of m.omega and m.theta at t=0 instead of being specified as extra equality constraints. Also notice that the differential equations are specified without using Constraint.Skip to skip enforcement at t=0. The Simulator cannot simulate any constraints that contain if-statements in their construction rules.

Previous conversations have specified that the prohibition on if-statements extends only to if-then statements conditioned on time.

Nevertheless, this is a potential problem for compatibility with Pyomo simulators, and we need to figure out the best way to address it. The method Pyomo would rather have us use is to deactivate constraints at the first time point and fix the corresponding capacity variables. Do we want to do that automatically for IDAES models with `dynamic=True`?

eslickj:
[...]

PR #796, opened 2022-04-07
Fix PID First Time Step

eslickj:
## Fixes issue #794

This fixes the PID controller model to work with Pyomo integrators.  Also fixes an issue with the derivative estimation code where a derivative calculation could start with an uninitialized value and fail.  

-
-

### Legal Acknowledgement


idaes/core/solvers/petsc.py
```diff
@@ -544,6 +544,7 @@ def calculate_time_derivatives(m, time):
                 for i, v in var.items():
                     try:
                         if disc_eq[i].active:
+                            v.value = 0 # Make sure there is a value
```
andrewlee94, line 547: Would it be worth doing:

```
if v.value is None:
    v.value = 0
```

As it stands, this will overwrite any value that is already there, even if it is already a good guess.

dallan-keylogic (approved):
The changes look good to me.

One thing I've been thinking is adding some `capacity_vars` attribute to unit models that contains a consistent set of state variables that can be fixed to provide an initial condition.  This wouldn't be unique, but it might help users determine the degrees of freedom that need to be fixed at the first time point.

idaes/core/solvers/petsc.py
eslickj: Yeah, I wasn't sure how to handle this.  I don't really know what's good and what isn't.  The calculation is pretty straight forward, but somehow the derivative vars have random junk that is maybe NaN.  I think it's a side affect of not returning derivative values from PETSc, but having derivative variables.

I guess the best solution to this is to make PETSc estimate and return derivatives then use this routing to adjust them given that the time steps in the descretized problem are different.  I guess this should be okay until I can update the solver.

PR #798, opened 2022-04-12
More BooleanVar in serializer

eslickj:
## Fixes

This adds BooleanVar wherever there is Var, and for all the canned what-to-store classes, explicitly adds Constraint and Blocks rather than falling thorough the a generic Component.  The fall through approach could cause trouble if new components are added that don't have the "active" attribute. 

### Legal Acknowledgement


jsiirola (dismissed):
Looks reasonable, pending successful tests.

eslickj:
It looks like I managed to uncover an existing bug with this change.  Should be fixed.

jsiirola (dismissed):

eslickj:
Well this is embarrassing but the bug I found opened a whole can of worms.  I think it's all fixed now.  I removed the skip_classes argument for the StoreSpec class.  The intended use would usually break the serializer.  You have to write the blocks, but after fixing the bug, the blocks weren't necessarily included.  Also if you want to write suffixes you need to write the components that have suffixes.  The bug was causing pretty much all components to be written so it was covering up a lot of stuff.

eslickj:
@jsiirola and @andrewlee94, I kept messing this up somehow, but I'm pretty sure it's good to go now.

idaes/core/util/model_serializer.py
```diff
@@ -262,34 +252,41 @@ def __init__(
...
-        self.classes = [i[0] for i in classes]
+        skip_classes = []
+        # Convert to lists, so we can add things
+        classes = list(classes)
+        data_classes = list(data_classes)
+        # Block and BlockData are required for model structure
+        if Block not in classes:
+            classes.append((Block, ()))
+        if Block._ComponentDataClass not in data_classes:
+            data_classes.append((Block._ComponentDataClass, ()))
```
jsiirola, line 263: Out of curiosity, why are these lists and not dicts?  The `in` here is a linear scan of the list and could be quite slow.  A dict would support faster lookup and avoid ambiguity if the class appears more than once in the list.

eslickj: I wrote this a very long time ago, before normal dictionaries preserved order.  Since the order of the classes is important, I guess I just went with lists or tuples.  It's been so long I don't have a clear memory of why I did what I did.  I may take another look at this when I fix the typo.  Dicts are at least more aesthetically pleasing.  I did think about this being slow, but this is in a class that's pretty simple and gets created up front, so the efficiency of this methods won't have any measurable impact on overall speed.

jsiirola (approved):

andrewlee94 (approved):
Rubber stamping.

PR #799, opened 2022-04-12
Dh and scaling

dallan-keylogic:
- Add additional scaling diagnostic tools to `idaes.core.util.scaling`
- Add logic in case the user creates variables on `Ports` with a rule that isn't Pyomo's default `Equality` rule.
- Change `DegeneracyHunter` to use the scaled Jacobian instead of the raw Jacobian
- Add an option to use Scipy's dense SVD because the sparse SVD `svds` is unreliable in calculating small singular values
- Add a function to `DegeneracyHunter` to simply list the variables and constraints with large components in the right and left singular vectors, because I've had nothing but trouble trying to use cbc as a solver to find the minimal degenerate set. 
- Bundled scaling for `log_mole_frac_phase_comp` in the generic property package into this PR because it's a small change.

### Legal Acknowledgement


idaes/core/util/model_diagnostics.py
```diff
@@ -17,102 +17,103 @@
 
 __author__ = "Alexander Dowling"
```
adowling2, line 18: Add @dallan-keylogic here? I am okay with whatever you decide

idaes/core/util/model_diagnostics.py
```diff
@@ -199,36 +201,31 @@ def check_variable_bounds(
...
         return vnbs
-
-    def check_rank_equality_constraints(self, tol=1e-6):
+    
+    def check_rank_equality_constraints(self, tol=1E-6, dense=False):
         """
         Method to check the rank of the Jacobian of the equality constraints
-
+        
         Args:
```
adowling2, line 209: Need to add `dense` to the doc string

idaes/core/util/model_diagnostics.py
```diff
@@ -238,26 +235,26 @@ def check_rank_equality_constraints(self, tol=1e-6):
             # TODO: Make this an exception
             print("Model needs at least 2 equality constraints to check rank.")
```
adowling2, line 236: Should this be a ValueError?

idaes/core/util/model_diagnostics.py
```diff
@@ -500,64 +493,112 @@ def _find_candidate_eqs(candidates_milp, solver, eq_con_list=None, tee=False):
...
             # Thus U is a n_eq x n_eq matrix
             # And V is a n_var x n_var
             # (U or V may be smaller in economy mode)
             # Thus we really only care about U
-            u, s, v = svds(self.jac_eq, k=n_sv, which="SM")
-
+            
+            if dense:
+                u, s, vT = svd(self.jac_eq.todense(),full_matrices=False)
+                u = np.flip(u[:,-n_sv:], axis=1)
```
adowling2, line 533: Add a comment to explain why we need to use np.flip

idaes/core/util/model_diagnostics.py
```diff
@@ -500,64 +493,112 @@ def _find_candidate_eqs(candidates_milp, solver, eq_con_list=None, tee=False):
...
+                # Maybe the warning is for singular values of nearly equal
+                # magnitude or a similar edge case?
+                u, s, vT = svds(self.jac_eq, k = n_sv, which='SM')#, solver='lobpcg')
+            
             # Save results
             self.u = u
             self.s = s
-            self.v = v
-
+            self.v = vT.transpose()
```
adowling2, line 546: Is this transpose valid for both dense=True and False?

idaes/core/util/model_diagnostics.py
```diff
@@ -500,64 +493,112 @@ def _find_candidate_eqs(candidates_milp, solver, eq_con_list=None, tee=False):
...
+            starting from 1, to calculate associated constraints and variables
+            tol: Size below which to ignore constraints and variables in
+            the singular vector
+
+        Returns
+        -------
+        None.
+
+        """
+        if self.s is None:
```
adowling2, line 570: Why do we also compute the 10 smallest? What happens if you apply to a small problem with less than 10 variables or constraints?

idaes/core/util/scaling.py
```diff
@@ -611,6 +646,60 @@ def extreme_jacobian_entries(
...
+    Args:
+        m: model
+        scaled: if true use scaled Jacobian
+        large: >= to this value is consdered large
+        small: <= to this is consdered small
+
+    Returns:
+        (list of tuples), Jacobian entry, Constraint
+    """
+    if jac is None or nlp is None:
```
adowling2, line 664: Suggestion: sprinkle in a few comments to make this easier to maintain

idaes/core/util/scaling.py
```diff
@@ -611,6 +646,60 @@ def extreme_jacobian_entries(
...
+        jac, nlp = get_jacobian(m, scaled)
+    el = []
+    for i, c in enumerate(nlp.clist):
+        norm = 0
+        for j in jac[i].indices:
+            norm += jac[i, j]**2
```
adowling2, line 670: For example, add a comment that this is computing an L2 norm (which helps explain the **2)

idaes/core/util/scaling.py
```diff
@@ -611,6 +646,60 @@ def extreme_jacobian_entries(
...
+    Args:
+        m: model
+        scaled: if true use scaled Jacobian
+        large: >= to this value is consdered large
+        small: <= to this is consdered small
+
+    Returns:
+        (list of tuples), Jacobian entry, Variable
+    """
+    if jac is None or nlp is None:
```
adowling2, line 691: Again, sprinkle in a few lines of comments in the code block below.

adowling2 (commented):
@dallan-keylogic Looks good. I suggested a few placed below to improve the comments. I also might have identified a few edge cases to think about.

If you have a spare 30 minutes, it would be helpful to expand the tests to get ride of my of the codecov complaints. Mainly testing that the correct exceptions are thrown. (I realize many of these codecov complaints are not for the sections of the code that you modified.)

idaes/core/util/model_diagnostics.py
dallan-keylogic: Yes, both `svd` and `svds` return the transpose (or Hermitian transpose) of the right singular vectors.

idaes/core/util/model_diagnostics.py
dallan-keylogic: Choosing to compute 10 is somewhat arbitrary, but was the default elsewhere. Applying it to a problem with fewer than 10 variables worked just fine (and is in fact tested), but risked a slightly confusing error message telling the user to rerun `svd_analysis` if they asked for the fifth or sixth singular values in a 5x5 problem. I overhauled the error messages to avoid that slightly confusing situation. As a result, if the user explicitly asks for more singular values than svds can give it, a `ValueError` is raised instead of just computing as many singular values as possible.

adowling2 (approved):

adowling2:
@eslickj Can you give this a quick look? I reviewed the degeneracy hunter parts. It would be good to get your quick input (review) on the scaling parts (which is more of a black box to me than I'd like to admit).

idaes/core/util/scaling.py
```diff
@@ -462,6 +481,10 @@ def badly_scaled_var_generator(
     their current scale factors and values. For each potentially poorly scaled
     variable it returns the var and its current scaled value.
 
+    Note that while this method is a reasonable heuristic for nonnegative
```
eslickj, line 484: We're looking at the magnitude here, I don't think nonnegative maters.

eslickj (approved):

idaes/core/util/scaling.py
dallan-keylogic: As you might be able to tell, I don't think this is a particularly reliable method for finding badly scaled variables. "Nonnegative" is trying to get at one of the failure conditions of this method: when you have a `Constraint` constraining a variable to equal zero. You have those with no-flux boundary conditions in the SOEC, and you might also get those if you want to have no pressure drop and constraint `DeltaP` to equal zero (for the moment, set aside whether or not fixing the variable to equal zero would be a better method). If you solve the problem and look at the scale of the variable at the solution, it's absolutely not going to give you a good scaling factor for the variable.

You also encounter this with enthalpy as a state variable. It has some zero point determined by the reference state, but can take on a wide range of negative and positive values. If you happen to have an enthalpy near zero, the scaling factor you need to give it isn't any different than the rest of the enthalpies.

adowling2:
@dallan-keylogic By running examples, do you mean the example notebook or examples within the tests?

The easier solution is to just add an extra variable and add a linear equality constraint that “calculates” that variable. This will not change the degrees of freedom or the optimal solution.

And this does not work because an irreducible degenerate set does not make sense with a single variable, correct?

dallan-keylogic:
Yes, I mean the examples notebook. The example in the test file works fine because the rank isn't checked.

There's nothing intrinsically wrong with just checking the Jacobian rank with one equality constraint, but `svds` does not work on 1 x n matrices.

To be honest, `check_rank_equality_constraints` is not a particularly accurate description of what that function does. What it does do is give you the number of singular values below a certain magnitude, but it doesn't actually check the rank, which also requires the _largest_ singular value.

dallan-keylogic:
The path of least resistance was just adding in special handling to get the only singular value for `n_eq = 1`. The actual solution we want is to add another function for checking the rank of the active set Jacobian, but that's not something I want to mess with at the present.

eslickj (approved):

adowling2 (approved):

adowling2:
@dallan-keylogic I just approved this PR.

I agree that SVD for `n_eq = 1` does not make that much sense. I am not sure what it means for the NLP to be degenerate if there is only one equality constraint.

On Slack, you mentioned applying the analysis to the active set of the Jacobian. I believe DH already has this capability. (Or if not here, the original GAMS and Julia versions did!) The challenge is how to compute the active set, especially for inequality constraints that are weakly active, i.e., their residuals are (near) zero but the multipliers are also (near) zero. I left this as an option for the user where or not to include weakly active inequality constraints. I am not sure what is "correct" from a theory perspective.

PR #800, opened 2022-04-12
Improving some properties tests

andrewlee94:
## Summary/Motivation:
As part of looking into ways to improve and standardize `initialize` methods across all IDAES components, I found a few property tests that did not index the StateBlocks they created. Due to the way `StateBlocks` are defined, this results in some unexpected behaviours when `initialize` is called on an unindexed StateBlock.

Seeing as `StateBlocks` are always indexed within a flowsheet, it is probably most important to test more important to test their behaviour when indexed than when scalar, so all tests should focus on indexed blocks.

As a note going forward, at least in the near term, it is likely that calling `initialize` on un-indexed `StateBlocks` will start raising exceptions.

- Add nominal indexing set to tests for `StateBlocks` which lacked these.
- Broke tests for activity coefficient models into parts to separate different sets of phases which were previously lumped into a smaller number of tests.

### Legal Acknowledgement


eslickj (dismissed):

bpaul4 (approved):
Everything looks good to me, thanks for updating these files. It appears the lone test failure is not related to this PR.

chineduobiora (approved):

bpaul4 (approved):

bpaul4 (approved):

eslickj (approved):

PR #806, opened 2022-04-14
PETSc trajectory concatination

### LINKED ISSUE #757: Improvements for IDAES interface to PETSc

eslickj:
## Fixes #757

This adds a few things.  1) You can specify a subset of points to integrate between from the fully integrated model.  This should let you pick up changes in the model, like say a variable is fixed in one section, ramps in another section and is fixed at another value in another section.  2) instead of saving the trajectory in sections when splitting up the time domain, this concatenates them into one trajectory file.  3) You can optionally interpolate the skipped points from the trajectory data.  4) You can optionally estimate time derivatives at the end of the solve. 5) Remove WSL code.

By default now the interpolation and derivative estimation will happen.

I know some stuff about the trajectory data handling isn't super efficient, but I want to get this and then interpolation done solidly then come back for performance if necessary.

### Legal Acknowledgement


idaes/core/solvers/petsc.py
```diff
@@ -414,13 +415,34 @@ def petsc_dae_by_time_element(
             interpret the trajectory data.
         trajectory_save_prefix (str or None): If a string is provided the
             trajectory data will be saved as gzipped json
+        between (list or tuple): List of time points to integrate between.  If
+            None use all time points in the model.  If the first or last time
+            point are not inclued, they will be added.
```
dallan-keylogic, line 420: > If the first or last time point are not inclued, they will be added.

1) inclued -> included

2) Requiring the first and last timepoints seems unnecessarily restrictive. Users may want to change the integrator settings for integrating during different time periods. If we don't want to support that now but come back to it later, we can raise a `NotImplementedError` if both those time points aren't included.

idaes/core/solvers/petsc.py
```diff
@@ -515,13 +538,39 @@ def petsc_dae_by_time_element(
                     options={"--ts_init_time": tprev, "--ts_max_time": t},
                 )
             if trajectory_save_prefix is not None:
+                tj_prev = tj
                 tj = PetscTrajectory(
                     stub=vars_stub, delete_on_read=True, unscale=t_block
                 )
-                tj.to_json(f"{trajectory_save_prefix}_{count}.json.gz")
+                if tj_prev is not None:
+                    # due to the way variables is generated we know varaibles
```
dallan-keylogic, line 546: varaibles -> variables

idaes/core/solvers/petsc.py
```diff
@@ -515,13 +538,39 @@ def petsc_dae_by_time_element(
...
+                        except KeyError:
+                            vec_prev = [pyo.value(vp)]*len(tj_prev.time)
+                        tj._set_vec(v, vec_prev + vec)
+                    tj._set_time_vec(tj_prev.time + tj.time)
+                variables_prev = variables
             tprev = t
-            count += 1
             res_list.append(res)
+        if tj is not None:
+            tj.to_json(f"{trajectory_save_prefix}_1.json.gz")
```
dallan-keylogic, line 573: Do we need the "_1" anymore?

idaes/core/solvers/tests/test_petsc.py
```diff
@@ -579,3 +636,7 @@ def test_rp_example4():
     )
     assert pyo.value(m.u[10]) == pytest.approx(398)
     assert pyo.value(m.x[10]) == pytest.approx(20)
+
+
+if __name__ == "__main__":
+    test_petsc_read_trajectory_parts()
```
dallan-keylogic, line 642: Probably remove this part in the final commit.

dallan-keylogic (commented):
Overall looks good and would be a major help for the full discretization work.

1) You should probably add a test in which different variables are fixed in different sections of the problem, to ensure the fixing-and-unfixing works as expected.
2) Can users choose to activate and deactivate constraints during different sections of the problem to manipulate the degrees of freedom?

idaes/core/solvers/tests/test_petsc.py
eslickj: Yeah.

idaes/core/solvers/petsc.py
eslickj: No, I just didn't want to break anyone's stuff.

dallan-keylogic: Better to make the change now, while there are only a few users.

idaes/core/solvers/petsc.py
eslickj: I guess that's technically right, but the first point could be different for the initial conditions, so not including it may be looking for trouble.  The last point is probably safe to omit, but then you can't interpolate everything.  I went back and forth on this a bit couldn't really decide.

dallan-keylogic: I'd rather give users the flexibility to try something fancy if they want.

eslickj: I guess that fits with our general philosophy.

eslickj: Okay, made this change.

idaes/core/solvers/petsc.py
```diff
@@ -360,6 +357,19 @@ def _sub_problem_scaling_suffix(m, t_block):
             if c in m.scaling_factor:
                 t_block.scaling_factor[c] = m.scaling_factor[c]
 
+class PetscDAEResults(object):
+    """This class stores the results of ``petsc_dae_by_time_element()``, it has.
```
blnicho, line 361: The period after "it has" is not needed.

idaes/core/solvers/petsc.py
```diff
@@ -360,6 +357,19 @@ def _sub_problem_scaling_suffix(m, t_block):
...
+    two attributes ``results`` and ``trajectory``.  Results is a list of pyomo
```
blnicho, line 362: Capitalize "Pyomo"

idaes/core/solvers/petsc.py
```diff
@@ -360,6 +357,19 @@ def _sub_problem_scaling_suffix(m, t_block):
...
+    solver results objects.  Since this petsc DAE solver utility solves the
```
blnicho, line 363: Fix capitalization of "PETSc" to be consistent

idaes/core/solvers/petsc.py
```diff
@@ -360,6 +357,19 @@ def _sub_problem_scaling_suffix(m, t_block):
...
+    initial conditions first then may integrate over the time domain in one or
+    more steps, the results of multiple solves are returned.  The trajectory is
+    a PetscTrajectory object which gives the full trajectory of the solution
```
blnicho, line 366: Recommend adding back-ticks around ``PetscTrajectory`` to be consistent with the formatting used in the rest of the docstring

idaes/core/solvers/petsc.py
```diff
@@ -360,6 +357,19 @@ def _sub_problem_scaling_suffix(m, t_block):
...
+    for all time steps taken by the PETSc solver. This is generally finer than
+    the PyomoDAE discretization.
```
blnicho, line 368: Missing period in "Pyomo.DAE"

idaes/core/solvers/petsc.py
```diff
@@ -360,6 +357,19 @@ def _sub_problem_scaling_suffix(m, t_block):
             if c in m.scaling_factor:
                 t_block.scaling_factor[c] = m.scaling_factor[c]
 
+class PetscDAEResults(object):
```
blnicho, line 360: Is this class really necessary? Why not just return a dictionary from ``petsc_dae_by_time_element``?

idaes/core/solvers/petsc.py
```diff
@@ -403,51 +413,71 @@ def petsc_dae_by_time_element(
...
-
-    Returns:
-        List of solver results objects from each solve. If there are initial
-        condition constraints and they are not skipped, the first object will
-        be from the initial condition solve.  Then there should be one for each
-        time element for each TS solve.
+        between (list or tuple): List of time points to integrate between. If
+            None use all time points in the model. Generally the list should
+            start with the first time point and end with the last time point;
+            however, this is not a requirement and the are situations where you
```
blnicho, line 423: Typo: "**there** are situations..."

idaes/core/solvers/petsc.py
```diff
@@ -371,11 +381,11 @@ def petsc_dae_by_time_element(
...
     snes_options=None,
     ts_options=None,
-    wsl=None,
     keepfiles=False,
     symbolic_solver_labels=True,
-    vars_stub=None,
-    trajectory_save_prefix=None,
+    between=None,
+    interpolate=True,
+    calculate_derivaties=True,
```
blnicho, line 388: Typo: calculate_derivatives

idaes/core/solvers/petsc.py
```diff
@@ -403,51 +413,71 @@ def petsc_dae_by_time_element(
...
+            None use all time points in the model. Generally the list should
+            start with the first time point and end with the last time point;
+            however, this is not a requirement and the are situations where you
+            may not want to include them. If you are not including the first
+            time point, you should take extra care with the initial conditions.
+            If the initial conditions are already correct consider using the
+            ``skip_initial`` option if the first time point is not included.
+        interpolate (bool): if True and trajectory is read, interpolate model
+            values from the trajectory
+        calculate_derivaties: (bool) if True, calculate the derivative values
```
blnicho, line 430: Typo: calculate_derivatives

idaes/core/solvers/petsc.py
```diff
@@ -403,51 +413,71 @@ def petsc_dae_by_time_element(
...
     """
+    tj = None
+    if between is None:
+        between = time
+    else:
+        for t in between:
+            try:
+                assert t in time
+            except AssertionError:
+                raise RuntimeError("Elements of between must be in the time set")
```
blnicho, line 445: Maybe add quotes or some other notation to make it clear that "between" is a keyword argument in this context.

idaes/core/solvers/petsc.py
```diff
@@ -403,51 +413,71 @@ def petsc_dae_by_time_element(
...
+        See PetscDAEResults documentation for more informations.
     """
+    tj = None
+    if between is None:
+        between = time
+    else:
+        for t in between:
+            try:
+                assert t in time
+            except AssertionError:
```
blnicho, line 444: Another way to do this is using Python `set` operations. Not sure if this is more performant but it avoids the loop:
```
unknown_time_points = set(between) - set(time)
if unknown_time_points:
    raise RunTimeError()
```

idaes/core/solvers/petsc.py
```diff
@@ -403,51 +413,71 @@ def petsc_dae_by_time_element(
...
+                raise RuntimeError("Elements of between must be in the time set")
+        between = pyo.Set(initialize=sorted(between))
+        between.construct()
```
blnicho, line 447: I would recommend just working with the sorted Python list of values instead of creating a floating Pyomo `Set` component that isn't attached to a model.

idaes/core/solvers/petsc.py
```diff
@@ -514,16 +543,84 @@ def petsc_dae_by_time_element(
...
+            for var in time_vars:
+                if id(var[tlast]) in no_repeat:
+                    continue
+                no_repeat.add(id(var[tlast]))
+                if isinstance(var[t0].parent_component(), pyodae.DerivativeVar):
+                    continue # skip derivative vars
+                try:
+                    vec = tj.interpolate_vec(itime, var[tlast])
+                except KeyError:
+                    # variable not in trajectory, this happens if the problem
```
blnicho, line 606: Incomplete comment

idaes/core/solvers/petsc.py
```diff
@@ -514,16 +543,84 @@ def petsc_dae_by_time_element(
...
+                    continue
+                for i, (t, v) in enumerate(var.items()):
+                    if t < t0 or t > tlast or t in between:
+                        # Time is outside the range or already set
+                        continue
+                    if not v.fixed:
+                        # May not have trajectory from fixed variables and they
+                        # shouldn't change anyway, so only set not fixed vars
+                        v.value = vec[i]
+        if calculate_derivaties:
```
blnicho, line 616: Typo: calculate_derivatives

idaes/core/solvers/petsc.py
```diff
@@ -514,16 +543,84 @@ def petsc_dae_by_time_element(
...
+            # the petsc solver interface does not currently return time
+            # derivatives, and if it did, they whould be estimated based on a
+            # smaller time step.  This option uses PyomoDAE's discretization
```
blnicho, line 619: Typo: Pyomo.DAE

idaes/core/solvers/petsc.py
```diff
@@ -654,7 +771,30 @@ def get_dt(self):
             dt[i] = self.time[i + 1] - self.time[i]
         return dt
 
-    def interpolate_vecs(self, times):
+    def interpolate(self, times):
+        """Create a new vector dictionary interpolated at times. This method
+        will also extraplote values outside the original time range, so care
```
blnicho, line 776: Typo: extrapolate

idaes/core/solvers/petsc.py
eslickj: I think the reason I did this was to use first(), last(), and at(), which I was using for the full time set.  I guess alternatively, I could us a list here are make a list from the full time set in the other if-branch.

eslickj:
@dallan-keylogic to answer you question yes example here https://github.com/IDAES/examples-pse/pull/112 (last example in notebook).  I will add related test and while I fix the comments.

idaes/core/solvers/petsc.py
```diff
@@ -514,16 +543,84 @@ def petsc_dae_by_time_element(
...
-                    stub=vars_stub, delete_on_read=True, unscale=t_block
+                    stub="tmp_vars_stub",
+                    delete_on_read=True,
+                    unscale=True,
+                    model=t_block,
                 )
-                tj.to_json(f"{trajectory_save_prefix}_{count}.json.gz")
+                # add fixed vars to the trajectory. this does two things 1)
+                # helps users looking for fixed var trajectory and 2) lets
+                # us concatonate trajectories with section that are mixed fixed
```
dallan-keylogic, line 556: concatonate -> concatenate

idaes/core/solvers/petsc.py
```diff
@@ -514,16 +543,84 @@ def petsc_dae_by_time_element(
...
+                        tj._set_vec(v, vec_prev + vec)
+                    tj._set_time_vec(tj_prev.time + tj.time)
+                variables_prev = variables
             tprev = t
-            count += 1
             res_list.append(res)
-    return res_list
-
+        # If the interpolation option is True and the trajectory is available
+        # interpoate the values any skipped time points from the trajectory
```
dallan-keylogic, line 587: interpoate -> interpolate

dallan-keylogic (approved):

idaes/core/solvers/petsc.py
eslickj: Not really necessary, just wanted to use attributes rather than keys.

idaes/core/solvers/petsc.py
eslickj: Fancy!  Plus there are already sets I think, so even nicer.

eslickj:
Okay, this is almost ready, but I need to add a test where we have a variables fixed of free and constraint active and in active in different parts of the time domain and integrate it in parts.  I have this in an example notebook, but I should really add a simple test for it.

eslickj:
@blnicho and @dallan-keylogic, I added a couple tests for integrating in sections.  Should be all set now.

dallan-keylogic (approved):

blnicho (approved):

PR #809, opened 2022-04-15
[RFC] Draft proposal for new initialization API

andrewlee94:
## Summary/Motivation:
As part of developing the new costing API and the general move to IDAES v2.0, there has been some discussion of better/different ways we could handle initialization of IDAES Components to allow more flexibility (notably, the ability to have "add-ons" which are attached to a `Block` but not constructed by the `build` method, e.g. costing) whilst also forcing more standardization (e.g. raising exceptions, checking degrees of freedom).

This PR demonstrates one proposal for how we could implement a standard `initialize` method across all components which standardizes the main initialization workflow and calls out to model specific sub-methods as required. Currently, the example is only applied to the `UnitModelBlockData` base class to limit the flow-on effects for the time being.

A summary of the propose API is as follows:

* A new `_initialization_order` list would be added in `ProcessBlockData` which will be used to record the order in which add-ons should be initialized. The first entry in `self._initialization_order` would be `self` (i.e. the current block).
* A general purpose `initialize` method would be implemented in `ProcessBlockData` which would hopefully handle all use cases.  The `initialize` method would perform the following steps:

1. Call a component specific `fix_state` method which would fix all necessary variables to result in a square model.
2. Iterate through `self._initialization_order` and call `initialze_prepare` on each add-on (skipping `self`), which would perform any necessary steps to allow the current component to be initialized. In most cases, this would just deactivate the add-on block. 
3. Check that the degrees of freedom on the block are zero (currently, many models assume this but do not explicitly check).
4. Iterate through `self._initialization_order` and call either `initialize_build` is `self` or `initialize` if an add-on. These method would perform the necessary steps to ensure each block is properly initialized.
5. Iterate through `self._initialization_order` in reverse and call `initialize_final` on all add-ons (skipping `self`), which would revert any changes made in step 2. At this point ,we should be confident that the model is square and well initialized so that we can call a sovler.
6. Call a solver on `self`, which should solve the current block with all add-ons.
7. Call `self.revert_state` to undo changes from step 1 and return the model state to where it started.
8. Check the solver status from the final solve (set 6, or step 4 if no add-ons are present) and raise an `InitializationError` if not optimal (this will force the check, so model developers cannot forget it).

Obviously, all models would need to be updated to implement the `fix_state`, `revert_state`, `initialize_build`, `initialize_prepare` and `initialize_final` methods, however as can be seen from this PR the changes are generally fairly minimal, and in many cases standard methods in `UnitModelBlockData` are sufficient.

- Covert `UnitModelBlockData` to use the proposed API.
- Add new methods to Control Volume classes to support new API (old methods have currently been retained for backwards compatibility).
- Update most unit models to implement the necessary methods.

## Further Changes:

If we decide this API is an improvement and worth the effort to implement, the next steps would be to move the standard `initialize` method form `UnitModelBlockData` to `ProcessBlockData` and to update all other component types to support the necessary methods. Hopefully, much of this can be done in the base classes without needing to update most of the derived models.

### Legal Acknowledgement

10. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.


andrewlee94:
Development of these ideas is continuing, but this PR is too far out of date and ideas have changed so I will close it for now.

PR #810, opened 2022-04-18
FixedBed0D Updates Scaling and Initialization

bpaul4:
## Fixes
This completes internal scaling for the FB0D model based on recent updates to the MB and BFB models.

- Expanded model scaling for constraints and variables
- New scaling tests
- Minor initialization changes

### Legal Acknowledgement

### DIFF
--- a/idaes/models_extra/gas_solid_contactors/unit_models/fixed_bed_0D.py
+++ b/idaes/models_extra/gas_solid_contactors/unit_models/fixed_bed_0D.py
@@ -15,7 +15,11 @@ IDAES 0D Fixed Bed Reactor model.
 """
 
 # Import Pyomo libraries
-from pyomo.environ import Constraint, Var, units as pyunits
+from pyomo.environ import (Constraint,
+                           Var,
+                           value,
+                           check_optimal_termination,
+                           units as pyunits)
 from pyomo.dae import DerivativeVar
 from pyomo.common.config import ConfigBlock, ConfigValue, In
 
@@ -26,6 +30,7 @@ from idaes.core import (
     UnitModelBlockData,
 )
 from idaes.core.util.config import is_physical_parameter_block
+from idaes.core.util.exceptions import InitializationError
 from idaes.core.util.constants import Constants as constants
 import idaes.logger as idaeslog
 from idaes.core.util import scaling as iscale
@@ -380,7 +385,7 @@ see reaction package for documentation.}""",
                 else:
                     return b.solids[t].temperature == b.solids[0].temperature
 
-    def initialize(
+    def initialize_build(
         blk,
         gas_phase_state_args=None,
         solid_phase_state_args=None,
@@ -411,7 +416,8 @@ see reaction package for documentation.}""",
         """
 
         # Set solver options
-        init_log = idaeslog.getInitLogger(blk.name, outlvl)
+        init_log = idaeslog.getInitLogger(blk.name, outlvl, tag="unit")
+        solve_log = idaeslog.getSolveLogger(blk.name, outlvl, tag="unit")
         opt = get_solver(solver, optarg)  # create solver
         opt.options = optarg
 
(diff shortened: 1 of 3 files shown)


idaes/models_extra/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -26,6 +30,7 @@
     UnitModelBlockData,
 )
 from idaes.core.util.config import is_physical_parameter_block
+from idaes.core.util.exceptions import InitializationError
```
andrewlee94, line 33: I do not see this being used later in the new code.

idaes/models_extra/gas_solid_contactors/unit_models/fixed_bed_0D.py
```diff
@@ -446,10 +452,97 @@ def initialize(
...
+        # Get units meta data from property packages (only solid needed)
+        units_meta_solid = (
+            self.config.solid_property_package.get_metadata().get_derived_units
+            )
+
+        # scale some variables
+        if hasattr(self, "bed_diameter"):
+            if iscale.get_scaling_factor(self.bed_diameter) is None:
+                sf = 1 / value(self.bed_diameter)
+                iscale.set_scaling_factor(self.bed_diameter, 1e-1 * sf)
```
andrewlee94, line 464: It seems a bit odd that your are multiplying the inverse of the current value by 0.1 to get the scaling factor. A comment on why htis is necessary would probably be useful.

idaes/models_extra/gas_solid_contactors/unit_models/tests/test_FB0D.py
```diff
@@ -176,6 +179,179 @@ def test_build(self, iron_oc):
...
+
+        return m
+
+    @pytest.mark.solver
+    @pytest.mark.skipif(solver is None, reason="Solver not available")
+    @pytest.mark.component
+    def test_initialize_unscaled(self, iron_oc_unscaled):
+        optarg = {
+            "bound_push": 1e-8,
+            "halt_on_ampl_error": "yes",
```
andrewlee94, line 239: A test should probably not have `halt_on_ampl_error` set to `"yes"`. Similarly, I would not set a specific linear solver here (especially seeing as MA27 is the default anyway).

Finally, it it necessary to set `bound_push` like this? Sometimes it is, but it should be avoided if not necessary (and scaling often resolves many of these issues).

idaes/models_extra/gas_solid_contactors/unit_models/tests/test_FB0D.py
```diff
@@ -176,6 +179,179 @@ def test_build(self, iron_oc):
...
+            "linear_solver": "ma27",
+        }
+
+        initialization_tester(iron_oc_unscaled)
+
+        solver = get_solver("ipopt", optarg)  # create solver
+
+        initialize_by_time_element(iron_oc_unscaled.fs,
```
andrewlee94, line 247: First, this looks like it should be separate from the `initialization_tester` above.

Secondly, there should be some assert statement following this to ensure that it actually did what was intended.

I think this also applies to some other similar tests later in this file.

idaes/models_extra/gas_solid_contactors/unit_models/tests/test_FB0D.py
```diff
@@ -176,6 +179,179 @@ def test_build(self, iron_oc):
...
+
+    @pytest.mark.solver
+    @pytest.mark.skipif(solver is None, reason="Solver not available")
+    @pytest.mark.component
+    def test_solve_unscaled(self, iron_oc_unscaled):
+        results = solver.solve(iron_oc_unscaled)
+
+        # Check for optimal solution
+        assert check_optimal_termination(results)
+        assert results.solver.status == SolverStatus.ok
```
andrewlee94, line 259: This is not necessary - the `check_optimal_termination` method handles this.

andrewlee94 (changes requested):
A few general comments, but I think only one of them is critical.

idaes/models_extra/gas_solid_contactors/unit_models/fixed_bed_0D.py
bpaul4: Ah, thanks - I'll clean up the imports. I had attempted to implement an `InitializationError` check similar to the `PressureChanger` initialization on [lines 759-777](https://github.com/IDAES/idaes-pse/blob/a7680abb68c179e7c3abd27ddcbfe5a3f39747c3/idaes/models/unit_models/pressure_changer.py#L759), but found that inserting an extra solve statement in order to check the termination condition spawned other issues. Please let me know if you think the FB0D should use the `InitializationError` or if it is simple enough not to need it.

idaes/models_extra/gas_solid_contactors/unit_models/tests/test_FB0D.py
bpaul4: See earlier comment on `InitializationError`.

idaes/models_extra/gas_solid_contactors/unit_models/fixed_bed_0D.py
andrewlee94: You should have the `InitialziationError`. However, do you need an additional solve to do it? An initialization routine should always finish with ha solve of the full model (unless it happens to have no active constraints), so there should generally be a solution for you to check. The only case where you can skip the check is if you have no active constraints (and even then it should be conditional on there being no active constraints).

bpaul4: Currently, the initialization routine does not have any full model solves - it initializes the gas/solid thermophysical properties and freezes them, initializes the reactions properties and releases the property states. I would need to add a Step 2 which is a full solve similar to the linked PressureChanger script.

bpaul4:
I've addressed the majority of the review comments, including adding the `InitializationError` back in and separating the `initialize_by_time` into it's own test for each of the four cases (scaled and unscaled, both with and without energy balance).

The only remaining failing test is `test_initialize_unscaled` for the case with an energy balance and it's unclear why constraint 1731 is failing or which equation it is (I added some flags to the unit model and test code so the exact error will appear in the workflow run log, I'll clean them up later). It seems the constraint indexing is not the same between the test and example problems, so I checked every constraint in the model. The only expression that exponentiates variables is the rate constant equation through the temperature variable and I would be surprised if this is causing the issue. I also was not able to reproduce the solver error when modifying the TGA flowsheet example to remove scaling and add an energy balance, though they should be the same formulation.

andrewlee94:
@bpaul4 There is a way to turn on "symbolic_solver_labels" which will give your the name of the constraint when a failure occurs. I think you can provide this as an argument to the solver call: `solver.solve(model, symbolic_solver_labels=True)`

idaes/models_extra/gas_solid_contactors/unit_models/tests/test_FB0D.py
```diff
@@ -345,31 +365,49 @@ def test_scaling(self, iron_oc):
             ) == iscale.get_constraint_transform_applied_scaling_factor(c)
         for t, c in FB0D.solids_energy_holdup_constraints.items():
             assert pytest.approx(
-                1.27323E-9, abs=1e-11
+                1.27324E-7, abs=1e-9
```
andrewlee94, line 368: This is a very loose tolerance to be checking too - generally you should check to near solver tolerance, thus I tend to use `rel=1e-5`.

bpaul4:
Thank you @andrewlee94 . I implemented the flag above and found that surprisingly it IS the rate constant equation that is failing to evaluate due to a math overflow error `exp(987.723)` which would yield `9.18E+428`. I reformulated the equation in the reaction package to scale down the exponential argument.

idaes/models_extra/gas_solid_contactors/properties/methane_iron_OC_reduction/hetero_reactions.py
```diff
@@ -456,8 +456,9 @@ def rate_constant_eqn(b, j):
                             )
                             * self.solid_state_ref.temperature
                         )
-                    )
+                    ) * 1E-3)
```
chineduobiora, line 459: Manually scaling the constraints isn't needed. For the failing  test, which is "test_initialize_unscaled" (I think), I suggest reducing the time_set to 360s and nfe to 10 in the iron_oc_unscaled method. Ideally, at the given initial conditions, and time horizon of 3600 you will want to apply a scaling transformation before solving this model.

idaes/models_extra/gas_solid_contactors/unit_models/fixed_bed_0D.py
chineduobiora: @bpaul4 it is okay to remove this.

idaes/models_extra/gas_solid_contactors/unit_models/tests/test_FB0D.py
```diff
@@ -176,21 +178,231 @@ def test_build(self, iron_oc):
...
+                "gas_property_package": m.fs.gas_props,
+                "solid_property_package": m.fs.solid_props,
+                "reaction_package": m.fs.solid_rxns,
+            }
+        )
+
+        # Discretize time domain
+        m.discretizer = TransformationFactory("dae.finite_difference")
+        m.discretizer.apply_to(m, nfe=100, wrt=m.fs.time, scheme="BACKWARD")
+
```
chineduobiora, line 208: I suggest reducing the time_set to 360s and nfe to 10 in the iron_oc_unscaled method. This will resolve the issue with the failing "test_initialize_unscaled" test. Ideally, at the given initial conditions, and time horizon of 3600 you will want to apply a scaling transformation before solving this model.

chineduobiora (changes requested):

bpaul4:
Thanks @chineduobiora, that resolved the issue.

bpaul4:
Thanks @chineduobiora, that resolved the issue.

chineduobiora (approved):
Looks good to me.

bpaul4:
@andrewlee94 this is ready for your re-review, please let me know if you have further suggestions before merging this.

idaes/models_extra/gas_solid_contactors/unit_models/tests/test_FB0D.py
```diff
@@ -233,13 +233,9 @@ def iron_oc_unscaled(self):
     @pytest.mark.skipif(solver is None, reason="Solver not available")
     @pytest.mark.component
```
andrewlee94, line 234: This looks like it should be an integration test - it looks like you are solving a (simple) dynamic problem, which is generally more than we put in `component` tests. Note that component tests should take no longer than 10 seconds to run (with very few exceptions).

andrewlee94 (approved):
I had one comment, but I don't think it should hold up this PR for now.

PR #813, opened 2022-04-19
Nonisothermal solid oxide cell model

dallan-keylogic:
## Summary/Motivation:
Bring current version of nonisothermal solid oxide fuel/electrolytic cell into main

### Legal Acknowledgement


dallan-keylogic:
Have a bug with a leaky solver config. Fixing that and adding the header stuff I forgot at the moment.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
```diff
@@ -0,0 +1,3997 @@
...
+
+
+class CV_Interpolation(enum.Enum):
+    UDS = 1  # Upwind difference scheme, exit face same as node center
+    CDS = 2  # Linear interpolation from upstream and downstream node centers
+    QUICK = 3  # Quadratic upwind interpolation, quadratic from two upwind
+    # centers and one downwind
+
+
+def _set_default_factor(c, s):
```
andrewlee94, line 63: These really look like they should be general purpose functions in the scaling toolbox.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
```diff
@@ -0,0 +1,3997 @@
...
+                        idx_set=(tset,iznodes,comps),
+                        units=pyo.units.mol/(pyo.units.s*pyo.units.m**2))
+        _create_if_none(unit, 
+                        "xflux_x1",
+                        idx_set=(tset,iznodes,comps),
+                        units=pyo.units.mol/(pyo.units.s*pyo.units.m**2))
+    
+@declare_process_block_class("SocChannel")
+class SocChannelData(UnitModelBlockData):
+    CONFIG = ConfigBlock()
```
andrewlee94, line 802: Is there a reason you are declaring a new `ConfigBlock` rather than inheriting the one from the base class? From what I can see, you are just replicating the things that are in the base class config anyway (`dynamic` and `has_holdup`).

idaes/models_extra/power_generation/unit_models/soc_submodels.py
eslickj, line 63: I know these are private, but a one or two sentence doc string about what they do may help people looking at the code.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
```diff
@@ -0,0 +1,3997 @@
...
+                **True** - set as a dynamic model,
+                **False** - set as a steady-state model.}""",
+        ),
+    )
+    CONFIG.declare(
+        "has_holdup",
+        ConfigValue(domain=In([useDefault, True, False]), default=useDefault),
+    )
+    CONFIG.declare(
+        "comp_list",
```
andrewlee94, line 822: I would recommend long-form names for configuration arguments to be consistent with IDAES standards.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
```diff
@@ -0,0 +1,3997 @@
...
+        ConfigValue(default=["H2", "H2O"], description="List of components"),
+    )
+    
+    CONFIG.declare(
+        "cv_zfaces",
+        ConfigValue(description="CV boundary set, should start with 0 and end with 1."),
```
andrewlee94, line 828: This looks like it should have a domain of some sort to make sure you get the type of input you are expecting. I think you should also be clearer about exactly what form the input should be in the doc string (is this a Python `list`, or a Pyomo `Set`)?

This applies to other config arguments below.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
```diff
@@ -0,0 +1,3997 @@
...
+        ),
+    )
+
+    def build(self):
+        super().build()
+
+        # Set up some sets for the space and time indexing
+        dynamic = self.config.dynamic
+        tset = self.flowsheet().config.time
+        self.comps = pyo.Set(initialize=self.config.comp_list)
```
andrewlee94, line 871: I would recommend the standard name of `component_list` here too.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
```diff
@@ -0,0 +1,3997 @@
...
+                        units=units,
+                        )
+            )
+    else:
+        setattr(blk, var, pyo.Reference(attr))
+
+def _init_dof_error(blk):
+    if not mstat.degrees_of_freedom(blk) == 0:
+        raise InitializationError(
+            f"{blk.name} encountered a nonsquare optimization problem during "
```
eslickj, line 103: Maybe just "nonsquare problem."  I guess if you expect it to be square, it's not optimization.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
```diff
@@ -0,0 +1,3997 @@
...
+        # z coordinates for nodes and faces
+        self.zfaces = pyo.Set(initialize=self.config.cv_zfaces)
+        self.znodes = pyo.Set(
+            initialize=[
+                (self.zfaces.at(i) + self.zfaces.at(i + 1)) / 2.0
+                for i in range(1, len(self.zfaces))
+            ]
+        )
+        # This sets provide an integer index for nodes and faces
+        self.izfaces = pyo.Set(initialize=range(1, len(self.zfaces) + 1))
```
andrewlee94, line 881: Do you need these? I think there are Pyomo methods for Sets that would let you achieve this using the existing `Sets`.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
```diff
@@ -0,0 +1,3997 @@
...
+        )
+        self.Dtemp = pyo.Var(
+            tset,
+            iznodes,
+            doc="Deviation of temperature at node centers "
+                "from temperature_z",
+            units=pyo.units.K,
+            bounds = (-1000,1000)
+        )
+        self.enth_mol = pyo.Var(
```
andrewlee94, line 977: A lot of this looks like it is doing things in the unit model that would normally be done in a `StateBlock`. Whilst it is more work to set up, I would suggest considering creating a (very simple) property package that will support these and using the more standard IDAES approach in order to facilitate future maintenance of the code and eventual promotion to the core model libraries.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
```diff
@@ -0,0 +1,3997 @@
...
+    
+            @self.Constraint(tset, ixnodes, iznodes)
+            def int_energy_density_solid_eqn(b, t, ix, iz):
+                return b.int_energy_density_solid[
+                    t, ix, iz
+                ] == b.solid_heat_capacity * b.solid_density * (
+                    b.temperature[t, ix, iz] - 1000 * pyo.units.K
+                )
+
+        # @self.Constraint(tset, ixnodes, iznodes)
```
andrewlee94, line 1936: Commented code.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
```diff
@@ -0,0 +1,3997 @@
...
+                - b.zface_area[ix]
+                * sum(
+                    b.zflux[t, ix, iz + 1, i]
+                    * comp_enthalpy_expr(b.temperature_zfaces[t, ix, iz + 1], i)
+                    for i in comps
+                )
+            )
+        if dynamic:
+            self.energy_balance_solid_eqn[tset.first(),:,:].deactivate()
+    def initialize_build(
```
andrewlee94, line 2203: Missing line break. I would suggest running `black` on all this code before the PR is merged - this is going to be enforced in the near future.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
```diff
@@ -0,0 +1,3997 @@
...
+                    for j in self.comps:
+                        sy = sgsf(self.mole_frac_comp[t,ix,iz,j],sy_def)
+                        cst(self.conc_eqn[t,ix,iz,j],sy*sP)
+                        ssf(self.Dconc[t,ix,iz,j],sxflux[j]*sLx/sD)
+                        
+                        cst(self.material_balance_eqn[t,ix,iz,j],
+                            sxflux[j]*sLy*sLz)
+                        
+               
+@declare_process_block_class("SocTriplePhaseBoundary")
```
andrewlee94, line 2443: With three of these classes in a single module, this is ends up being rather long. I would suggest you consider breaking this into separate modules for each class to help keep things shorter and easier to work throguh.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
```diff
@@ -0,0 +1,3997 @@
...
+                hg[it][-1][-1] = None
+            for ix, x in enumerate(xnodes):
+                zg[it][-1][ix + offset] = 1.0
+                xg[it][-1][ix + offset] = x
+                hg[it][-1][ix + offset] = pyo.value(right[t, ix + 1])
+
+    return zg, xg, hg
+
+
+element_list = ["Ar","H","O","C","N","S"]
```
andrewlee94, line 321: This is all related to thermophysical properties, which normally would not go in a unit model file.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
```diff
@@ -0,0 +1,3997 @@
...
+        cf = faces.at(ic)
+        lambf = (cd - cf) / (cd - cu)
+        return (1 - lambf) * phi_func(icu) + lambf * phi_func(icd)
+    else:
+        # Since we are doing linear interpolation derivative is the slope
+        # between node centers even if they are not evenly spaced
+        return (phi_func(icd) - phi_func(icu)) / (cd - cu)
+
+
+def contour_grid_data(var, time, xnodes, znodes, left=None, right=None, top=None, bottom=None):
```
eslickj, line 226: Needs docstring

andrewlee94 (changes requested):
Looking through the code, the first thing that becomes apparent is that these models are not so much IDAES models but rather a Pyomo models - i.e. they not use `StateBlocks` or `ControlVolumes`.  Whilst this does not mean the models are wrong, it does mean they are very different from anything else and they would need to be completely rebuilt to move them into the core model library. I would strongly suggest that these models should be changed to follow a more IDAES-like form before being merged into the main IDAES repository (if necessary they could be placed in `workspace` so that they can be shared amongst the team whilst this is happening).

idaes/models_extra/power_generation/unit_models/soc_submodels.py
```diff
@@ -0,0 +1,3997 @@
...
+    return (
+        0.002666
+        * cm2_to_m2
+        * temperature ** (3 / 2)
+        / p
+        / Pa_to_bar
+        / mab ** 0.5
+        / sab ** 2
+        / omega
+    )
```
eslickj, line 558: Does this need units?

idaes/models_extra/power_generation/unit_models/soc_submodels.py
eslickj: Unfortunately, that's my fault.  Given the complexity of this 2D model and slowness of some general properties, I decided for efficiency and robustness to hard code the properties in this one.  It's just abnormally complicated and I wanted tight control of everything.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
eslickj: pyo.RangeSet()?

idaes/models_extra/power_generation/unit_models/soc_submodels.py
eslickj: While I don't disagree that may need to be saved for the future.  We can probably swap that put pretty seamlessly by making a new default prop pack.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
andrewlee94: I was thinking `pyo.Set.at()` and the like.

dallan-keylogic:
> Looking through the code, the first thing that becomes apparent is that these models are not so much IDAES models but rather a Pyomo models - i.e. they not use `StateBlocks` or `ControlVolumes`. Whilst this does not mean the models are wrong, it does mean they are very different from anything else and they would need to be completely rebuilt to move them into the core model library. I would strongly suggest that these models should be changed to follow a more IDAES-like form before being merged into the main IDAES repository (if necessary they could be placed in `workspace` so that they can be shared amongst the team whilst this is happening).

The biggest problem with trying to use a more standard IDAES construction is that we don't yet have a two-dimensional `ControlVolume` class. Conceivably, the channels could be moved over to use a `ControlVolume1D`, but most other things would remain unchanged.

The ultimate goal is to create an SOC module level model where scaling between cell scale quantities and module scale quantities would happen, heat loss equations would live, and full-fledged `PropertyBlocks` would exist to allow different state variables to connect directly to the module. However, there are a lot of design questions that I don't want to answer at the present, so my current method is to use translator blocks at the flowsheet level to scale between cell-scale and module-scale and translate between FTPx state variables and whatever the user is doing in the main flowsheet.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
eslickj: @dallan-keylogic if you do this, I suggest creating a like soc_common for the common content then one module for each submodel.  Maybe that's obvious, just pointing out that it doesn't need to be much work.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
```diff
@@ -0,0 +1,3999 @@
...
+        self.znodes = pyo.Set(
+            initialize=[
+                (zfaces[i] + zfaces[i+1]) / 2.0
+                for i in range(len(zfaces)-1)
+            ]
+        )
+        comps = self.comps = pyo.Set(initialize=self.config.comp_list)
+        self.tpb_stoich = copy.copy(self.config.tpb_stoich_dict)
+        # TODO maybe let user specify inert species directly? Floating point
+        # equalities make me nervous---Doug
```
eslickj, line 2511: Floating point equalities should make you nervous, but the stochiometric coefficients aren't calculated, so if they are 0 we should enter an integer 0 in the dict.  Even 0.0 is probably okay as long as it's a constant.  Calculating coefficients could get you into trouble here, but that shouldn't be an issue.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
```diff
@@ -0,0 +1,3999 @@
...
+                    sy = sgsf(self.mole_frac_comp[t,iz,j], sy_def)
+                    ssf(self.log_mole_frac_comp[t,iz,j], 1)
+                    cst(self.mole_frac_comp_eqn[t,iz,j], sy)
+                    cst(self.log_mole_frac_comp_eqn[t,iz,j], sy)
+                    sxflux = sgsf(self.xflux[t,iz,j], 1e-2)
+                    cst(self.xflux_eqn[t,iz,j], sxflux)
+
+@declare_process_block_class("SocContactResistor")
+class SocContactResistorData(UnitModelBlockData):
+    CONFIG = ConfigBlock()
```
eslickj, line 2760: If possible inherit UnitModelBlockData config

idaes/models_extra/power_generation/unit_models/soc_submodels.py
dallan-keylogic: My preferred solution is to add an `overwrite=True/False` option to `set_scaling_factor` to handle the "if None` part, then make the "iterating over the index to set individual scaling factors" aspect default behavior of `set_scaling_factor`.  Indexed variables shouldn't have scaling factors distinct from their indexed parts.  `_set_and_get_scaling_factor` might still be a function then, but the other three are eliminated.

eslickj: I'm pretty sure we can add the overwrite option, and default to whatever it does now without hurting anything.

dallan-keylogic: Yeah, that change is pretty small. It's the other two that are more impactful. I maintain that calling `get_scaling_factor` on an indexed variable without specifying the indices should return an error: otherwise the user could specify scaling factors at each index yet calling `get_scaling_factor` on the variable for an unspecified index wouldn't return any factor.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
dallan-keylogic: I'd prefer to future-proof the code if possible. How about assuming that components that do not have a stoichiometric coefficient specified are inert? The problem is that then users could not specify a coefficient of zero, which may lead to confusion.

The best method is probably just adding an `inert_list` to the config.

Complicating matters is attempting to take into account solid-phase species like O^2- ions and vacancies. I don't think we have cause to worry about inert solid-phase species, though, so assuming the inerts are a subset of gas species should be fine.

idaes/models_extra/power_generation/unit_models/soc_submodels.py
dallan-keylogic: The big question to me is "how do we want users to reference variables in the model?" Initially I was going to change things to use Pyomo's idiosyncratic "index continuum variables by floats" method in order to be consistent with what's implemented in `ControlVolume1D`. However, the only discretization methods supported are finite differences and collocation, in which equations are written for the points at `zfaces` (as well as additional, collocation points for collocation). However, a finite volume method is used here, and `znodes` are centered at the midpoint between the points specified in the faces. I don't know that users, having passed `zfaces = [0, 0.25, 0.5, 0.75, 1]` will be thrilled having to index `temperature_z` using `znodes=[0.125, 0.375, 0.625, 0.875]`.

eslickj:
@dallan-keylogic, I'm not sure why, but it seems to be complaining about get_solver not being there.  It's right it did move, but the old deprecated import path still works.  My guess is update to idaes.core.solvers.get_solver and it will be ok.

bpaul4:
@eslickj @dallan-keylogic I recently came across similar issues importing reorganized modules although I had pulled the latest from the main IDAES repository, and updating via `pip install -U idaes-pse` and then `pip install -U .` resolved the issues.

dallan-keylogic:
Thanks @eslickj  and @bpaul4. The issue is now resolved since I updated imports.

idaes/models_extra/power_generation/unit_models/soc_submodels/channel.py
```diff
@@ -0,0 +1,839 @@
...
+    CONFIG.declare(
+        "opposite_flow",
+        ConfigValue(default=False, description="If True assume velocity is negative"),
+    )
+    CONFIG.declare(
+        "below_electrode",
+        ConfigValue(
+            domain=In([True, False]),
+            description="Decides whether or not to create material "
+            "flux terms above or below the channel.",
```
eslickj, line 82: Maybe for description "Set the electrode boundary (material flux) location." There is a minor disconnect between the description and the option name since the description doesn't mention the electrode.

eslickj (commented):
This looks pretty good to me.  It's pretty long and complex, so I don't know that I could see any issues with the equations by reading through, but the structure and everything looks really nice.  I have a couple more files to finish up and I'm going to try to use this so I can get a better feel for it.  I can't see any major issues, so I think it's safe for @andrewlee94 to start reviewing as I finish up.

idaes/models_extra/power_generation/unit_models/soc_submodels/channel.py
andrewlee94: Further - the options here are "True" and "False" but the text (to a naïve reader at least) implies it should be "Above" or "below".

idaes/models_extra/power_generation/unit_models/soc_submodels/conductive_slab.py
```diff
@@ -0,0 +1,315 @@
...
+                == b.xface_area[iz]
+                * (b.heat_flux_x[t, ix, iz] - b.heat_flux_x[t, ix + 1, iz])
+                + b.zface_area[ix]
+                * (b.heat_flux_z[t, ix, iz] - b.heat_flux_z[t, ix, iz + 1])
+                + b.joule_heating[t, ix, iz]
+            )
+
+        if dynamic:
+            self.energy_balance_solid_eqn[tset.first(), :, :].deactivate()
+
```
andrewlee94, line 257: Does this model need to overload the `initialize_build` method? At the moment, I think calling `initialize` on this model would result in attribute errors due to the absence of `self.control_volume`. If this model is not intended to be solved/initialized in isolation, I would suggest having `initialize_build` raise a `NotImplementedError("This model is not intended to be initialized in isolation")`

idaes/models_extra/power_generation/unit_models/soc_submodels/electrode.py
```diff
@@ -0,0 +1,952 @@
...
+                            self.int_energy_mol[t, ix, iz] / self.vol_mol[t, ix, iz],
+                        )
+                for i in comps:
+                    _set_if_unfixed(
+                        self.conc_mol_comp_deviation_x1[t, iz, i],
+                        self.conc_mol_comp_deviation_x[t, self.ixnodes.last(), iz, i],
+                    )
+
+        opt = get_solver(solver, optarg)
+        common._init_solve_block(self, opt, solve_log)
```
andrewlee94, line 751: I am not sure if `_init_solve_block` does this already, but this should raise an `InitializationError` if if fails to solve. I would suggest that it should return a solver status that you can check for optimal termination and then raise an exception if required.

idaes/models_extra/power_generation/unit_models/soc_submodels/solid_oxide_cell.py
```diff
@@ -0,0 +1,994 @@
...
+        ConfigValue(
+            default={"H2": -0.5, "H2O": 0.5, "e^-": 1.0},
+            description="Dictionary with species as keys and stoichiometric coefficients as values "
+            "for the redox reaction that occurs at the fuel-side triple phase boundary",
+        ),
+    )
+    CONFIG.declare(
+        "inert_fuel_species_triple_phase_boundary",
+        ConfigValue(
+            default=[],
```
andrewlee94, line 95: I think this might cause unexpected problems, see https://github.com/IDAES/idaes-pse/issues/330.

I think it would be better to make the `None`, and then later check for this and handle appropriately.

idaes/models_extra/power_generation/unit_models/soc_submodels/solid_oxide_cell.py
```diff
@@ -0,0 +1,994 @@
...
+        common._init_solve_block(self, opt, solve_log)
+
+        self.potential_eqn.activate()
+        # TODO---does the user ever have a reason to fix current density
+        # besides initialization and initial conditions?
+        self.current_density.unfix()
+        self.potential.fix()
+
+        init_log.info_high("Solving cell with potential equations active")
+        common._init_solve_block(self, opt, solve_log)
```
andrewlee94, line 743: As before regarding `InitializationErrors`.

andrewlee94 (commented):
There is a lot to go through here and I am not that familiar with the models so my review was at a fairly high level. My biggest comment is that there probably needs to be some sort of documentation to go with this and explain what all these different models are for and how they are intended to be used (plus highlighting anything non-standard like the hard coded component lists and properties).

Full documentation would be preferred, but for now some more doc strings and maybe a short README might suffice.

idaes/models_extra/power_generation/unit_models/helm/compressor.py
```diff
@@ -158,7 +158,7 @@ def initialize_build(
         solve_log = idaeslog.getSolveLogger(self.name, outlvl, tag="unit")
 
         # Create solver
-        slvr = get_solver(solver, optarg)
+        opt = get_solver(solver, optarg)
```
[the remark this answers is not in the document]
dallan-keylogic: I actually did a find-and-replace on `slvr` to change it to `opt` in my project. I didn't see that the search scope wasn't limited at first, so this got changed too. I can change it back.

I preferred `opt` because I thought having both `solver` and `slvr` might be confusing. If we're doing `slvr`, it can get changed back.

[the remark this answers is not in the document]
andrewlee94: I tend to use `solver_obj` in these cases to make it clear what it is. I agree that also having the `solver` argument makes this a bit confusing, but `opt` does not tell a user what this object is and also implied "optimization", whereas in this case we are solving a square problem.

idaes/models_extra/power_generation/unit_models/soc_submodels/channel.py
```diff
@@ -0,0 +1,839 @@
...
+                comps,
+                doc="Material flux from channel to electrode surface "
+                "(positive is out)",
+                initialize=0,
+                units=pyo.units.mol / pyo.units.m**2 / pyo.units.s,
+            )
+
+            @self.Expression(tset, iznodes, comps)
+            def conc_mol_comp_deviation_x0(b, t, iz, j):
+                return 0
```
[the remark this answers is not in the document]
dallan-keylogic: It has to do with the `below_electrode` flag. In one channel, the electrode is located at the x=1 boundary, while in the other it is located at the x=0 boundary. For the side of the channel not facing the electrode, there is no need to create `conc_mol_comp_deviation` and `material_flux` variables, but other equations might expect them to exist (`material_flux` shows up in the material balance, I don't think `conc_mol_comp_deviation` shows up in another equation). I preferred creating `Expressions` with value zero rather than write conditional logic in other places that expect them to exist.

idaes/models_extra/power_generation/unit_models/soc_submodels/channel.py
```diff
@@ -0,0 +1,839 @@
...
+            doc="Component concentration at node centers",
+            units=pyo.units.mol / pyo.units.m**3,
+            bounds=(0, None),
+        )
+        self.temperature_deviation_x = pyo.Var(
+            tset,
+            iznodes,
+            doc="Deviation of temperature at node centers " "from temperature_z",
+            units=pyo.units.K,
+            bounds=(-1000, 1000),
```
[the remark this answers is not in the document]
dallan-keylogic: Yeah, I expect them to mostly be on the order of 10 Kelvin, but don't want to assume that a priori. The reason for temperature bounds was to prevent IPOPT from taking enormous steps, then getting AMPL errors when the (absolute) temperature is negative. If it violates a bound, it still does a line search, but it doesn't fill your console with output.

idaes/models_extra/power_generation/unit_models/soc_submodels/electrode.py
dallan-keylogic: I created this helper function in part because I kept writing these tiny blocks of code raising initialization errors 1) if the problem was nonsquare somehow or 2) if the solve failed.  Now it does both of those things (and redirects the output to `solve_log`.

eslickj:
@andrewlee94 and @dallan-keylogic, I'm a bit biased here because I'm somewhat responsible for the structure of this model.  In this case, with a model this complex, I think this is the correct route.  Externally the model appears to be an IDAES model.  Internally I don't think the way we choose to write the model should be an issue.  I'm completely okay with this forever being in ```models-extra```.

andrewlee94:
@eslickj To be clear, I am fine with it being in `models_extra` too, but I think we need some (short) documentation somewhere explaining how this model is structured (and highlight that it is not structured the same way as most models).

idaes/models_extra/power_generation/unit_models/soc_submodels/solid_oxide_cell.py
dallan-keylogic: Changing the default to `None` is easy enough, but I just realized that we would have the same issue anywhere we include a `dict` or `list` as a default config argument, empty or not. I looked at the implementation of `pyomo.common.config.ConfigBase` and it doesn't appear to create a copy when a default value is given. I think I don't ever mutate these lists or dictionaries in any submodule, so it wouldn't be a major problem that they all share the same default list or dictionary, but that might not be true in all IDAES or Pyomo models. 

Just making every `list`/`dict` default argument `None`, then filling in actual default values later would work, but it would be a lot more confusing to the user than using `None` to represent an empty `list`/`dict`. @jsiirola , any thoughts?

data/config.yaml
```diff
@@ -1,8 +1,8 @@
 _id: d779730c7cc94155877a24e2e8be2ad1
-created: '2022-04-19T11:53:46.882828'
+created: '2022-02-09T14:59:17.579871'
```
andrewlee94, line 2: I am seeing this file change in each commit still. Firstly, merging the latest main branch should fix the issue of the tests changing this all the time. Ideally we would also remove this file from the PR, but I don't think it hurts anything to leave it in and working out how to undo it might be more effort than it is worth,

andrewlee94 (approved):
Only tow minor comments now, and I don't think either of them should prevent this from being merged.

data/config.yaml
dallan-keylogic: If you're just reviewing changes since you last reviewed, that might be the reason. Note that the timestamp here is _older_ than the one it replaced. I don't see it in the overall diff.

idaes/models_extra/power_generation/unit_models/soc_submodels/common.py
```diff
@@ -243,8 +241,59 @@ def _interpolate_2D(
...
+    Parameters
+    ----------
+    domain: Container
+        The container that specifies the allowable values.  Incoming
+        values are passed to ``domain.__contains__()``, and if ``True``
+        is returned, the value is accepted and returned.
+    """
+    #  TODO Need to determine what to do about repeated entries before this can become a Pyomo PR
+    def __new__(cls, domain=None, cast=None):
+        return super(_SubsetOf, cls).__new__(cls)
```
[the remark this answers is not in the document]
dallan-keylogic: This section was derived from `In` in Pyomo, so that's why it's currently that way. Since I hope to ultimately make this a Pyomo PR and then get rid of the implementation here, we can sort out that detail later.

data/config.yaml
andrewlee94: Ah, that is probably the reason.

andrewlee94 (approved):

eslickj (approved):

PR #814, opened 2022-04-19
Adding units of measurement to IDAES reporting tools

### LINKED ISSUE #44: Display units of measurement in model outputs
Now that IDAES models support explicit declaration of units of measurement on all components, we should update our reporting methods to include these in the report outputs and include the ability for users to provide a set of preferred units of measurement for the output results. However, we don't want to have to force users to declare preferred units unless they want to, so we need to have a fall back for these cases.

For single values this is straight forward: check to see if the user has provided a preferred set of unit. If they have, convert the value and print the result followed by the preferred units, otherwise take the value as is and print the declared units of the component. Pyomo will probably want to implement the underlying methods for this, as this is general to any Pyomo model using units, and we can just use this method in our reporting functions.

However, this becomes more complicated for aggregate outputs such as stream tables as now we have multiple use-cases to consider. Ideally, the user would declare a set of preferred units for their table and all values in the streams would have units, so we could just put a summary of units in the row headers and convert all the values to the desired units and be done. However, we need to consider the following:

1. Flowsheets can have multiple property packages which use different units. If a user does not declare a set of preferred units, we have to take the values as they are, but how do we report units? We cannot just use the row headers anymore as there is no guarantee of consistent units, so do we need to put units on each value (this would get crowded fast)? Also how do we determine when we can put units in the row header and when we can't?
2. What happens in a property package doesn't include units but user specified a set of units to use? Convert methods will now fail due to incompatible units, so do we just raise the exception and tell users to give the properties units, or do we just report the value as is and somehow flag units as undefined?

We would also need to define an API for specifying the preferred units for a flowsheet. As this is a user-output level, I think we need to provide the ability to specify units by quantity, rather than just the 5 base quantities - e.g. users need to be able to say report enthalpies in kJ/kmol and entropies in BTU/lb.R. We could provide pre-defined output units sets (maybe via the DMF) that provide a basis that users could modify if desired.

andrewlee94:
## Summary/Motivation:
Adding units of measurement to the IDAES reporting tools is long overdue.

- Adds a new `reporting_units` entry in the IDAES config for defining desired reporting units of derived quantities,
- Adds new `units_of_measurement` module to `ideas.core.util` with methods for converting quantities to IDAES reporting units,

### Legal Acknowledgement

### DIFF
--- a/idaes/config.py
+++ b/idaes/config.py
@@ -72,6 +72,17 @@ orig_environ = {
     "LD_LIBRARY_PATH": os.environ.get("LD_LIBRARY_PATH", ""),
     "DYLD_LIBRARY_PATH": os.environ.get("DYLD_LIBRARY_PATH", ""),
 }
+# Define set of default units of measurmenet for common derived quantities
+default_uom = {
+    "pressure": "Pa",
+    "energy": "J",
+    "energy_mol": "J/mol",
+    "energy_mass": "J/kg",
+    "entropy": "J/K",
+    "entropy_mol": "J/mol/K",
+    "entropy_mass": "J/kg/K",
+    "power": "W",
+}
 
 def canonical_arch(arch):
     """Get the offical machine type in {x86_64, aarch64} if possible, otherwise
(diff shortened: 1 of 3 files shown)


eslickj:
@jsiirola and @andrewlee94, is core.util overused?  Maybe we should have core.units?  core.scaling and a few others?  I mean util sounds like some random stuff that you may find useful.  As it is there is some really important stuff buried in there.

eslickj (approved):

dangunter (approved):
Looks good. Documentation of where/when to set reporting units is needed, if not elsewhere already.

elbashandy (approved):
Looks good

PR #821, opened 2022-04-21
MEA solvent column model

anujad95:
## Fixes
None

## Summary/Motivation:
The MEA solvent column has been included as a unit model.

- Addition of the MEA solvent column model.

### Legal Acknowledgement


idaes/models_extra/column_models/MEAsolvent_column.py
```diff
@@ -0,0 +1,1483 @@
...
+            res = opt.solve(blk, tee=slc.tee)
+
+        init_log.info_high(
+            "Step 5 complete: {}.".format(idaeslog.condition(res)))
+
+        # ---------------------------------------------------------------------
+        blk.vapor_phase.release_state(flags=vflag)
+        blk.liquid_phase.release_state(flags=lflag)
+
+        if (res.solver.termination_condition != TerminationCondition.optimal or
```
andrewlee94, line 1282: You should use `check_optimal_termination` from `pyomo.environ` here instead of these two conditions. It does the same thing but is more compact, and also is capable of handling solvers other than IPOPT with different exit codes.

idaes/models_extra/column_models/MEAsolvent_column.py
```diff
@@ -0,0 +1,1483 @@
...
+        
+        # ---------------------------------------------------------------------
+        init_log.info('Step 6: Velocity constraints')
+        blk.velocity_vap.unfix()
+        blk.velocity_liq.unfix()
+        
+        for c in velocity_constraints:
+            getattr(blk, c).activate()
+            
+        use_idaes_solver_configuration_defaults()
```
andrewlee94, line 1299: I do not think you should be changing solver options in the middle of a solve like this (unless it is absolutely necessary). Could you have just set these from the start?

idaes/models_extra/column_models/MEAsolvent_column.py
```diff
@@ -0,0 +1,1483 @@
...
+        optarg = {
+                "tol": 2e-8,
+                "max_iter": 150,
```
andrewlee94, line 1302: 150 iterations seems excessive for an initialization routine which should be a square problem (I would generally set the maximum to 50, and even that is generally too many). Do you really need to set them limit this high?

idaes/models_extra/column_models/MEAsolvent_column.py
```diff
@@ -0,0 +1,1483 @@
...
+                "bound_push":1e-10}
+        opt.options = optarg
+        opt.solve(blk, tee=True)
```
andrewlee94, line 1305: All of these solver calls should be wrapped in the IDAES solver logging context manager.

See line 971 in https://github.com/IDAES/idaes-pse/blob/main/idaes/models/unit_models/pressure_changer.py if you need an example.

idaes/models_extra/column_models/MEAsolvent_column.py
```diff
@@ -0,0 +1,1483 @@
...
+            iscale.set_scaling_factor(v, 1/value(blk.holdup_liq[t, x]))
+            
+        # Scale constraint
+        for (t, x), v in blk.holdup_liq_constraint.items():
+            iscale.constraint_scaling_transform(
+            v, iscale.get_scaling_factor(blk.holdup_liq[t, x]))
+        
+        # ---------------------------------------------------------------------    
+        init_log.info('Step 9: Vapor phase mass transfer coefficient constraint')
+        init_log.info('Initializing mass transfer coefficient (vapor phase) - degrees_of_freedom = {}'.format(degrees_of_freedom(blk)))
```
andrewlee94, line 1354: Should we be logging the degrees of freedom, or doing a check and raising and exception if not 0?

idaes/models_extra/column_models/MEAsolvent_column.py
```diff
@@ -0,0 +1,1483 @@
...
+            
+        # ---------------------------------------------------------------------    
+        init_log.info('Step 13: Flooding constraint')
+        init_log.info('Initializing flooding calculations - degrees_of_freedom = {}'.format(degrees_of_freedom(blk)))   
+        
+        blk.floodfraction.unfix()
+        blk.flood_fraction_constr.activate()
+        
+        opt.solve(blk, tee=True)
+        init_log.info('Step 14: Initialization completed - degrees_of_freedom = {}'.format(degrees_of_freedom(blk)))
```
andrewlee94, line 1482: 1. You are missing the step to unfix the inlet conditions using the flags from the property/control volume initializations.
2. You need to do a final check of the solver status and raise an Initialization error if the conditions are not optimal.

idaes/models_extra/column_models/MEAsolvent_column.py
```diff
@@ -0,0 +1,1483 @@
...
+            else:
+                pass
+            
+        for (t, x), v in blk.E3_eqn.items():
+            if x!=1:
+                iscale.constraint_scaling_transform(
+                v, 1)
+            else:
+                pass
+        optarg = {
```
andrewlee94, line 1432: Once gain, I am wary of hard coding solver options like this.

andrewlee94 (changes requested):
I have a lot of comments, and there are a few things missing from this PR. First and foremost, there need to be unit tests for this model as well as some documentation.

andrewlee94 (changes requested):
I think this is looking good, but it needs tests for the MEA column model.

idaes/models_extra/column_models/tests/test_MEAsolvent_column.py
```diff
@@ -0,0 +1,182 @@
...
+# Import Pyomo libraries
+from pyomo.environ import (ConcreteModel,
+                           value,
+                           Constraint,
+                           check_optimal_termination,
+                           units as pyunits)
+
+# Import IDAES Libraries
+from idaes.core import FlowsheetBlock
+from idaes.models_extra.column_models.solvent_column \
```
andrewlee94, line 29: Is this import required?

idaes/models_extra/column_models/tests/test_MEAsolvent_column.py
```diff
@@ -0,0 +1,182 @@
...
+    import PackedColumn
+from idaes.models_extra.column_models.MEAsolvent_column import MEAColumn
+from idaes.models.properties.modular_properties.base.generic_property import (
+        GenericParameterBlock)
+from idaes.models_extra.column_models.properties.MEA_vapor \
+    import flue_gas as vaporconfig_absorber
```
andrewlee94, line 35: Be consistent with use of underscores - I would suggest this should be `vapor_config` (I don't think the "absorber" part is necessary unless you also do the striper part (which wouldn't be a bad idea).

idaes/models_extra/column_models/tests/test_MEAsolvent_column.py
```diff
@@ -0,0 +1,182 @@
...
+solver.options["bound_push"] = 1e-10
+
+class TestAbsorber:
+    @pytest.fixture(scope="class")
+    def model(self):
+        m = ConcreteModel()
+        m.fs = FlowsheetBlock(default={"dynamic": False})
+    
+        # Set up molar flow bounds
+        vaporconfig_absorber["state_bounds"]["flow_mol"] = (0, 1, 1e6, pyunits.mol / pyunits.s)
```
andrewlee94, line 58: You should create a copy of the config dict before doing this - otherwise you have just changed the original version which might affect other tests which are run later.

idaes/models_extra/column_models/tests/test_MEAsolvent_column.py
```diff
@@ -0,0 +1,182 @@
...
+        liquidconfig["state_bounds"]["flow_mol"] = (0, 1, 1e6, pyunits.mol / pyunits.s)
+    
+        # Set up property package
+        m.fs.vapor_properties = GenericParameterBlock(default=vaporconfig_absorber)
+        m.fs.liquid_properties = GenericParameterBlock(default=liquidconfig)
+    
+        # Number of finite elements and finite element list in the spatial domain
+        x_nfe = 40
+        x_nfe_list = [i / x_nfe for i in range(x_nfe + 1)]
```
andrewlee94, line 67: This line is not necessary (you only need to do this is you are doing a non-uniform mesh).

idaes/models_extra/column_models/tests/test_MEAsolvent_column.py
```diff
@@ -0,0 +1,182 @@
...
+        m.fs.unit.liquid_inlet.pressure.fix(100000) # Pa
+        m.fs.unit.liquid_inlet.mole_frac_comp[0,"CO2"].fix(0.022)
+        m.fs.unit.liquid_inlet.mole_frac_comp[0,"H2O"].fix(0.868)
+        m.fs.unit.liquid_inlet.mole_frac_comp[0,"MEA"].fix(0.11)
+
+        return m
+
+    @pytest.mark.unit
+    def test_degrees_of_freedom(self, model):
+        assert degrees_of_freedom(model) == 0
```
andrewlee94, line 108: Could you add a check for unit consistency please.

andrewlee94 (commented):
A few minor things needed, but they should be easy to address.

idaes/models_extra/column_models/tests/test_MEAsolvent_column.py
```diff
@@ -55,21 +54,21 @@ def model(self):
         m.fs = FlowsheetBlock(default={"dynamic": False})
     
         # Set up molar flow bounds
-        vaporconfig_absorber["state_bounds"]["flow_mol"] = (0, 1, 1e6, pyunits.mol / pyunits.s)
-        liquidconfig["state_bounds"]["flow_mol"] = (0, 1, 1e6, pyunits.mol / pyunits.s)
+        vaporconfig_copy = vaporconfig.copy()
```
andrewlee94, line 57: I think you need to do `deepcopy` here - hopefully that will fix the issues with the solvent condenser test.

andrewlee94 (approved):

PR #787, opened 2022-04-04
Grid integration double loop part 1

xiangao1:
## Summary/Motivation:
Open this PR to replace #638. 

## Changes proposed in this PR:
- Copied necessary files from #638 
- The new PR was needed because #638 originated from a branch on @Robbybp's fork which included some experimental changes to Caprese, which should not be included in this PR.

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #824, opened 2022-04-25
Pass static generator parameters to Prescient from bidder

xiangao1:
## Fixes
Pass static generator parameters to Prescient from bidder.

## Summary/Motivation:
This PR enables users to pass static generator parameters, e.g., Pmax, Pmin, to Prescient through double loop. 

First, I am proposing a class that holds the parameters. 

How the class will be used?
1. An instance of this class is required to create a `bidder` / `tracker`. (This can be part of the effort to address this: https://github.com/IDAES/idaes-pse/pull/787#discussion_r848854374)
2. When bidding into the market, these values will be passed to Prescient by `bidder`.
3. In this way, we don't have to maintain a lot of individual properties, e.g., pmin, pmax, in a `bidder` object.

Then I implemented functionalities to pass these static parameters using the model data object described above through the coordinator.

- Implement the data structure and corresponding validators.
- Unit test for the data structure.
- Implement functionalities to pass static parameters into Prescient.
- Update tests and examples correspondingly.

### Legal Acknowledgement

5. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.


bknueven (commented):
Given that this code is *doing* validating, I think it would be nice to have at least the `ValueError` exceptions tested.

This code also does floating point comparisons. I wonder if things like `value < self.min_value` need to be more like `value < self.min_value and not math.isclose(value, self.max_value` (e.g., "much less than" https://github.com/grid-parity-exchange/Egret/blob/828bd72236bced3a2c063bf63665fb569f4dfe3f/egret/model_library/unit_commitment/params.py#L1108)

If the validators need to be a bit more forgiving, e.g., because they're validating the output of an optimization model, it may be best for them to *return* the validated value so it can be adjusted, if needed, e.g., if `value < self.min_value` but `value` was close to `self.min_value` then `self.min_value` could be used instead.

xiangao1:
> Given that this code is _doing_ validating, I think it would be nice to have at least the `ValueError` exceptions tested.
> 
> This code also does floating point comparisons. I wonder if things like `value < self.min_value` need to be more like `value < self.min_value and not math.isclose(value, self.max_value` (e.g., "much less than" https://github.com/grid-parity-exchange/Egret/blob/828bd72236bced3a2c063bf63665fb569f4dfe3f/egret/model_library/unit_commitment/params.py#L1108)
> 
> If the validators need to be a bit more forgiving, e.g., because they're validating the output of an optimization model, it may be best for them to _return_ the validated value so it can be adjusted, if needed, e.g., if `value < self.min_value` but `value` was close to `self.min_value` then `self.min_value` could be used instead.

@bknueven Thanks! I have added more tests and applied "much less than".

Right now I think this model data structure will only be used to store the static data. The user will "manually" create one instance of this, and then use it to create `bidder` and `tracker`. So I think it's ok that the validators are not forgiving. If the bid comes from optimization problem solutions, we have some code to adjust values, e.g., [here](https://github.com/IDAES/idaes-pse/blob/9abeb3bd34cc03c21120637cecb2d34ab1961bb0/idaes/apps/grid_integration/bidder.py#L359).

xiangao1:
Hi @bknueven @adowling2, I have refactored this PR for passing static parameters from bidder to Prescient. Could you review the changes? Thanks!

idaes/apps/grid_integration/model_data.py
```diff
@@ -0,0 +1,348 @@
...
+
+        if production_cost_bid_pairs[0][0] != self.p_min:
+            raise ValueError(
+                f"The first power output in the bid should be the Pmin {self.p_min}, but {production_cost_bid_pairs[0][0]} is provided."
+            )
+
+        if production_cost_bid_pairs[-1][0] != self.p_max:
+            raise ValueError(
+                f"The last power output in the bid should be the Pmax {self.p_max}, but {production_cost_bid_pairs[-1][0]} is provided."
+            )
```
bknueven, line 322: I think its fine to do this validation, but its no longer required for Egret/Prescient.

idaes/apps/grid_integration/coordinator.py
```diff
@@ -353,6 +356,64 @@ def _clone_tracking_model(self):
...
+        for param, value in self.bidder.bidding_model_object.model_data:
+            if param != "gen_name":
+                continue
+            elif param == "p_cost":
+                curve_value = convert_marginal_costs_to_actual_costs(value)
+                gen_dict[param] = {
+                    "data_type": "cost_curve",
+                    "cost_curve_type": "piecewise",
+                    "values": curve_value,
+                }
```
bknueven, line 380: Is this covered in an integration test? seems like important code to cover for Prescient interaction.

bknueven (commented):
Looks pretty good -- I have a question about testing/coverage.

ksbeattie:
@adowling2 is this needed for the DISPATCHES April/May (internal) or the June (milestone) release?

adowling2:
This will be helpful for two of the DISPATCHES case studies. It is definitely nice to have for June; I am not sure if the case studies will need it.

idaes/apps/grid_integration/coordinator.py
xiangao1: Thanks @bknueven ! I fixed it!

ksbeattie:
If this is needed for the DISPATCHES June release, then it needs to get into the idaes-pse May release, so this month.

idaes/apps/grid_integration/coordinator.py
```diff
@@ -353,6 +356,64 @@ def _clone_tracking_model(self):
...
+                if "p_fuel" in gen_dict:
+                    gen_dict.pop("p_fuel")
+            else:
+                gen_dict[param] = value
```
adowling2, line 384: Add a comment here explaining what this is doing?

idaes/apps/grid_integration/coordinator.py
```diff
@@ -353,6 +356,64 @@ def _clone_tracking_model(self):
...
+
+            if param == "startup_cost" and "startup_fuel" in gen_dict:
```
adowling2, line 386: Is there a reason this is outside the if, elif, else statement above? If so, just add a one-line comment explaining

adowling2 (changes requested):
Small request for small explanation/extra comments.

xiangao1:
Thanks, @adowling2 ! I have addressed your comments.

adowling2 (approved):
Looks good; my earlier comments have been addressed.

xiangao1:
@ksbeattie @lbianchi-lbl This is ready to merge.

PR #825, opened 2022-04-25
Color variable types

elbashandy:
## Summary/Motivation:
Adds visual indicators in the UI for different Variable Types in the Stream table. Supported Variable types include: fixed, unfixed, Parameter, and Expression

To test it properly please run the ngcc example here: 

https://github.com/IDAES/examples-pse/blob/main/src/Examples/Flowsheets/ngcc/ngcc.ipynb

It should show variable types in the Stream table similar to this:

Screen Shot 2022-04-25 at 5 11 19 PM

-
-

### Legal Acknowledgement


dangunter:
When I test this on the HDA flowsheet (HDA_flowsheet_with_distillation_solution), only pressure shows as fixed even though the temperature and associated `flow_mol_phase_comp` indexed vars are also fixed in the flowsheet. So, for example, if you visualize the flowsheet right before "Initialization", you do not see all the fixed vars in cells 23 and 24 above it. (Note: I am working with version 1.13.0 of the examples (`idaes get-examples -U -V 1.13.0`) although I think the HDA flowsheet is pretty stable so this probably also applies to examples-pse main.)

elbashandy:
@dangunter I fixed the issue. You can give it another try

idaes/core/ui/flowsheet.py
```diff
@@ -226,6 +228,8 @@ def _identify_arcs(self):
             self._known_endpoints.add(component.source.parent_block())
             self._known_endpoints.add(component.dest.parent_block())
             self._ordered_stream_names.append(component.getname())
+            if self.add_variable_type:
```
dangunter, line 231: I don't understand the approach here. Why not put the variable type in its own data structure, e.g. `self._ordered_stream_types`?

idaes/core/ui/fsvis/static/js/stream_table.js
```diff
@@ -15,6 +43,59 @@ export class StreamTable {
...
+            const elem_vartype = document.createElement('span'); // Parent node
+            elem_vartype.className = stream_table_class;
+
+            // Create dot with the right color and the right variable type text
+            const elem_dot = document.createElement('span');
+            const elem_text = document.createElement('span');
+            elem_text.className = 'streamtable-vartype-text';
+            switch (var_type) {
+                case this.UNFIXED.num:
+                    console.info(`Unfixed variables don't have a visual indicator`);
```
dangunter, line 74: It seems pretty "chatty" to output this every time you render

idaes/core/ui/fsvis/static/js/stream_table.js
```diff
@@ -87,13 +170,40 @@ export class StreamTable {
                         row_object[variable_col] = row_object[variable_col] + '<span class="streamtable-units">&ndash;</span>';
                     }
                 }
+                else if (columns[col_index].includes("_vartype")) {
```
dangunter, line 173: ok, so you're decoding your own encoding of the column name -- again, wouldn't this be easier by just passing an extra list of types that matches the columns; or alternatively, a dictionary of `{column_name: type}` ?

idaes/core/util/tables.py
```diff
@@ -250,13 +256,93 @@ def create_stream_table_dataframe(
...
+            (default) where streams are displayed as columns, or 'index' where
+            stream are displayed as rows.
+        precision: rounding the floating numbers to the give precision. Default
+            is 5 digits after the floating point.
+
+    Returns:
+        A pandas DataFrame containing the stream table data.
+    """
+    # Variable Types:
+    class VariableTypes():
```
dangunter, line 307: Don't need parentheses after class name.

idaes/core/ui/fsvis/static/js/stream_table.js
```diff
@@ -15,6 +25,61 @@ export class StreamTable {
...
+            const elem_dot = document.createElement('span');
+            const elem_text = document.createElement('span');
+            elem_text.className = 'streamtable-vartype-text';
+            const var_type_class = `streamtable-vartype-${var_type}`;
+            switch (var_type) {
+                case this.UNFIXED:
+                    // This will execute once since this.existing_var_types is a set
+                    console.debug(`Unfixed variables don't have a visual indicator`);
+                    break;
+                case this.FIXED:
```
dangunter, line 60: Since all these cases do the same exact thing, you can have only one body:
```
case this.FIXED:
case this.PARAMETER:
case this.EXPRESSION:
    elem_dot.className = var_type_class;
    elem_dot.title = var_type;
    elem_text.innerHTML = var_type;
    break;
default: ..etc..
```

idaes/core/ui/fsvis/static/js/stream_table.js
```diff
@@ -87,13 +154,34 @@ export class StreamTable {
...
                     }
                 }
-                else {
+                else if (columns[col_index] === "Variable") {
                     row_object[columns[col_index]] = data[col_index];
                 }
+                else {
+                    var [value, type] = data[col_index];
+                    switch (type) {
+                        case this.FIXED:
```
dangunter, line 163: See comment for L60 above, again you can compress to one body:
```
this.existing_var_types.add(type)
```

dangunter: Note that if you defined an object whose keys were these four values, you could also do:
```
if type in this.VAR_TYPES:
    this.existing_var_types.add(type)
else:
   console.warn(..) etc.
```
This would also simplify the other case statement(s).

dangunter (changes requested):
A few suggestions, none of them very critical. Just respond to them and I can approve quickly the next time.

ksbeattie (approved):
LGTM

dangunter (approved):
ok looks good

PR #827, opened 2022-04-26
Fix serializing with top level BlockData

### LINKED ISSUE #826: Inlet states are not unfixed after initialization of indexed blocks
@eslickj  The changes in this commit https://github.com/IDAES/idaes-pse/commit/7c6bb17ddd26b0abf4e8e88b51b8684db2b2a8a5 are causing the following issue. For indexed blocks, the `initialize` method does not unfix the inlet state after initialization. I have attached a script that demonstrates this issue. The function `build_fs1` builds a flowsheet containing indexed turbines (turbine[1] and turbine[2]) and the function `build_fs2` builds a flowsheet containing non-indexed turbines (turbine_1 and turbine_2). In both cases, the degrees of freedom must remain zero after initialization. In the second case, the degrees of freedom is zero after initialization, but in the first case, it becomes -6 since the inlet states are not unfixed.

[script.zip](https://github.com/IDAES/idaes-pse/files/8565620/script.zip)
 
@nareshsusarla

eslickj:
## Fixes #826

_BlockData is not in the components to serialize (since it is component data), so if you wanted to serialize a block data object as the root object, it wouldn't work since _BlockData isn't in the list of component classes to serialize.  I fixed this by catching _BlockData and pretending it's a Block when looking at the list of components to serialize.  This should only happen when a _BlockData object is the root object.  Otherwise, it only shows up as component data.  Sorry I know that's confusing, but it fixes #826

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/model_serializer.py
+++ b/idaes/core/util/model_serializer.py
@@ -317,6 +317,12 @@ class StoreSpec(object):
                 alist = self.classes[cl][0]
                 ff = self.classes[cl][1]
                 break
+        if isinstance(o, Block._ComponentDataClass):
+            # If you're here you are trying to serialize an element of an
+            # indexed block at the top level.  We do want to allow that, so
+            # we'll pretend it's a block.
+            alist = self.classes[Block][0]
+            ff = self.classes[Block][1]
         return (alist, ff)
 
     def get_data_class_attr_list(self, o):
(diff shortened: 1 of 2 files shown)


andrewlee94 (approved):
Looks good to me.

eslickj:
@andrewlee94, sorry I left a print in the test again.  Just removed it.  Could you review again?

andrewlee94 (approved):

radhakrishnatg (approved):
It fixed the issue. Thanks, John!

nareshsusarla:
Yes it works. Thanks

PR #834, opened 2022-04-29
Remove CEOS tests for unused external functions

eslickj:
## Fixes

This removes test for CEOS external functions that are no longer used.  This will let me remove them from the external function.

### Legal Acknowledgement


idaes/config.py
```diff
@@ -482,13 +482,16 @@ def create_dir(d):
...
 def get_data_directory():
     """Return the standard data directory for idaes, based on the OS."""
-    try:
-        if os.name == 'nt':  # Windows
-            data_directory = os.path.join(os.environ['LOCALAPPDATA'], "idaes")
-        else:  # any other OS
-            data_directory = os.path.join(os.environ['HOME'], ".idaes")
-    except AttributeError:
-        data_directory = None
+    if 'IDAES_DATA' in os.environ and os.environ['IDAES_DATA']:
```
dallan-keylogic, line 485: Might be cleaner to expand `os.environ['IDAES_DATA']` to `os.environ['IDAES_DATA'] is not None` to clarify what's going on here.

dallan-keylogic (approved):

andrewlee94 (approved):
I will assume this is OK (it is hard to tell without seeing the changes in the cubic code).

idaes/config.py
eslickj: I was actually checking for a blank string, but maybe I should check that up one level is a valid directory.  I'll make that change.

eslickj:
@andrewlee94 the changes to the cubic code are in the last PR merged into IDAES-EXT.  The only two function currently used remain basically unchanged, other than slight reorganization of the code.  The real change is that I deleted the unused ones and added a couple still unused alternatives (fake root and NAN for the nonexistent root).  Those new ones are for some new formulation ideas @viiibhav is working on.

andrewlee94 (approved):

dallan-keylogic (approved):

eslickj:
I added a check and log warning if the data_directory doesn't exist, and now that I can actually test this, fixed up some things that fail if there is no data directory.  By the way, the reason I included the data directory thing in this PR is so that you can have multiple versions of the IDAES directory around.  It will let you have different versions of the binaries (which could be important if you want to work with older versions of IDAES especially with functions used by old versions removed).  An unrelated, use is bypassing a modified global config file for testing by creating an IDAES data directory just for tests.  Part of getting you IDAES environment can be setting the IDAES_DATA environment var.

eslickj:
@andrewlee94, I tested the IDAES_DATA environment variable myself, but testing it in CI would (I think) require a couple tests run separately.  I guess we'll either have to adim merge this or I'll have to figure out how to run tests in another python instance.

dallan-keylogic (approved):

andrewlee94 (approved):

PR #836, opened 2022-05-02
Moving NGFC To Examples

bpaul4:
## Fixes

- <strike>Improved variable and constraint scaling to increase tractability during initialization</strike>
- <strike>Modified solve scheme to initialize unit models and solve flowsheet to ensure robustness solution</strike>
- <strike>Serialized solved points post-initialization and post-solve</strike>
- <strike>Associated fixes to property packages and updates to tests</strike>
- Update: this flowsheet is moving to the examples repository via https://github.com/IDAES/examples-pse/pull/119 and files are removed from the main repository in this PR

## Changes proposed in this PR:
-
-

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #847, opened 2022-05-05
New Binary Version (More ARM64)

eslickj:
## Summary/Motivation:

This mainly adds macOS binaries for M1-based Macs.  Since I also have a convenient way to build and test ARM64 Linux binaries now that doesn't take all day (doesn't rely on QEMU), I threw more of those in too.  (For binary changes see https://github.com/IDAES/idaes-ext/releases/ 2.6.1 and 2.6.2)

This (well not this PR but the new binaries) also adds some extra cubic root finding functions for Cubic EoSs.  These aren't used anywhere in IDAES yet, but they are there to try out some new formulations on the side.  This also removes some older cubic EoS functions that are no longer used.

For testing these, I had to test the ARM64 versions locally.  For linux, I used docker and did the test the same way we do Intel on GitHub actions.  For macOS, I used a clean VM with UTM and just manually installed IDAES and ran the tests.

### Legal Acknowledgement


andrewlee94 (approved):
LGTM

dallan-keylogic (dismissed):

dallan-keylogic (approved):

eslickj:
If the NGFC test fails again on the integration tests, we may consider skipping it and bringing it back with PR #836.  I'd like to get some testing in with these binaries before we cut the release, and it would be nice to get the partial macOS support out.  Unfortunaly with every new platform and set of tests I add, I discover new fragile tests.

eslickj:
Well I messed up again.  The tests pass but, but I see some extra error messages that don't cause anything to fail, but are probably not goo.  Looks like I'll have to make a 2.6.3 binary release before merging this.

eslickj:
Okay, I broke down and resorted using the debugger to find the cause of metis crashing.  It's a GKfree function issue with variadic functions.  Here are details https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms, I'll just have to patch the metis 4.0.3 on macOS.

dallan-keylogic (approved):

andrewlee94 (approved):
LGTM

lbianchi-lbl:
@eslickj as I understand, the integration tests that are currently failing are related to #836. Should we try to get #836 merged before this one, to see if it's enough to address the test failures? (apologies if this has been asked already during the call today, I might have missed it if so).

andrewlee94:
@lbianchi-lbl Unfortunately, it looks like at least half of the failures are real failures (and the same sort of thing @eslickj has been fighting all week with these binaries).

eslickj:
@lbianchi-lbl and @andrewlee94, I think I just fixed the one test that was failing by adjusting the solver setting.  If you can review my last commit and the tests pass this should be good to go.  I think it would be good to have this before #836, because #836 has the same issue.  Don't want to spend a lot of time fixing it there to only need to figure out how to fix it again so we can merge this.

andrewlee94 (approved):

dallan-keylogic (approved):

PR #848, opened 2022-05-05
Adding better testing of unit model report method

andrewlee94:
## Summary/Motivation:
This PR improves the testing of the unit model `report` method by actually checking the output.

-
-

### Legal Acknowledgement


eslickj (approved):

idaes/models/unit_models/tests/test_equilibrium_reactor.py
```diff
@@ -242,9 +242,39 @@ def test_conservation(self, sapon):
...
+    Molar Concentration NaOH             mole / meter ** 3     100.00  0.0026894
+    Molar Concentration EthylAcetate     mole / meter ** 3     100.00  0.0026894
+    Molar Concentration SodiumAcetate    mole / meter ** 3     0.0000     99.997
+    Molar Concentration Ethanol          mole / meter ** 3     0.0000     99.997
+    Temperature                                     kelvin     303.15     304.32
+    Pressure                                        pascal 1.0132e+05 1.0132e+05
+====================================================================================
+"""
+
+        assert output in stream.getvalue()
```
lbianchi-lbl, line 277: - Defining intermediate variables could help pytest generate more readable diffs in case the assertion fails
- If the only reason we're using `in` instead of `==` in the assertion is the possible presence of newlines in either of the strings, we could `.strip()` both of them before comparing (I'm not sure if this has an effect on the failed assertion message)

```suggestion
        expected = """
====================================================================================
Unit : fs.unit                                                             Time: 0.0
------------------------------------------------------------------------------------
    Unit Performance

    Variables: 

    Key             : Value  : Units  : Fixed : Bounds
          Heat Duty : 0.0000 :   watt :  True : (None, None)
    Pressure Change : 0.0000 : pascal :  True : (None, None)

------------------------------------------------------------------------------------
    Stream Table
                                             Units            Inlet     Outlet  
    Volumetric Flowrate                meter ** 3 / second  0.0010000  0.0010000
    Molar Concentration H2O              mole / meter ** 3     55388.     55388.
    Molar Concentration NaOH             mole / meter ** 3     100.00  0.0026894
    Molar Concentration EthylAcetate     mole / meter ** 3     100.00  0.0026894
    Molar Concentration SodiumAcetate    mole / meter ** 3     0.0000     99.997
    Molar Concentration Ethanol          mole / meter ** 3     0.0000     99.997
    Temperature                                     kelvin     303.15     304.32
    Pressure                                        pascal 1.0132e+05 1.0132e+05
====================================================================================
"""
        generated = str(stream.getvalue())
        assert generated.strip() == expected.strip()
```

lbianchi-lbl (commented):
Aside from the (known and already discussed) drawbacks of relying on comparing human-readable strings in tests, this looks fine.

I've left a couple of suggestions (using one of the tests as an example, but they apply to all other similar tests). Implementing these could result in clearer error messages for failing tests, but they aren't critical.

idaes/models/unit_models/tests/test_equilibrium_reactor.py
bpaul4: I'm running into a similar issue debugging output checks in #711 , and am wary of spacing differences inside the string between lines as well. If we mainly wish to test the content we could remove the newlines and spacing before comparing, such as `assert generated.replace(" ", "").replace("\n", "") == expected.replace(" ", "").replace("\n", "")`. There's probably a more aesthetic way of doing this but it would ensure any string comparison failures are due to incorrect values and not formatting.

bpaul4: I went ahead and implemented the approach above in my tests for the PR linked above, and the tests are more stable for long outputs. Additionally, when there are no new lines or spaces the error report will actually skip identical leading or trailing characters as below, which greatly speeds up debugging and yields more helpful failure messages. In this case, a few of the values changed by tiny amounts which altered the final string output (e.g. -59.350**06318307154** should be corrected to -59.350**063183071555**):

```
>           assert output_string == stream_string
E           AssertionError: assert 'Extentofreac...=============' == 'Extentofreac...============='
E             Skipping 222 identical leading characters in diff, use -v to show
E             Skipping 2613 identical trailing characters in diff, use -v to show
E             - 063183071555DutyfromReaction(MW)):28.21686516526679Compressorwork(MW):-5.785653099883788e-25Turbinework(MW):-2.4215308912405678FeedMixeroutlettemperature(C)):20.051714213753655RecycleMixeroutlettemperature(C)):40.21858118534789FeedCompressoroutlettemperature(C)):40.21858118534789FeedCompressoroutletpressure(Pa)):5100000.0Heateroutlettemperature(C)):215.0Reactoroutlettemperature(C)):231.8500047840716Turbineoutlettemperature(C)):142.0419469...
E
E             ...Full output truncated (4 lines hidden), use '-vv' to show
```

andrewlee94: @bpaul4 I am working on a better way to test this which avoids the string comparisons. This PR will be updated shrotly.

andrewlee94:
@elbashandy I needed to update the serialized json files for the UI tests as part of this (as I corrected some errors in the underlying reporting functions). You should probably take a quick look to make sure I didn't do anything I shouldn't have.

idaes/models/unit_models/tests/test_cstr.py
```diff
@@ -252,39 +252,14 @@ def test_conservation(self, sapon):
...
-    Molar Concentration EthylAcetate     mole / meter ** 3     100.00     20.316
-    Molar Concentration SodiumAcetate    mole / meter ** 3     0.0000     79.684
-    Molar Concentration Ethanol          mole / meter ** 3     0.0000     79.684
-    Temperature                                     kelvin     303.15     304.09
-    Pressure                                        pascal 1.0132e+05 1.0132e+05
-====================================================================================
-"""
-
-        assert output in stream.getvalue()
+    def test_get_performance_contents(self, sapon):
```
bpaul4, line 255: Is it important to check the stream table output, as well?

bpaul4 (dismissed):
Just a minor comment on the CSTR test module, otherwise I think this looks good.

bpaul4:
It seems the check is setting the dictionary using `pytest.approx` but checks for exact strings/values in the final results. I think this would be resolved by using list comprehension, e.g. have `expected` contain `Inlet` entries in the form `'Molar Flow (mol/s)': 100` and then check the results with something like:
`for key, value in stable.to_dict['Units']:
    `assert value == expected['Units'][key]` (since the unit containers should be equal
`for key, value in stable.to_dict['Inlet']:
    `assert value == pytest.approx(expected['Inlet'][key], rel=1e-4)` (since the stream values should be equal within a tolerance)

bpaul4 (approved):
As long as it passes all the tests now, the new syntax looks good to me.

idaes/models/unit_models/tests/test_heat_exchanger.py
```diff
@@ -39,7 +40,7 @@
 from idaes.models.unit_models.heat_exchanger import (
     delta_temperature_lmtd_callback,
     delta_temperature_lmtd2_callback,
-    delta_temperature_lmtd3_callback,
+    # delta_temperature_lmtd3_callback,
```
jsiirola, line 43: Can this be deleted?

jsiirola (approved):
Looks good.  One minor cleanup.

idaes/models/unit_models/tests/test_heat_exchanger.py
andrewlee94: I thought I had deleted that...

PR #851, opened 2022-05-06
Saving model onchange

elbashandy:
## Summary/Motivation:
Enhances the way we save the model. Instead of saving it every time the mouse leaves the paper. We save it now when the graph is actually modified

-
-

### Legal Acknowledgement


idaes/core/ui/fsvis/static/js/main.js
```diff
@@ -135,7 +143,66 @@ export class App {
...
+     */
+    getSaveTimeInterval() {
+        let settings_url = "/setting?setting_key=".concat(this._save_time_interval_key);
+
+        let save_time_interval = this._default_save_time_interval;
+
+        $.ajax({url: settings_url, type: 'GET', contentType: "application/json"})
+            // On failure inform user and stop
+            .fail(error => this.informUser(
+                2, "Fatal error: cannot get setting value: " + error))
```
dangunter, line 156: not sure this should stop the application, since you can have a default value and survive.

dangunter (approved):
I would have just pushed the settings at the start instead of adding a new endpoint to retrieve them one at a time, but it looks like this works so whatever.

ksbeattie (approved):
LGTM

PR #854, opened 2022-05-06
Configure GHA run serialization to automatically cancel redundant workflow runs

lbianchi-lbl:
## Summary/Motivation:

- At some point in the last few months (thanks to @andrewlee94 for the most recent suggestion to implement this) GitHub Actions added the functionality to tag workflows so that only one run is allowed at any given time
- If the `cancel-in-progress` option is set to `true`, already running workflows will be canceled
- This is very useful for us since it reduces the need of "manual corralling" workflow runs when merging several PRs one after the other
- By choosing the concurrency `group` opportunely, we can require that at most one workflow run per PR should be running at a given time, so whenever a new commit is pushed to the PR head branch (e.g. by merging the current `main` into it), outdated and therefore unneeded workflow runs are cancelled automatically

-
-

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -25,6 +25,17 @@ on:
       # synchronize occurs whenever commits are pushed to the PR branch
       - synchronize
 
+concurrency:
+  # NOTE: the value of `group` should be chosen carefully,
+  # otherwise we might end up over- or under-canceling workflow runs
+  # e.g. if we want to have Codecov results for each commit on `main`,
+  # we should use something `github.sha` instead of `github.ref_name`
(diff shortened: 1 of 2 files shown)


lbianchi-lbl:
It seems to be working as expected.

First, when pushing a commit while a job for the same PR is running, the most recent run is queued:

image

Then, the original run is canceled automatically, resulting in the most recent run being active:

image

andrewlee94 (approved):
I will assume @lbianchi-lbl knows what he is doing.

ksbeattie (approved):
LGTM

PR #867, opened 2022-05-18
Bug in initialization of FpcTP property packages

andrewlee94:
## Summary/Motivation:
The WaterTAP project identified a bug in the initialization of modular property packages using FpcTP state variables with phase equilibrium. Fundamentally, this system cannot be initialized with the phase equilibrium constraints active without unfixing some of the flowrate state variables, which would run against the normal assumptions we make for the initialization workflow.

For now, this PR addresses the higher level issues of ensuring zero degrees of freedom at each solve step by adding logic to skip activation of the phase equilibrium constraints for these systems, as well as adding in the flowrate-mole fraction constraints for this system (which had been missed previously). This gets the state block to initialize correctly, but leaves in in a non-equilibrium state which might cause issues at the unit model level (this is a good case for providing initial guesses via state args).

- Adds logic to skip activation of the phase equilibrium constraint for Fpc systems.
- Adds in missing activation of the mole fraction constraints for Fpc and Fc systems
- Adds a test for initializing FpcTP systems with phase equilibrium based on the example from WaterTAP.
- Runs `black` on a couple of files.

### Legal Acknowledgement

### DIFF
--- a/idaes/models/properties/modular_properties/base/generic_property.py
+++ b/idaes/models/properties/modular_properties/base/generic_property.py
@@ -1506,6 +1506,8 @@ class _GenericStateBlock(StateBlock):
                         "component_flow_balances",
                         "sum_mole_frac",
                         "phase_fraction_constraint",
+                        "mole_frac_phase_comp_eq",
+                        "mole_frac_comp_eq",
                     ):
                         c.activate()
                     if c.local_name == "log_mole_frac_phase_comp_eqn":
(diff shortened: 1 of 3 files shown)


andrewlee94:
@lbianchi-lbl @ksbeattie This PR is required for WaterTAP to allow them to use the latest version of IDAES.

dallan-keylogic (dismissed):

dallan-keylogic (approved):

eslickj (approved):

Issue #868, opened 2022-05-18
Be able to test the UI from a docker container


elbashandy:
`localhost` is hardcoded in the UI code. To connect to the server from a docker container, it should be read as 0.0.0.0 from the docker container's perspective.

PR #875, opened 2022-05-20
Allowing user to make the server global

elbashandy:
## Fixes
Addresses #868 

## Summary/Motivation:
The user should be able to specify if the server should run as a localhost or as `0.0.0.0`

-
-

### Legal Acknowledgement


dangunter (changes requested):
The web page needs to still have localhost, or an actual IP of the server, instead of 0.0.0.0; javascript can't connect to a server at 0.0.0.0 so the following script pops up a webpage that fails:
```
from idaes.core.ui.fsvis import visualize
from idaes_examples.Tutorials.Basics.visualizer_tutorial import create_model
m = create_model()
m.fs.visualize('foo', localhost=False)
```

PR #884, opened 2022-05-27
Surrogate Visualization Tests and Documentation

bpaul4:
## Fixes
- Tests and documentation for the surrogate visualization module
- Most of the "additions" in this PR are from the CSV file; we could discuss importing the data from `examples-pse` or another location to keep this update small
-
-

### Legal Acknowledgement


bpaul4:
1. We could mention 'matplotlib' rather than the submodules. I thought it would be useful to mention the submodules as they are responsible for different aspects of the methods.
2. Do the three surrogate methods produce different objects? If they produce the same object type and data structure, we may not need to test all three.
3. I considered testing the 'show' argument, but found that a plot viewer appears while running in the console which pauses the tests.

OOAmusat (approved):
LGTM

PR #887, opened 2022-06-01
Add Delay to Raising Logger Error

dallan-keylogic:
## Summary/Motivation:
I had an application where PETSc was raising an `ApplicationError` because of a failed solve without any solver output logged. It seems that the reason why was that the logger is on a separate thread than the script file, but the exception was shutting down execution before the output could be captured. Introducing a one-second delay before reraising the exception fixed the issue in my application. I'm not sure how to write a test for this, though.

### Legal Acknowledgement


eslickj (dismissed):

jsiirola:
Would it be better to just explicitly flush the log stream on exit rather than to add an arbitrary delay?

eslickj:
@jsiirola, I thought that was already happening, but not working.  Let me take another look at this.

eslickj:
@dallan-keylogic, can you see if it works to the following to the top of the thread class log_value method:
```python
if self.stop.isSet():
   self.stream.flush()
```

dallan-keylogic:
@eslickj That modification did not work.

idaes/logger.py
```diff
@@ -297,6 +298,7 @@ def run(self):
             self.log_value()
             self.stop.wait(self.sleep)
             if self.stop.isSet():
+                time.sleep(2)
```
eslickj, line 301: we can probably update 2 to self.sleep

eslickj (approved):

idaes/logger.py
dallan-keylogic: I tried that, but got a test failure, I assume because the output was logged in a different order.

eslickj: What test failed?

eslickj: Ah, yeah it looks like the delay for self.sleep (1 s I think) is too short still.  I'll create an issue for potentially revisiting this that explains the problem more thoroughly.  I think we should merge this now to avoid the problem where the solver fails and we don't know why due to missing output, and we can discuss if there is a better solution later.

jsiirola:
My suspicion is that this is an issue with Windows stream buffering (as that is the only platform that we are seeing it on).  Pyomo faces something similar when working through the TeeStream infrastructure.  As I was digging through this, is there a particular reason that `solver_log` isn't implemented using:

```python
@contextmanager
def solver_log(logger, level=logging.ERROR):
    tee = logger.isEnabledFor(level)
    if not solver_capture():
        yield SolverLogInfo(tee=tee)
    else:
        with capture_output(LogStream(level, logger)) as s:
            yield SolverLogInfo(tee=tee)
```

@dallan-keylogic, can you try replacing `solver_logger` with this and see if the problem goes away?

ksbeattie:
@dallan-keylogic, will try the above suggestion from @jsiirola .

dallan-keylogic:
I merged the branches the wrong way when trying to get the new fix onto this branch. Closing this PR in favor of #1098.

PR #905, opened 2022-07-05
SSLW Compressor Exceptions

bpaul4:
- Add exception for compressors that do not assume isentropic thermodynamics
- Add tests for this exception

### Legal Acknowledgement


MAZamarripa (approved):
lgtm

idaes/models/costing/SSLW.py
```diff
@@ -1016,13 +1016,24 @@ def cost_compressor(
...
+        # (costing not needed)
+        elif hasattr(blk.unit_model.config, "thermodynamic_assumption"):
+            if (blk.unit_model.config.thermodynamic_assumption.name) != "isentropic":
+                # PYLINT-TODO-FIX fix exception message with correct number of arguments
+                raise ValueError(
+                    "{} - pressure changers without isentrpoic "  # pylint: disable=E1305
+                    "assumption are too simple to be costed. ".format(
+                        blk.unit_model.name
+                    )
+                )
```
lbianchi-lbl, line 1036:
```suggestion
            if blk.unit_model.config.thermodynamic_assumption.name != "isentropic":
                raise ValueError(
                    f"{blk.unit_model.name} - pressure changers without isentropic "
                    "assumption are too simple to be costed."
                )
```

idaes/models/costing/tests/test_SSLW_costing.py
```diff
@@ -851,13 +852,46 @@ def test_pfr(self, model):
...
+    def test_isothermal_compressor(self, model):
+        # Test exception for non-supported compressor flags
+        model.fs.unit = PressureChanger(
+            default={
+                "property_package": model.fs.pparams,
+                "compressor": True,
+                "thermodynamic_assumption": ThermodynamicAssumption.isothermal,
+            }
+        )
+        with pytest.raises(ValueError):
```
lbianchi-lbl, line 876: This is only a suggestion for a possible enhancement: to make this test more specific, you can specify a substring or pattern that the exception message is expected to contain using the `match` kwarg: https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-raises.

idaes/models/costing/SSLW.py
lbianchi-lbl: Suggestions for minor enhancements:

- I don't think the `# pylint:` directives here are needed anymore, so we can get rid of them
- Using f-strings instead of `.format()` is more readable
- Typo: `isentrpoic` in exception message

idaes/models/costing/tests/test_SSLW_costing.py
bpaul4: This is very useful, thank you for sharing!

idaes/models/costing/SSLW.py
```diff
@@ -1016,13 +1016,21 @@ def cost_compressor(
...
         if not blk.unit_model.config.compressor:
             raise TypeError(
                 "cost_compressor method is only appropriate for "
                 "pressure changers with the compressor argument "
                 "equal to True."
             )
+        # compressor = True, and using non-isentropic assumption
+        # (costing not needed)
+        elif hasattr(blk.unit_model.config, "thermodynamic_assumption"):
+            if (blk.unit_model.config.thermodynamic_assumption.name) != "isentropic":
```
lbianchi-lbl, line 1029: Nit: I think parentheses are not needed here.

bpaul4:
Codecov thinks line 1030 is uncovered when two tests execute this line. I support manually merging this PR, since the change is small and confirmed to work as expected.

idaes/models/costing/SSLW.py
```diff
@@ -1016,13 +1016,21 @@ def cost_compressor(
...
         if not blk.unit_model.config.compressor:
             raise TypeError(
                 "cost_compressor method is only appropriate for "
                 "pressure changers with the compressor argument "
                 "equal to True."
             )
+        # compressor = True, and using non-isentropic assumption
+        # (costing not needed)
+        elif hasattr(blk.unit_model.config, "thermodynamic_assumption"):
+            if blk.unit_model.config.thermodynamic_assumption.name != "isentropic":
```
andrewlee94, line 1029: I am not sure this is quite comprehensive enough - the firs thing that came to mind is what happens if the user is trying to use this on a Pump? In this case, the error message would be misleading - it is not that it can't be costed, but rather that the user is using the wrong costing method. I think we should do one of tow things:

1. Check more comprehensively for the differ types and return a more useful message.
2. Be more permissive, and only raise an error for those cases that don't work (i.e. fail for isothermal, but allow most others to proceeded perhaps with a warning).

idaes/models/costing/SSLW.py
```diff
@@ -1016,13 +1016,21 @@ def cost_compressor(
...
+        # Confirm that unit is a compressor
         if not blk.unit_model.config.compressor:
             raise TypeError(
                 "cost_compressor method is only appropriate for "
                 "pressure changers with the compressor argument "
                 "equal to True."
             )
+        # compressor = True, and using non-isentropic assumption
+        # (costing not needed)
+        elif hasattr(blk.unit_model.config, "thermodynamic_assumption"):
```
andrewlee94, line 1028: 1. Do we need the `hasattr` check here? What models do we expect not to have this attribute?
2. In the following line, I would check against the `Enum` object itself, not the name attribute of the `Enum`. I.e. ` if blk.unit_model.config.thermodynamic_assumption.name != ThermodynamicAssumption.isentropic`

andrewlee94 (changes requested):
A couple of comments from me.

idaes/models/costing/SSLW.py
bpaul4: That's a good point. This would only be an issue when users pass `cost_compressor` as the costing method for PressureChangers, since the `unit_mapping` will see Pumps and use the correct costing method if one is not provided. I agree with your first suggestion and will make the checks more explicit

idaes/models/costing/SSLW.py
```diff
@@ -1016,13 +1016,32 @@ def cost_compressor(
...
+        if (
+            blk.unit_model.config.thermodynamic_assumption
+            == ThermodynamicAssumption.pump
+        ):
+            raise ValueError(
+                f"{blk.unit_model.name} - pressure changers with the pump "
+                "assumption should use the cost_pump method."
+            )
+        # isothermal or adiabatic compressors are too simple to cost
+        elif (
```
andrewlee94, line 1037: For future expansion (e.g. if we ever add a polytropic option), it might be better to test for the exceptions instead. I.e.
```
elif blk.unit_model.config.thermodynamic_assumption in [ThermodynamicAssumption.isothermal, ThermodynamicAssumption.adiabatic]:

**Review by andrewlee94 — APPROVED** [review]
Looks good (assuming tests pass).

**Comment by dallan-keylogic** [comment]
What's the point of adding these exceptions? When I wanted to use these costing methods for a Helmholtz compressor, I had to add a bunch of stuff to the Config dictionary after the fact to get past a bunch these exceptions
```
 # Have to call method because Helmholtz model doesn't automatically call it
    fs.water_compressor.costing = pyo.Block()
    # Despite me specifying the mover type, it's still going to check the ThermodynamicAssumption
    fs.water_compressor.config.declare(
        "thermodynamic_assumption",
        ConfigValue(
            default=ThermodynamicAssumption.isentropic
        ),
    )
    fs.water_compressor.config.declare(
        "compressor",
        ConfigValue(
            default=True
        ),
    )
    pressure_changer_costing(fs.water_compressor.costing, mover_type="compressor", compressor_type="screw")
    add_total_plant_cost(fs.water_compressor)
```

**Review by dallan-keylogic — CHANGES_REQUESTED** [review]
Protecting the user from themselves adds a lot of overhead. I'm not convinced it's worth it in this case.

File: idaes/models/costing/SSLW.py [hunk 7]
```diff
@@ -1016,13 +1016,32 @@ def cost_compressor(
...
                 an integer or not (default = True).
         """
-        # Confirm that unit is a turbine
+        # Confirm that unit is a compressor
         if not blk.unit_model.config.compressor:
             raise TypeError(
                 "cost_compressor method is only appropriate for "
                 "pressure changers with the compressor argument "
                 "equal to True."
             )
```
dallan-keylogic, line 1025: If you call the `cost_compressor` function on a unit model that's not a compressor, the user deserves what they get.

idaes/models/costing/SSLW.py
```diff
@@ -1016,13 +1016,32 @@ def cost_compressor(
...
+            )
+        # isothermal or adiabatic compressors are too simple to cost
+        elif blk.unit_model.config.thermodynamic_assumption in [
+            ThermodynamicAssumption.isothermal,
+            ThermodynamicAssumption.adiabatic,
+        ]:
+            raise ValueError(
+                f"{blk.unit_model.name} - pressure changers without isentropic "
+                "assumption are too simple to be costed."
+            )
```
dallan-keylogic, line 1044: I've found cases where an isothermal compressor adequately describes a multi-stage isentropic compressor with intercooling.
```
        self.cmp04 = gum.Compressor(
            default={"property_package": self.h2_pure_prop_params,
                     "thermodynamic_assumption": gum.pressure_changer \
                         .ThermodynamicAssumption.isothermal
                     }
        )
        self.cmp04.efficiency_isentropic = pyo.Var(self.time, initialize=1,
                                                   units=pyo.units.dimensionless)
...
        @self.cmp04.Expression(self.time)
        def work_mechanical(b, t):
            return (b.outlet.flow_mol[t] *
                    (b.control_volume.properties_out[t].gibbs_mol - b.control_volume.properties_in[t].gibbs_mol)
                    / b.efficiency_isentropic[t])

        @self.cmp04.Expression(self.time)
        def heat_duty(b, t):
            return ((b.control_volume.properties_out[t].entr_mol
                     - b.control_volume.properties_in[t].entr_mol)
                    * b.outlet.temperature[t] * b.outlet.flow_mol[t]
                    - (1 - b.efficiency_isentropic[t]) * b.work_mechanical[t])
```

idaes/models/costing/SSLW.py
andrewlee94: It might be best to log a `warning` in these cases. I agree there are situations where you want to use a different unit model with the compressor costing.

dallan-keylogic (approved):
LGTM

PR #917, opened 2022-07-19
Fix stream brushing when column is hidden and reshown

### LINKED ISSUE #907: Hiding stream table columns breaks brushing
If you hide stream table columns the "brushing" stops working with some of the streams.

elbashandy:
## Fixes
Closes #907 

-
-

### Legal Acknowledgement


dangunter (changes requested):
The issue per se is fixed -- but why is "Variable" one of the columns that can be hidden? I don't think that makes much sense; that one should not be hide-able.

dangunter (approved):
Looks good

ksbeattie (approved):
LGTM, though to be honest, I couldn't reproduce the original problem.

ksbeattie:
Also, it seems little strange to me how when a column is highlighted and unhighlighted the contents of the column jump in and out a bit with the highlighting.

PR #922, opened 2022-07-26
Switch ProcessBlock 'default' argument to just use keyword arguments

### LINKED ISSUE #729: Remove need for "default" argument when declaring models with config arguments
It would be good to remove the need for the `default` argument when creating instances of IDAES mdoels. Currently, this argument is used to separate the IDAES-specific arguments (that go to the config block) from the Pyomo-specific arguments (that are used during object construction).

@jsiirola suggested that we should be able to set up an API to retrieve the expected arguments from the Pyomo objects so that we could explicitly separate these from the IDAES config arguments without the need for `default`.

eslickj:
## Summary/Motivation:

Instead of supplying the default argument to ProcessBlock init, this makes it so you can just supply keyword args.  This also deprecates the default argument.  

Tests will fail until Pyomo is updated (requires Pyomo/pyomo#2471). 

### Legal Acknowledgement


idaes/core/base/process_block.py
```diff
@@ -19,9 +19,13 @@
 
 import sys
 import logging
+import copy
```
eslickj, line 22: Remove copy import

data/config.yaml
```diff
@@ -1,8 +1,8 @@
 _id: d779730c7cc94155877a24e2e8be2ad1
```
eslickj, line 1: Don't change this file.

idaes/core/base/process_block.py
```diff
@@ -66,22 +70,50 @@ def _rule_default(b, *args):
             ..
 """
 
+def _get_pyomo_block_kwargs():
+    funcs=get_overloads_for(Block.__init__)
```
andrewlee94, line 74: A short comment might be helpful to remind us what this does in the future.

idaes/core/base/process_block.py
```diff
@@ -66,22 +70,50 @@ def _rule_default(b, *args):
...
 def _process_kwargs(o, kwargs):
     kwargs.setdefault("rule", _rule_default)
-    o._block_data_config_default = kwargs.pop("default", None)
     o._block_data_config_initialize = ConfigBlock(implicit=True)
     o._block_data_config_initialize.set_value(kwargs.pop("initialize", None))
     o._idx_map = kwargs.pop("idx_map", None)
+    _default = kwargs.pop("default", None)
+    if _default is not None:
+        deprecation_warning(
+            "The default argument for the ProcessBlock class is deprecated"
```
andrewlee94, line 90: It might be worth saying something like "Arguments can now be passed directly as keyword arguments" so users know how to fix this,

andrewlee94 (commented):
Two minor comment about documenting what is happening, but otherwise looks good to me.

data/config.yaml
```diff
@@ -1,8 +1,8 @@
...
-created: '2022-07-26T12:13:07.493686'
+created: '2022-04-18T14:36:30.754421'
```
andrewlee94, line 2: Not this PR, but we really need to work out what keeps changing this and fix it. @dangunter @lbianchi-lbl

andrewlee94 (approved):

eslickj:
I guess there are few tests still to fix that choked on the added warnings.  I'll fix them soon.

data/config.yaml
eslickj: I think its running test that include something data related.

eslickj:
@andrewlee94 and @jsiirola, I finally changed enough that the documentation tests pass.  I hope this is ready to go.

docs/reference_guides/core/process_block.rst
```diff
@@ -28,32 +28,32 @@ ProcessBlockData class.
...
+key is the block index and the value is the configuration dictionary.
 
-.. testcode:: 
+.. testcode::
 
     m = ConcreteModel()
-    m.b = MyBlock([0,1,2,3,4], default={"xinit":1, "yinit":2},
-                  initialize={0:{"xinit":1, "yinit":2}})
+    m.b = MyBlock(
+        [0,1,2,3,4], xinit=1, yinit=2, initialize={0:{"xinit":1, "yinit":2}}
```
andrewlee94, line 55: Whilst we are here, is `initialize` the best name for this? It might well be, but this is the time to ask the question. Given that "initialize" also has a very important (and different meaning) for solving the model, I wonder if we should avoid using it here to prevent any chance of confusion,.

That said, the only ideas I have had so far are some variation on "indexed_config" or similar, which I am not sure is really any better.

andrewlee94 (commented):
I didn't look through all the files, but what I saw looked good.

I only had one question, and it is a big one: is `initialize` the right name for passing indexed config argument?

docs/reference_guides/core/process_block.rst
eslickj: It is consistent with other promo components and they are arguments for object initialization.

eslickj: That said, it would be a heck of a lot easier to change than default.

eslickj: If we do change initialize, let's do it in another PR.

data/config.yaml
bpaul4: There are two DMF tests that create new config.yaml files, it's possible one or both are not actually creating temporary directories as expected: [test_datasets](https://github.com/IDAES/idaes-pse/blob/main/idaes/core/dmf/tests/test_datasets.py) [test_dmfbase](https://github.com/IDAES/idaes-pse/blob/main/idaes/core/dmf/tests/test_dmfbase.py)

docs/explanations/components/property_package/general_reactions/generic_reactions.rst
```diff
@@ -27,7 +27,7 @@ The most common way to use the Generic Reaction Package Framework is to create a
 
     m.fs.thermo_properties = PhysicalParameterBlock()
 
-    m.fs.reaction_properties = GenericReactionParameterBlock(default={"property_package": m.fs.thermo_properties, config_dict})
+    m.fs.reaction_properties = GenericReactionParameterBlock(property_package=m.fs.thermo_properties, **config_dict)
```
bpaul4, line 30: Why were the asterisks not needed before? Is this changed required now?

bpaul4, line 30: Why were the asterisks not needed before? Is this changed required now?

docs/reference_guides/core/process_block.rst
```diff
@@ -28,32 +28,32 @@ ProcessBlockData class.
           self.y = Var(initialize=self.config.yinit)
 
 The following example demonstrates creating a scalar instance of the new class.
-The ``default`` key word argument is used to pass information on the the
-MyBlockData ConfigBlock.
+The keyword arguments are used to pass information on the the MyBlockData
```
bpaul4, line 31: Typo, remove one 'the'

docs/reference_guides/model_libraries/power_generation/unit_models/soec_design.rst
```diff
@@ -56,19 +56,17 @@ are not necessarily realistic.
       }
 
       m.fs.o2_side_prop_params = GenericParameterBlock(
-          default=get_prop(sweep_comp, {"Vap"}, eos=eos),
+          **get_prop(sweep_comp, {"Vap"}, eos=eos),
```
bpaul4, line 59: These two instances, same question as above regarding '**' arguments

bpaul4 (approved):
A few comments, nothing critical though.

docs/explanations/components/property_package/general_reactions/generic_reactions.rst
eslickj: The dictionary ``config_dict`` was passed to a keyword argument called ``default`` as a dictionary.  Now we are pass in the same information directly as keyword args.  The asterisks expand the dict into keyword arguments.

lbianchi-lbl:
@eslickj the failing jobs are in the environments where IDAES is installed without the `[prerelease]` tag, and therefore it uses the latest released version of Pyomo instead of the `IDAES/Pyomo` tag containing the latest changes. My suggestion would be to use the URL to the `IDAES/Pyomo` tag in the `install_requires` list as well (i.e. in addition to the `prerelease` `extras_require` target, which I'd like to get rid of at some point in the near future):

```py
setup(
    ...
    install_requires=[
        # idaes core / dmf
        "backports.shutil_get_terminal_size",
        "bunch",
        "click>=8",
        "colorama",
        "distro",  # help identify linux distros for binary downloads
        "flask",  # for ui/fsvis
        "flask-cors",
        "jupyter",
        "lxml",
        "matplotlib",
        "nbconvert",
        "nbformat",
        "numpy",
        "networkx",
        "omlt==0.3.1",  # fix the version for now as package evolves
        "pandas",
        "pint",
        "psutil",
        "pyomo @ https://github.com/IDAES/pyomo/archive/6.4.1.idaes.2022.07.27.zip",
        "pytest",
        "pyyaml",
        "requests",  # for ui/fsvis
        "python-slugify",  # for ui/fsvis
        "scipy",
        "sympy",
        "tinydb",
        "rbfopt",
        "xlrd",  # for DMF read of old .xls Excel files
        "openpyxl",  # for DMF read of new .xls Excel files
        # lbianchi-lbl: see https://github.com/IDAES/idaes-pse/issues/661
        "ipython<8.0.0",
    ],
    ...
)
```

PR #925, opened 2022-08-04
Create work term for Pressure Changer with EnergyBalanceType.none

adam-a-a:
## Summary/Motivation:
In WaterTAP, we noticed that having a property model with `EnergyBalanceType.none` would be incompatible with a Pump unit. Thus, to be compatible with the Pump, the property model currently needs to have  `EnergyBalanceType.enthalpyTotal` set. However, enthalpy is not necessary for the Pump; we only really care about volumetric flow rate, pressure (head) and efficiency to then calculate work. 

In short, the `work` variable is only ever created on a ControlVolume Block when `EnergyBalanceType.enthalpyTotal` is set which shouldn't always be the case. For more details, please have a look at this discussion in our WaterTAP repo: https://github.com/watertap-org/watertap/issues/651

- add conditional check before Reference connecting `control_volume.work` to `self.work_mechanical` in pressure_changer unit
- if the conditional is true, then `work` variable is created on control volume
- the conditional ensures the following is true before creating `work` on control volume within pressure_changer unit:
1. `work` does not exist on control volume
2. `EnergyBalanceType` is none
3. `ThermodynamicAssumption` is `Pump`

### Legal Acknowledgement

### DIFF
--- a/idaes/models/unit_models/pressure_changer.py
+++ b/idaes/models/unit_models/pressure_changer.py
@@ -105,8 +105,8 @@ class IsentropicPerformanceCurveData(ProcessBlockData):
             try:
 
                 @self.Expression(self.flowsheet().time)
-                def head_isentropic(b, t):  # units are energy/mass
-                    b = b.parent_block()
+                def head_isentropic(self, t):  # units are energy/mass
+                    b = self.parent_block()
                     if hasattr(b.control_volume.properties_in[t], "flow_mass"):
                         return (
                             b.work_isentropic[t]
@@ -120,8 +120,8 @@ class IsentropicPerformanceCurveData(ProcessBlockData):
                         )
 
                 @self.Expression(self.flowsheet().time)
-                def head(b, t):  # units are energy/mass
-                    b = b.parent_block()
+                def head(self, t):  # units are energy/mass
+                    b = self.parent_block()
                     if hasattr(b.control_volume.properties_in[t], "flow_mass"):
                         return (
                             b.work_mechanical[t]
@@ -367,10 +367,10 @@ see property package for documentation.}""",
 
         # Pressure Ratio
         @self.Constraint(self.flowsheet().time, doc="Pressure ratio constraint")
-        def ratioP_calculation(b, t):
+        def ratioP_calculation(self, t):
             return (
-                b.ratioP[t] * b.control_volume.properties_in[t].pressure
-                == b.control_volume.properties_out[t].pressure
+                self.ratioP[t] * self.control_volume.properties_in[t].pressure
+                == self.control_volume.properties_out[t].pressure
             )
 
         # Construct equations for thermodynamic assumption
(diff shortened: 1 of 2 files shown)


idaes/models/unit_models/pressure_changer.py
```diff
@@ -342,6 +343,20 @@ def build(self):
...
         # Set references to balance terms at unit level
         # Add Work transfer variable 'work'
+        # If the 'work' variable wasn't already built on the control volume but is needed, create it now.
+        if (
+            not hasattr(self.control_volume, "work")
+            and self.config.thermodynamic_assumption == ThermodynamicAssumption.pump
+            and eb is None
+        ):
+            units = self.config.property_package.get_metadata().get_derived_units
+            self.control_volume.work = Var(
```
andrewlee94, line 353: Should we be adding the `work` term to the control volume in this case, or just create `work_mechanical` as a `Var` at the unit level?

idaes/models/unit_models/pressure_changer.py
```diff
@@ -444,7 +459,7 @@ def add_adiabatic(self):
 
         @self.Constraint(self.flowsheet().time)
         def zero_work_equation(b, t):
-            return self.control_volume.work[t] == 0
+            return b.control_volume.work[t] == 0
```
andrewlee94, line 462: I would argue this is the reverse of what you should do - it would probably be better to have `self` as the argument in both cases.

Also, whilst not part of this PR, this option is misnamed. `adiabatic` means no heat transfer (which is true for all our options), not no work transfer.

idaes/models/unit_models/tests/test_pressure_changer.py
```diff
@@ -845,6 +844,38 @@ def test_config(self):
...
+            default={
+                "property_package": m.fs.properties,
+                "energy_balance_type": EnergyBalanceType.none,
+            }
+        )
+
+        assert m.fs.unit.config.energy_balance_type == EnergyBalanceType.none
+        assert hasattr(m.fs.unit.control_volume, "work")
+
+        m2 = ConcreteModel()
```
andrewlee94, line 865: This looks like it should be two separate tests.

idaes/models/unit_models/tests/test_pressure_changer.py
```diff
@@ -845,6 +844,38 @@ def test_config(self):
...
+
+        m.fs.unit = Pump(
+            default={
+                "property_package": m.fs.properties,
+                "energy_balance_type": EnergyBalanceType.none,
+            }
+        )
+
+        assert m.fs.unit.config.energy_balance_type == EnergyBalanceType.none
+        assert hasattr(m.fs.unit.control_volume, "work")
```
andrewlee94, line 863: You should probably also check the reference here too.

andrewlee94 (changes requested):
A few minor comments.

One thing not covered here - we should check that the documentation discussed what happens in these cases.

idaes/models/unit_models/pressure_changer.py
adam-a-a: I agree. All I did was update `self` to `b` on a whim and has no relevance to the PR, so I can just change it back.

idaes/models/unit_models/pressure_changer.py
adam-a-a: I started with that approach, but turned out to be a bit cleaner without cascading changes necessary.

adam-a-a: Left this one open for the time being, subject to other reviewers' thoughts.

eslickj: I'm fine with adding it to the control volume I think.  I suppose it is the expected location for it.

eslickj (approved):

MAZamarripa (approved):
lgtm

andrewlee94 (approved):

idaes/models/unit_models/pressure_changer.py
lbibl: @andrewlee94 ,  I looked over this discussion and understood either way above would do the job well.  But while I was reading a initialization API design proposal you wrote,  I would like to say that I think it would be better to only create `work_mechanical`,  as you suggested, by the principle that a var associated with one block should stay with this block, rather than being placed into another block like what is currently.  This way will also be documented more clearly, but I think the doc PR I just opened described the current version well.    Just a thought.  I would be happy to help change it if the IDAES team decides to, otherwise I look forward to concluding the doc PR I opened.  Thanks.

PR #926, opened 2022-08-05
Initial implementation of IDAES Performance Testing Suite

andrewlee94:
## Summary/Motivation:
This PR adds the first content for the IDAES performance testing suite. Work still needs to be done to determine how these will be run as part of the CI suite.

- Add `main.py` and `compare.py` files from Pyomo performance testing suite. These should be removed once Pyomo makes their code easily available.
- Refactor `HC_PR` tests to create general functions for the performance testing steps.
- Adds a general performance test runner/template for IDAES models
- Implements `HC_PR` as first performance test
- Add new `pytest` mark for performance tests and config methods to mark these to run only on demand.

## How to Run Performance Tests

From the `idaes-pse` root folder, run the following:

`python [pyomo]\scripts\performance\main.py --project idaes -v -d [output_dir] . --performance`

(the `.` is the path to the performance testing folder). 
A log file (json) will be created in `[output_dir]` containing the performance record.

### Legal Acknowledgement


idaes/models/properties/modular_properties/examples/tests/test_HC_PR.py
```diff
@@ -229,32 +269,7 @@ def test_build(self):
...
-        model.props[1].mole_frac_comp["ethylene"].fix(0.077)
-        model.props[1].mole_frac_comp["propene"].fix(0.077)
-        model.props[1].mole_frac_comp["butene"].fix(0.077)
-        model.props[1].mole_frac_comp["pentene"].fix(0.077)
-        model.props[1].mole_frac_comp["hexene"].fix(0.077)
-        model.props[1].mole_frac_comp["heptene"].fix(0.077)
-        model.props[1].mole_frac_comp["octene"].fix(0.076)
-
-        assert degrees_of_freedom(model.props[1]) == 0
+        build_model()
```
jsiirola, line 272: Shouldn't this be `return build_model()`?

jsiirola (changes requested):
Looks pretty good.  One change...

FYI: the PR in Pyomo to move the `main.py` and `compare.py` scripts into the main Pyomo repo is up (Pyomo/pyomo#2476) and should be merged once tests pass.

Robbybp:
I intend to try this out and give it a thorough review next week

jsiirola:
FYI: Pyomo/pyomo#2476 was merged and will be in the next Pyomo release (supporting the next IDAES release)

idaes/tests/performance/test_models.py
```diff
@@ -0,0 +1,74 @@
...
+    def _run_test(self, model_lib, init_model, solve_model):
+        gc.collect()
+        timer = TicTocTimer()
+
+        model = model_lib()
+        self.recordData("build model", timer.toc("build model"))
+
+        gc.collect()
+        timer.tic(None)
+        assert_units_consistent(model)
```
Robbybp, line 48: This call to `assert_units_consistent` seems to ensure that tests that build a dynamic model will fail unless the time discretization transformation is performed as part of `init_model`. Is this intentional? I would prefer time discretization to be either part of "build model" or its own category, but don't feel that strongly (for now) about it being lumped in with "initialize".

idaes/tests/performance/test_models.py
```diff
@@ -0,0 +1,74 @@
...
+        self.recordData("initialize", timer.toc("initialize"))
+
+        gc.collect()
+        timer.tic(None)
+        solve_model(model)
+        self.recordData("final solve", timer.toc("final solve"))
+
+
+@pytest.mark.performance
+class TestMisc(TestModel):
```
Robbybp, line 63: This should probably have a more descriptive name, e.g. `TestIdaesPerformanceSuite`

Robbybp (changes requested):
Looks pretty good so far. My main request is for someway to more easily construct different "test methods", e.g. a `_run_test`-like method for dynamic models that times the discretization transformation.

Also, when implementing a performance test, we have to (a) implement three suitable methods in a test file somewhere in IDAES and (b) import these methods and call them in a new test method in `TestMisc`. It would be nice if we could somehow just implement a test class (something like `IdaesPerformanceTest`), that we then mark with `@performance`, that we inherit from and implement the three methods that we need. Then the "test driver method" can just call these three methods. That way, to write a performance test, we only have to implement one class.

Robbybp:
Also, see my PR https://github.com/andrewlee94/idaes-pse/pull/16 into this branch for a couple of performance tests with the moving bed model.

idaes/tests/performance/test_models.py
andrewlee94: We had not thought about this issue, and ultimately this is not an IDAES issue but rather a Pyomo issue that needs to be fixed (and we need to keep prodding them to get it elevated in priority). However, also see more more general response below.

andrewlee94:
@Robbybp The `_run_test` method and associated class is not intended to be universal - it is meant to be a helper class for common tests based off of existing IDAES test examples (where the build, initialize and solve methods more-or-less already exist). For cases where this is not as convenient or appropriate we can write custom test methods that record the necessary data.

This also addresses the concerns about dynamic examples and unit consistency - these examples could skip the unit consistency tests for now (until Pyomo fixes that problem).

idaes/tests/performance/test_models.py
```diff
@@ -0,0 +1,74 @@
...
+        gc.collect()
+        timer.tic(None)
+        init_model(model)
+        self.recordData("initialize", timer.toc("initialize"))
+
+        gc.collect()
+        timer.tic(None)
+        solve_model(model)
+        self.recordData("final solve", timer.toc("final solve"))
+
```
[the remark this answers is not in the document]
andrewlee94: No, that is for the component/integration tests to check. Here we only want to record how long it took to build and solve the model.

idaes/tests/performance/test_models.py
```diff
@@ -0,0 +1,74 @@
...
+        tmp = getattr(self, "testdata", None)
+        if not tmp is None:
+            tmp[name] = value
+
+    def _run_test(self, model_lib, init_model, solve_model):
+        gc.collect()
+        timer = TicTocTimer()
+
+        model = model_lib()
+        self.recordData("build model", timer.toc("build model"))
```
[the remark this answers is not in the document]
andrewlee94: For the most part, I believe these are just labels and not much more - developers can use whatever they feel like here. Note that performance results are categorized by test class first and then label (rather than label then test class), so there is no requirement for consistency between test classes (beside convenience for interpreting results).

idaes/tests/performance/test_models.py
```diff
@@ -0,0 +1,74 @@
...
+
+        This method is only meaningful when running this TestCase with
+        'nose', using the TestData plugin.
+
+        """
+        tmp = getattr(self, "testdata", None)
+        if not tmp is None:
+            tmp[name] = value
+
+    def _run_test(self, model_lib, init_model, solve_model):
```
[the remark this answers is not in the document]
andrewlee94: This was copied from the Pyomo test suite and I stuck with the `model_lib` name to avoid confusing with any actual method names `build_model`.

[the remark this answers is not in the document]
jsiirola: Yeah - and I  am not a huge fan of `model_lib` (that was mostly to cut down on diffs when Pyomo moved to the new performance test harness).  I think `create_model` or `generate_model` might be better?

Robbybp:
@andrewlee94 So for a different type of test, we should implement a new class, perhaps `TestDynamicModel`, with its own `_run_test` method?

Robbybp:
I'm also thinking about an application where, for all models in the IDAES performance suite, we collect some statistics, then initialize and construct a `PyomoNLP` and collect some more statistics. For an application like this, it would be necessary to have a standard way to access the `build_model`, `initialize_model`, and `solve_model` functions. I think an `IdaesPerformanceTest` base class that performance test cases must inherit from would be one way to do this. Does this idea of a common API for "performance test cases" seem like a good idea that would be worth supporting?

andrewlee94:
@Robbybp The first thing to note is that Pyomo is actually collecting a lot of information in the background that is not immediately apparent. E.g., we get construction times for every Pyomo component in the model and every transformation performed automatically. So, chances as most of the statistics you are thinking about are already being captured.

The things called out in the `_run_test` method are just there to give us some high-level summaries.

andrewlee94:
@Robbybp r.e. different types of tests; for now I would hope that we can make a general testing class that covers most cases (once the unit consistency issue is resolved, and I have a work around for that that I will do soon). However, any test should feel free to write its own test class if it has things it wants to log separately (e.g. a complex flowsheet might want to  log build and initialization times by sub-section). TLDR: the test class here is intended to be a helper class for cases where it is applicable, not to enforce an arbitrary standard form (I'll add comments to this extent).

Robbybp:
I am thinking more about statistics related to the algebraic/NLP models, rather than timing statistics. For instance, tracking the frequency and types of nonlinearities, size of expressions, condition numbers of various matrices, etc. This doesn't really have anything to do with performance in the sense we're discussing here, but I think it would be nice to use this collection of IDAES models as a set of fairly large, fairly practical algebraic/NLP models to help motivate research into preprocessing/solution algorithms for these types of problems. And I think it would be fairly easy/nonintrusive to require a common API to build, initialize, and solve our performance test models, at least those that use this standard workflow.

andrewlee94:
@Robbybp Ah - that was part of the intent behind the three methods (build, initialize and solve) - so that we could reuse the same code between multiple tests. I would not be restrictive however, but encourage it wherever possible - there will inevitably be cases whee it doesn't work.

I am going to rework things to have the single common class for each test case which has standard names for the three sub-methods, so I think that will address a large part of what you are suggesting.

andrewlee94:
@Robbybp I've added some changes so that you only need to provide a single test class for each case which is expected to have the methods to build, initialize and solve the model. I also added an option to flag to allow for tests which we know fail unit consistency - we still do the assertion, but we check that it fails in these cases (so we know to go back and fix this later).

idaes/models/properties/modular_properties/examples/tests/test_HC_PR.py
```diff
@@ -58,6 +58,47 @@ def _as_quantity(x):
 
 # Test for configuration dictionaries with parameters from Properties of Gases
 # and liquids 4th edition
+class HC_PR_Model:
```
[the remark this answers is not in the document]
andrewlee94: They could (and I considered it), but (for now at least) I wondered about how much value it would add.

Robbybp:
I just played around with importing and operating on all IDAES performance suite models. Long comment incoming.
Recall that the purpose of this is to be able to answer questions like:
- What is the average number of variables/constraints in an IDAES performance suite model?
- What percentage of constraints in each model are nonlinear?
- What is the condition number after initialization of each constraint Jacobian?
- etc.

If these models all inherit from some base class like `IdaesPerformanceTest` that I suggest above, this amounts to importing everything from every module in IDAES and checking if it is an instance of `IdaesPerformanceTest`. This seems slow and cumbersome and I don't really know how to do it. If there is a good way to do this, I would be interested to learn.

An alternative I have come up with is to have the performance test base class inherit from `unittest.TestCase`:
```python
import pyomo.common.unittest as unittest

class IdaesPerformanceTest(unittest.TestCase):
    def build_model(self):
        raise NotImplementedError()
    def initialize_model(self, model):
        raise NotImplementedError()
    def solve_model(self, model):
        raise NotImplementedError()
    def test_dummy(self):
        pass
```
Then we can iterate over performance tests with a small script using `TestLoader.discover` from `unittest`:
```python
import os
from idaes.tests.performance.idaes_performance_base import IdaesPerformanceTest
import pyomo.common.unittest as unittest
import idaes


def generate_test_cases(test_suite, discovered=None):
    if discovered is None:
        discovered = set()
    for test in test_suite:
        if isinstance(test, unittest.TestCase):
            if test.__class__ not in discovered:
                # Sometimes we find the same test class twice for some reason...
                discovered.add(test.__class__)
                yield test
        elif isinstance(test, unittest.TestSuite):
            yield from generate_test_cases(test, discovered=discovered)


def flatten_test_suite(test_suite):
    return list(generate_test_cases(test_suite))


def main():
    loader = unittest.defaultTestLoader
    idaes_dir = os.path.dirname(idaes.__path__[0])
    test_suite = loader.discover(idaes_dir)
    test_cases = flatten_test_suite(test_suite)
    for test in test_cases:
        if isinstance(test, IdaesPerformanceTest):
            print(test)


if __name__ == "__main__":
    main()
```
which displays:
```console
<a bunch of deprecation warnings>
test_dummy (idaes.models.properties.modular_properties.examples.tests.test_HC_PR.HC_PR_Model)
test_dummy (idaes.tests.performance.idaes_performance_base.IdaesPerformanceTest)
test_dummy (idaes.models.unit_models.tests.test_heat_exchanger_1D.HX1D_Model)
```
Note that the `test_dummy` method is required, otherwise these test cases will have no tests, and `discover` will just pick up an empty `TestSuite` rather than the `TestCase`s we're looking for.

We can get around this with a slight change to the test infrastructure:
- `TestModel` now serves as a base class and implements a test that calls `_run_test`
- Specific test classes inherit from both `TestModel` and `TestCase`

In `test_models.py`:
```python
class TestModel(object):
    def recordData(self, name, value):
        ...
    def _run_test(self, consistent_units=True):
        ...
    def build_model(self):
        raise NotImplementedError()
    def initialize_model(self, model):
        raise NotImplementedError()
    def solve_model(self, model):
        raise NotImplementedError()
    def test_build_initialize_solve(self):
        self._run_test()
```
In the modules for the specific tests:
```python
import pytest
import pyomo.common.unittest as unittest
from idaes.tests.performance.test_models import TestModel

@pytest.mark.performance
class HX1D_Model(TestModel, unittest.TestCase):
    ...
```
```python
import pytest
import pyomo.common.unittest as unittest
from idaes.tests.performance.test_models import TestModel

@pytest.mark.performance
class HC_PR_Model(TestModel, unittest.TestCase):
    ...
```
Then the same script as above (with `TestModel` instead of `IdaesPerformanceTest`) can iterate over TestCases and identify those that are also instances of TestModel, with output:
```console
test_build_initialize_solve (idaes.models.properties.modular_properties.examples.tests.test_HC_PR.HC_PR_Model)
test_build_initialize_solve (idaes.models.unit_models.tests.test_heat_exchanger_1D.HX1D_Model)
```
To run the tests from this `idaes/tests/performance/` directory, we now need to import them in to a test file, e.g. `test_performance_suite.py`:
```python
from idaes.models.properties.modular_properties.examples.tests.test_HC_PR import (
    HC_PR_Model,
)
from idaes.models.unit_models.tests.test_heat_exchanger_1D import HX1D_Model
```
and the test suite can be run from `main.py` as before.
I have implemented this in my `perf_testing-discoverable` branch, for what it's worth.

andrewlee94:
@Robbybp So, a few comments in response (there is a lot there so I will probably miss something.

1. First, I think we are talking about different things here. The purpose of these classes is not  to be the test themselves but to hold methods that are common in other tests. The tests can then leverage these methods rather than duplicating code. Note that in the examples I have so far the `build_model` methods, etc. are use in the unit tests for each model as well as the performance tests.
2. The actual test methods are then written elsewhere and have `pytest` marks that can be used for test discovery. `pytest` has ways of marking tests to be skipped by default, so we can set up tests that will only be run if called explicitly (which I have done for the performance tests). So my thought is that you would want to write a separate set of tests to count constraints, etc. which leverage these common methods.
3. Finally, have we considered what we would use these tests your are proposing for in the overall CI environment? These are definitely things of interest, but I am wondering what we would use these for.

Robbybp:
@andrewlee94 My goal is to be able to iterate over all Performance Test Suite models in a Python script and use a common API to build, initialize, and solve. Ideally, I would be able to do this automatically, without needing to know what the test suite models are or where they are located. This is accomplished with the script I show above.

Responding to your points:
1. In my proposal, a performance test model doubles as a test and utility class whose methods can still be used in other tests. The motivation for making these classes tests is so they are easily discoverable (for iteration in a Python script) with `unittest.TestLoader.discover`
2. Rather than discovering just the test methods, I would like to discover the underlying Performance Test Models that implement build and initialize methods. The tests are still decorated with `@performance` and can be discovered with pytest
3. I do not plan to iterate over Performance Test Suite models in any tests or CI environment. I would like to do so to generate statistics that justify the use of these models to benchmark optimization solvers, as I think "models like IDAES models" are under-represented as benchmark problems. I would like to collect statistics that help me define more precisely what I mean by "models like IDAES models"

andrewlee94:
@Robbybp OK, I think I see the source of the problem here - the real issue is that we do not have a formal, easily discoverable definition of what constitutes the performance test suite. At the moment, the list of models is defined implicitly as part of the performance test runner which makes it hard to repurpose for other uses.

I think the immediate solution is fairly straight forward, and actually makes your life easier too - we need to create a separate module which defines the models that make up the test suite, and then have the test runners work from that (where it is the timing tools in this PR or your tools for counting constraints, etc.). The only catch I see with this is that it does start to push us towards a standard form for all these to take (i.e. the performance test base class) which I had been hoping to avoid (to allow for more custom performance tests if required), but we can deal with those when they arise.

My first thoughts are thus:

1. Define a standard form for performance tests via the base class.
2. Have a "performance test suite module" that imports these classes from their various locations and puts them in a dict for easy discovery. This will give us a neat, unambiguous definition of what constitutes the performance test suite.
3. The test runner can then use the above dict to parameterize the performance tests, which saves us having to write test methods for each model in the test suite as a bonus.

This way, we don't have the do things like import all of ideas and search for instances of specific classes.

However, as I said this does force us into a standard form for the performance test models, which might limit us in future. Notably, I wanted to leave things open for us to write more granular performance logging in cases where it is appropriate (e.g. flowsheets). However, there is probably a way to make this work - I just need to think a bit longer.

andrewlee94:
@Robbybp I think I have the solution: the `PerformanceTestClass` should be merged with the existing `TestModel` class and has only two methods that it must define: `build_model` and `run_test`. This way the performance test suite can just grab the `run_test` method, which can be customized for a specific model or left as is, and the `build_model` should return an instance of the model to be tested (for use in your tools or others).

The `run_model` method may or may not make use of the `build_model` method as required by the test case (i.e. simple cases will just call the `build_model`, `initialize_model` and `solve_model` methods, but more complex cases might have fully customized `run_model` methods).

Can you see any holes in this idea?

Robbybp:
@andrewlee94 My main question is about how you envision the "performance test suite" to look. A test class that iterates over a dict, creating a new `test_*` method for every item in the dict? And these test methods just call `run_test`?

My other comments are:
- I would like the base class for the standard performance test to have methods for initialization and solve as well.
- Different types of models can have different classes for their performance tests. E.g. `DynamicPerformanceTest` or `FlowsheetPerformanceTest`
- I would actually prefer to not have some global dict that defines the "IDAES Performance Suite". My main reason is so that, to implement a performance test, you only have to touch one file (the test file where you subclass `PerformanceTest` and `TestCase` for a specific model). Having a global data structure to define the performance tests strikes me as "storing the same piece of data in two locations"

andrewlee94:
@Robbybp I think I have something working that doesn't require the dict of test classes and just uses a common base class and `pytest` discovery. This does mean however that you need to do some sort of search to find the performance tests.

I envision most performance test classes to have `initialize_mode` and `solve_model` methods (I actually have a default `solve_model` method in the base class), but I was looking at them as only being required if you used the default `run_test` method (i.e. if you wrote your own then you might not need the initialize and solve methods).

Finally, there is nothing to stop you from having multiple performance tests for a given model - just create as many performance test calsses as you want for it (or if you want to be fancy, you can embed multiple test runs in a single class too).

Robbybp:
Have you looked at what I implemented in https://github.com/andrewlee94/idaes-pse/pull/17 ? It is a working implementation of what I've described here. I'm interested to see your implementation as well.

andrewlee94:
@Robbybp What you have is very similar to what I have implemented, the main differences being that I am working on the issue you had with being able to run tests in isolation.

andrewlee94:
@Robbybp Here is my latest version. you should be able to use the following from the `idaes-pse` folder to run the performance tests alone:

```
python idaes/tests/performance/main.py -v -d [output_dir] . --performance -m performance
```
I still need to remove the need for the copy of `main.py` (as it is now in the base Pyomo package), but I am having some issues with the pytest arguments and am waiting for some help from @jsiirola .

Robbybp:
@andrewlee94 Out of curiosity, why are `-m performance` and `--performance` both required?

andrewlee94:
@Robbybp They might not be at this point - I am still working on those details. I suspect (but have not tried) the `-m performance` might be sufficient.

idaes/models/unit_models/tests/test_heat_exchanger_1D.py
```diff
@@ -703,44 +703,57 @@ def test_conservation(self, btx):
...
+    m.fs.unit.tube_inlet.flow_mol[0].fix(5)
+    m.fs.unit.tube_inlet.enth_mol[0].fix(7000)
+    m.fs.unit.tube_inlet.pressure[0].fix(101325)
 
-        return m
+    return m
+
+
+@pytest.mark.performance
+class Test_HX1D_Performance(PerformanceBaseClass):
```
Robbybp, line 743: Can these test classes also inherit from `unittest.TestCase`, e.g.
```python
class Test_HX1D_Performance(PerformanceBaseClass, unittest.TestCase):
    ...
```
This way they are discoverable with `unittest.TestLoader.discover`. The reason to put the inheritance on the classes for specific models is so that the "discovery script" I posted above generates only these classes, rather than `PerformanceBaseClass` as well. This is inelegant and could be avoided if we could perform this discovery without relying on unittest. Presumably this is possible with pytest, probably via hooks/plugins (`pytest_collection`?) or the `pytest.Collector` class, but I wasn't able to get it working in a couple hours of trying.

idaes/tests/performance/performance_test_base.py
```diff
@@ -0,0 +1,149 @@
...
+from pyomo.environ import assert_optimal_termination
+
+from idaes.core.solvers import get_solver
+from idaes.core.util.model_statistics import degrees_of_freedom
+
+# Get default solver for testing
+solver = get_solver()
+
+
+class PerformanceBaseClass:  # unittest.TestCase):
```
Robbybp, line 30: Inheriting from TestCase here causes this base class to show up when using `unittest.TestLoader.discover`, so I think it should not happen (and the comment can be removed). Presumably we could do discovery without unittest, but I have not figured it out.

idaes/tests/performance/performance_test_base.py
```diff
@@ -0,0 +1,149 @@
...
+    * solve_model - (optional) a method to solve the model of interest. This method is called as part
+        of the standard test_performance method, and a standard method is provided by the base class.
+
+    Developers may add additional test methods to the derived class to run additional performance tests if desired.
+    These methods should ass start with 'test_' in order ot be discovered by pytest.
+
+    Usage:
+
+    Performance tests should inherit from this based class and decorate the derived class with the 'performance'
+    pytest mark. This will automatically categorize all tests in this class as performance tests which will only be run
```
Robbybp, line 52: Will these line lengths be shortened when `black` is run?

Robbybp (commented):
Assuming we either inherit from `unittest.TestCase` or figure out pytest test discovery, I am fine with this. Given the nuanced discussion we have had so far, this Performance Test Suite could use some documentation that developers can follow to add their favorite models to the suite. I am happy to start this, which I will provide as a PR into this branch.

idaes/models/unit_models/tests/test_heat_exchanger_1D.py
andrewlee94: Yes, that should be possible. I had intended to do this after the change in the comment below.

idaes/tests/performance/performance_test_base.py
andrewlee94: Yes - this is why this was commented out. I just hadn't cleaned it up yet.

idaes/tests/performance/performance_test_base.py
andrewlee94: Probably not - they are either within the limit we are setting for `black` or are ignored (my pre-commit hook would have already fixed this).

andrewlee94:
@Robbybp I've addressed all your comments, although I am still looking to see if there is a way to get the base class to inherit from `unittest.TestCase` but not get run by `pytest` (so that we don't need to remember to have all the derived classes do multiple inheritance).

Robbybp:
@andrewlee94 I believe inheriting from `TestCase` doesn't change what `pytest` will run, but it does change what `unittest` will run. The only reason this is an issue is that I can't figure out test discovery with pytest.

Robbybp:
Another comment is that the HC_PR test takes about 5 minutes to run on my laptop. Is this too long?

andrewlee94:
@Robbybp 5 minutes is the expected time for the HC_PR test, which is why it was included. It involved a lot of deep expressions and unit coversions so hopefully represents one of our best cases for performance improvement.

andrewlee94:
@Robbybp @jsiirola I have implemented the `__init_subclass__` method to force inheritance from `TestCase` and hopefully fixed the last lingering test failures. This should be ready for review.

andrewlee94:
Also, the latest (final) way to run the performance tests is (updated in original comment too):

`python [pyomo]\scripts\performance\main.py --project idaes -v -d [output_dir] . --performance`

idaes/core/util/performance.py
```diff
@@ -0,0 +1,169 @@
...
+    performance tests if desired. These methods should use standard pytest marks and
+    naming conventions.
+
+    Usage:
+
+    Performance tests should inherit from this based class and unittest.TestCase.
+    Inheritance from TestCase is used to make the derived classes easily discoverable
+    for use outside the performance testing infrastructure. The derived class should
+    be decorated with the 'performance' pytest mark. This will automatically categorize
+    all tests in this class as performance tests which will only be run when explicitly
```
jsiirola, line 59: "which will" -> "that will"

andrewlee94:
@Robbybp Would you have time to review this?

idaes/core/util/performance.py
```diff
@@ -0,0 +1,169 @@
...
+    called for.
+
+    Raises:
+        TypeError is derived class does not inherit from unittest.TestCase.
+    """
+
+    def __init_subclass__(cls):
+        if unittest.TestCase not in cls.mro():
```
jsiirola, line 67: Note that this prevents creating a helper class that inherits from this, and from which multiple test suites then inherit.  That said, I don't thin it's a big deal, as the helper class can be implemented as a mixin class.

idaes/core/util/performance.py
jsiirola, line 59: "which will" -> "that will"

idaes/core/util/performance.py
```diff
@@ -0,0 +1,169 @@
...
+                "Classes derived from PerformanceBaseClass must also inherit from "
+                "unittest.TestCase."
+            )
+
+    def recordData(self, name, value):
+        """
+        A method for recording data associated with a test.
+        """
+        tmp = getattr(self, "testdata", None)
+        if not tmp is None:
```
jsiirola, line 78: PEP8: `if tmp is not None`

jsiirola (approved):
Some minor comments, but I think this looks pretty good.

Robbybp:
I'll make sure this works for my use case and review shortly

idaes/core/util/performance.py
Robbybp: My current plan is to implement a `DynamicPerformanceTest` subclass in order to override `test_performance` to (a) not check unit consistency and (b) potentially record the discretization transformation time as well. I would like to inherit from the base class so that I can still identify dynamic performance tests with `isinstance`. I think this restriction is enough that I would prefer this check to go in `__init__`.

idaes/core/util/performance.py
```diff
@@ -0,0 +1,169 @@
...
+    be decorated with the 'performance' pytest mark. This will automatically categorize
+    all tests in this class as performance tests that will only be run when explicitly
+    called for.
+
+    Raises:
+        TypeError is derived class does not inherit from unittest.TestCase.
+    """
+
+    def __init_subclass__(cls):
+        if unittest.TestCase not in cls.mro():
```
Robbybp, line 67: Minor nitpick: I think `if not issubclass(cls, unittest.TestCase):` is cleaner than directly checking the MRO. Let me know if there is any technical reason to check the MRO instead.

idaes/core/util/performance.py
andrewlee94: A couple of comments:
1. For dynamic models, hopefully Pyomo can get a fix in for the underlying problem, which would remove the need for this distinction.
2. I believe the transformation time should already be captured without the need for us to record it. Note that the Pyomo performance testing infrastructure logs a lot of information behind the scenes.

Robbybp (commented):
See my comment above about `__init_subclass_` vs `__init__`, otherwise, this looks good.

idaes/core/util/performance.py
jsiirola: @Robbybp, why would a mixin class not work?  You don't need `DynamicPerformanceTest` to inherit from `PerformanceBaseClass`, do you?  It should be sufficient for your dynamic tests to:
```python
class TestMyThing(DynamicPerformanceTest, PerformanceBaseClass, unittest.TestCase):
```

idaes/core/util/performance.py
andrewlee94: The reason for this is that when @jsiirola suggested this approach he mentioned checking the MRO, so I just went with that rather than looking at alternative approaches.

idaes/core/util/performance.py
Robbybp: Regardless, it would be useful to have an explicit distinction for dynamic models (and flowsheet models, and maybe other types of models as well) as we may want to benchmark different initialization or solution strategies that are only valid for these types of models.

idaes/core/util/performance.py
Robbybp: This doesn't need to block the PR, but I like `issubclass` because I think most Python users know what `issubclass` does, but (probably?) don't know what `mro()` does.

idaes/core/util/performance.py
andrewlee94: I am not sure if this is the best way to categorise things like that - I would suggest a set of special pytest marks would be better in this case (at least from the timing perspective at least). This would make it easy to run different subsets of the performance tests.

Secondly, I would also suggest that in at least some of these cases there should be multiple performance tests in a single test class (e.g. steady-state case, dynamic case, initialization method 1, initialization method 2, etc.).

Robbybp: @jsiirola No reason that doesn't work, just didn't think of it and didn't know what a mixin class was. That is a little more complicated, but probably worth it for the benefit of failing at import time rather than runtime.

andrewlee94: I was guessing that was the intent - I just wonder if we should have some other way to distinguish things like that that we could use for multiple purposes.

jsiirola: That sounds like Robby doesn't want different performance metrics, but rather he wants to collect diagnostics on performance test models?  These things (#vars, condition #, etc) are not expected to change over time, right?  It feels lie we are confounding performance tests with other test suite diagnostics...

idaes/core/util/performance.py
jsiirola: I agree: `issubclass` is a cleaner implementation.  I believe the class is sufficiently constructed at this point that it should work.

idaes/core/util/performance.py
andrewlee94: Yes, to an extent we are. I think we have 3 things here:

1. A standard suite of models we use for this type of testing.
2. A set of performance tests based on the model suite.
3. A set of diagnostics (#vars, etc.) on these models.

When we started 1 and 2 were intertwined, and we are trying to disentangle them now.

Robbybp: Those assessments sound accurate. Maybe we need a list of "test models" (with an API for build/initialize/solve) somewhere, then, separately, a way to run a standard `test_performance` on each of them.

Robbybp:
This works for now, but per John's comment above maybe we should discuss a better way to separate the definition of the test models from the performance testing of them.

andrewlee94:
@Robbybp That was what @jsiirola and I decide on. We wanted to get the basics in so we could start testing (and mark off the upcoming deliverable), and we thought we needed a longer (separate) discussion about what the API we wanted looked like going forward.

PR #931, opened 2022-08-10
Adding multi select box for the ui

elbashandy:
## Summary/Motivation:
Adding the feature of selecting multi UnitModels in the UI and move them all together.

1. Selecting unit models as a group:
Screen Shot 2022-08-18 at 2 30 18 PM

2. Moving the group together by holding left click on one of the components:
Screen Shot 2022-08-18 at 2 28 58 PM

-
-

### Legal Acknowledgement


dangunter (approved):
It works great! Still might be a little better if default pointer on the page were an arrow instead of a hand, but the move-plus icon is consistent whether you're doing groups or single units, so it's good enough for me.

ksbeattie (approved):
Looks really good!  I've found some odd behavior when selecting units with the option key selected then moving the mouse a little bit. I'll leave that for another issue/PR though.

PR #933, opened 2022-08-14
Fix multistage turbine test

eslickj:
## Fixes

This fixes up the multistage steam turbine test.  It's required for new IAPWS binaries to pass, but it's also an improvement in general.  The main change is including the reheater temperature constraint before initializing the reheater.  This keeps the turbine from seeing a weird temperature.  It also cleans up a little out dated code.

### Legal Acknowledgement

### DIFF
--- a/idaes/models_extra/power_generation/unit_models/helm/tests/test_turbine_multistage.py
+++ b/idaes/models_extra/power_generation/unit_models/helm/tests/test_turbine_multistage.py
@@ -32,7 +32,7 @@ import idaes.logger as idaeslog
 from idaes.core.solvers import get_solver
 
 # Set up solver
-solver = get_solver(options={"max_iter":20})
+solver = get_solver(options={"max_iter": 20})
 
 
 @pytest.mark.unit
@@ -126,7 +126,7 @@ def test_initialize():
 
     pyo.TransformationFactory("network.expand_arcs").apply_to(m)
     iscale.calculate_scaling_factors(m)
-    turb.initialize(outlvl=idaeslog.ERROR, optarg={"max_iter":20})
+    turb.initialize(outlvl=idaeslog.ERROR, optarg={"max_iter": 20})
     turb.ip_stages[1].inlet.unfix()
 
     for t in m.fs.time:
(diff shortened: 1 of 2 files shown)


idaes/models_extra/power_generation/unit_models/helm/tests/test_turbine_multistage.py
```diff
@@ -165,11 +176,11 @@ def test_initialize_calc_cf():
 
     # Set the inlet of the ip section, which is disconnected
     # here to insert reheater
-    p = 7.802e06
+    p = 1.4e06
     hin = pyo.value(iapws95.htpx(T=880 * pyo.units.K, P=p * pyo.units.Pa))
     m.fs.turb.ip_stages[1].inlet.enth_mol[0].value = hin
     # m.fs.turb.ip_stages[1].inlet.flow_mol[0].value = 25220.0
-    m.fs.turb.ip_stages[1].inlet.pressure[0].value = p
+    # m.fs.turb.ip_stages[1].inlet.pressure[0].value = p
```
andrewlee94, line 183: Do we need these commented lines? If so, a reason might help for future reference.

andrewlee94 (approved):
Looks OK to me - just one comment on some commented lines but nothing that needs to hold this up.

idaes/models_extra/power_generation/unit_models/helm/tests/test_turbine_multistage.py
bpaul4: Why did the `p` value change?

eslickj: The original was just an initial guess for initialization.  I set the inlet stage pressure ratio to 0.6, and in this case use that to calculate the inlet stage flow coefficient.  So the original pressure was really far off.  I think I did the original test in a hurry and despite the bad initialization setup, it passed and I didn't give it any more thought.

dallan-keylogic (approved):
LGTM

idaes/models_extra/power_generation/unit_models/helm/tests/test_turbine_multistage.py
bpaul4: Okay, thanks for clarifying. I trust your judgement here.

bpaul4 (approved):

PR #936, opened 2022-08-18
Bump bin version to 3.0.0

eslickj:
## Fixes

Update the binaries version.  This doesn't change much.  It adds the new functions for the general Helmholtz EoS, but they are not used yet in the current main branch.  The only thing that may affect anything is that there is a bug fix for a missing derivative in the old IAPWS95 and Span-Wagner functions.  The new binaries also include headers and such to build against the IPOPT shared libraries.  I'll work out new install methods for that later.  For now you'll find a tar.gz file in the bin directory.  The plan in the future is to put it in .idaes/local and extract it, so you get ./daes/local/include, .idaes/local/lib, ...

### Legal Acknowledgement


andrewlee94 (approved):
LGTM.

bpaul4 (approved):

PR #937, opened 2022-08-23
Remove redirections to modules moved in v2.0

andrewlee94:
## Summary/Motivation:
This PR removes the backwards compatibility redirections for the `gas_solids_contactors`, `generic_models`, `power_generation` and `surrogates` folders and associated code that has been moved in the new repo structure.

This PR also fixes some imports pointing to the old locations that were missed previously.

-
-

### Legal Acknowledgement


bpaul4 (commented):
This looks good, thanks for your time and efforts here. It looks to me like the doc error comes from `docs/explanations/modeling_extensions/surrogate/plotting/index.rst`, there are a few places where `automethod` references a deprecated import path.

eslickj (approved):
Looks good to me although I did not check each file.  I think the examples tests are a subset of the full examples tests.  Should we also make sure all the examples work before merging this?

andrewlee94:
@eslickj Yes, we need to test the examples before this is merged (hence the draft PR and WIP title). Plus, we won;t be merging this until the August release is finished.

idaes/models/properties/modular_properties/eos/tests/test_ceos_CO2_H2O.py
```diff
@@ -1,144 +0,0 @@
-#################################################################################
```
bpaul4, line 1: Why is this file being removed?

andrewlee94: It is being removed as it does not belong here - this is a test depends on a file in `models_extra` and core tests should not depend on non-core models. I think I moved these tests somewhere else.

bpaul4 (approved):

PR #938, opened 2022-08-23
Only aggregating flows with at least one registered flow

bknueven:
## Summary/Motivation:
Defining additional unused flows in a costing package can cause additional variables and constraints to be passed to Ipopt for flowsheets not utilizing those flows. This can sometime cause "spooky action at a distance" on sensitive flowsheets. See watertap-org/watertap#684 for more detail.

- ~~Fix unneeded `aggregate_flow_costs[ftype]` variables on the flowsheet costing block to `0` instead of constraining them to be `0`, eliminating (a) unneeded variable(s)/constraint(s) from the model.~~
- Define a new set, `used_flows` of flows which have been registered with the costing package. This allows for the elimination of unneeded variable(s) and constraint(s) as well as two `try/except` blocks in the costing package.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/base/costing_base.py
+++ b/idaes/core/base/costing_base.py
@@ -145,6 +145,7 @@ class FlowsheetCostingBlockData(ProcessBlockData):
         # Set up attributes for registering units and flows
         self._registered_unit_costing = []
         self.flow_types = pyo.Set()
+        self.used_flows = pyo.Set()
         self._registered_flows = {}
         self.defined_flows = {}
 
@@ -307,18 +308,15 @@ class FlowsheetCostingBlockData(ProcessBlockData):
             calculate_variable_from_constraint(var, cons)
 
         # Initialize aggregate flows and costs
-        for f in self.flow_types:
-            try:
-                fvar = getattr(self, f"aggregate_flow_{f}")
-                fconst = getattr(self, f"aggregate_flow_{f}_constraint")
-                calculate_variable_from_constraint(fvar, fconst)
+        for f in self.used_flows:
+            fvar = getattr(self, f"aggregate_flow_{f}")
+            fconst = getattr(self, f"aggregate_flow_{f}_constraint")
+            calculate_variable_from_constraint(fvar, fconst)
 
-                calculate_variable_from_constraint(
-                    self.aggregate_flow_costs[f],
-                    self.aggregate_flow_costs_constraint[f],
-                )
-            except AttributeError:
-                self.aggregate_flow_costs[f].set_value(0)
+            calculate_variable_from_constraint(
+                self.aggregate_flow_costs[f],
+                self.aggregate_flow_costs_constraint[f],
+            )
 
         # Call costing package initialization
         try:
@@ -418,6 +416,7 @@ class FlowsheetCostingBlockData(ProcessBlockData):
         for f in self.flow_types:
             # We will use the first costed flow as representative of the whole
             if len(self._registered_flows[f]) > 0:
+                self.used_flows.add(f)
                 f1 = self._registered_flows[f][0]
                 funits = pyo.units.get_units(f1)
                 agg_var = pyo.Var(units=funits, doc=f"Aggregate flow for {f}")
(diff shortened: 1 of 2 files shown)


TimBartholomew (approved):
LGTM

idaes/core/base/costing_base.py
```diff
@@ -448,7 +448,8 @@ def aggregate_flow_costs_constraint(blk, ftype):
                     pyo.units.convert(agg_var * cost_var, to_units=c_units / t_units)
                 )
             except AttributeError:
-                return blk.aggregate_flow_costs[ftype] == 0 * c_units / t_units
+                blk.aggregate_flow_costs[ftype].fix(0 * c_units / t_units)
```
andrewlee94, line 451: I am not sure that this is the best way to handle whatever issue you are facing. First, I am always hesitant to fix a variable a part of the code and secondly this results in different behavior between cases where the variable exists and when it does not.

This looks to me like we should either make sure `self.flow_types` only contains defined flows, or make `aggregate_flow_costs` an `Expression` instead.

andrewlee94 (changes requested):
I am not sure this is the best way to go about this. However, it would help to know exactly what the issues you have been seeing are - I tried following the link in the discussion, but that just lead to another PR which in turn linked further onwards.

idaes/core/base/costing_base.py
bknueven: The issue in WaterTAP is we were defining flows using the `defined_flow` attribute. When we add new defined flows, (e.g., as in watertap-org/watertap#684), the costing package will add a new variable and constraint for that aggregated flow. This affects our preexisting flowsheets (which do not utilize the new flow type), as this new variable and constraint now gets carried through to the solver.

I did think through a few other solutions, not all of which require changes in IDAES.
1. Be more careful about the indexing of `blk.aggregate_flow_costs` in this method.
2. Only define flows that are needed for a particular flowsheet within WaterTAP (watertap-org/watertap#686)
3. Fixing WaterTAP sensitive regression tests.

I am happy to go with (2) for now, as it reflects changes we want to make in WaterTAP anyways. But for others developing costing packages the current behavior might eventually cause them the same trouble we ran into.

andrewlee94: @bknueven So what we are seeing then is a conflict between wanting to have a single, central package that does everything versus case-specific implementations, and how to find a balance between these. This was not something that was considered in great depth when designing the costing framework - it was more-or-less based on the case-specific implementation, and assuming any unused terms would not have a great effect.

First, I assume you are seeing actual solver failures that can be attributed to this (i.e. it is not just an issue of seeing more variables and constraints than are strictly necessary)?

From what I can see, the best solution would be to build only those terms that are necessary for a given flowsheet - i.e. separate the `defined_flows` from the `used_flows`, and only index variables/constraints/expressions by `used_flows`. Of course, these requires determining which flows are used first, however that can be achieved by iterating through the list of flows. I.e. there is an overhead, but not much.

bknueven: > First, I assume you are seeing actual solver failures that can be attributed to this (i.e. it is not just an issue of seeing more variables and constraints than are strictly necessary)?
Yes, we do see either solver failures or changes in solution.

> From what I can see, the best solution would be to build only those terms that are necessary for a given flowsheet - i.e. separate the `defined_flows` from the `used_flows`, and only index variables/constraints/expressions by `used_flows`. Of course, these requires determining which flows are used first, however that can be achieved by iterating through the list of flows. I.e. there is an overhead, but not much.

There's probably near zero overhead, because we already do this a few lines above. We just need to build the `used_flows` set in this for loop: https://github.com/IDAES/idaes-pse/blob/3de93b17d99bdf7f6b3ca0afa0f7fa8824a94946/idaes/core/base/costing_base.py#L418-L435

bknueven: This change would also simply, in addition to the constraint in question, the initialization as well: https://github.com/IDAES/idaes-pse/blob/3de93b17d99bdf7f6b3ca0afa0f7fa8824a94946/idaes/core/base/costing_base.py#L310-L321

ksbeattie:
@bknueven, I've put this on the Nov release board with normal priority.  Let me know if that's not what you want.

andrewlee94 (approved):

PR #944, opened 2022-09-07
[WIP] Add PBR (packed bed reactor) model to IDAES

kanishka-ghosh:
## Summary/Motivation:
This PR will add a packed bed reactor (PBR) unit model to IDAES.  The PBR model is based on the PFR model that already exists in IDAES.  The PBR model can be used to model packed-bed catalytic reactions.

[WIP] Tests for the PBR model

Possible unit inconsistency: Units for `heat_duty` obtained using the PBR are W/m whereas in [this PFR example](https://github.com/IDAES/examples-pse/blob/main/src/Examples/UnitModels/Reactors/plug_flow_reactor_testing.ipynb), the units of `heat_duty` are assumed to be W for the PFR (PBR model is based on the PBR)

- Add PBR unit model
- Add tests for PBR unit model

### Legal Acknowledgement


idaes/models/unit_models/packed_bed_reactor.py
```diff
@@ -0,0 +1,337 @@
+##############################################################################
```
andrewlee94, line 1: As a first observation, this might be better placed in `models_extra` for now. The requirements for getting code merged there are significantly lower than for the main `models` folder. This might be suitable for merging in `models`, but would documentation and thorough testing.

idaes/models/unit_models/packed_bed_reactor.py
```diff
@@ -0,0 +1,337 @@
...
+                        UnitModelBlockData,
+                        useDefault)
+from idaes.core.util.config import (is_physical_parameter_block,
+                                    is_reaction_parameter_block)
+
+from idaes.core.util.misc import add_object_reference
+import idaes.core.util.unit_costing as costing
+from idaes.core.util.constants import Constants as const
+
+__author__ = "Andrew Lee, John Eslick, Alejandro Garciadego, Kanishka Ghosh"
```
andrewlee94, line 35: I do not think that John Eslick and I need to be on this author list (not as sure about Alejandro).

idaes/models/unit_models/packed_bed_reactor.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+@declare_process_block_class("PBR")
+class PBRData(UnitModelBlockData):
+    """
+    Standard Plug Flow Reactor Unit Model Class
```
andrewlee94, line 41: Need to update docs string here.

idaes/models/unit_models/packed_bed_reactor.py
```diff
@@ -0,0 +1,337 @@
...
+                                    is_reaction_parameter_block)
+
+from idaes.core.util.misc import add_object_reference
+import idaes.core.util.unit_costing as costing
+from idaes.core.util.constants import Constants as const
+
+__author__ = "Andrew Lee, John Eslick, Alejandro Garciadego, Kanishka Ghosh"
+
+
+@declare_process_block_class("PBR")
```
andrewlee94, line 38: I would recommend `PackedBedReactor` as a better name - we normal err on the side of verbose.

idaes/models/unit_models/packed_bed_reactor.py
```diff
@@ -0,0 +1,337 @@
...
+        doc="""Number of finite elements to use when transforming length
+domain,
+**default** - 20."""))
+    CONFIG.declare("collocation_points", ConfigValue(
+        default=3,
+        description="No. collocation points to use for DAE transformation",
+        doc="""Number of collocation points to use when transforming length
+domain,
+**default** - 3."""))
+    CONFIG.declare("mass_catalyst", ConfigValue(
```
andrewlee94, line 200: I do not think these are things that should be set as config arguments - without having seen the following code these look like they are just default values for parameters that will be constructed later.

For cases like this, we prefer to get the user to set the values directly after the model is constructed.

idaes/models/unit_models/packed_bed_reactor.py
```diff
@@ -0,0 +1,337 @@
...
+        default=1,
+        description="true density of the catalyst",
+        doc="""true density of the catalyst,
+**default** - 1.1"""))
+    CONFIG.declare("bulk_density", ConfigValue(
+        default=1,
+        description="bulk density of the catalyst",
+        doc="""bulk density of the catalyst,
+**default** - 1.1"""))
+    def build(self):
```
andrewlee94, line 215: You will want to run `black` on this code and check for PEP-8 formatting - there should be a line break before the declaration of a new method.

idaes/models/unit_models/packed_bed_reactor.py
```diff
@@ -0,0 +1,337 @@
...
+        Begin building model (pre-DAE transformation).
+
+        Args:
+            None
+
+        Returns:
+            None
+        """
+        # Call UnitModel.build to setup dynamics
+        super(PBRData, self).build()
```
andrewlee94, line 226: In Python 3, you only need `super().build()`.

idaes/models/unit_models/packed_bed_reactor.py
```diff
@@ -0,0 +1,337 @@
...
+        add_object_reference(self, "length", self.control_volume.length)
+
+        # Add volume variable for full reactor
+        units = self.config.property_package.get_metadata()
+        self.volume = Var(initialize=1,
+                          doc="Reactor Volume",
+                          units=units.get_derived_units("volume"))
+
+        self.area = Var(initialize=1,
+                        doc="area",
```
andrewlee94, line 281: I would suggest "Total Cross Sectional Area" to clarify and distinguish this from the control volume area.

idaes/models/unit_models/packed_bed_reactor.py
```diff
@@ -0,0 +1,337 @@
...
+                        units=units.get_derived_units("area"))
+
+        self.geometry = Constraint(expr=self.volume == self.config.mass_catalyst/self.config.bulk_density)
```
andrewlee94, line 284: You should never use entries from the config block in an expression - if you do this it means that they can never be changed from the initial value in the config block as they will be interpreted a constants.

As mentioned earlier, what you should do is:

1. Remove these entries from the config block.
2. Create new `Var` components for these on the model, along with appropriate units (e.g. for density)
3. Use these new components when constructing the required expressions.
4. Users will then need to fix the values for these `Vars` as inputs when solving their model, which also allows them to be adjusted if needed or (even optimized in future if appropriate).

idaes/models/unit_models/packed_bed_reactor.py
```diff
@@ -0,0 +1,337 @@
...
+
+        self.area_from_volume = Constraint(expr=self.area == self.volume/self.length)
+
+        def calculate_void_fraction(self):
+            return 1 - self.config.bulk_density/self.config.catalyst_density
+        self.epsilon = Expression(rule=calculate_void_fraction)
+
+        self.real_area = Constraint(expr=self.control_volume.area == self.area*
```
andrewlee94, line 292: I would suggest a different name for this, or at least append `_eq` to the end so that users do not mistake this for a variable giving the "real area".

idaes/models/unit_models/packed_bed_reactor.py
```diff
@@ -0,0 +1,337 @@
...
+        # Add PBR performance equation
+        @self.Constraint(self.flowsheet().config.time,
+                         self.control_volume.length_domain,
+                         self.config.reaction_package.rate_reaction_idx,
+                         doc="PBR performance equation")
+        def performance_eqn(b, t, x, r):
+            if x == 0.0:
+               return Constraint.Skip
+            else:
+                return b.control_volume.rate_reaction_extent[t, x, r] == (
```
andrewlee94, line 304: This looks like it makes some assumptions about how the reaction rate is defined which need to be clarified by documentation (notably that it is rate per unit mass). Due to this case-specific assumption, I think this model needs to go in `models_extra` for now.

idaes/models/unit_models/packed_bed_reactor.py
```diff
@@ -0,0 +1,337 @@
...
+            self.deltaP = Reference(self.control_volume.deltaP[...])
+
+    def _get_performance_contents(self, time_point=0):
+        var_dict = {"Volume": self.volume}
+        var_dict = {"Length": self.length}
+        var_dict = {"Area": self.area}
+
+        return {"vars": var_dict}
+
+    def get_costing(self, year=None, module=costing, **kwargs):
```
andrewlee94, line 324: You can remove this `get_costing` stuff - it is in the process of being deprecated and will no longer work after November.

idaes/models/unit_models/tests/test_pbr.py
```diff
@@ -0,0 +1,298 @@
+#################################################################################
```
andrewlee94, line 1: I assume this is still WIP, but this model looks like it would need a case-specific set of tests with their own properties (due to the assumptions about the units for the reaction rate).

andrewlee94 (changes requested):
@kanishka-ghosh Thank you for your contribution!

I have taken a look though the code as it is and had a number of comments and suggestions on how to improve it. I think most of them are fairly simple to address, although I think you will find that these actually bring up another unit consistency issue that you have overlooked so far.

Regarding you comment on heat duty - the units of heat in the control volume are W/m due to the discretization of the length domain, and this will be inherited by the unit model (and thus needs to be addressed there). Note thatThis would need to be integrated over the length of the reactor to get the overall heat duty (I am not sure if Pyomo has added this feature yet - it was missing when we first wrote the PFR model years ago).

chineduobiora:
> @kanishka-ghosh I have thought about this a little bit more, and I am beginning to wonder if this is the wrong way to go about modeling these systems. Notably, this approach assumes the reaction kinetics and a the catalyst material are separate - i.e. the catalyst proprieties are set on the unit level an the reaction properties are in the separate reaction package. I can see a number of issues that then arise from this, such as unit and thermodynamic consistency. I.e. the reaction kinetics are all based on a specific catalyst (activation energy and pre-exponential factor) which are at least partially coupled to the catalyst voidage and bulk density.
> 
> I wonder if the better approach is to instead build the catalyst properties into the reaction package and just use the standard reactor models (e.g. PFR). This keeps the catalyst and reaction properties coupled to ensure consistency is maintained. Basically, my proposal is that the reaction package should contain the catalyst parameters (bulk density and voidage) and the necessary calculations to take the experimental reaction rate in `mol/s.kg_catalyst` and provide the corrected volumetric reaction rate (in `mol/s.m^3`) - basically just move all the calculations to the reaction package.
> 
> @aostace01 @chineduobiora Do you have any thoughts on this?

@kanishka-ghosh, I agree with @andrewlee94. I think the model is relatively simple and close enough to the PFR that that’s a reasonable way to go.

adowling2:
This is a great discussion. I asked @kanishka-ghosh to make a draft PR to get feedback just like this! @agarciadiego also helped with this example before his graduation. (Want to give credit and a shout-out.)

bpaul4:
@kanishka-ghosh this is an interesting contribution, thanks for your efforts here. The IDAES library includes three gas-solid contactor models (FixedBed0D, MovingBed and BubblingFluidizedBed) that may be of interest to you. Relevant model, property/reaction package and flowsheet files may be found here: https://github.com/IDAES/idaes-pse/tree/main/idaes/models_extra/gas_solid_contactors.

kanishka-ghosh:
@andrewlee94 I have not been working on this PR recently.  I plan to work on this towards the end of this month.

ksbeattie:
@adowling2, @kanishka-ghosh is this still active?

adowling2:
Yes, my intention is for this to get contributed.

andrewlee94:
For now, there does not seem to be much activity on this PR. I will close it for now to reduce clutter on the PR board and help us see what is outstanding for the upcoming release, but we can re-open this once activity start up again.

adowling2:
@kanishka-ghosh Sounds good.

PR #947, opened 2022-09-09
Fixing some old deprecations

andrewlee94:
## Summary/Motivation:
This PR removes some older deprecation warnings that appear to have been fully resolved.

## Changes proposed in this PR:
-
-

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #948, opened 2022-09-09
Addressing Deprecation Messages

MarcusHolly:
## Fixes

Addresses some deprecation warnings

## Summary/Motivation:
Continuation of https://github.com/IDAES/idaes-pse/pull/947

- Addresses svg_tag warning
- Adresses stitch_dynamic warning

### Legal Acknowledgement


MarcusHolly:
Not going to work on this again until next week, but I have a few questions:

1.) Are all of the deprecation warnings I should be addressing marked by the "@deprecated("
- svg_tag
- TagReference
- sipopt
- stitch_dynamic
- kaug
- get_costing

2.) Are there any that I should hold off on trying to address (like TagReference)?

3.) @andrewlee94 I haven't deleted any of the @deprecated messages yet, but how did you run the pytest again to make sure the changes actually addressed the issue? When I ran "pytest -m "not integration"", it started going through all the files and would've taken hours to run. So how was your pytest so short? Did you only test the individual files that were changed?

andrewlee94:
@MarcusHolly 

1&2. For now, focus on svg_tag and TagReference. I have separate PRs that deal with costing and a lot of the others, and the remainder will require a lot more effort.
3. There is a bug with running the tests right now such that the marks are ignored (my fault) - I am working with @lbianchi-lbl  to fix this at the moment.

data/config.yaml
```diff
@@ -1,8 +1,8 @@
 _id: d779730c7cc94155877a24e2e8be2ad1
-created: '2022-04-18T14:36:30.754421'
+created: '2022-09-09T13:05:49.279830'
```
andrewlee94, line 2: This file should not get checked in for PRs - running the tests keep changing this for some reason, and we haven't managed to fix that yet.

idaes/core/util/tests/test_tag_svg.py
```diff
@@ -22,7 +22,8 @@
 from pyomo.environ import ConcreteModel, Set, Block, Var
 from pyomo.network import Port
 
-from idaes.core.util.misc import TagReference, svg_tag
+from idaes.core.util.misc import TagReference
+from idaes.core.util.tags import svg_tag
```
andrewlee94, line 26: If this was done right, the new `svg_tag` should have its own tests so we should be able to remove this instead. If not, then we should move these tests over to the `test_tags.py` file.

andrewlee94 (changes requested):

idaes/core/util/tests/test_tag_svg.py
MarcusHolly: I don't think svg_tag really has its own test since it is only referenced in "lazy testing" where it confirms the data from the xml files. I just moved everything from this file into test_tags.py . Should I try to write new tests for svg_tag or leave it to someone else?

data/config.yaml
MarcusHolly: So is there a way to remove this file from the PR or were you just noting that, ideally, changes in this file shouldn't be picked up by PRs?

andrewlee94: There are ways to revert changes to a file after a commit has been made (git can do just about anything). @lbianchi-lbl is probably the best person to ask on how to do it (there are many ways it can be done, but there are different side-effects).

andrewlee94: Note however- this file is meant to be there, it just should not be changed by a PR (i.e. you want to revert your changes, not delete the file).

lbianchi-lbl: > There are ways to revert changes to a file after a commit has been made (git can do just about anything). @lbianchi-lbl is probably the best person to ask on how to do it (there are many ways it can be done, but there are different side-effects).

The following should work (where 0309eec is the last commit on `main` off of which this branch was created):

```sh
git checkout 0309eec -- data/config.yaml
git commit -m "Revert changes to data/config.yaml"
```

MarcusHolly:
> @MarcusHolly Something else to be aware of for IDAES - you should now run `black` on the code prior to committing it (`black` is an automatic code formatter).
> 
> To install: `pip install black` To run `black [dir_or_file]`
> 
> You can also set up a pre-commit hook in `git` that will automatically run `black` before any commit and let you know if there are any failures: https://github.com/IDAES/idaes-pse/wiki/howto-pre-commit

Just tried the pre-commit hook to automatically run black, but it looks like that test is still failing... I'll run black manually on my next commit

idaes/core/util/tests/test_tag_svg.py
andrewlee94: The existing (moved) tests should be sufficient for now.

andrewlee94 (approved):

eslickj (approved):

PR #973, opened 2022-09-23
Get rid of default in documentation

### LINKED ISSUE #972: Deprecate default in documentation
Although `default` has been excised from the repo's code, it's still showing up in automatically generated unit model documentation:

image

dallan-keylogic:
## Fixes
Remove references to the `default` dictionary in process block documentation. I'm not sure the best way to convey to the reader that they're meant to input config entries like normal function args, but this is what it looks like now:

image

Fixes #972 .

### Legal Acknowledgement


eslickj (approved):

andrewlee94:
@dallan-keylogic What is the status on this PR? We probably want to get this into the November release so that our documentation is up-to-date for the transition.

dallan-keylogic:
@andrewlee94 I can take a look at it today. I still need to figure out what sort of substitution is going on when documents are built (probably via scripts @eslickj wrote), and then what is the best way to convey to the user that they should input config arguments like normal function keyword arguments.

eslickj:
I guess without generating the documentation, you should be able to print the `__doc__` string for an IDAES model class and make adjustments.  I can probably make it look right, so let me know if you have any trouble and I'll mess with it.

dallan-keylogic:
@andrewlee94 , I'm taking another look at this one today. Managed to smooth out the config arguments, but still need to change formatting from headings to bullet points:

image

Looking at this documentation again, I don't think we _ever_ want the user to use any of the first three arguments. Below the config arguments, we have another couple of arguments where we map config dictionaries to indices when creating an indexed unit. 

image


Obviously the "default" reference needs to go, but in order for these arguments to make sense we need to maintain some sort of distinction between config arguments and non-config keyword arguments.

andrewlee94:
@dallan-keylogic Whilst we do want to get the formatting sorted out, for this release I think we should focus on making sure we get rid of the references to default so that out docs are correct (if poorly formatted).

dallan-keylogic:
I reverted us back to having Config be a special header for now, but got rid of the additional "default" hiding in the arguments below the config section.

image

PR #976, opened 2022-09-30
Implement transport properties for modular property framework

dallan-keylogic:
## Summary/Motivation:
Currently there aren't generic methods to generate viscosity and thermal conductivity in the modular property framework. This PR will implement some common methods from the Properties of Gases and Liquids and test them against sample calculations contained therein.

### Legal Acknowledgement


idaes/models/properties/modular_properties/base/tests/test_generic_property.py
```diff
@@ -1308,11 +1314,11 @@ def test_build_all(self, frame):
                     frame.props[1], frame.params.get_metadata().properties[p]["method"]
                 )
                 continue
-            elif p in ["diffus_phase_comp"]:
+            elif p in ["diffus_phase_comp", "visc_d_phase_comp"]:
                 # phase indexed properties - these will be tested separately.
                 continue
             else:
-                assert hasattr(frame.props[1], p)
+                hasattr(frame.props[1], p)
```
andrewlee94, line 1321: Shouldn't the `assert` remain here, otherwise the test will not fail if the attribute is not present; `hasattr` will return `False` but `pytest` won't know that that is a failure.

idaes/models/properties/modular_properties/eos/ceos.py
```diff
@@ -227,12 +228,24 @@ def func_b(m, j):
         )
 
         if mixing_rule_a == MixingRuleA.default:
+            units = b.params.get_metadata().derived_units
```
andrewlee94, line 231: As a heads up, you might want to talk to @viiibhav to coordinate your PR with his (https://github.com/IDAES/idaes-pse/pull/977) - there are pretty much guaranteed to conflict, so it would help to at least be aware of what each of you are doing.

idaes/models/properties/modular_properties/pure/tests/test_ChapmanEnskog.py
```diff
@@ -0,0 +1,166 @@
...
+    m.params.config.parameter_data = {
+        "lennard_jones_sigma": (param_dict["sigma"], pyunits.angstrom),
+        "lennard_jones_epsilon_reduced": (
+            param_dict["epsilon"],
+            pyunits.K,
+        ),
+    }
+
+    # Also need to dummy configblock on the model for the test
+    # m.config = ConfigBlock(implicit=True)
```
andrewlee94, line 51: Commented code.

idaes/models/properties/modular_properties/pure/Eucken.py
```diff
@@ -0,0 +1,76 @@
...
+# Gas viscosity at low pressures from Lennard Jones paramters. Note that LJ parameters
+# are underdetermined when estimated from viscosity data (see The Indeterminacy of
+# the Values of Potential Parameters as Derived from Transport and Virial Coefficient
+# by Reichenberg D., 1973 for more information) so it's important to use LJ parameters
+# from the same source.
+class Eucken(object):
+    class therm_cond_phase_comp(object):
+        @staticmethod
+        def build_parameters(cobj, p):
+            # pobj = cobj.parent_block().get_phase(p)
```
andrewlee94, line 35: Commented code.

idaes/models/properties/modular_properties/base/generic_property.py
```diff
@@ -827,12 +831,30 @@ def build(self):
...
-                                meth.build_parameters(cobj, p)
-                            except AttributeError:
-                                # No build_parameters method
-                                # Assume it is not needed and continue
-                                continue
+                            if hasattr(meth, "build_parameters"):
+                                build_parameters = meth.build_parameters
+                            else:
+                                # If not, guess meth is a class holding property subclasses
+                                try:
```
andrewlee94, line 838: Just to be sure, could you explain why this was needed (as a comment on the PR, not in the code)?

idaes/models/properties/modular_properties/base/generic_property.py
```diff
@@ -1100,6 +1125,7 @@ def define_metadata(cls, obj):
                 "mw": {"method": "_mw"},
                 "mw_comp": {"method": "_mw_comp"},
                 "mw_phase": {"method": "_mw_phase"},
+                "number_prandtl_phase": {"method": "_number_prandtl_phase"},
```
andrewlee94, line 1128: I would name this `Prandlt_number_phase` - I don't think we have a standard for this yet (but could be wrong).

idaes/models/properties/modular_properties/base/generic_property.py
```diff
@@ -3715,33 +3857,121 @@ def rule_surf_tens_phase(b, p):
...
+            )
 
             def rule_therm_cond_phase(b, p):
-                return get_phase_method(b, "therm_cond_phase", p)(b, p)
+                try:
+                    return b.therm_cond_phase[p] == get_phase_method(
+                        b, "therm_cond_phase", p
+                    )(b, p)
+                except GenericPropertyPackageError:
+                    # What's the best way to communicate to the user that therm_cond_phase isn't meaningful here?
```
andrewlee94, line 3879: I am not sure there is a good way - there comes a point where we have to rely on user-knowledge, and if the user lacks the knowledge the issue is more "why are you using IDAES" than with specific subparts of the tool itself.

andrewlee94 (changes requested):
There is a lot to go through, but I didn't see anything that stood out as wrong. Just a few minor comments.

idaes/models/properties/modular_properties/base/generic_property.py
dallan-keylogic: The way `diffus_phase_comp` is implemented now, `build_parameters` is a method directly on the parent class. However, in order to support providing `visc_d_phase_comp` and `therm_cond_phase_comp` methods through the `ConstantProperty` class, we need to also support looking up subclasses with the name, e.g., `visc_d_phase_comp` on the parent class.

andrewlee94: OK, I think that is a good enough reason. We could insist they provide the sub-class from ConstantProperties, but I don't see any reason not to support this API either (I can also see some good reasons to do it to avoid unintended conflicts, e.g. getting cp from one module but enthalpy from another).

bpaul4 (approved):
One minor suggestion, not necessary though.

dallan-keylogic:
@ksbeattie @lbianchi-lbl , we seem to get test failures from Python 3.8 when CodeCov can't upload coverage reports frequently. Is there any way to fix that?

idaes/models/properties/modular_properties/base/generic_property.py
```diff
@@ -3718,32 +3821,95 @@ def _therm_cond_phase(self):
...
-                return get_phase_method(b, "therm_cond_phase", p)(b, p)
+                try:
+                    return get_phase_method(b, "therm_cond_phase", p)(b, p)
+                except GenericPropertyPackageError:
+                    # Handle case where thermal conductivity isn't defined for a phase
+                    return Expression.Skip
 
             self.therm_cond_phase = Expression(
                 self.phase_list,
-                doc="Thermal conductivity for each phase",
```
andrewlee94, line 3725: Any particular reason you dropped the doc string?

idaes/models/properties/modular_properties/base/generic_property.py
```diff
@@ -3718,32 +3821,95 @@ def _therm_cond_phase(self):
...
+                    cobj.config.therm_cond_phase_comp is not None
+                    and p in cobj.config.therm_cond_phase_comp
+                    and cobj.config.therm_cond_phase_comp[p] is not None
+                ):
+                    return cobj.config.therm_cond_phase_comp[
+                        p
+                    ].therm_cond_phase_comp.return_expression(b, cobj, p, b.temperature)
+                else:
+                    # Handle case where thermal conductivity isn't defined for a particular phase-component pair
+                    return Expression.Skip
```
andrewlee94, line 3853: Do we really want to do this? This leaves a potential gap for users who might not be aware of this. My initially feeling is that if someone want to define a set of phase-component properties then they need to define all valid-phase component properties.

idaes/models/properties/modular_properties/base/generic_property.py
```diff
@@ -3718,32 +3821,95 @@ def _therm_cond_phase(self):
...
+
+            self.therm_cond_phase_comp = Expression(
+                self.phase_list,
```
andrewlee94, line 3856: You should use the `phase_component_set` here - this will automatically exclude any invalid phase-component pairs (i.e., any component that was declared to not exist within a given phase).

andrewlee94 (changes requested):
There is a lot to go through ,so I probably missed a few things, but I only had a couple of comments (and only one major one).

The major comment is regarding properties indexed by phase and component - you should use the `phase_component_set` to index these rather than `(phase_list, component_list)`, as htis will eliminate nay invalid combinations.

idaes/models/properties/modular_properties/base/generic_property.py
andrewlee94: Is this still necessary after the change to using `phase_component_set`? If not we should remove it, otherwise I think we need to have a discussion about what we should expect in these cases.

idaes/models/properties/modular_properties/base/generic_property.py
```diff
@@ -3715,7 +3818,11 @@ def _therm_cond_phase(self):
         try:
 
             def rule_therm_cond_phase(b, p):
-                return get_phase_method(b, "therm_cond_phase", p)(b, p)
+                try:
+                    return get_phase_method(b, "therm_cond_phase", p)(b, p)
+                except GenericPropertyPackageError:
+                    # Handle case where thermal conductivity isn't defined for a phase
```
andrewlee94, line 3824: I am a little bit cautious about this - what are the implications of having this missing, and what error message would a user receive if they tried accessing a missing entry? This is probably the right thing to do, but what happens in a unit model if it tries to get a missing thermal conductivity?

idaes/models/properties/modular_properties/base/generic_property.py
andrewlee94: I assume this is building off the conditional in `therm_cond_phase`. It is probably OK to assume that thermal conductivity is not defined for a specific phase, but my initial feeling is that it should be defined for all components within a given phase. If so, we should probably put a check in place for this.

andrewlee94 (changes requested):
A couple of comments on the code, but most of this looks good.

However, we do need some documentation of the new libraries and methods.

idaes/models/properties/modular_properties/base/generic_property.py
andrewlee94: Yeah - that error is not going to be helpful to the general user, especially one who is taking a property definition and unit model they did not write themselves.

We have a problem here that this is general purpose code, and so it needs to work gracefully in all circumstances, which this unfortunately does not. I'll need to think more about this, but I definitely do not like leaving these as `Expression.Skip`.

idaes/models/properties/modular_properties/pure/Eucken.py
```diff
@@ -0,0 +1,67 @@
...
+                    f"{cobj.local_name}."
+                )
+
+            if hasattr(cp_mol_ig_comp, "return_expression"):
+                cp_func = cp_mol_ig_comp.return_expression
+            else:
+                cp_func = cp_mol_ig_comp.cp_mol_ig_comp.return_expression
+
+            therm_cond = (
+                b._visc_d_phase_comp[p, cobj.local_name]
```
lbianchi-lbl, line 62: This should get rid of the Pylint error on this file:

```suggestion
                b._visc_d_phase_comp[p, cobj.local_name]  # pylint: disable=protected-access
```

lbianchi-lbl:
@dallan-keylogic If Pylint and Black are playing the annoying game of "moving the comment to where it doesn't work", you have a couple of alternatives:

- Use the numeric code (`W0212`) instead of the full name `protected-access` so that the line is below the length limit enforced by Black
- Pylint directive apply either at the line level or at the scope level, so you could also move the directive (`# pylint: disable=protected-access`) to the containing block
  - Note that this would disable the check for the entire scope (unless you add the corresponding enabling directive `# pylint: enable=protected-access`), but since this check is more of an "API user beware" that some potentially bug-inducing behavior, it'd probably be fine

idaes/models/properties/modular_properties/base/generic_property.py
andrewlee94: I am still concerned about `Expression.Skip` in these cases - for the phase-component properties I think we are OK as we are making those private attributes now, but for these public facing ones I think we need to be more careful.

I think we either need to provide a meaningful value for that property in all phases, or we need to make sure that a meaningful error is returned (which I don't think is possible as that occurs outside the property package). Thus, my initial inclination is that we need to insist that users define a method for all phases, even in cases where they know for sure it won;t be used (they can always assign a constant value). This also helps make it clear to any downstream users, as they can see from the config that a constant value is used (whereas an absent value could be overlooked).

The alternative is to have some way of making sure the user gets a useful message - i.e. PropertyNotSupportedError, and probably a logger warning when the Expression is skipped as well. However, I am not sure how we would actually catch that.

idaes/models/properties/modular_properties/pure/tests/test_ChapmanEnskog.py
```diff
@@ -0,0 +1,166 @@
...
+        m.props[1], m.params, "Vap", m.props[1].temperature
+    )
+    expr_micropoise = pyunits.convert(expr, pyunits.micropoise)
+
+    # Pulled from Table 9.2, Properties of Gases and Liquids 5th Ed.
+    T_list = [10, 100, 300, 700]  # Gas temperature in C
+    visc_list = [120, 163, 246, 376]  # Experimental viscosities in millipoise
+    for i in range(4):
+        m.props[1].temperature.value = T_list[i] + 273.15
+        assert pyo.value(expr_micropoise) == pytest.approx(visc_list[i], rel=0.03)
```
andrewlee94, line 111: Minor comment: this a fairly loose tolerance (3%) and I see it used in other tests too, although I suspect that is unavoidable due to the uncertainty in the data. Just wanted to check to make sure these could not be tightened.

idaes/models/properties/modular_properties/transport_properties/tests/test_shell_and_tube_1D_transport.py
```diff
@@ -0,0 +1,516 @@
...
+            )
+    return solver.solve(m, tee=True)
+
+
+class Test_transport_properties_ideal(object):
+    @pytest.fixture(scope="class")
+    def hx(self):
+        return build_model(eos=EosType.IDEAL)
+
+    @pytest.mark.component
```
andrewlee94, line 286: These are really `intergration` tests, as they involve combining a unit model and a property package. This will also help keep test times more manageable. This also ties into a TODO comment I noted later about changing the marks when the new NL writer was the default.

idaes/models_extra/power_generation/properties/natural_gas_PR.py
```diff
@@ -75,18 +87,27 @@ class EosType(enum.Enum):
...
         "equation_of_state": Cubic,
         "equation_of_state_options": {"type": CubicType.PR},
+        "visc_d_phase": ViscosityWilke,
+        "therm_cond_phase": ThermalConductivityWMS,
     },
     "Liq": {
         "type": LiquidPhase,
         "equation_of_state": Cubic,
         "equation_of_state_options": {"type": CubicType.PR},
+        "visc_d_phase": None,
```
andrewlee94, line 97: This made me think of a possible option for the issue of skipping properties in certain phases - maybe we could have a `NoMethod` option instead of `None`. We would still insist that users provide a method for all properties in all phases, but for cases like this it would let the user flag a given phase to be skipped - the benefit being that it is explicit in the config dictionary.

andrewlee94 (changes requested):
A few comment, mostly minor. The one big concern is the case of properties being skipped in some phases (I am OK with the solution for the phase-component case). However, I did think of a possible solution that tries to balance skipping expressions for some indices whilst being explicit about the fact while looking at the natural gas example.

docs/reference_guides/model_libraries/generic/property_models/index.rst
```diff
@@ -8,3 +8,4 @@ Property Models
     activity_coefficient
     helmholtz
     interrogator
+    transport_properties/index
```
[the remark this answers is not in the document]
andrewlee94: The modular property package docs are located here: https://idaes-pse.readthedocs.io/en/stable/explanations/components/property_package/general/index.html#generic-property-package-framework

However, we obviously need to add a cross-link or reorganize the docs, as it took me a few minutes to find these.

idaes/models/properties/modular_properties/pure/ChapmanEnskog.py
```diff
@@ -0,0 +1,110 @@
...
+        """Method to build Lennard Jones parameters"""
+        units = cobj.parent_block().get_metadata().derived_units
+        if not hasattr(cobj, "lennard_jones_sigma"):
+            cobj.lennard_jones_sigma = Var(
+                doc="Parameter sigma from Lennard Jones potential",
+                units=units["length"],
+            )
+            set_param_from_config(cobj, param="lennard_jones_sigma")
+
+        if not hasattr(cobj, "lennard_jones_epsilon_reduced"):
```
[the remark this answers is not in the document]
dallan-keylogic: Presently, yes, but I'm not ruling out a future case in which one is created but the other isn't.

idaes/models/properties/modular_properties/pure/tests/test_ChapmanEnskog.py
dallan-keylogic: We're testing against experimental values here, and that sort of tolerance is just what we get. The thermal conductivity methods are even worse, sometimes getting 15-30% error. At a certain point, getting the right order of magnitude is victory enough.

idaes/models_extra/power_generation/properties/natural_gas_PR.py
andrewlee94: That is probably less important, as they are semi-private at least, although being explicit is always good.

andrewlee94: I think those are best left private - they have questionable physical meaning anyway, even when they are all defined.

idaes/core/base/phases.py
```diff
@@ -107,6 +107,12 @@ class PhaseData(ProcessBlockData):
         "visc_d_phase",
         ConfigValue(description="Method to calculate dynamic viscosity of phase"),
     )
+    CONFIG.declare(
+        "transport_property_options",
+        ConfigValue(
+            description="Options for transport properties like viscosity, surface tension, and thermal conductivity"
```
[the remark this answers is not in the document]
andrewlee94: `ConfigBlocks` can be accessed as dicts too, so we can keep things in a way that supports both if necessary.

The advantage of the ConfigBlock is that it lets us defined expected keys to help catch errors (assuming there is sufficient consistency in the keys).

idaes/models_extra/power_generation/properties/natural_gas_PR.py
dallan-keylogic: Wait, this doesn't work the way I thought it would. I was hoping to be able to do something like:

```
    class therm_cond_phase(object):
        """No method for thermal conductivity"""

        @staticmethod
        def build_parameters(pobj):
            """No parameters needed"""
            pass

        @staticmethod
        def return_expression(b, p):
            """Raise error when therm_cond_phase is called for in phase p"""
            raise PropertyNotSupportedError(
                f"Property package does not support thermal conductivity in phase {p}"
            )
```

Unfortunately, that error is just going to get raised during `Expression` creation, not when the user tries to access it. If we're just hiding the `Expression.Skip` inside a new class, is that so much better than what we had before?

andrewlee94: What I had in mind was this:

```
 class therm_cond_phase(object):
        """No method for thermal conductivity"""

        @staticmethod
        def build_parameters(pobj):
            """No parameters needed"""
            pass

        @staticmethod
        def return_expression(b, p):
            """Raise error when therm_cond_phase is called for in phase p"""
            _log.debug(f"Skipping construction of thermal conductivty for phase {p}")
            return Expression.Skip
```

Yes, it does do exactly the same thing, and just buries the `Expression.Skip` deeper. However, it does add two things:

1. It adds a logger message to help users who run into issue because of this (if they care to look).
2. More importantly, it makes the decision to skip that phase explicit in the configuration dictionary; i.e., the developer has to clearly state that they are skipping this phase. this is better than a system where it can be skipped implicitly by just not declaring anything.

Basically, the point is to force the developer/user to acknowledge the choice, and to provide hints to later users if something goes wrong. Unfortunately, I do not think we can (easily) deal with the actually error that will occur if/when someone tries to access a skipped property.

dallan-keylogic: (2) was true of the old method as well. (1) is potentially useful, but I think any user who pays attention to their log is probably going to be able to figure out the Pyomo error anyway.

andrewlee94: The issue with using `None` is that is can be done implicitly is you are not careful (e.g. it is the default value for `ConfigValues`, and is generally accepted as a value in any `ConfigValue` as well). It is better to use something explicit that cannot be confused for not providing any value.

As for 1, you over estimate the ability of our general users. However, the main purpose here is not to provide a hint; it is so we have something to point to when someone complains (i.e. the logger told you what was wrong, but you ignored it).

dallan-keylogic: Made a [feature request for Pyomo](https://github.com/Pyomo/pyomo/issues/2770#issue-1629760340) to maybe let us attach additional information to the `KeyError` raised, or maybe substitute our own exception. In the meantime, we'll log it.

idaes/models/properties/modular_properties/transport_properties/no_method.py
```diff
@@ -0,0 +1,63 @@
...
+from pyomo.environ import Expression
+import idaes.logger as idaeslog
+
+_log = idaeslog.getLogger(__name__)
+
+
+class NoMethod(object):
+    """Class to pass when user wants to deliberately not assign a property to a given phase"""
+
+    class visc_d_phase(object):
```
andrewlee94, line 33: Do we need the nested classes here? I think we should be able to just use a single common class for all properties which would make this far more extensible (so we don't need to modify this every time we add a new property).

idaes/models/properties/modular_properties/transport_properties/no_method.py
```diff
@@ -0,0 +1,63 @@
...
+# by the software owners: The Regents of the University of California, through
+# Lawrence Berkeley National Laboratory,  National Technology & Engineering
+# Solutions of Sandia, LLC, Carnegie Mellon University, West Virginia University
+# Research Corporation, et al.  All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and
+# license information.
+#################################################################################
+"""
+"Argh!" shouted the cyclops, "IDAES is throwing an error!"
```
andrewlee94, line 14: Whilst this is a very clever reference to Odysseus and Polyphemus, I am not sure it is something we really want in the code base (unfortunately).

idaes/models/properties/modular_properties/transport_properties/tests/test_no_method.py
```diff
@@ -0,0 +1,132 @@
...
+def define_state(b):
+    b.state_defined = True
+
+
+# Dummy method to avoid errors when setting metadata dict
+def set_metadata(b):
+    pass
+
+
+def construct_dummy_model(component_dict):
```
andrewlee94, line 54: Does this actually get used in these tests or can it be removed?

idaes/models/properties/modular_properties/transport_properties/tests/test_no_method.py
```diff
@@ -0,0 +1,132 @@
...
+    m = construct_dummy_model(component_dict)
+    caplog.clear()
+
+    NoMethod.therm_cond_phase.build_parameters(m.params.Vap)
+    assert len(caplog.records) == 1
+    assert (
+        caplog.records[0].message
+        == "Skipping construction of thermal conductivity for phase Vap"
+    )
+    NoMethod.therm_cond_phase.return_expression(m.props[1], "Vap")
```
andrewlee94, line 131: It would also be good to check that this returns `Expression.Skip`.

idaes/models/properties/modular_properties/transport_properties/no_method.py
dallan-keylogic: Looking at `get_phase_method`, it's possible we'd be able to get away with one large class except for one hangup: we need to log which property we're skipping construction of.  Right now, that's being done in the `build_parameters` method of each individual `visc_d_phase`, etc, changing the logged warning as appropriate. It's not obvious to me how we could log an informative message while also using a single common class.

idaes/models/properties/modular_properties/transport_properties/tests/test_no_method.py
dallan-keylogic: We do not need a real state block to log the error message, but we do need a parameter block. The most I'd want to do is eliminate the `component_dict` argument, I think

idaes/models/properties/modular_properties/transport_properties/no_method.py
andrewlee94: I see the issue now - the `build_parameters` method does not get the property as an argument so it can't add the name. Whilst the `get_expression` method does have the necessary information, you don't want to spam this message every time we do a skip (which could be multiple times for the same `IndexedExpression`).

idaes/models/properties/modular_properties/transport_properties/tests/test_no_method.py
andrewlee94: Ah - I missed the use of this method in the tests.

idaes/models/properties/modular_properties/transport_properties/no_method.py
dallan-keylogic: I don't think it would spam the log each time each time the property was accessed, just the first time the property was attempted to be constructed (in any phase, not just the one being skipped). However, for something like the MEA column, that would result in a log message for each finite element, so hundreds of log messages. We could get around this by demoting the logged message from "Warning" to "Debug", though.

andrewlee94: It would still put out a log message for each index as it is constructed, which could be a lot. But maybe that is OK if it was a debug message.

We could consider both - a high level, generic "There is a property with a skipped expression" message at info or warning level (emitted once by `build_parameters`), and a more specific "This index was skipped for property Y" at the debug level (emitted by `get_expression`). That could be done with a single, common class.

idaes/models_extra/power_generation/unit_models/soc_submodels/common.py
```diff
@@ -477,7 +477,7 @@ def __call__(self, possible_subset):
         "G": 0.0,
         "H": 0.0,
     },
-    "O^2-": {  # Chosen to match O2 coeffs besides enthalpy and entropy of formation
+    "O^2-": {  # Chosen to match O2 coeffs
```
dallan-keylogic, line 480: I didn't intend to commit this, but this correction to the comment is correct for the version of the model in the repo, so let's just leave it in.

andrewlee94 (approved):
Looks good, thanks for all the hard work.

PR #978, opened 2022-10-04
Improved Power Plant Costing Methods

bpaul4:
- Adds Nov 2008 and Sep 2019 CEPCI factors to the registered currency units dictionary
- Significant improvement to physical units handling for process and currency parameters, including allowing users to specify a base year with automatic currency units conversion
- Ensures units consistency across power plant costing methods, and employs new `UnitModelCostingBlock` syntax
- Variables costs were changed from units of $MM/production (implied for one year) to actual units of $MM/year. Only power production is considered, not hydrogen.

Some other changes:
- Combines costing parameter and exponent dictionaries, and adds bare erected cost currency units for a specific reference year and prefix
- For example, dictionaries containing `"BEC_units": "K$2018"` will tell the costing methods to set the reference costs in `kUSD_2018` (Pyomo unit container) and convert to the base year chosen by users.
- Similarly, scaled parameters now automatically get variable units and reference parameter accounts units must use correct Pyomo units syntax.

- Finally, the tests were updated, including the ones using the old costing methods. We will remove those in a future update.
- This will break the current NGCC SOEC example. A fix for this exists in # and should be merged ASAP after merging this PR.

### Legal Acknowledgement


bpaul4:
When integration tests are run, the only notebook that should fail is `ngcc_soec.ipynb` due to changes to the costing parameter dictionary. This is expected, and is corrected in the accompanying `examples-pse` update linked above.

idaes/core/base/costing_base.py
```diff
@@ -70,6 +70,7 @@ def register_idaes_currency_units():
                 "USD_2006 = 500/499.6 * USD_CE500",
                 "USD_2007 = 500/525.4 * USD_CE500",
                 "USD_2008 = 500/575.4 * USD_CE500",
+                "USD_2008_Nov = 500/566.2 * USD_CE500",
```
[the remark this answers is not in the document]
bpaul4: I understand your point more clearly and appreciate the comment. Let's connect offline to discuss implementation options.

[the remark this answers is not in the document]
MAZamarripa: declaring global units causes issues. One case: If you import the costing package in two instances and both declare global units.

idaes/models_extra/power_generation/costing/costing_dictionaries.py
```diff
@@ -66,3 +66,42 @@
 
 with open(os.path.join(directory, "sCO2_costing_parameters.json"), "r") as file:
     sCO2_costing_params = json.load(file)
+
+if not os.path.exists(
```
AlexNoring, line 70: I like that the exponents and other parameters are being combined into one dictionary. Does it make sense to just make the new dictionary offline and add it to the PR? That way we could remove the two json files being zipped together and the zipping code to keep the codebase cleaner.

idaes/models_extra/power_generation/costing/power_plant_capcost.py
```diff
@@ -0,0 +1,1995 @@
...
+import idaes.core.util.scaling as iscale
+from idaes.core import register_idaes_currency_units
+from idaes.models_extra.power_generation.costing.costing_dictionaries import (
+    BB_costing_params,
+    sCO2_costing_params,
+)
+
+from idaes.core import (
+    FlowsheetCostingBlockData,
+    register_idaes_currency_units,
```
AlexNoring, line 65: `register_idaes_currency_units` is imported twice.

idaes/models_extra/power_generation/costing/power_plant_capcost.py
```diff
@@ -0,0 +1,1995 @@
...
+# -----------------------------------------------------------------------------
+
+import pyomo.environ as pyo
+from idaes.core import declare_process_block_class
+
+
+@declare_process_block_class("QGESSCosting")
+class QGESSCostingData(FlowsheetCostingBlockData):
+    # Register currency and conversion rates based on CE Index
+    register_idaes_currency_units()
```
AlexNoring, line 88: If `register_idaes_currency_units()` is already called when this module is loaded is there any reason to call it here?

idaes/models_extra/power_generation/costing/power_plant_capcost.py
```diff
@@ -0,0 +1,1995 @@
...
+            self.fixed_charge_factor = Param(
+                initialize=0.0707,
+                mutable=True,
+                doc="Fixed charge rate from Exhibit 3-5 based on CRF values, real for three/five years 0.0707, nominal for three/five years 0.0886",
+            )
+            self.annualized_cost = Expression(
+                expr=self.fixed_charge_factor * self.total_as_spent_cost
+            )
+
+            self.add_COE_cost = Param(
```
AlexNoring, line 300: Maybe `add_COE_cost `should be a Var for added flexibility.

idaes/models_extra/power_generation/costing/power_plant_capcost_custom_dict.py
```diff
@@ -0,0 +1,676 @@
+#################################################################################
```
AlexNoring, line 1: I think the naming convention for this file should be more specific since we might need more custom dictionaries in the future, unless the plan is to add all new accounts to this file.

idaes/models_extra/power_generation/costing/power_plant_capcost.py
```diff
@@ -0,0 +1,1995 @@
...
+                    "Eng Fee"
+                ]
+                process_contingencies[account] = BB_costing_params[str(tech)][ccs][
+                    account
+                ]["Process Contingency"]
+                project_contingencies[account] = BB_costing_params[str(tech)][ccs][
+                    account
+                ]["Project Contingency"]
+            except KeyError:
+                try:  # next look for data in custom dictionaries
```
AlexNoring, line 652: If it is possible to merge the original and custom dictionaries we will only need to look in one location and can cut down on duplicated code.

idaes/models_extra/power_generation/costing/power_plant_capcost.py
```diff
@@ -0,0 +1,1995 @@
...
+        # check to see if a costing block already exists
+        if hasattr(self, "costing"):
+            raise AttributeError(
+                "{} already has an attribute costing. "
+                "Check that you are not calling get_costing"
+                " twice on the same model".format(self.name)
+            )
+
+        # create a costing Block
+        self.costing = Block()
```
AlexNoring, line 1060: Since this is all being built in a `FlowsheetCostingBlock` I don't think we need the self.costing block anymore, everything can be stored in self.

idaes/models_extra/power_generation/costing/power_plant_capcost.py
```diff
@@ -0,0 +1,1995 @@
...
+            "Reference Parameter": 13078,
+            "Exponent": 0.7,
+            "Eng Fee": 0.097,
+            "Process": 0,
+            "Project": 0.110,
+        }
+
+        # check to see if a costing block already exists
+        if hasattr(self, "costing"):
+            raise AttributeError(
```
AlexNoring, line 1308: Same comment about self.costing not being needed anymore

idaes/models_extra/power_generation/costing/power_plant_capcost.py
```diff
@@ -0,0 +1,1995 @@
...
+            raise Exception(
+                "CE_index_year %s is not a valid currency base option. "
+                "Valid CE index options include CE500, CE394 and years from "
+                "1990 to 2020." % (CE_index_year)
+            )
+
+        TPC_list = []
+        for o in b.parent_block().component_objects(descend_into=True):
+            # look for costing blocks
+            if hasattr(o, "costing") and hasattr(o.costing, "total_plant_cost"):
```
AlexNoring, line 1948: If the user doesn't name a `UnitModelCostingBlock` m.fs.something.costing we won't be able to find it with this method.  Same with the other methods here that search for blocks named costing.

idaes/models_extra/power_generation/costing/costing_dictionaries.py
MAZamarripa: That will be a different PR (for backwards compatibility).

idaes/models_extra/power_generation/costing/power_plant_capcost.py
andrewlee94: `register_idaes_currency_units` should not be called at the module level.

idaes/models_extra/power_generation/costing/power_plant_capcost.py
andrewlee94: The FlowsheetCostingBlock registers all costing blocks attached to it, so you should search for things via the list of registered units, not by name.

idaes/models_extra/power_generation/costing/power_plant_capcost.py
andrewlee94: Agreed - we are already creating separate blocks to hold all of the costing information, so there is no need to add further blocks.

idaes/models_extra/power_generation/costing/power_plant_capcost.py
Daison2102: I agree, and maybe use a more descriptive name for it to keep consistence,  for example "additional_cost_of_electricity"

idaes/models_extra/power_generation/costing/power_plant_capcost_custom_dict.py
bpaul4: I renamed the file `cansolv_mea_capcost_custom_dict.py` to better reflect the accounts present.

idaes/models_extra/power_generation/costing/power_plant_capcost.py
bpaul4: I like this suggestion, as long as we don't mind keeping the bituminous baseline and cansolv/mea accounts in the same place. @MAZamarripa what are your thoughts on this? We could wait until the cleanup to merge them, since we will remove the old dictionaries at that point as well.

idaes/models_extra/power_generation/costing/power_plant_capcost.py
bpaul4: Thanks for the suggestion, I will update places manually looking for "costing" attributes to look in the `_registered_unit_costing` list instead.

idaes/models_extra/power_generation/costing/power_plant_capcost.py
```diff
@@ -0,0 +1,1995 @@
...
+                if not hasattr(self, "tonne_CO2_capture"):
+                    self.tonne_CO2_capture = Param(
+                        initialize=tonne_CO2_capture, mutable=True, units=pyunits.ton
+                    )
+                self.cost_of_capture = Expression(
+                    expr=(
+                        (
+                            self.annualized_cost / pyunits.year
+                            + self.total_fixed_OM_cost / pyunits.year
+                            + self.total_variable_OM_cost[0] * self.capacity_factor
```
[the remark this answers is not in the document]
andrewlee94: Unit conversions should all be handled by `pint`, so it should (generally) not be necessary to put them in here.

idaes/models_extra/power_generation/costing/power_plant_capcost.py
Daison2102, line 300: Maybe using a more descriptive name for it to keep consistence, for example "additional_cost_of_electricity"

idaes/core/base/costing_base.py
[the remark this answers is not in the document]
andrewlee94: I have talked this over with other members of the core dev team and we agree that these units should not be defined as part of the core unit definition. To prevent issues with re-declaring units, you should write a method that first checks to see if the units are present (and maybe check the conversion factor as well) and only register the units if they do not already exist.

I am looking into ways we can handle this during testing to make sure we do not pollute the global unit registry for the tests with these as well.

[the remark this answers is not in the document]
andrewlee94: The following code at the module level of any tests for this code will hopefully work:

```
@pytest.fixture(scope="module", autouse=True)
def _register_units_for_this_module():
    orig = pyo.units.pint_registry()
    tmp = pint_module.UnitRegistry()
    pyo.units.set_pint_registry(tmp)
    yield
    pyo.units.set_pint_registry(orig)
```

This should temporarily switch out the Pyomo unit registry for the duration of tests within the module and then switch it back when it is done (you could also write a context manager to do this). Then, you can register the units you want to use within these tests without having to worry about any conflicting units being defined previously, or having your unit definitions stick around and cause conflicts in later tests.

idaes/models_extra/power_generation/costing/power_plant_capcost.py
```diff
@@ -0,0 +1,1995 @@
...
+            )
+
+            # only build COE for power plants
+            if net_power is not None:
+                self.cost_of_electricity = Expression(
+                    expr=(
+                        (
+                            self.annualized_cost / pyunits.year
+                            + self.total_fixed_OM_cost / pyunits.year
+                            + self.total_variable_OM_cost[0] * self.capacity_factor
```
Daison2102, line 314: I just noticed that in the method `get_variable_OM_costs`, all individuals `variable_operating_costs` already are multiplied by `capacity_factor`, line 1778. Thus I believe that there is not need to multiply the `total_variable_OM_cost` here by the capacity_factor. Double check this.

idaes/models_extra/power_generation/costing/power_plant_capcost.py
Daison2102, line 336: Same comment in regards to multiplying the `total_variable_OM_cost` by the `capacity_factor`. Double check this

idaes/models_extra/power_generation/costing/power_plant_capcost.py
```diff
@@ -0,0 +1,1995 @@
...
+                    else 0 * CE_index_units
+                )
+                + self.total_TPC
+                * self.pct_TPC  # other owners costs (other + spare parts + financing + other pre-production)
+            )
+
+            self.tasc_toc_factor = Param(
+                initialize=1.093,
+                mutable=True,
+                doc="TASC/TOC factor from Exhibit 3-7 reference 1, real for three years 1.093, real for five years 1.154, nominal for three years 1.242, nominal for five years 1.289",
```
Daison2102, line 284: Line too long, tray to break this into different lines

idaes/models_extra/power_generation/costing/power_plant_capcost.py
```diff
@@ -0,0 +1,1995 @@
...
+            )
+
+            self.total_as_spent_cost = Expression(
+                expr=self.total_overnight_capital * self.tasc_toc_factor
+            )
+
+            self.fixed_charge_factor = Param(
+                initialize=0.0707,
+                mutable=True,
+                doc="Fixed charge rate from Exhibit 3-5 based on CRF values, real for three/five years 0.0707, nominal for three/five years 0.0886",
```
Daison2102, line 294: Line too long, tray to break this into different lines

idaes/models_extra/power_generation/costing/power_plant_capcost.py
```diff
@@ -0,0 +1,1995 @@
...
+                "CE_index_year %s is not a valid currency base option. "
+                "Valid CE index options include CE500, CE394 and years from "
+                "1990 to 2020." % (CE_index_year)
+            )
+
+        param_dict = sCO2_costing_params[equipment]
+
+        if custom_accounts is not None:
+            # adding custom params to the dictionary
+            for key, value in custom_accounts.items():
```
Daison2102, line 1077: Consider to change `value` for `val` in this loop as the method value is being imported from pyomo and this may cause some confusion

idaes/models_extra/power_generation/costing/power_plant_capcost.py
```diff
@@ -0,0 +1,1995 @@
...
+
+        if (
+            hasattr(self, "annualized_cost")
+            and hasattr(self, "total_fixed_OM_cost")
+            and hasattr(self, "total_variable_OM_cost")
+        ):
+            var_dict["Total Annualized Cost [$MM/year]"] = (
+                value(self.annualized_cost)
+                + value(self.total_fixed_OM_cost)
+                + self.capacity_factor * value(self.total_variable_OM_cost[0])
```
Daison2102, line 389: Same comment in regards to multiplying the `total_variable_OM_cost` by the `capacity_factor`. Double check this

idaes/models_extra/power_generation/costing/power_plant_capcost.py
```diff
@@ -0,0 +1,1995 @@
...
+
+        # create a costing Block
+        self.costing = Block()
+        self.costing.library = "ASU"
+
+        # find flowsheet block to create global costing parameters
+        try:
+            fs = self.flowsheet()
+        except AttributeError:
+            fs = self.parent_block()
```
Daison2102, line 1322: Is this necessary? `fs`  is never being used

idaes/models_extra/power_generation/costing/power_plant_capcost.py
```diff
@@ -0,0 +1,1995 @@
...
+            units=CE_index_units,
+        )
+
+        @b.Constraint()
+        def total_TPC_eq(c):
+            return c.total_TPC == sum(TPC_list)
+
+    def display_flowsheet_cost(b):
+        print("\n")
+        print("Total flowsheet cost: $%.3f Million" % value(b.flowsheet_cost))
```
Daison2102, line 1967: It seems `flowsheet_cost` is not defined. Double check it

idaes/models_extra/power_generation/costing/power_plant_capcost.py
```diff
@@ -0,0 +1,1995 @@
...
+                    TPC
+                    * c.maintenance_material_TPC_split
+                    * c.maintenance_material_percent
+                    / (capacity_factor * pyunits.year)
+                    * (capacity_factor)
+                )
+
+    def get_variable_OM_costs(
+        self,
+        production_rate,
```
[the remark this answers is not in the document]
bpaul4: You're correct, I'll remove it - it was left over from the old method.

idaes/core/base/costing_base.py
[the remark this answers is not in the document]
bpaul4: Thank you for providing the code. Would a user be able to similarly define scoped custom currency units within a script, flowsheet or notebook?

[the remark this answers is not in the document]
andrewlee94: Yes - IDAES has a function (somewhere) for registering new units which is just a wrapper around `pint`'s own interface (it includes the check for if something is already defined if I recall).

Note that I missed a line above to register your new units, but that should probably occur as part of a test anyway.

idaes/models_extra/power_generation/costing/power_plant_capcost.py
MAZamarripa: I prefer to keep them separated: i) as an example to use "custom" costing correlations, ii) to keep "baseline" costing correlations isolated from others (custom, private, generic, public, etc.)

idaes/models_extra/power_generation/costing/costing_dictionaries.py
```diff
@@ -66,3 +66,42 @@
 
 with open(os.path.join(directory, "sCO2_costing_parameters.json"), "r") as file:
```
andrewlee94, line 67: This code appears to be written at the module level, which means it will be run immediately on import (e.g. during testing). This is not a good idea and can have unexpected side effects. You should create a simple method to wrap this code which returns the costing dictionary.

idaes/models_extra/power_generation/costing/generic_ccs_capcost_custom_dict.py
```diff
@@ -0,0 +1,676 @@
...
+# the BEC_units happen to be in thousands of 2016 USD for these accounts
+# for the power plant methods to work, BEC_units must be "$year", "K$year" or
+# "$Myear" for USD, thousands USD and millions USD, respectively
+import os
+import json
+from pyomo.common.fileutils import this_file_dir
+
+directory = this_file_dir()
+
+if not os.path.exists(
```
andrewlee94, line 640: As above - you should not have code which executes at the module level.

idaes/models_extra/power_generation/costing/power_plant_capcost.py
```diff
@@ -0,0 +1,2015 @@
...
+        prices=None,
+        fixed_OM=True,
+        variable_OM=False,
+        fuel=None,
+        chemicals=None,
+        waste=None,
+        tonne_CO2_capture=None,
+        CE_index_year="2018",
+    ):
+        """
```
andrewlee94, line 124: You should probably update this to be more specific to the current case. This looks like it is just taken from the base class/example.

For one, you should document all the arguments to this function in the doctstring (which can then be used to automatically generate docs for the method as well).

andrewlee94 (changes requested):
In addition to my request to move the definition of the new costing units from `costing_base.py`, I also saw a few cases where you have code that executes as the module level which needs to be addressed (just wrap it all in a simple method).

idaes/models_extra/power_generation/costing/power_plant_capcost.py
AlexNoring: I don't think we need to store all the accounts in the same place to do this. Earlier in this method we move entries from `additional_costing_params` to `custom_costing_params`. Could we then move the entries from `custom_costing_params` into `process_params`?

bpaul4: Yes, that is possible, thank you for the suggestion @AlexNoring . I will update the method accordingly.

MAZamarripa (approved):
lgtm

MAZamarripa (approved):
lgtm

idaes/models_extra/power_generation/costing/tests/test_CCS_capcost.py
```diff
@@ -43,15 +47,14 @@
...
-    dict_path = os.path.join(
-        os.path.join(directory, ".."), "generic_ccs_costing_data.json"
-    )
-    assert os.path.exists(dict_path)
+def test_regenerate_ccs_dictionaries():
+    # Switch to a temporary directory to test regenerating the dictionaries
+    _temp_context = TempfileManager.new_context()
+    wrkdir = _temp_context.create_tempdir()
+
+    load_generic_ccs_costing_dictionary(path=wrkdir)
```
andrewlee94, line 56: There needs to be some check of what is generated here. You should at least check that the expected files were generated, and maybe even check that they look right.

Also, if the `load_generic_ccs_costing_dictionary` method returns something, you should check what is returned.

andrewlee94 (commented):
One more relatively simple request - could you do at least some minimal check of the files that are generated by load_generic_ccs_costing_dictionary (at least check that they were created).

Daison2102 (approved):

PR #983, opened 2022-10-05
Enable derived costing package to define a flow cost

bknueven:
## Summary/Motivation:
In the WaterTAP costing package, we have names like `electricity_base_cost` for predefining flow costs. It would be less brittle and error-prone if the WaterTAP costing package could just use the name `electricity_cost` directly:
https://github.com/watertap-org/watertap/blob/c71b898cf490b7dc6f7a0115e95b8dbb6a771adf/watertap/costing/watertap_costing_package.py#L122-L128. Here, changing this mutable Param doesn't actually do anything, because its value is extracted when the `electricity_cost` `Var` is created.

- When registering a flow, if the name `{flow_type}_cost` is already on the global costing block, just use it instead of creating a new `Var`.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/base/costing_base.py
+++ b/idaes/core/base/costing_base.py
@@ -337,13 +337,19 @@ class FlowsheetCostingBlockData(ProcessBlockData):
         """
         self.flow_types.add(flow_type)
 
-        # Create a Var to hold the cost
-        # Units will be different between flows, so have to use scalar Vars
-        fvar = pyo.Var(
-            units=pyo.units.get_units(cost), doc=f"Cost parameter for {flow_type} flow"
-        )
-        self.add_component(f"{flow_type}_cost", fvar)
-        fvar.fix(cost)
+        name = f"{flow_type}_cost"
+        current_component = self.component(name)
+        if current_component is not None and current_component is cost:
+            pass
+        else:
+            # Create a Var to hold the cost
+            # Units will be different between flows, so have to use scalar Vars
+            fvar = pyo.Var(
+                units=pyo.units.get_units(cost),
+                doc=f"Cost parameter for {flow_type} flow",
+            )
+            self.add_component(name, fvar)
+            fvar.fix(cost)
 
         self._registered_flows[flow_type] = []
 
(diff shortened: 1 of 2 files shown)


adam-a-a (approved):
LGTM

idaes/core/base/costing_base.py
```diff
@@ -337,13 +337,19 @@ def register_flow_type(self, flow_type, cost):
...
-        fvar = pyo.Var(
-            units=pyo.units.get_units(cost), doc=f"Cost parameter for {flow_type} flow"
-        )
-        self.add_component(f"{flow_type}_cost", fvar)
-        fvar.fix(cost)
+        name = f"{flow_type}_cost"
+        current_component = self.component(name)
+        if current_component is not None and current_component is cost:
+            pass
+        else:
```
andrewlee94, line 344: I think I see an edge case here that needs to be caught - what happens if a component `name` already exists but is not `cost`? I think this would implicitly replace the existing component, so I think we need to catch this and raise an `Exception`.

andrewlee94 (changes requested):
I think there is an edge case that we need to address, plus we should probably note something about this in the docs (docs/reference_guides/core/costing/costing_framework.rst).

idaes/core/base/costing_base.py
bknueven: I agree -- I was just making a minimal change to existing behavior. Happy to add an `Exception` here though.

andrewlee94: I think an `Exception` is the best approach - it is easy to add in here, and I generally never try to second guess the user. I.e. if they do something wrong it is best to tell them to fix it rather than make assumptions.

bknueven:
> I think there is an edge case that we need to address, plus we should probably note something about this in the docs (docs/reference_guides/core/costing/costing_framework.rst).

I've added a `RuntimeError` (along with a test).

For the documentation, I've opted to extend the docstring for this function. The costing framework documentation doesn't seem to mention flow costing.

andrewlee94 (approved):
Looks good.

PR #991, opened 2022-10-17
Adding UnitSet object to manage definition of base and derived units

andrewlee94:
## Summary/Motivation:
As a start towards removing a number of "magic string constants" in the IDAES codebase, this PR introduces a new `UnitSet` object which defines the set of known quantities in IDAES as properties along with their define units of measurement. For backwards compatibility, `UnitSets` allow quantities to be accessed in a dict-like fashion using a string for the quantity name, but as this is rolled out we will begin replacing instances of this in the codebase with direct references to the properties on the `UnitSet`.

- Add new `UnitSet` object, and replaces existent `_default_units` and `_derived_units` dicts with this in `properties_meta`.
- Adds standalone tests for `UnitSets`.
- Updates tests which relied on the old dict format to use `UnitSets` instead.

### Legal Acknowledgement


idaes/core/base/reaction_base.py
```diff
@@ -214,7 +214,15 @@ def _validate_property_parameter_units(self):
...
-        for u in r_units:
+        for u in [
+            "time",
+            "length",
+            "mass",
+            "amount",
+            "temperature",
+            "current",
+            "luminous_intensity",
+        ]:
```
jsiirola, line 225: Can we add a `fundamental_units` list to the `UnitSet` so we don't need this explicit embedded constant here?

jsiirola (commented):
A couple questions (I think we can make this implementation simpler)...

idaes/core/base/property_meta.py
```diff
@@ -104,60 +104,79 @@ class UnitSet(object):
...
+    def __init__(
+        self,
+        amount: _PyomoUnit = units.mol,
+        current: _PyomoUnit = units.watt,
+        length: _PyomoUnit = units.meter,
+        luminous_intensity: _PyomoUnit = units.candela,
+        mass: _PyomoUnit = units.kilogram,
+        temperature: _PyomoUnit = units.kelvin,
+        time: _PyomoUnit = units.seconds,
+        **kwds,
```
jsiirola, line 132: If you are going to put all the base units explicitly, then can we drop `**kwds` (at least until you are going to start supporting user-specified derived units)?

andrewlee94: I was debating that - I wondered if it would be useful to have a more useful error message for extra arguments, but there does come a point where you can't fix stupid.

andrewlee94:
@dangunter @jsiirola I think this is ready for the next round of reviews, and is hopefully almost ready to merge (I'll save updating the rest of the code base for a separate PR).

idaes/core/base/property_meta.py
```diff
@@ -97,27 +97,251 @@ def define_metadata(cls, pcm):
...
+
+            # Check for expected dimensionality
+            expected_dimensionality = {
+                "amount": units.mol,
+                "current": units.watt,
+                "length": units.meter,
+                "luminous_intensity": units.candela,
+                "mass": units.kilogram,
+                "temperature": units.kelvin,
+                "time": units.seconds,
```
jsiirola, line 158: You could simplify this by making the `base_quantities` list this dict.  Reduces redundancy...

andrewlee94: I kept them separate as this also defines some "units" - I didn't want a user to accidentally take the units from here and use them somewhere else.

jsiirola (approved):

eslickj (approved):

PR #995, opened 2022-10-24
Improving properties metadata

andrewlee94:
## Summary/Motivation:
This PR introduces a new PropertyMetadata and PropertySet object for recording what thermophysical properties are supported by a property package. The eventual goal is to have a concrete definition of the standard properties used in IDAES that can be accessed programmatically and does not rely on string constants, as well as formalizing what metadata is stored for each.

-
-

### Legal Acknowledgement


andrewlee94:
Note that test failures are expected for now - there are some tests that are using the property metadata for unexpected purposes.

idaes/core/base/property_set.py
```diff
@@ -0,0 +1,2099 @@
...
+    def __iter__(self):
+        for a in dir(self):
+            if not a.startswith("_"):
+                # Ignore anything starting with "_" as they will show up twice otherwise
+                # This will also skip private attributes of object class
+                aobj = getattr(self, a)
+                if isinstance(aobj, PropertyMetadata):
+                    yield aobj
+
+    def define_property(
```
dangunter, line 200: The docs say name is required, yet the code seems to silently use the default value of None. Suggest taking `=None` off the name argument to enforce this.

eslickj (approved):

ksbeattie:
@andrewlee94 is this needed for the Nov release?

andrewlee94:
@ksbeattie No, it is not required (although it would be nice to get it finished).

eslickj:
I found another problem.  Producing hydrogen lowers the cooling duty in the NGCC + SOEC since it lower the condenser duty by a lot since steam is extracted from the steam cycle.  I need to fix some cooling duty parameters at the NGCC (no SEOC) values to get a cooling system sized for the SOEC to be able to turn off.  Working on that now.

jsiirola (approved):

PR #997, opened 2022-10-27
Scaling Issues with New Pyomo Solver Writer

MarcusHolly:
## Fixes
- Test failures associated with switching the pyomo solver writer from v1 to v2

- Sets scaling factors and default scaling in the flowsheet
-

### Legal Acknowledgement


MarcusHolly:
A few comments on this one:

1.) I can't see if these changes actually resolved the test_demo_flowsheet failures since this test has always been successful when I run it, even when I updated to v2.

2.) There are a few warnings associated with the iscale.calculate_scaling_factors that can be fixed by getting rid of lines 116-119, but I'm not sure if this is a good idea.
image

3.) It seems like I don't have the ability to assign myself or reviewers to PRs...

MarcusHolly:
@andrewlee94 I think I've made all the appropriate changes to see if fixing the scaling would resolve the failures with the activity coefficient files, but the result is still exactly the same (resulting in failures). This could be because my changes are still not being applied to the failing test for whatever reason or because scaling wasn't the issue in the first place. Either way, I will start looking into the procedure for writing NL files and will probably pick this up again on Monday.

idaes/models/properties/activity_coeff_models/tests/test_ideal_ideal_FcTP.py
```diff
@@ -69,6 +72,8 @@ def test_build(self, model):
         assert not hasattr(model.fs.state_block_ideal_vl[0], "eq_activity_coeff")
         assert not hasattr(model.fs.state_block_ideal_vl[0], "eq_mol_frac_out")
 
+        iscale.calculate_scaling_factors(model.fs.state_block_ideal_vl[0])
```
andrewlee94, line 75: This needs to go in the fixture, not the build test. The reason being that the build test is marked as a unit test, but the tests involving solvers are marked as component tests; thus if you try to run only the component tests, this line will not be executed and the solver tests will be missing scaling.

idaes/models/properties/activity_coeff_models/activity_coeff_prop_pack.py
```diff
@@ -1800,3 +1826,88 @@ def default_material_balance_type(self):
...
+                    iscale.constraint_scaling_transform(c, s, overwrite=False)
+                else:
+                    s = iscale.get_scaling_factor(
+                        self.mole_frac_comp[i], default=1, warning=True
+                    )
+                    iscale.constraint_scaling_transform(c, s, overwrite=False)
+
+        if self.is_property_constructed("density_mol"):
+            for c in self.eq_density_mol.values():
+                iscale.constraint_scaling_transform(c, sf_P, overwrite=False)
```
andrewlee94, line 1878: Why is molar density being scaled using the pressure scaling factor?

MarcusHolly: dens_mol_phase from the cubic prop pack was scaled using the pressure scaling factor, so I just carried it over w/o really double-checking it. But looking at it now, it doesn't make much intuitive sense why this would be the case, so I'll change it.

idaes/models/flowsheets/demo_flowsheet.py
```diff
@@ -70,9 +190,12 @@ def set_dof(m):
     m.fs.F03.heat_duty.fix(0)
     m.fs.F03.deltaP.fix(0)
 
+    iscale.calculate_scaling_factors(m)
+
 
 def initialize_flowsheet(m):
-    m.fs.M01.initialize(outlvl=idaeslog.WARNING)
+
+    m.fs.M01.initialize(outlvl=idaeslog.INFO_HIGH)
```
andrewlee94, line 198: Need to change these back once things are working.

idaes/models/flowsheets/demo_flowsheet.py
```diff
@@ -49,19 +51,137 @@ def build_flowsheet():
 
     TransformationFactory("network.expand_arcs").apply_to(m.fs)
 
+    m.fs.BT_props.set_default_scaling(
```
andrewlee94, line 54: For clarity, it might be helpful to break all of this out as a `set_scaling` method that is called by `main`.

idaes/models/flowsheets/demo_flowsheet.py
```diff
@@ -18,10 +18,12 @@
 """
 
 from pyomo.environ import ConcreteModel, TransformationFactory
-from pyomo.network import Arc
+from pyomo.network import Arc, SequentialDecomposition
```
andrewlee94, line 21: I don't think you are using this.

idaes/models/flowsheets/tests/test_demo_flowsheet.py
```diff
@@ -95,7 +95,7 @@ def test_solve_flowsheet(model):
         0.5, 1e-4
     )
     assert model.fs.M01.outlet.pressure[0].value == pytest.approx(101325, 1e-4)
-    assert model.fs.M01.outlet.temperature[0].value == pytest.approx(368.2, 1e-4)
+    assert model.fs.M01.outlet.temperature[0].value == pytest.approx(370.79185, 1e-4)
```
andrewlee94, line 98: This is a rather large change to be observing from just scaling.

idaes/models/properties/activity_coeff_models/activity_coeff_prop_pack.py
```diff
@@ -209,6 +210,31 @@ def build(self):
...
+        self.set_default_scaling("enth_mol_phase_comp", 1e-4, index="Vap")
+        self.set_default_scaling("enth_mol_phase", 1e-3, index="Liq")
+        self.set_default_scaling("enth_mol_phase", 1e-4, index="Vap")
+        self.set_default_scaling("entr_mol_phase_comp", 1e-2)
+        self.set_default_scaling("entr_mol_phase", 1e-2)
+        self.set_default_scaling("mw", 100)
+        self.set_default_scaling("mw_phase", 100)
+        self.set_default_scaling("gibbs_mol_phase_comp", 1e-3)
+        # Check values for these 3
+        self.set_default_scaling("fug_vap", 1)
```
andrewlee94, line 235: This needs the same scaling as pressure (more or less). It is `pressure*fugacity_coefficient`.

eslickj (approved):

PR #1000, opened 2022-10-31
Remove old Helmholtz EoS and switch to data files for component definitions.

eslickj:
The new binaries are out, so this is ready to go.

## Summary/Motivation:

Add things need to support a new version of the Helmholtz EoS that reads component data from files.  Remove old versions of the Helmholtz EoS for better maintainability.  The new AMPL user-defined function interface takes an optional argument as the last argument that is the data directory path.  It is also possible to set the data directory through an environment variable, but at this point the argument seems like the more solid option.  Once the data files are read on the first call the data directory argument is ignored, so other than passing the string around there is not much overhead in reading data files.  We can switch the the environment variable at any point if it makes more sense.  Users won't know the difference unless the directly call functions. 

- Remove old Helmholtz EoS packages (iapws95 and swco2)
- Add compatibility wrapper for iapws95 and swco2 to general Helmholtz EoS
- Now the general Helmholtz EoS takes data files to add new pure components

### Legal Acknowledgement


data/config.yaml
```diff
@@ -1,8 +1,8 @@
 _id: d779730c7cc94155877a24e2e8be2ad1
-created: '2022-04-18T14:36:30.754421'
+created: '2022-10-28T15:23:03.285808'
```
andrewlee94, line 2: Note really part of this PR, but we really need to decide what to do with this file - the tests should not be changing it, and it definitely shouldn't be something that shows up in PRs.

idaes/models/properties/general_helmholtz/components/__init__.py
```diff
@@ -26,4 +26,8 @@
     "co2": co2,
     "r134a": r134a,
     "r1234ze": r1234ze,
+    "H2O": h2o,
```
andrewlee94, line 29: Is there a way we could catch this later? E.g. wherever the key is used do `key.lower()` to catch any form of capitalization?

idaes/models/properties/general_helmholtz/helmholtz_state.py
```diff
@@ -1037,7 +1108,7 @@ def define_display_vars(self):
                 "P": self.pressure,
                 "Vapor Fraction": self.vapor_frac,
                 "Molar Enthalpy": self.enth_mol,
-                "Molar Entropy": self.entr_mol,
+                # "Molar Entropy": self.entr_mol,
```
andrewlee94, line 1111: Is there a reason this was left commented out? Also below.

idaes/models/properties/helmholtz/helmholtz.py
```diff
@@ -1,1793 +1 @@
...
-                iscale.set_scaling_factor(v, sf_inte_p * sf_dens_p)
-        try:
-            iscale.set_scaling_factor(self.eq_sat, sf_pres / 1000.0)
-        except AttributeError:
-            pass  # may not have eq_sat, and that's ok
-        try:
-            iscale.set_scaling_factor(self.eq_complementarity, sf_pres / 10)
-        except AttributeError:
-            pass  # may not have eq_complementarity which is fine
+from idaes.models.properties.general_helmholtz import *
```
andrewlee94, line 1: Should we make the a deprecated path so that we can eventually remove this entirely?

idaes/models/unit_models/tests/test_feed_flash.py
```diff
@@ -273,7 +271,7 @@ def test_initialize(self, iapws):
     @pytest.mark.skipif(solver is None, reason="Solver not available")
     @pytest.mark.component
     def test_solve(self, iapws):
-        results = solver.solve(iapws)
+        results = solver.solve(iapws, tee=True)
```
andrewlee94, line 274: Should we be setting `tee=True` in a test? I suppose it can help if the test fails, but otherwise it will just get hidden by pytest.

idaes/models/unit_models/tests/test_heat_exchanger_1D.py
```diff
@@ -868,49 +868,44 @@ def test_get_stream_table_contents(self, iapws):
...
             },
             "Cold Side Outlet": {
-                "Molar Flow (mol/s)": pytest.approx(1, rel=1e-4),
-                "Mass Flow (kg/s)": pytest.approx(1.8015e-2, rel=1e-4),
-                "T (K)": pytest.approx(286.34, rel=1e-4),
-                "P (Pa)": pytest.approx(1e5, rel=1e-4),
+                "Molar Flow": pytest.approx(1, rel=1e-4),
+                "Mass Flow": pytest.approx(1.8015e-2, rel=1e-4),
+                "T": pytest.approx(270.4877112932641, rel=1e-4),
+                "P": pytest.approx(11032305.8275, rel=1e-4),
```
andrewlee94, line 906: Are these changes in values expected? I notice that P has increased significantly.

idaes/models/unit_models/tests/test_heat_exchanger_1D.py
```diff
@@ -1093,49 +1088,44 @@ def test_get_stream_table_contents(self, iapws):
...
             },
             "Hot Side Outlet": {
-                "Molar Flow (mol/s)": pytest.approx(1, rel=1e-4),
-                "Mass Flow (kg/s)": pytest.approx(1.8015e-2, rel=1e-4),
-                "T (K)": pytest.approx(286.34, rel=1e-4),
-                "P (Pa)": pytest.approx(1e5, rel=1e-4),
+                "Molar Flow": pytest.approx(1, rel=1e-4),
+                "Mass Flow": pytest.approx(1.8015e-2, rel=1e-4),
+                "T": pytest.approx(270.4877112932641, rel=1e-4),
+                "P": pytest.approx(11032305.8275, rel=1e-4),
```
andrewlee94, line 1110: As above.

idaes/models/unit_models/tests/test_mixer.py
```diff
@@ -1128,44 +1129,36 @@ def test_get_stream_table_contents(self, iapws):
...
                 },
                 "Outlet": {
-                    "Molar Flow (mol/s)": 1,
-                    "Mass Flow (kg/s)": 1.8015e-2,
-                    "T (K)": 286.34,
-                    "P (Pa)": 1e5,
+                    "Molar Flow": 1,
+                    "Mass Flow": 1.8015e-2,
+                    "T": 270.4877112932641,
+                    "P": 11032305.8275,
```
andrewlee94, line 1159: As above.

idaes/models/unit_models/tests/test_separator.py
```diff
@@ -1432,53 +1432,44 @@ def test_get_stream_table_contents(self, iapws):
...
                 },
                 "outlet_1": {
-                    "Molar Flow (mol/s)": 1,
-                    "Mass Flow (kg/s)": 1.8015e-2,
-                    "T (K)": 286.34,
-                    "P (Pa)": 1e5,
+                    "Molar Flow": 1,
+                    "Mass Flow": 1.8015e-2,
+                    "T": 270.4877112932641,
+                    "P": 11032305.8275,
```
andrewlee94, line 1454: As above.

idaes/models_extra/power_generation/flowsheets/subcritical_power_plant/subcritical_boiler_flowsheet.py
```diff
@@ -963,8 +964,8 @@ def initialize(m):
 
     if m.dynamic is False:
         _log.info("Solving boiler steady-state problem...")
-        with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
-            res = solver.solve(fs, tee=slc.tee)
+        # with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
```
andrewlee94, line 967: Commented line

idaes/models_extra/power_generation/unit_models/boiler_heat_exchanger_2D.py
```diff
@@ -3165,7 +3171,12 @@ def initialize_build(
             blk.head_hin_eqn.deactivate()
 
         with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
-            res = opt.solve(blk, tee=slc.tee)
+            res = opt.solve(
+                blk,
+                tee=slc.tee,
+                options={"halt_on_ampl_error": "yes"},
```
andrewlee94, line 3177: Should we really have these options in a working flowsheet (that we hope will be an example for users)?

idaes/models/properties/general_helmholtz/components/__init__.py
eslickj: I can add a function call as an additional layer.  I guess I was just being lazy here.

idaes/models/properties/general_helmholtz/helmholtz_state.py
eslickj: I was thinking about bringing it back but forget.  Will remove.

idaes/models/properties/helmholtz/helmholtz.py
eslickj: Yeah.  That's my plan.  I wan't sure what to do.  I don't want to update everything to use the new property package yet, so I didn't want to add bunch of new deprecation warnings.  There are too many moving parts here.  Maybe after this PR I can update our models and tests and add the deprecation warning.

idaes/models/unit_models/tests/test_heat_exchanger_1D.py
eslickj: They aren't results.  These are pre solve unit tests and the new property package has different default values.  They should be overwritten in initialization, but it hasn't been run for these tests.

idaes/models_extra/power_generation/unit_models/boiler_heat_exchanger_2D.py
eslickj: Oops, left over from debugging.

idaes/models/properties/helmholtz/helmholtz.py
andrewlee94: Lets wait until after the v2.0 release so that we can have one release with no deprecations.

andrewlee94 (approved):

idaes/models/properties/general_helmholtz/helmholtz_functions.py
```diff
@@ -1236,7 +1287,7 @@ def _create_component_and_phase_objects(self):
     def build(self):
         super().build()
         # Check if the specified compoent is supported
-        if self.config.pure_component not in supported_components:
+        if get_component_module(self.config.pure_component) is None:
             raise RuntimeError(f"Component {self.config.pure_component} not supported.")
```
dallan-keylogic, line 1291: This could probably be a `ConfigurationError`.

idaes/models_extra/power_generation/flowsheets/subcritical_power_plant/subcritical_boiler_flowsheet.py
```diff
@@ -782,9 +784,8 @@ def initialize(m):
     fs.aECON.shell_inlet.temperature[:].value = 861.3
     fs.aECON.shell_inlet.pressure[:].value = 100000
     if m.dynamic is False:
-        fs.aECON.initialize(outlvl=outlvl)
+        fs.aECON.initialize(outlvl=idaeslog.DEBUG)
```
dallan-keylogic, line 787: Change this back to `outlvl`

idaes/models_extra/power_generation/unit_models/boiler_heat_exchanger_2D.py
```diff
@@ -1320,7 +1321,8 @@ def N_Re_tube_eqn(b, t, x):
         )
         def friction_factor_tube_eqn(b, t, x):
             return (
-                b.tube_friction_factor[t, x] * b.tube_N_Re[t, x] ** 0.25
+                b.tube_friction_factor[t, x]
+                * smooth_max(b.tube_N_Re[t, x], 1, 1e-5) ** 0.25
```
dallan-keylogic, line 1325: I take it these `smooth_max` terms are here to avoid IPOPT taking the Reynolds number somewhere weird (like negative numbers)?

eslickj: That's right.  The Reynolds number wants to go negative a lot resulting in a ton of evaluation errors on the way to the solution.  This makes it solve a lot cleaner and faster.

dallan-keylogic: My preferred solution would be to just bound it with `bounds=(0, None)`. There are advantages and disadvantages to both approaches. If bounds are used, step sizes are reduced but there's less nonlinearity.

dallan-keylogic (approved):

PR #1007, opened 2022-11-04
Set and get scaling factor

dallan-keylogic:
## Summary/Motivation:
When scaling a model, it's often desirable to scale constraints and variables based on the values the user set while having a default scaling factor if unset. This results in calling `set_scaling_factor` and `get_scaling_factor` back to back on the same component in order to get the scaling factor that's ultimately on the variable (whether set by a user or assigned a default value` in order to further scale other components. This repetition gets annoying when multiplied by 20 variable/equation pairs through a unit model.

The utility function `set_and_get_scaling_factor` was created to assign a default scaling value to a component if the user didn't specify any, then returns the scaling factor ultimately on the component, whether from a user or a default value.

- Add function `set_and_get_scaling_factor`
- Replace usage of private function in SOC submodels with `set_and_get_scaling_factor` from `core\util\scaling.py`

### Legal Acknowledgement


eslickj (commented):
set_scaling_factor currently returns None.  Why not just have set scaling factor return the scaling factor?

dallan-keylogic:
I looked into letting `set_scaling_factor` return the current scaling factor. The problem is what happens when `set_scaling_factor` is called on indexed components. Right now the default behavior is to set a scaling suffix on the indexed component, then propagate the scaling factors to all the subcomponents of the indexed component. I don't like scaling suffices on indexed components---I think standard practice should be to get scaling factors directly from their subcomponents by iterating through their indices. I still think `set_scaling_factor` should work on indexed components, but just as shorthand to iterating over their indices to set the subcomponent scaling factors.

So which scaling factor should `set_scaling_factor` return if called on an indexed component? I don't want there to be a scaling factor directly on the indexed components, and the subcomponents may have different scaling factors depending on index. The rationale for the separate function, then, is to restrict `set_and_get_scaling_factor` to base level components while raising an error on indexed components.

eslickj:
@dallan-keylogic, good point.  I'm not sure.  I guess the thing that makes the most sense maybe is that an indexed component should return a dictionary of scale factor values.  I'm okay with the extra function for now, but we're probably overdue to a scaling refresh.

eslickj (approved):

radhakrishnatg (approved):
Everything looks good, Doug!

PR #1008, opened 2022-11-04
Allowing an InitializationError to be caught

bknueven:
## Summary/Motivation:
If a modeler want to catch a unit model `InitializationError`, the current `initialize` method can leave the unit in a modified state. This PR would ensure changes the unit are cleaned up before an `InitializationError` is raised.

- Add `try/except/finally` around the call to `initialize_build`

### Legal Acknowledgement

### DIFF
--- a/idaes/core/base/unit_model.py
+++ b/idaes/core/base/unit_model.py
@@ -526,14 +526,19 @@ Must be True if dynamic = True,
             c.deactivate()
 
         # Remember to collect flags for fixed vars
-        flags = blk.initialize_build(*args, **kwargs)
+        try:
+            flags = blk.initialize_build(*args, **kwargs)
+        except InitializationError:
+            raise
+        finally:
+            # In case the InitializationError is caught,
+            # leave the unit unmodified
+            # If costing block exists, activate and initialize
+            for c in init_order:
+                c.activate()
 
-        # If costing block exists, activate and initialize
-        for c in init_order:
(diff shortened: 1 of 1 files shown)


idaes/core/base/unit_model.py
```diff
@@ -526,14 +526,19 @@ def initialize(blk, *args, **kwargs):
...
-                c.initialize(**cost_args)
+        try:
+            flags = blk.initialize_build(*args, **kwargs)
+        except InitializationError:
+            raise
+        finally:
+            # In case the InitializationError is caught,
+            # leave the unit unmodified
+            # If costing block exists, activate and initialize
+            for c in init_order:
```
andrewlee94, line 537: We definitely want to leave the model in an unmodified state if an InitializationError is encountered, but I think this issue is somewhat broader than this.

1. We probably need to update control volumes and state blocks to do the same checks to make sure they reset themselves after an initialization failure. Control volumes are easy enough, but state blocks would have to be done on an individual basis (for now).
2. Similarly, costing blocks would also need to handle initialization failures gracefully.
3. Finally, does it make sense to try to initialize the costing block given that the unit model failed? I can see arguments both ways.

Note that we intend to address many of these in a new initialization API that is currently in the planning stages.

bknueven:
We'll apply this patch in WaterTAP instead watertap-org/watertap#839.

PR #1014, opened 2022-11-09
Using flow cost as it is in costing base

bknueven:
## Summary/Motivation:
Creating an intermediate `Var` for flow costing can create issues when the flow cost is defined by intermediate (mutable) parameters. This PR would just use the object provided by the user directly, with some sanity checks.

- In `FlowsheetCostingBlockData.register_flow_type`, instead of creating an intermediate variable, put an appropriately named reference to the passed in object on the costing block with the appropriate name.
- Check if the passed in object is fixed, and raise an error if it can't be fixed easily.

### Legal Acknowledgement


andrewlee94:
@bknueven For our information, could you comment on the issues caused by intermediate Params?

bknueven:
> @bknueven For our information, could you comment on the issues caused by intermediate Params?

Sure -- here's some WaterTAP code:
```python
def build_naocl_cost_param_block(blk):

    blk.cost = pyo.Param(
        initialize=0.23,
        doc="NaOCl cost",
        units=pyo.units.USD_2018 / pyo.units.kg,
    )
    blk.purity = pyo.Param(
        mutable=True,
        initialize=0.15,
        doc="NaOCl purity",
        units=pyo.units.dimensionless,
    )

    costing = blk.parent_block()
    costing.defined_flows["NaOCl"] = blk.cost / blk.purity
```
Here costing is an instance of `FlowsheetCostingBlockData`, and this is a build rule for a parameter block on the flowsheet costing block.

As it is, the `register_flow_type` method will read the value of `blk.cost / blk.purity` and create a new fixed variable with that value. Any subsequent update to the mutable parameter `blk.purity` has no effect on the model.

I think the costing package probably was written with the intent of always using constants for defined flow costs. However, WaterTAP as a need, and I think it is a general one, to conduct sensitivity analyses on input cost parameters. This make using either mutable parameters, fixed variables, or even combinations thereof natural.

idaes/core/base/costing_base.py
```diff
@@ -345,22 +345,24 @@ def register_flow_type(self, flow_type, cost):
...
         else:
-            # Create a Var to hold the cost
-            # Units will be different between flows, so have to use scalar Vars
-            fvar = pyo.Var(
-                units=pyo.units.get_units(cost),
-                doc=f"Cost parameter for {flow_type} flow",
-            )
-            self.add_component(name, fvar)
-            fvar.fix(cost)
+            add_object_reference(self, name, cost)
```
andrewlee94, line 356: The original intent was that `cost` would be a `float` which would be used to initialize the new `Var` - this change would break that usage as no `Var` would be created (the new `self.name` attribute would just be a pointer to the `float`).

andrewlee94 (changes requested):

andrewlee94:
@bknueven The original intent was that there would be a single fixed `Var` for the cost which could be used for sensitivity studies - it was not envisioned that this would itself be a function of other things. That said, I cannot think of a reason why we could not support this, although it would need to be clearly documented.

However, as noted in my comment, the proposed fix would break the original intended usage, so that would need to be addressed before this could be merged.

bknueven:
> @bknueven The original intent was that there would be a single fixed `Var` for the cost which could be used for sensitivity studies - it was not envisioned that this would itself be a function of other things. That said, I cannot think of a reason why we could not support this, although it would need to be clearly documented.
> 
> However, as noted in my comment, the proposed fix would break the original intended usage, so that would need to be addressed before this could be merged.

Due to the way units are handled in the Pyomo expression system, this was nontrivial. However, I think I have a working solution.

idaes/core/base/costing_base.py
```diff
@@ -345,22 +345,33 @@ def register_flow_type(self, flow_type, cost):
...
+                    doc=f"Cost parameter for {flow_type} flow",
+                )
+                self.add_component(name, fvar)
+                fvar.fix(cost)
+            else:
+                add_object_reference(self, name, cost)
+                if not pyo.is_fixed(cost):
+                    try:
+                        cost.fix()
+                        _log.warn(f"fixing flow cost for flow {flow_type}")
```
andrewlee94, line 370: Typo: Capital f.

idaes/core/base/costing_base.py
```diff
@@ -345,22 +345,33 @@ def register_flow_type(self, flow_type, cost):
...
+                # Create a Var to hold the cost
+                # Units will be different between flows, so have to use scalar Vars
+                fvar = pyo.Var(
+                    units=pyo.units.get_units(cost),
+                    doc=f"Cost parameter for {flow_type} flow",
+                )
+                self.add_component(name, fvar)
+                fvar.fix(cost)
+            else:
+                add_object_reference(self, name, cost)
```
andrewlee94, line 366: I wonder if the is the best way to do this. Notably, this allows users to assign unnamed `expressions` for the cost but I wonder if it would be better to insist on named `Expressions` instead? This would save us from needing to do the `is_constant_unit_to_units` check, and might also be more robust to future changes.

I.e instead of allowing `costing.defined_flows["NaOCl"] = blk.cost / blk.purity` require:

```
blk.NaOCl_cost = pyo.Expression(expr=blk.cost / blk.purity)
costing.defined_flows["NaOCl"] = blk.NaOCl_cost
```

bknueven: We could force the user to prove named `Expressions`, or create one ourselves here.

I'm not sure that sidesteps the need for the `is_constant_up_to_units` check though. Wouldn't we still want to catch the case when we're evaluating the `value` of a non-constant `expression`?

andrewlee94: I would probably just do:

```
if isinstance(cost, [Var, Param, Expression]):
   add_object_reference(self, name, cost)
else:
   # Assume it is a constant and make a Var
```

Maybe add an `elif` to check for an unnamed `expressions` and raise a `TypeError`.

bknueven: > Maybe add an `elif` to check for an unnamed `expressions` and raise a `TypeError`.

Ah, that's exactly the rub though -- `0.2 * pyunits.J` (the original intended usage here) *is* and unnamed `expression`. Hence the gymnastics for evaluating unnamed `expression`s.

andrewlee94: Ah.... I see. Maybe we would still need the `is_constant_up_to_units`.

I am hoping @jsiirola might have some time to take a quick look at this and provide some thoughts.

jsiirola: I am not sure why `register_flow_type` doesn't just put whatever the user handed it into an `Expression` and attach that to the block.  Does it matter if the cost is variable?  (I don't think so)

andrewlee94: @jsiirola If the user were to pass `0.2*units` as the argument however, this would make the cost a constant with no way to change it without changing the `expression` itself. The original API took this and created a `Var` so that users could change this latter for sensitivity studies or parameter estimation (or taking an existing costing package and changing values to suit a new use case).

jsiirola: ...but being able to change the `Expression` *after* constructing expressions is exactly what `Expression` components are for.  They are mutable named placeholders (or references, depending on how you want to look at it) in expressions -- and are the *only* mutable things in the expression system.

andrewlee94: That wouldn't work in a parameter estimation case however for example. Yes, the user could just have provided a `Var` upfront, but that would mean the current API no longer works (and I really don;t want to deprecated anything right now with IDAES 2.0 still not quite done).

andrewlee94:
@bknueven @jsiirola and I just had a phone call to discuss this, and we eventually decided that the best solution (and what we should have done we originally wrote the API) was to take the `cost` argument and put it in a named `Expression` rather than a `Var`.

However, this would change the behaviour of current examples (of which there are a lot) with use the original workflow of providing a constant multiplied by units. The best solution would be to go update all the example to provide `Vars` instead of constants here (and I would suggest we do that as much as possible anyway), but given how many uses of costing we have already, we probably need to maintain backwards compatibility to the old workflow as well.

So, we propose the idea of a `SettableExpressionWithUnits` object that does the following:

1. Separates the units from the expression and stores these separately so that we can maintain unit consistency.
2. If the users does `SEWU.set_value(v)`, we convert `v` to the stored units and then replace `SEWU._expr` with the new value.
3. If a user does `SEWU.fix()` or `SEWU.unfix()` we replace the `Expression` with a `Var` with the correct value and units.

This does create a possible edge case where the `Expression` is a function of other `Vars` or `Params` and a user tries to `fix` or `unfix` the `Expression`, thus separating the `Expression` from the original components. However, seeing as the user had to originally provide the underlying components when creating the `Expression`, one would expect that they would fix or unfix those instead.

@MAZamarripa This discussion has implications for the power plant costing work so it would be good to hear your thoughts on this.

bknueven:
watertap-org/watertap#865 side-steps this issue for WaterTAP -- this PR can be closed as far as WaterTAP is concerned.

andrewlee94:
WaterTap seems to have found a work around for now, and seeing as this turns out to be harder than expected I will close this for now.

PR #1015, opened 2022-11-09
Add dens_star_mol to Helmholtz Property Block

eslickj:
## Fixes

I was missing this property I need in some viscosity or thermal conductivity calculations.  I do have tests for those expressions, but somehow this got through. 

### Legal Acknowledgement

### DIFF
--- a/idaes/models/properties/general_helmholtz/helmholtz_state.py
+++ b/idaes/models/properties/general_helmholtz/helmholtz_state.py
@@ -570,6 +570,10 @@ class HelmholtzStateBlockData(StateBlockData):
         self.dens_mol_crit = pyo.Expression(
             expr=params.dens_mass_crit / params.mw, doc="critical mole density"
         )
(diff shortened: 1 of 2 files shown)


idaes/models/properties/general_helmholtz/helmholtz_state.py
```diff
@@ -570,6 +570,10 @@ def mole_frac_comp_rule(b, i):
         self.dens_mol_crit = pyo.Expression(
             expr=params.dens_mass_crit / params.mw, doc="critical mole density"
         )
+        self.dens_mol_star = pyo.Expression(
```
andrewlee94, line 573: This looks OK, but I think you need to add `dens_mol_star` to the metadata as well.

andrewlee94 (changes requested):

andrewlee94 (approved):

bpaul4 (approved):

PR #1018, opened 2022-11-10
Fix supercritical_steam_cycle PFD output

eslickj:
## Fix 

Fix mix up between file path string and sig string.

### Legal Acknowledgement


bpaul4 (approved):
Thank you for quickly opening the fix, @eslickj.

andrewlee94 (approved):

PR #1019, opened 2022-11-15
Update grid integration

dguittet:
## Fixes
- `idaes/apps/grid_integration/bidder.py`: bidder should not add the original (pmax, marginal_cost) point to each optimized bid curve, since the pmax is purposefully being modified by the bid optimization. Also, for RE with variable energy, the pmax is not fulfillable but is a function of time
- `idaes/apps/grid_integration/model_data.py`: add function that allows modifying any Prescient generator parameter via `**kwargs` input; also update enforcement of pmax in default production cost curve for above

### Legal Acknowledgement


ksbeattie:
@dguittet when is this expected for IDAES?  It's too late for the Nov release, fyi.

adowling2:
@dguittet Is it okay for this to get merged into IDAES in February?

ksbeattie:
I've placed this on the Feb release board with normal priority, please speak up if that's incorrect.

bknueven (approved):
The code looks good.

I do wonder if we need to be doing something fundamentally different (it looks like we're adding another epicycle here), but that should be saved for a subsequent discussion and PR.

Xinhe-Chen (approved):
Looks good to me. Thank you, Darice.

andrewlee94:
@dguittet Is this PR required for the February IDAES release? If so, we need to get the issue with test resolved - it looks like the diff cover of the changes is low, so we either need to add more tests or at least have an explanation of why it is OK.

dguittet:
It should be merged. I'll add some tests today. Thanks for checking

Issue #1016, opened 2022-11-09
Helmholtz Available Components Check


eslickj:
I need to modify the why the Helmholtz property package checks if a component is available.  Currently I assume if it has transport properties on the IDAES side it should be available.  That makes is hard to add new components without modifying IDAES code, especially if you don't care about transport properties. 

I'm not sure how to handle this, but I need to figure it out soon.

andrewlee94:
My first thought is that this sounds like there should be a register of available components, which we populate with the standard ones by default and users can add new one if they have them. Maybe a case for the DMF.

eslickj:
@andrewlee94, yeah the DMF could be a good place to handle this.  That would give a place to put the data files too.  I guess I can add a registry system, and we can take more time to consider the DMF and maybe improving the way we deal with the data files.

PR #1021, opened 2022-11-15
Add a way to register new helmholtz components

eslickj:
## Fixes Issue #1016

This lets you register new components.  Right now if you get new data files, you need to add them to the directory with the rest of the Helmholtz data files that get installed with the binaries, but when I get a chance, I want to let you specify alternate directories and maybe alternate names when you register components.  Also would be good to look into using the DMF, but I need a quick fix for now.

This also lets you have thermo properties for a component without implementing transport properties.

I have tested this with other components that we are not currently making available and don't have transport properties and it works.  I know I need to rework the Helmholtz docs, so I'll do that in another new PR ASAP. 

-
-

### Legal Acknowledgement

### DIFF
--- a/idaes/models/properties/general_helmholtz/components/__init__.py
+++ b/idaes/models/properties/general_helmholtz/components/__init__.py
@@ -16,13 +16,13 @@ some properties not implimented as external functions.
 
 __author__ = "John Eslick"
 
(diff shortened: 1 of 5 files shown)


andrewlee94 (approved):
LGTM

dallan-keylogic (approved):
LGTM

PR #1023, opened 2022-11-17
Adding scaling expression walker

andrewlee94:
## Part of improving scaling tools

**Note:** This PR bumps the Pyomo tag as it requires some improvements to the Pyomo scaling transformation.

## Summary/Motivation:
This PR adds a new expression walker that will extract the scaling factors in the expression and return a list of aggregate scaling factors for each additive term. This will allow us to automatically estimate scaling in expressions and hopefully reduce the need for manual scaling methods.

Scaling of Vars and Params is determined via the associated scaling suffix, and if no suffix value is present for a given component then a value of 1 is assumed. The expression walker can be set to log warnings and/or raise exceptions when missing scaling factors are encountered.

-
-

### Legal Acknowledgement


idaes/core/util/scaling.py
```diff
@@ -1041,3 +1047,165 @@ def set_derivative_factor_from_state(self, deriv, nominal_wrt=1.0):
...
+
+        self.warning = warning
+        self.exception = exception
+
+    def _get_scale_factor_for_sum(self, node, child_scale_factors):
+        sf = []
+        for i in child_scale_factors:
+            for j in i:
+                sf.append(j)
+        return sf
```
dallan-keylogic, line 1090: For a sum of _strictly positive numbers_, a good estimate of the scaling factor of the sum is the harmonic sum of the scaling factors of the numbers (like resistors in parallel). For example, let, a ~ O(1), b ~ O(10), and c ~O(0.1). We have scaling factors s_a = 1, s_b = 0.1, and s_c = 10. The harmonic sum of these scaling factors is 1/(1/1 + 1/0.1 + 1/10) ~= 0.09. It's dominated by the smallest scaling factor (s_b) because the actual sum is dominated by the largest quantity (b). 

Say you then have a ~ O(1), b ~ O(1), and c ~ O(1), and so s_a = 1, s_b =1 , and s_c =1. You'd then have for a scaling factor for the sum 1/(1+1+1) = 1/3.

I'm not sure what exact structure `child_scaling_factor` is supposed to have, nor what structure the `sf` list is supposed to have, so I cannot evaluate this code on that.

idaes/core/util/scaling.py
```diff
@@ -1041,3 +1047,165 @@ def set_derivative_factor_from_state(self, deriv, nominal_wrt=1.0):
...
+        EXPR.SumExpression: _get_scale_factor_for_sum,
+        EXPR.NPV_SumExpression: _get_scale_factor_for_sum,
+        EXPR.ProductExpression: _get_scale_factor_for_product,
+        EXPR.MonomialTermExpression: _get_scale_factor_for_product,
+        EXPR.NPV_ProductExpression: _get_scale_factor_for_product,
+        EXPR.DivisionExpression: _get_scale_factor_for_division,
+        EXPR.NPV_DivisionExpression: _get_scale_factor_for_division,
+        EXPR.PowExpression: _get_scale_factor_for_power,
+        EXPR.NPV_PowExpression: _get_scale_factor_for_power,
+        EXPR.NegationExpression: _get_scale_factor_single_child,
```
dallan-keylogic, line 1160: One of the shortcomings of this system is that both sums and differences are represented equivalently. Sums of strictly positive quantities can be scaled without user intervention, but when negative numbers are involved [catastrophic cancellation](https://en.wikipedia.org/wiki/Catastrophic_cancellation) can occur, which causes problems for both scaling and numerical methods.

An area in which this regularly occurs in IDAES is with temperature. We deal with absolute temperatures from 300 to 1000K, but temperature differences for heat conduction can be on the order of 1 or 0.1 K.

idaes/core/util/scaling.py
```diff
@@ -1041,3 +1047,165 @@ def set_derivative_factor_from_state(self, deriv, nominal_wrt=1.0):
...
+
+    def _get_scale_factor_single_child(self, node, child_scale_factors):
+        assert len(child_scale_factors) == 1
+        return child_scale_factors[0]
+
+    def _get_scale_factor_for_unary_function(self, node, child_scale_factors):
+        assert len(child_scale_factors) == 1
+        func_name = node.getname()
+        func = getattr(math, func_name)
+        return [func(i) for i in child_scale_factors[0]]
```
dallan-keylogic, line 1129: If I'm reading this correctly, this is saying the scaling factor for f(x), s_f(x), is the function f evaluated at the scaling factor of x, f(s_x). This is implicitly assuming that f(s_x*x) = f(s_x)*f(x), which is tantamount to saying that f(x) = x^a for some power a.

This is not going to be true for exponential functions: exp(10) ~= 22000, exp(1/10) ~= 1.11., logarithms (unless you're doing stat mech, output of logarithms is generally going to be O(1)) or trig functions.

idaes/core/util/scaling.py
```diff
@@ -1041,3 +1047,165 @@ def set_derivative_factor_from_state(self, deriv, nominal_wrt=1.0):
...
+    }
+
+    def exitNode(self, node, data):
+        """Callback for :class:`pyomo.core.current.StreamBasedExpressionVisitor`. This
+        method is called when moving back up the tree in a depth first search."""
+
+        # first check if the node is a leaf
+        nodetype = type(node)
+
+        # TODO: What do we do with an indexed component? Fail outright?
```
dallan-keylogic, line 1179: Can indexed components (as opposed to their subcomponents evaluated at particular indices) even appear in expressions? I've wanted to get rid of scaling factors on indexed components for a while now.

idaes/core/util/scaling.py
```diff
@@ -1041,3 +1047,165 @@ def set_derivative_factor_from_state(self, deriv, nominal_wrt=1.0):
...
+
+        if nodetype in native_types or nodetype in pyomo_constant_types:
+            if node == 0:
+                return [1]  # Catch cases of zero and use default scaling of 1
```
dallan-keylogic, line 1183: In line with what I said about summations, positive infinity would be the best scaling factor to assign to zero, because then it wouldn't throw off the harmonic sum of scaling factors (say, if you were summing 0.001 + 0.01 + 0). Not sure whether that would cause problems down the line elsewhere, though.

idaes/core/util/scaling.py
```diff
@@ -1041,3 +1047,165 @@ def set_derivative_factor_from_state(self, deriv, nominal_wrt=1.0):
...
+        assert type(node) is IndexTemplate
+        return [1]
+
+    def _get_scale_factor_external_function(self, node, child_units):
+        # TODO: Work out what to do here
+        # Maybe pass child_units into the external function and get the result?
+        # For now, assume we know nothing about scaling
+        return [1]
+
+    node_type_method_map = {
```
eslickj, line 1147: I think there are a few options for directly evaluating the expression with the scaling values for variables rather than the variables that should make this much much simpler.  Maybe the simplest is to swap in the scaling factor for variables values and set it back when you are done.

dallan-keylogic: Evaluating the expression with the scaling factors will fail quite frequently. If you have the expression a+b+c with a~O(1), b~O(1), and c~O(100), you have s_a = 1, s_b = 1 and s_c = 0.01. Evaluating the expression, you'd get a scaling factor of 2.01, when the expression is O(100) because c is O(100).

idaes/core/util/scaling.py
andrewlee94: The idea here was just to get the scaling factors for each term - we would then decide on what to do with them later. The aim was to let us support a range of different approaches (maximum, minimum, harmonic sum, etc.).

idaes/core/util/scaling.py
andrewlee94: Again, this is why the walker just returns the scaling factor for each term in the sum so the user can go from there.

idaes/core/util/scaling.py
andrewlee94: This should not matter in most cases, a Pyomo will eliminate the 0 term anyway - however we need to have a real number as the scaling factor to avoid evaluation errors elsewhere. There is also a potential case of 0 actually being `False`, in which case `1` is probably the best scaling factor.

idaes/core/util/scaling.py
eslickj: Not directly evaluating the original expression, evaluating the subexpressions.

idaes/core/util/scaling.py
andrewlee94: No, that is not quite what this is saying. As written, `sf_x` is effectively the expected magnitude of `x` and thus the expected magnitude of `f(x) ~= f(sf_x)`. If we go and apply the variable scaling factor, we get:

```
x_prime = x*sf_x
expr_prime =f(x_prime/sf_x)
```

I.e. the variable scaling factor gets cancelled out in the end, but the magnitude of the function should be `O(f(sf_x))`.

idaes/core/util/scaling.py
andrewlee94: As @eslickj  said, the result is `[1, 1, 0.01]`, not `2.01`. The aim here is to get the magnitude of all the additive terms separately, so that we can then work out what the best overall scaling would be,

idaes/core/util/scaling.py
dallan-keylogic: > As written, `sf_x` is effectively the expected magnitude of `x`

For what it's worth, I intended `sf_x` to be the _inverse_ of the expected magnitude of `x`, so that `sf_x * x = O(1)`. It's my understanding that's what IDAES scaling factors are, and if they're not that the documentation needs to be changed.

 That doesn't affect the larger point, though. I'm challenging that `f(x) ~= f(sf_x)` (or `f(1/sf_x)` if we used my convention) in the first place. That's true for powers, but not true for many other classes of function. Exponential functions send nearby values to different orders of magnitude. Logarithms take everything to O(1) (except the sort of inconceivably huge numbers you encounter in Stat Mech). Sine and cosine take things to O(1) without qualification, except you might get unlucky where you evaluate it and get zero out instead.

andrewlee94: That is true, but we have no way of dealing with that - the only information we have is what the user gave us as the variable scaling factor. Note that this tool is intended to be used to fill in information with the best guesses available based on the variable scaling factors - it is not intended to be a magic bullet. Also, user and model developers can (and should) provide their own constraint scaling where possible - this is just to help us fill in the gaps automatically.

I will also note that sin and cos do not always give O(1) - they have a _maximum_ magnitude of O(1) but can be anything smaller than that too.

idaes/core/util/scaling.py
eslickj: I think just to avoid confusion we should shift to expected magnitude for user-facing stuff.  Scale factors are still a little ambiguous.

dallan-keylogic: They're called scaling _factors_, not scaling _divisors_.

We want to be sure that whatever system we decide on, we're consistent between `set_scaling_factor` and `constraint_scaling_transform`

idaes/core/util/scaling.py
```diff
@@ -1041,3 +1047,292 @@ def set_derivative_factor_from_state(self, deriv, nominal_wrt=1.0):
...
+
+def set_constraint_scaling_max_magnitude(
+    component, warning: bool = True, overwrite: bool = False, descend_into: bool = True
+):
+    """
+    Set scaling factors for constraints using maximum expected magnitude of additive terms in expression.
+    Scaling factor for constraints will be 1 / max(abs(nominal value)).
+
+    Args:
+        component: a Pyomo component to set constraint scaling factors for.
```
dallan-keylogic, line 1232: This section flips back and forth between "component" and "constraint". We'd generally prefer `Constraint` scaling to happen with `constraint_scaling_transform` rather than `set_scaling_factor`, because `constraint_scaling_transform` also affects the size of the primal residual when IPOPT evaluates how close to satisfied a constraint is, while `set_scaling_factor` affects only the matrix inversion.

andrewlee94: No, this was built explicitly as the first step to replace `constraint_scaling_transform` and to transposition instead to using the Pyomo scaling transformation (and to not pass any suffixes to IPOPT).

Also, the use of component and constraint is deliberate- the method can take Pyomo Blocks, in which case it will scale all the constraints within the Block.

eslickj (approved):

dallan-keylogic (approved):

idaes/core/base/process_base.py
```diff
@@ -115,6 +115,79 @@ def build(self):
...
+
+    def set_default_scaling(self, attribute, value, index=None):
+        """Set a default scaling factor for a property.
+
+        Args:
+            attribute: property attribute name
+            value: default scaling factor
+            index: for indexed properties, if this is not provided the scaling
+                factor default applies to all indexed elements where specific
+                indexes are no specifically specified.
```
dallan-keylogic, line 148: "scaling factor default" -> default scaling factor and "no" -> "not"

idaes/core/util/scaling.py
eslickj: Yeah, I know we multiply by scaling factor (I have seen them used to divide though and I agree it's counter intuitive).  It doesn't necessarily change anything for this PR, but for a final user interface would be to allow the end user to provide expected magnitude as a way to set the scaling factor.  Like we could have `set_scaling_factor` and/or `set_expected_magnitude()`.  It makes very little difference, but magnitude seems more clear and easier to thing about.

dallan-keylogic: I guess the main thing is to make sure the name is consistent with how we're using them. We can support one or both methods to set them.

PR #1027, opened 2022-11-18
Add in costing nonmethod

dallan-keylogic:
## Summary/Motivation:
Presently, [in a lot of the costing related to the SOFC-IES work](https://github.com/IDAES/examples-pse/blob/63dd7143d0dde9c8e345971b51b588f582d4ef94/src/Examples/Flowsheets/power_generation/ngcc/ngcc_soec_costing.py#L787), we're required to create a costing block on a model because other methods are looking for one, but then the cost produced by the costing method is superseded by one we specifically create. For example:

```
    m.fs.soec.soec_module.diameter = pyo.Var(initialize=0, units=pyunits.m)
    m.fs.soec.soec_module.length = pyo.Var(initialize=0, units=pyunits.m)
    m.fs.soec.soec_module.diameter.fix(0.1)
    m.fs.soec.soec_module.length.fix(0.1)
    m.fs.soec.soec_module.costing = UnitModelCostingBlock(
        flowsheet_costing_block=m.fs.costing,
        costing_method=SSLWCostingData.cost_vessel,
        costing_method_arguments={
            "include_platforms_ladders": False,
        },)

    @m.fs.soec.soec_module.costing.Expression()
    def total_plant_cost(c):
        extra_installed_area = 1.10  # accounts for cell degradation
        return pyunits.convert(60.87 * m.fs.soec.soec_module.number_cells *
                               extra_installed_area * pyunits.USD_2018,
                               CE_index_units)
```
There, `total_plant_cost` is the only thing ultimately used for capital cost estimation. 

As a result, I've added in a costing method that does nothing so we don't have to create these unnecessary variables and constraints.

### Legal Acknowledgement

### DIFF
--- a/idaes/models/costing/SSLW.py
+++ b/idaes/models/costing/SSLW.py
@@ -19,7 +19,7 @@ Costing package based on methods from:
     Chapter 22. Cost Accounting and Capital Cost Estimation
     22.2 Cost Indexes and Capital Investment
 
-Curently, this costing package only includes methods for capital costing of
+Currently, this costing package only includes methods for capital costing of
 unit operations.
 """
 import pyomo.environ as pyo
@@ -238,12 +238,18 @@ class SSLWCostingData(FlowsheetCostingBlockData):
     @staticmethod
     def initialize_build(self):
         """
-        Here we can add intialization steps for the things we built in
+        Here we can add initialization steps for the things we built in
         build_process_costs.
 
         Note that the aggregate costs will be initialized by the framework.
         """
         # TODO: For now,  no additional process level costs to initialize
+        pass
+
+    def cost_none(blk, integer=True):
+        """
+        Placeholder costing method for those who want to write their own costing method.
+        """
         pass
 
     def cost_heat_exchanger(
--- a/docs/reference_guides/model_libraries/power_generation/costing/power_plant_costing_netl.rst
+++ b/docs/reference_guides/model_libraries/power_generation/costing/power_plant_costing_netl.rst
@@ -244,11 +244,11 @@ Below is an example of how to add cost correlations to a flowsheet including a h
         flow_pattern=HeatExchangerFlowPattern.countercurrent)
     # set inputs
     m.fs.unit.shell_inlet.flow_mol[0].fix(100*pyunits.mol/pyunits.s)
-    m.fs.unit.shell_inlet.enth_mol[0].fix(3500*pyunits.mol/pyunits.s)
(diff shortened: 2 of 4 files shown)


idaes/models/costing/SSLW.py
```diff
@@ -238,14 +238,20 @@ def build_process_costs(self):
...
-        Here we can add intialization steps for the things we built in
+        Here we can add initialization steps for the things we built in
         build_process_costs.
 
         Note that the aggregate costs will be initialized by the framework.
         """
         # TODO: For now,  no additional process level costs to initialize
         pass
 
+    def cost_none(blk, integer=True):
```
andrewlee94, line 249: Is this method actually required? When we wrote the new API, we explicitly allowed for users to provide their own methods during costing - i.e. when registering costing for a unit model, the method used does not need to come from the same library as the costing block, and users can (and should) provide their own methods where applicable.

Thus, this sounds like a case where the users should be doing the following:

```
model.fs.unit.costing = UnitModelCostingBlock(
        flowsheet_costing_block=model.fs.costing,
        costing_method=my_costing_method,
        costing_method_arguments={**my_args},
    )
```

andrewlee94 (changes requested):

idaes/models/costing/SSLW.py
bpaul4: @andrewlee94 The method is not required to use the costing framework. However, as currently implemented the model must build a group of unnecessary variables and constraints in order to define the `UnitModelCostingBlock`, such as in the code snippet in the PR summary. Adding a method that does nothing allows users to add a total plant cost constraint without fully defining the SSLW equipment cost arguments or having to write their own "empty" costing method.

andrewlee94: Using the example above, it would look something like:

```
def my_costing_method(blk, *args):
    @blk.Expression()
    def total_plant_cost(blk):
        extra_installed_area = 1.10  # accounts for cell degradation
        return pyunits.convert(60.87 * blk.unit_model.soec_module.number_cells *
                               extra_installed_area * pyunits.USD_2018,
                               CE_index_units)
```

dallan-keylogic: @andrewlee94 Would `my_costing_method` need to be attached to a `FlowsheetCostingBlock` class, or could it just exist as a regular function?

andrewlee94: It can be a stand-alone function, i.e. you can declare it in the flowsheet module immediately before using it.

andrewlee94:
@dallan-keylogic Is this PR still required, or can it be closed?

PR #1036, opened 2022-11-29
Add smooth LMTD approximation

eslickj:
## Summary/Motivation:

Add the smooth LMTD approximation from

Kazi, S., M. Short, A.J. Isafiade,
    L.T. Biegler. "Heat exchanger network synthesus with detailed exchanger
    designs - 2. Hybrid optimization strategy for synthesis of heat exchanger
    networks". AIChE Journal, 2020.

This handles the case where the temperature difference at both ends of the heat exchanger is the same, and it doesn't deviate much from the standard LMTD calculation.  It's potentially more robust even when you don't expect there to be an issue with LMTD since you may evaluate LMTD near the singularity while solving.

This only adds a delta T callback, test and docs.  It doesn't change anything else, so at the very least it won't break anything. 

### Legal Acknowledgement


andrewlee94 (approved):
Looks good, pending tests passing.

bpaul4 (approved):
Everything looks reasonable.

PR #1042, opened 2022-12-01
Helmholtz fix

dallan-keylogic:
Presently the Helmholtz EoS doesn't correctly hold the state when the property block is initialized. This PR fixes that and adds a small test to ensure that states are getting held and released appropriately.

Also lumped in is an option to specify a guess for the heat duty in the `HeatExchanger1D` like the `HeatExchanger0D`. Presently, I have it specified like the `HeatExchanger0D` as a `(value, units)` tuple, but is there a reason why we're doing things that way rather than just passing in a `value * units` object?

### Legal Acknowledgement


eslickj (approved):

bpaul4 (approved):
Looks good, pending the tests.

andrewlee94:
@dallan-keylogic Regarding the use of (value, unit) tuples, it is probably because the HX model was done in the early stages of the development of the units infrastructure and that this just hasn't been updated. For now, we should retain consistency, but we can look at updating that in the revised initialization APIs.

PR #1047, opened 2022-12-08
Add MomentumBalanceType config for Separator Unit

### LINKED ISSUE #1046: Separator unit model doesn't support MomentumBalanceType
I would like to use a separator unit model to split a solids stream into multiple solids streams.  Looking through the separator unit model code, this looks like it should be possible but when I try it in my model it doesn't work.  I get a message that says my properties package must support pressure, which my solids property package does not.  It looks like the separator unit model doesn't support the MomentumBalanceType property like other unit models do (HeatExchanger for instance).

luohezhiming:
## Summary/Motivation:
Add `MomentumBalanceType` config and it can be set as `None`

- Add `MomentumBalanceType`
- Revise corresponding test files

### Legal Acknowledgement


idaes/models/unit_models/separator.py
```diff
@@ -209,6 +210,22 @@ class SeparatorData(UnitModelBlockData):
...
+        "momentum_balance_type",
+        ConfigValue(
+            default=MomentumBalanceType.pressureTotal,
+            domain=In(MomentumBalanceType),
+            description="Momentum balance construction flag",
+            doc="""Indicates what type of momentum balance should be constructed,
+    **default** - MomentumBalanceType.pressureTotal.
+    **Valid values:** {
+    **MomentumBalanceType.none** - exclude momentum balances,
+    **MomentumBalanceType.pressureTotal** - single pressure balance for material,
```
andrewlee94, line 223: This should be updated to reflect what the Separator does - i.e. if momentum balance is none then no balance is written, otherwise pressure is equal in all outlets.

idaes/models/unit_models/tests/test_separator.py
```diff
@@ -664,6 +665,49 @@ def test_add_material_splitting_constraints_none_phase_component(self, build):
 
         assert not hasattr(build.fs.sep, "material_splitting_eqn")
 
+    @pytest.mark.unit
+    def test_add_momentum_splitting_constraints_none_total(self, build):
+        build.fs.sep.config.momentum_balance_type = MomentumBalanceType.none
+
+        build.fs.sep.add_split_fractions(build.outlet_list, build.fs.sep.mixed_state)
+
+        build.fs.sep.add_material_splitting_constraints(build.fs.sep.mixed_state)
```
andrewlee94, line 674: This should be calling `add_momentum_splitting_constraints`. Same applies below.

However, to make this even more thorough (and less repetitive), I suggest using a parameterized test:

```
@pytest.mark.parametrize("balance_type", MomemtumBalanceType)
@pytest.mark.parametrize("split_type", SplittingType)
@pytest.mark.unit
def test_momentum_splitting(self, build, balance_type, split_type):
    build.fs.sep.config.split_basis = split_type
    build.fs.sep.config.momentum_balance_type = balance_type

    build.fs.sep.add_split_fractions(build.outlet_list, build.fs.sep.mixed_state)
    build.fs.sep.add_momentum_splitting_constraints(build.fs.sep.mixed_state)

    if balance_type is MomentumBalanceType.none:
        assert not hasattr(build.fs.sep, "momentum_splitting_eqn")
    else:
        assert isinstance(build.fs.sep.momentum_splitting_eqn, Constraint)
```

This will cover all combinations in a single test and can replace all of your new tests and the existing one in one neat block of code.

andrewlee94 (changes requested):
Looking good - only three requests (two as comments).

The third request is to also update the Separator documentation to describe this behaviour.

andrewlee94:
@luohezhiming You also need to run `black` on the code (the main tests won;t run until that test passes).

idaes/models/unit_models/separator.py
luohezhiming: I added a comment (pressure in all outlets is equivalent) for each option except `none`.

idaes/models/unit_models/tests/test_separator.py
luohezhiming: Rewrote accordingly

idaes/models/unit_models/separator.py
```diff
@@ -209,6 +210,22 @@ class SeparatorData(UnitModelBlockData):
...
+        "momentum_balance_type",
+        ConfigValue(
+            default=MomentumBalanceType.pressureTotal,
+            domain=In(MomentumBalanceType),
+            description="Momentum balance construction flag",
+            doc="""Indicates what type of momentum balance should be constructed,
+    **default** - MomentumBalanceType.pressureTotal.
+    **Valid values:** {
+    **MomentumBalanceType.none** - exclude momentum balances,
+    **MomentumBalanceType.pressureTotal** - single pressure balance for material (pressure in all outlets is equivalent),
```
andrewlee94, line 223: I think we should just make these say "pressure in all outlets is equal". It might be even more correct to make anything but `none` and `pressureTotal` unsupported for now.

luohezhiming: I've updated. Since I click request your review again, the other reviewers disappeared, so please help me to add them again, thank you!

luohezhiming:
> Looking good - only three requests (two as comments).
> 
> The third request is to also update the Separator documentation to describe this behaviour.

I've updated the documentation.

andrewlee94 (approved):

bpaul4 (approved):

PR #1052, opened 2022-12-15
Add 2021 CEPCI Value

bpaul4:
- Adding `USD_2021` units to the IDAES currency dictionary, as these units will be required for applications in other subtasks.

### Legal Acknowledgement


andrewlee94 (changes requested):
1. Just to check, can you provide a reference for this value?
2. You need to update the tests to cover this new entry.

bpaul4:
1. I double checked the source, and updated it accordingly to use the full year aggregate and not just a partial year estimate. I added the source to the docstring.
2. I added the new entry to the test file.

andrewlee94 (approved):

anujad95 (approved):

PR #1055, opened 2022-12-16
Add global deprecation warning to 1.13 series to notify users before 2.0.0 is released

### LINKED ISSUE #1045: Add global deprecation warnings for IDAES v1
The idea is to create a patch release on the 1.x series (i.e. 1.13.1 or 1.13.2) notifying users that IDAES v1 is/will be soon unsupported, and that they should migrate to IDAES v2. This is because, until idaes-pse 2.0.0 is released, running the default command `pip install idaes-pse` will install the latest stable release.

Possible implementations include:

- [ ] Add a message in `setup.py` so that the deprecation warning is emitted
- [ ] Add the deprecation message to `idaes/__init__.py`, so that it's emitted whenever `idaes` is imported
  - If we do this, we'd want to have a non-nagging way for users to opt out of the warning if they intend to continue using the unsupported v1 release for their work, e.g. by creating a sentinel file in `~/.idaes`

lbianchi-lbl:
## Summary/Motivation:

- Until 2.0.0 (i.e. final) is released, running `pip install idaes-pse` will result in the latest stable release on the 1.x series (currently 1.13.0) to be installed
- This PR adds a deprecation message that is emitted whenever the root `idaes` module is imported (i.e. once per process)
- If users wish to continue using the unsupported v1 series without the message, they are provided instructions to install 1.13.0 (which is functionally identical, but without the warning)

- Add deprecation warning to be displayed whenever `idaes` is imported (once per process)

### Legal Acknowledgement

### DIFF
--- a/idaes/__init__.py
+++ b/idaes/__init__.py
@@ -25,6 +25,23 @@ from .ver import __version__  # noqa
 
 _log = logging.getLogger(__name__)
 
+
+_log.warning(
+f"""The v1 series of IDAES (including this version, {__version__}) is no longer supported.
+
+IDAES users should migrate their code to IDAES v2, which will become the default
+with the upcoming 2.0.0 final release.
+
+For more information, including step-by-step guides on how to migrate code to IDAES v2,
+visit https://github.com/IDAES/idaes-pse/wiki/idaes-v2.
+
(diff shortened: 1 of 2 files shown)


ksbeattie (approved):
LGTM

Issue #841, opened 2022-05-04
Pysmo fails on ARM64


eslickj:
On both macOS and all linux distress I tried with ARM64 processors Pysmo tests fails.  I can't really see a clear reason by the test output is below.

```
=================================== FAILURES ===================================
______________ TestPysmoSurrogate.test_evaluate_unisurrogate_poly ______________

self = <idaes.core.surrogate.tests.test_pysmo_surrogate.TestPysmoSurrogate object at 0xffff71c686a0>
pysmo_surr1 = <idaes.core.surrogate.pysmo_surrogate.PysmoSurrogate object at 0xffff8abae430>

    @pytest.mark.unit
    def test_evaluate_unisurrogate_poly(self, pysmo_surr1):
        # Test ``evaluate_surrogate`` for one output with interaction terms
        x = [
            -2,
            -1.8,
            -1.6,
            -1.4,
            -1.2,
            -1.0,
            -0.8,
            -0.6,
            -0.4,
            -0.2,
            0,
            0.2,
            0.4,
            0.6,
            0.8,
            1.0,
            1.2,
            1.4,
            1.6,
            1.8,
            2.0,
        ]
    
        inputs = np.array([np.tile(x, len(x)), np.repeat(x, len(x))])
        inputs = pd.DataFrame(inputs.transpose(), columns=["x1", "x2"])
    
        out = pysmo_surr1.evaluate_surrogate(inputs)
        for i in range(inputs.shape[0]):
>           assert pytest.approx(out["z1"][i], rel=1e-8) == (
                -75.26111111111476
                - 8.815277777775934 * inputs["x1"][i]
                + 18.81527777777826 * inputs["x2"][i]
                - 2.2556956302821618e-13 * (inputs["x2"][i] * inputs["x1"][i])
            )
E           assert -38.372222222227634 ± 3.8e-07 == -95.26111111112031
E             comparison failed
E             Obtained: -95.26111111112031
E             Expected: -38.372222222227634 ± 3.8e-07

core/surrogate/tests/test_pysmo_surrogate.py:1357: AssertionError
---------------------------- Captured stdout setup -----------------------------

===========================Polynomial Regression===============================================

Warning: solution.pickle already exists; previous file will be overwritten.

No iterations will be run.
Default parameter estimation method is used.
Parameter estimation method:  pyomo 

max_fraction_training_samples set at  0.5
Number of adaptive samples (no_adaptive_samples) set at  4
[...]

PR #1070, opened 2023-01-21
Add option for ALAMO model as expression

eslickj:
This adds an option so that when you build a Pyomo model from an ALAMO surrogate, you have the option to build it as expressions rather than constraints.

The surrogate tests are also very delicate.  The Pysmo tests included tests that had multiple equally good results and a lot of tests for both Pysomo and ALAMO depend on comparing floating point numbers as strings.  I have a few Pysmo tests for save/load pickled results that still fail for me, but if the CI tests pass, we can investigate that later. 

This may partly address issue #841

### Legal Acknowledgement


idaes/core/surrogate/alamopy.py
```diff
@@ -1226,25 +1226,43 @@ def evaluate_surrogate(self, inputs):
...
         """
-        Method to populate a Pyomo Block with surrogate model constraints.
+        Method to populate a Pyomo Block with surrogate model constraints
+        or expressions.
 
         Args:
             block: Pyomo Block component to be populated with constraints.
-            additional_options: None
-               No additional options are required for this surrogate object
+            additional_options (dict): as_expression (bool) is an optional
```
andrewlee94, line 1234: 1. Doc string does not match argument.
2. Why make this a dict of options rather than a list of separate arguments?

andrewlee94 (changes requested):
A couple of comments on the code, plus a few more general things:

1. Can you comment on the usage of the as_expressions option, and how that would actually get used in practice?
2. The new options need documentation, and probably need to comment on how they should be used.

idaes/core/surrogate/alamopy.py
eslickj: I didn't write it, it was already there and already a dict.  I think the dict comes from keyword args to ``build_model`` in the general class.  This isn't something users use directly.  I'll fix the doc string, but I'm not really sure how to properly document it.

I suppose the right way would be to add it to the build_model doc string, but it only works with ALAMO for now.

andrewlee94: Is this method called indirectly via another method on the actual SurrogateModel block (i.e. `build_model`)?

andrewlee94 (approved):
Looks OK to me, pending getting the Pysmo tests fixed.

eslickj:
Ok, I updated the tests for Pysmo saves to save then load to a new object and compare the two.  I think depending on strings to compare floating point results should be okay if you always stick to the same platform, but it's pretty delicate. 

I added a data point to the Pysmo tests, which should make the solution unique now.  I think that should fix #841.

idaes/core/surrogate/tests/test_pysmo_surrogate.py
```diff
@@ -1373,9 +1373,13 @@ def test_populate_block_unisurrogate_poly(self, pysmo_surr1):
...
         assert isinstance(blk.pysmo_constraint, Constraint)
         assert len(blk.pysmo_constraint) == 1
-        assert str(blk.pysmo_constraint["z1"].body) == (
-            "outputs[z1] - (-75.26111111111476 - 8.815277777775934*inputs[x1] + 18.81527777777826*inputs[x2] - 2.2556956302821618e-13*(inputs[x2]*inputs[x1]))"
-        )
+        cstr = str(blk.pysmo_constraint["z1"].body).replace("outputs[z1]", "10")
+        cstr = cstr.replace("inputs[x1]", "1")
+        cstr = cstr.replace("inputs[x2]", "5")
+        assert eval(cstr) == pytest.approx(0, abs=1e-8)
+        # assert str(blk.pysmo_constraint["z1"].body) == (
```
andrewlee94, line 1380: Is this commented code still necessary?

eslickj: No, I just forgot to delete the old code.  Will remove.

idaes/core/surrogate/tests/test_pysmo_surrogate.py
```diff
@@ -1373,9 +1371,10 @@ def test_populate_block_unisurrogate_poly(self, pysmo_surr1):
         assert blk.outputs["z1"].bounds == (None, None)
         assert isinstance(blk.pysmo_constraint, Constraint)
         assert len(blk.pysmo_constraint) == 1
-        assert str(blk.pysmo_constraint["z1"].body) == (
-            "outputs[z1] - (-75.26111111111476 - 8.815277777775934*inputs[x1] + 18.81527777777826*inputs[x2] - 2.2556956302821618e-13*(inputs[x2]*inputs[x1]))"
-        )
+        cstr = str(blk.pysmo_constraint["z1"].body).replace("outputs[z1]", "10")
```
OOAmusat, line 1374: @eslickj If I understand this correctly you're testing the value of the evaluated expression, not the expression itself? The purpose of this test was to ensure that the pyomo block was actually generating the correct/accurate constraint on the block. There is are separate tests for evaluating the surrogate(s).

eslickj: The issue is that the test compares a string, so the floating point number have to be exactly the same.  On arm64 processors at the very least there are slight differences.  The only way I can really think to test this is to parse our the coefficients from the string and compare.  It seemed like a lot of work.

eslickj: I guess this shouldn't be too hard with regular expressions.

OOAmusat: You're right, it probably isn't worth the effort. We can go with that you have.

OOAmusat (approved):
LGTM

PR #1072, opened 2023-01-26
Moving bed with bidirectional spatial discretization

Robbybp:
## Summary/Motivation:
For counter-current flow systems, spatial discretizations of the two phases should have "opposite direction", e.g. forward for the solid phase and backward for the gas. To approximate finite volume methods, these should be "upwind" with respect to direction of flow. This PR adds options to the moving bed unit model to allow users to specify discretizations of the two phases independently.

- Add `gas_transformation_scheme` and `solid_transformation_scheme` config options to the moving bed unit model. These are only supported if `dae.finite_difference` is used for `transformation_method`

### Legal Acknowledgement

### DIFF
--- a/idaes/models_extra/gas_solid_contactors/unit_models/moving_bed.py
+++ b/idaes/models_extra/gas_solid_contactors/unit_models/moving_bed.py
@@ -154,7 +154,10 @@ and to "LAGRANGE-RADAU" for collocation transformation method,
             domain=In([None, "BACKWARD", "FORWARD"]),
             description="Scheme to use for DAE transformation",
             doc="""Scheme to use when transforming length domain of the gas
-phase. See Pyomo documentation for supported schemes,
+phase. If this option is supplied, ``solid_transformation_scheme`` must be
+supplied as well. This cannot be set (other than to ``None``) if the
+``transformation_scheme`` option is set.
+See Pyomo documentation for supported schemes.
 **default** - None.
 **Valid values:** {
 **None** - defaults to the value of the ``transformation_scheme`` option,
@@ -169,7 +172,10 @@ phase. See Pyomo documentation for supported schemes,
             domain=In([None, "BACKWARD", "FORWARD"]),
             description="Scheme to use for DAE transformation",
             doc="""Scheme to use when transforming length domain of the solid
-phase. See Pyomo documentation for supported schemes,
+phase. If this option is supplied, ``gas_transformation_scheme`` must be
+supplied as well. This cannot be set (other than to ``None``) if the
+``transformation_scheme`` option is set.
+See Pyomo documentation for supported schemes,
 **default** - None.
 **Valid values:** {
 **None** - defaults to the value of the ``transformation_scheme`` option,
--- a/idaes/models_extra/gas_solid_contactors/unit_models/moving_bed.py
+++ b/idaes/models_extra/gas_solid_contactors/unit_models/moving_bed.py
@@ -721,13 +721,13 @@ see reaction package for documentation.}""",
                     self,
                     wrt=self.gas_length_domain,
                     nfe=self.config.finite_elements,
-                    scheme=self.config.transformation_scheme,
+                    scheme=self._gas_tr_scheme,
                 )
                 self.discretizer.apply_to(
                     self,
                     wrt=self.solid_length_domain,
                     nfe=self.config.finite_elements,
-                    scheme=self.config.transformation_scheme,
+                    scheme=self._solid_tr_scheme,
                 )
             else:
                 self.discretizer.apply_to(
--- a/idaes/models_extra/gas_solid_contactors/unit_models/moving_bed.py
+++ b/idaes/models_extra/gas_solid_contactors/unit_models/moving_bed.py
@@ -136,8 +136,10 @@ by the Pyomo TransformationFactory,
             default=None,
             domain=In([None, "BACKWARD", "FORWARD", "LAGRANGE-RADAU"]),
             description="Scheme to use for DAE transformation",
-            doc="""Scheme to use when transforming domain. See Pyomo
-documentation for supported schemes,
+            doc="""Scheme to use when transforming domain. If specified,
+this scheme is applied to discretizations in both the gas and solid phases.
+In this case, ``gas_transformation_scheme`` and ``solid_transformation_scheme``
+cannot be specified. See Pyomo documentation for supported schemes.
 **default** - None.
 **Valid values:** {
 **None** - defaults to "BACKWARD" for finite difference transformation method,
--- a/idaes/models_extra/gas_solid_contactors/unit_models/moving_bed.py
+++ b/idaes/models_extra/gas_solid_contactors/unit_models/moving_bed.py
@@ -410,21 +410,17 @@ see reaction package for documentation.}""",
                 "dae.collocation"
                 ".".format(self.name)
             )
-        elif (
-            self.config.transformation_method != "dae.finite_difference"
-            and (
-                self.config.gas_transformation_scheme is not None
-                or self.config.solid_transformation_scheme is not None
-            )
+        elif self.config.transformation_method != "dae.finite_difference" and (
+            self.config.gas_transformation_scheme is not None
+            or self.config.solid_transformation_scheme is not None
         ):
             raise ConfigurationError(
                 "Invalid configuration of {}. transformation_method must be"
-                " \"dae.finite_difference\" if gas_transformation_scheme or"
+                ' "dae.finite_difference" if gas_transformation_scheme or'
                 " solid_transformation_scheme are set".format(self.name)
             )
-        elif (
-            (self.config.gas_transformation_scheme is None)
-            != (self.config.solid_transformation_scheme is None)
+        elif (self.config.gas_transformation_scheme is None) != (
+            self.config.solid_transformation_scheme is None
         ):
             raise ConfigurationError(
                 "Invalid configuration of {}. Either both"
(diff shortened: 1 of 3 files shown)


bpaul4 (commented):
This looks very nice, and well commented with useful exceptions and error messages.

One suggestion, would it be more convenient for users to set a `bed_configuration=[COCURRENT, COUNTERCURRENT]` argument? This would be more intuitive and if users don't pass discretization flags the script could automatically set the discretization to `gas_transformation_scheme=BACKWARD` and `solid_transformation_scheme=FORWARD` for countercurrent beds.

idaes/models_extra/gas_solid_contactors/unit_models/moving_bed.py
```diff
@@ -143,6 +144,40 @@ class MBRData(UnitModelBlockData):
 and to "LAGRANGE-RADAU" for collocation transformation method,
 **"BACKWARD"** - Use a finite difference transformation method,
 **"FORWARD""** - use a finite difference transformation method,
+**"LAGRANGE-RADAU""** - use a collocation transformation method}""",
+        ),
+    )
+    CONFIG.declare(
+        "gas_transformation_scheme",
+        ConfigValue(
+            default=None,
```
andrewlee94, line 153: Seeing as this defaults to `"BACKWARD"` anyway, I would suggest that you just set the default value as such and remove `None` from the domain (that said, `None` is always a valid value for the domain, even if it is not explicitly declared).

idaes/models_extra/gas_solid_contactors/unit_models/moving_bed.py
```diff
@@ -143,6 +144,40 @@ class MBRData(UnitModelBlockData):
...
+**"BACKWARD"** - Use a finite difference transformation method,
+**"FORWARD""** - use a finite difference transformation method,
+**"LAGRANGE-RADAU""** - use a collocation transformation method}""",
+        ),
+    )
+    CONFIG.declare(
+        "solid_transformation_scheme",
+        ConfigValue(
+            default=None,
+            domain=In([None, "BACKWARD", "FORWARD"]),
```
andrewlee94, line 171: As above.

idaes/models_extra/gas_solid_contactors/unit_models/moving_bed.py
```diff
@@ -433,12 +498,49 @@ def build(self):
...
+        # transformation schemes are specified, or neither are specified.
+        if (
+            # solid and gas transformation schemes are specified
+            (self.config.solid_transformation_scheme is not None)
+            and (self.config.gas_transformation_scheme is not None)
+        ):
+            # Create length domain sets for gas and solid phases. This is
+            # necessary so we can apply different discretizations to the
+            # different phases.
+            self._gas_tr_scheme = self.config.gas_transformation_scheme
```
andrewlee94, line 511: Is it necessary to assign these aliases to the model object, as opposed to just using them locally within the code?

idaes/models_extra/gas_solid_contactors/unit_models/moving_bed.py
```diff
@@ -433,12 +498,49 @@ def build(self):
...
+            self.gas_length_domain = ContinuousSet(
+                bounds=(0.0, 1.0),
+                initialize=self.config.length_domain_set,
+                doc="Gas phase normalized length domain",
+            )
+            # Create an instance attribute "length_domain" for convenience.
+            # Arbitrarily, use the solid phase length domain.
+            #
+            # We do not use Reference as it does not work for Sets.
+            super(_BlockData, self).__setattr__(
```
andrewlee94, line 527: Minor point, but we would normally use `add_object_reference` for this (as this is what it does behind the scenes).

idaes/models_extra/gas_solid_contactors/unit_models/moving_bed.py
```diff
@@ -433,12 +498,49 @@ def build(self):
...
-            bounds=(0.0, 1.0),
-            initialize=self.config.length_domain_set,
-            doc="Normalized length domain",
-        )
+        # In this branch, we assume that either both gas and solid
+        # transformation schemes are specified, or neither are specified.
+        if (
+            # solid and gas transformation schemes are specified
+            (self.config.solid_transformation_scheme is not None)
+            and (self.config.gas_transformation_scheme is not None)
```
andrewlee94, line 506: A minor question, but would it be better to just check that the transformation schemes are different here - i.e. the real point is that if they are different then we need separate domains.

However, going a step further, I would suggest we consider just having two domains at all times (regardless of the transformations used). In the current approach, the model structure is different depending on the choices made, which may have unforeseen consequences, so it might be worth just using two domains at all times to avoid this.

idaes/models_extra/gas_solid_contactors/unit_models/tests/test_MB.py
```diff
@@ -1123,3 +1132,335 @@ def test_initialization_error(self, iron_oc):
...
+            gas_phase_config={"property_package": m.fs.gas_properties},
+            solid_phase_config={"property_package": m.fs.solid_properties},
+        )
+        with pytest.raises(
+            ConfigurationError,
+            match=r"Either both.*must be set",
+        ):
+            m.fs.unit2.build()
+
+        # with pytest.raises(
```
andrewlee94, line 1168: Commented code.

andrewlee94 (changes requested):

idaes/models_extra/gas_solid_contactors/unit_models/moving_bed.py
Robbybp: It is necessary as they need to be used in the `_apply_transformation` method

idaes/models_extra/gas_solid_contactors/unit_models/moving_bed.py
Robbybp: My thought was that, if schemes are provided for solid and gas, they should have different length domains. In this way, whether or not we have two length domains depends on whether these arguments were provided, rather than the value of the arguments. I don't feel strongly about this, however.

I would rather not *always* have two length domains, however, as this would break code that relies on the phases having the same length domain (I have written such code myself). I agree that always having two length domains is a cleaner solution, however.

Robbybp:
@bpaul4 
> This looks very nice, and well commented with useful exceptions and error messages.
> 
> One suggestion, would it be more convenient for users to set a `bed_configuration=[COCURRENT, COUNTERCURRENT]` argument? This would be more intuitive and if users don't pass discretization flags the script could automatically set the discretization to `gas_transformation_scheme=BACKWARD` and `solid_transformation_scheme=FORWARD` for countercurrent beds.

There currently is a `flow_type` option that must be set to `"counter_current"`. I think it would make sense to add an option for co-current, however, that is outside the scope of this PR. I also agree that it would be nice to automatically set the discretization schemes based on the "flow type", but would rather not do this here and now as it would be a breaking change. It is also worth noting that, as far as I know/remember, a "backward-backward" discretization (the current default) is fine (at least from a singularity perspective) for a steady state model.

idaes/models_extra/gas_solid_contactors/unit_models/moving_bed.py
Robbybp: This doesn't default to `"BACKWARD"`, it defaults to whatever `transformation_scheme` is. Having `None` as the default is important as it lets us check whether these arguments are set.

andrewlee94: Ah - I see now. However, the doc string mentions Lagrange-Radau, which is not in the `In()` declaration.

Also, now that there are schemes for each side separately, is the declaration above still required (an overall transformation scheme)?

Robbybp: Good catch. Lagrange-Radau is not supported if gas/solid are set individually. This is caught by the `transformation_method != "dae.finite_difference"` check on line 409, which is tested. I just updated the docstring.

`transformation_scheme` is still supported for backward compatibility, convenience, and to support collocation for the length domain (which may be useful for steady state).

idaes/models_extra/gas_solid_contactors/unit_models/moving_bed.py
Robbybp: Alternatively, the same thing could be accomplished by augmenting the call signature of `_apply_transformation`.

andrewlee94 (approved):

bpaul4 (approved):

Robbybp:
I don't think I can make the dev meeting tomorrow, but this is ready to merge, as far as I'm concerned. Can somebody let me know here if it's being held up for any reason (e.g. the failing tests).

andrewlee94:
I would like @aostace01 to take a look at this as someone with more knowledge of the model, but I just noticed that she does not appear to be tagged on this PR.

aostace01 (commented):
Robby, this is a great new feature for the moving bed model; thank you for contributing it. Everything looks good to me.
 
There is one thing that comes to mind though: the [documentation](https://github.com/IDAES/idaes-pse/blob/main/docs/reference_guides/model_libraries/gas_solid_contactors/unit_models/moving_bed.rst) - it should reflect this new option of having separate discretization schemes for the gas and solid phases. A few sentences explaining that having just one scheme for both (that could be useful for steady state studies) remains an option would be helpful, too; maybe as a comment in the code, or documentation, or both - not sure what the consensus is on this.

Robbybp:
@aostace01 I just updated the config docstrings to clarify the exclusivity and dependence of the `transformation_scheme`, `gas_transformation_scheme`, and `solid_transformation_scheme` options. This will appear in the "MBR Class" section of the docs. 

I also added a small "Discretization" section near the top of the doc page, which recommends using these new options for dynamic operation.

aostace01 (approved):
Robby, thank you for updating the config docstrings, and for adding a paragraph to the MBR documentation.

PR #1077, opened 2023-02-02
Mea column improvements

dallan-keylogic:
## Summary/Motivation:
Miscellaneous changes/fixes in order to improve scaling and convergence of column models.

### Legal Acknowledgement


idaes/core/base/property_base.py
```diff
@@ -761,11 +761,22 @@ def calculate_scaling_factors(self):
...
         ):
             if iscale.get_scaling_factor(v) is None:  # don't replace if set
                 name = v.getname().split("[")[0]
                 index = v.index()
                 sf = self.config.parameters.get_default_scaling(name, index)
                 if sf is not None:
                     iscale.set_scaling_factor(v, sf)
+
+        for con in self.component_data_objects(Constraint, descend_into=False):
+            # if con.name == 'stripper_section.stripper.liquid_phase.properties[0.0,0.8].appr_to_true_species[Liq,CO2]':
```
andrewlee94, line 774: This comment needs to be removed.

idaes/core/base/property_base.py
```diff
@@ -761,11 +761,22 @@ def calculate_scaling_factors(self):
...
+            #     import pdb; pdb.set_trace()
+            if iscale.get_constraint_transform_applied_scaling_factor(con) is None:
+                name = con.getname().split("[")[0]
+                index = con.index()
+                sf = self.config.parameters.get_default_scaling(name, index)
+                if sf is not None:
+                    iscale.constraint_scaling_transform(con, sf, overwrite=False)
```
andrewlee94, line 781: This is a fairly significant change to what this method does - we probably need to discuss the implications of this.

idaes/models_extra/column_models/solvent_condenser.py
```diff
@@ -321,11 +321,11 @@ def build(self):
         flow_basis = self.liquid_phase[t_init].get_material_flow_basis()
         if flow_basis == MaterialFlowBasis.molar:
             fb = "flow_mole"
-        elif flow_basis == MaterialFlowBasis.molar:
-            fb = "flow_mass"
+        # elif flow_basis == MaterialFlowBasis.mass:
```
andrewlee94, line 324: Commented code

andrewlee94 (changes requested):
A few comments on things that need to be fixed or discussed a bit more.

idaes/core/base/property_base.py
andrewlee94: Thinking a bit more (and repeating a comment from Slack for posterity) - I think this is the wrong way to go.

IPOPT should be considered a special case, not the general case, in which case we need to set the scaling suffix for the constraints and then let the user decide what to do with them. IPOPT users can use the `constraint_scaling_transformation` function if they wish, but the generic workflow would use the Pyomo scaling transformation (which relies on having the suffixes).

dallan-keylogic: Removing the default status of IPOPT and `constraint_scaling_transform` in the long run is fine, but it's also a complete revolution in how scaling is treated in IDAES models, which make heavy use of `constraint_scaling_transform`. If this change is reverted, I'll just as likely use `constraint_scaling_transform` somewhere else to do the desired scaling: probably in `electrolyte_states.py`.

andrewlee94: It is fine to still use `constraint_scaling_transform`, but it should not be done in the base classes (i.e. anything in `idaes.core.base`). Ideally, we would only do this in the model level `calculate_scaling_factors` method.

The point being this is not something we want baked in deep as it will be harder to change later - ideally we want to limit the use of this to the `calculate_scaling_factors` methods so that when we make the transition, we only need to change those (or if we are lucky, just switch to new methods and delete those as they are no longer needed).

lbianchi-lbl:
This model will eventually (soon) be accompanied by a surrogate model. This PR might or might not be merged in the meantime.

dallan-keylogic:
Superseded by #1608

PR #1079, opened 2023-02-03
New mixed phase property functions for Helmholtz EoS.

eslickj:
## Summary/Motivation:

This changes the mixed phase properties for Helmholtz EoS to directly call external functions for properties as a function of (h, P), (s, p), or (u, p) depending on the state varaible selection.  The goal of this is to prevent calls to functions for properties of phases that don't exist.  While I'm not aware of any issues calling functions for properties in phases that don't exist, this should be an improvement.  It may make things solver faster by reducing the number of external function calls, and potentially it could improve robustness.

The CI tests will fail until I get the new binaries up with the new functions calls.  I did test this myself, and everything seems to work.  It is probably okay to review despite the tests not running.

### Legal Acknowledgement


idaes/models/properties/general_helmholtz/helmholtz_functions.py
```diff
@@ -154,12 +154,165 @@ class AmountBasis(enum.Enum):
...
+    "w_func": {  # speed of sound
         "fname": "w",
         "units": pyo.units.m / pyo.units.s,
         "arg_units": [dimensionless, dimensionless, dimensionless],
         "doc": "w(comp, delta, tau)",
     },
+    "v_func": {  # specific volume
+        "fname": "v",
+        "units": pyo.units.m**3 / pyo.units.kg,
+        "arg_units": [dimensionless, pyo.units.kg / pyo.units.kg, pyo.units.kPa],
```
dallan-keylogic, line 166: Delta and Tau are some normalized (and possibly inverted?) versions of density and temperature, respectively, right? Assuming "arg_units" means, "units with which the function is called", we've got an issue here.

idaes/models/properties/general_helmholtz/helmholtz_functions.py
```diff
@@ -154,12 +154,165 @@ class AmountBasis(enum.Enum):
...
+    "h_sp_func": {  # enthalpy
+        "fname": "h_sp",
+        "units": pyo.units.kJ / pyo.units.kg,
+        "arg_units": [dimensionless, pyo.units.kJ / pyo.units.kg/ pyo.units.K, pyo.units.kPa],
+        "doc": "h(comp, entropy, pressure)",
+    },
+    "g_sp_func": {  # Gibbs free energy
+        "fname": "g_sp",
+        "units": pyo.units.kJ / pyo.units.kg,
+        "arg_units": [dimensionless, pyo.units.kJ / pyo.units.kg/ pyo.units.K, pyo.units.kPa],
```
dallan-keylogic, line 234: These units also look wrong

idaes/models/properties/general_helmholtz/helmholtz_functions.py
```diff
@@ -481,6 +634,31 @@ def _sv_str(**kwargs):
...
+                if self.amount_basis == AmountBasis.MOLE:
+                    s = kwargs["s"] * self.param.uc["J/mol/K to kJ/kg/K"]
+                else:
+                    s = kwargs["s"] * self.param.uc["J/kg/K to kJ/kg/K"]
+                return StateVars.PS, blk, c, s, p
+            if "u" in kwargs and kwargs["u"] is not None:
+                if self.amount_basis == AmountBasis.MOLE:
+                    u = kwargs["u"] * self.param.uc["J/mol to kJ/kg"]
+                else:
+                    u = kwargs["u"] * self.param.uc["J/kg to kJ/kg"]
```
dallan-keylogic, line 658: All these are indented a level beyond what they should, I think

idaes/models/properties/general_helmholtz/helmholtz_functions.py
```diff
@@ -481,6 +634,31 @@ def _sv_str(**kwargs):
     def add_funcs(self, names=None):
         add_helmholtz_external_functions(self.blk, names=names)
 
+    def _state_vars(self, **kwargs):
```
dallan-keylogic, line 637: Could you add a docstring describing what's supposed to be returned here? It's hard to follow the code further on without knowing what "c", "u", etc. are.

idaes/models/properties/general_helmholtz/helmholtz_functions.py
eslickj: Yeah, those units are wrong.  I don't actually use that function anywhere yet so it did cause a problem.  Will fix.

idaes/models/properties/general_helmholtz/helmholtz_functions.py
eslickj: I think these are right.  The args art component (dimensionless), entropy (kJ/kg/K), and pressure kPa.  The Gibbs free energy is kJ/kg.

idaes/models/properties/general_helmholtz/helmholtz_functions.py
eslickj: Yes

idaes/models/properties/general_helmholtz/helmholtz_functions.py
dallan-keylogic: The doc directly below says `"g(comp, delta, tau)"`. Maybe the variables are what are wrong?

eslickj: Oh yeah the doc string is wrong.

idaes/models/properties/general_helmholtz/helmholtz_functions.py
eslickj: I think it's right they are in the "if p is an args."  I will double check everything in the function again though.

dallan-keylogic: So all of those conversions should happen only if pressure is a state variable and also converted?

eslickj: I used to have (s, T), but as it is pressure is always a state variable. Could change in the future if we find a reason.

andrewlee94 (approved):

idaes/models/properties/general_helmholtz/helmholtz_functions.py
```diff
@@ -1438,14 +1768,17 @@ def build(self):
             "J/mol to J/kg": 1.0 / self.mw,
             "kJ/kg to J/mol": (pyo.units.J * 1000 / pyo.units.kJ) * self.mw,
             "J/mol/K to kJ/kg/K": (pyo.units.kJ / 1000 / pyo.units.J) / self.mw,
-            "J/mol/K to J/kg/K": 1.0 / self.mw,
+            "J/mol/K to J/kg/K": 1 / self.mw,
```
dallan-keylogic, line 1771: I seem to recall @andrewlee94 replacing a bunch of integer zeros or ones with floating point zeros or ones somewhere in core. Is this a place where that distinction matters?

idaes/models/properties/general_helmholtz/helmholtz_functions.py
```diff
@@ -563,12 +782,22 @@ def basic_calculations(
 
     def s(self, **kwargs):
         """Mixed phase entropy"""
-        blk, delta_liq, delta_vap, tau, x, c = self.basic_calculations(**kwargs)
-        self.add_funcs(names=["s_func"])
-        s = (
-            blk.s_func(c, delta_liq, tau, _data_dir) * (1 - x)
-            + blk.s_func(c, delta_vap, tau, _data_dir) * x
-        )
+        sv, blk, c, u, p = self._state_vars(**kwargs)
```
dallan-keylogic, line 785: It bugs me that the first state variable is always assigned to `u`, even though it could be any one of internal energy, entropy, or enthalpy. I think confusion around this point was why I gave up reviewing this PR the first time. It's doesn't affect the code's functionality, but it's bad for the maintainability to use confusing naming conventions.

dallan-keylogic (approved):

idaes/models/properties/general_helmholtz/helmholtz_functions.py
eslickj: I'm pretty sure it doesn't matter here since the division will make it a float anyway.

idaes/models/properties/general_helmholtz/helmholtz_functions.py
```diff
@@ -770,12 +1067,34 @@ def x(self, **kwargs):
...
+        if sv == StateVars.PH:
+            self.add_funcs(names=["tau_func"])
+            tau = blk.tau_func(c, u, p, _data_dir)
+        elif sv == StateVars.PU:
+            self.add_funcs(names=["tauu_func"])
+            tau = blk.tauu_func(c, u, p, _data_dir)
+        elif sv == StateVars.PS:
+            self.add_funcs(names=["taus_func"])
+            tau = blk.taus_func(c, u, p, _data_dir)
```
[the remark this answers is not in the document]
eslickj: I in the beginning there was only tau as a function of h and p.  I was just afraid to change it.  Maybe I'll do another PR to clean some things up.  Too many moving parts confuse me.

idaes/models/properties/general_helmholtz/helmholtz_functions.py
andrewlee94: It also doesn't really matter unless you want to look at the expression body itself.

idaes/models/properties/general_helmholtz/helmholtz_functions.py
eslickj: Noted.  I'm planing to follow this up soon that just uses direct property calls for everything without going through delta and tau.  I'll change it then.

PR #1080, opened 2023-02-03
Adding metadata for valid range in property packages

andrewlee94:
## Summary/Motivation:
Per a request from the UQ team, this PR adds a new metadata entry for properties which records the valid range of values for a given property (e.g., based in the range of data used to fit the property correlation). It also adds some utility functions to set bounds based on the valid range and to check for components with values outside the valid range.

This also adds a utility function to call IPOPT with "halt_on_ampl_error" and "symbolic_solver_labels" which was requested by some people and belongs in the model diagnostic tools.

- New `valid_range` metadata property
- Utility function to retrieve valid range from metadata
- Utility function to set bounds based on valid range
- Utility function to list all components with values outside valid range
- Utility function for IPOPT with halt on AMPL errors and symbolic solver labels
- Tests and documentation

### Legal Acknowledgement


idaes/core/util/model_diagnostics.py
```diff
@@ -694,10 +713,152 @@ def find_irreducible_degenerate_sets(self, verbose=True, tee=False):
...
     def print_variable_bounds(v):
-        """Print variable, bounds, and value
+        """
+        Print variable, bounds, and value
+
         Argument:
             v: variable
+
         Return:
             nothing
```
shermanjasonaf, line 723: Consider changing `Argument` to `Arguments` and `Return` to `Returns`? Docstring not rendered as expected in online pages.

idaes/core/util/model_diagnostics.py
```diff
@@ -581,9 +601,8 @@ def underdetermined_variables_and_constraints(self, n_calc=1, tol=0.1, dense=Fal
             dense: If True, use a dense svd to perform singular value analysis,
             which tends to be slower but more reliable than svds
```
shermanjasonaf, line 602: Documentation of the arguments to `underdetermined_variables_and_constraints` doesn't seem to render as expected in the online webpages. Might want to add tab indent to second line of documentation for each of the arguments `ncalc`, `tol`, and `dense`.

For example, tweak
```
            dense: If True, use a dense svd to perform singular value analysis,
            which tends to be slower but more reliable than svds
```
to
```
            dense: If True, use a dense svd to perform singular value analysis,
                which tends to be slower but more reliable than svds
```

idaes/core/util/model_diagnostics.py
```diff
@@ -694,10 +713,152 @@ def find_irreducible_degenerate_sets(self, verbose=True, tee=False):
...
         """
         print(v, "\t\t", v.lb, "\t", v.value, "\t", v.ub)
+
+
+def get_valid_range_of_component(component):
+    """
+    Return the valid range for a component as specified in the model metadata.
+
+    Args:
+        component - Pyomo component to get valid range for
```
shermanjasonaf, line 734: Consider changing to 
```
        component: Pyomo component to get valid range for
```
so that docstring renders as expected in online pages?

idaes/core/util/model_diagnostics.py
```diff
@@ -694,10 +713,152 @@ def find_irreducible_degenerate_sets(self, verbose=True, tee=False):
...
+
+    Returns:
+        valid range for component if found. This will either be a 2-tuple (low, high) or None.
+
+    Raises:
+        AttributeError if metadata object not found
```
shermanjasonaf, line 740: Change to: 
```
        AttributeError: if metadata object not found
```

idaes/core/util/model_diagnostics.py
```diff
@@ -694,10 +713,152 @@ def find_irreducible_degenerate_sets(self, verbose=True, tee=False):
...
+def set_bounds_from_valid_range(component, descend_into=True):
+    """
+    Set bounds on Pyomo components based on valid range recorded in model metadata.
+    WARNING - this function will overwrite any bounds already set on the component/model.
+
+    This function will iterate over component data objects in Blocks and indexed components.
+
+    Args:
+        component - Pyomo component to set bounds on. This can be a Block, Var or Param.
+        descend_into - (optional) Whether to descend into components on child Blocks (default=True)
```
shermanjasonaf, line 776: Change, for example, the documentation for `component` to:
```
        component : Pyomo component to set bounds on. This can be a Block, Var or Param.
```
(i.e. replace the `-` with a `:`). Similarly for `descend_into`.

idaes/core/util/model_diagnostics.py
```diff
@@ -694,10 +713,152 @@ def find_irreducible_degenerate_sets(self, verbose=True, tee=False):
...
+    """
+    Return a list of component objects with values outside the valid range specified in the model
+    metadata.
+
+    This function will iterate over component data objects in Blocks and indexed components.
+
+    Args:
+        component - Pyomo component to search for component outside of range on.
+            This can be a Block, Var or Param.
+        descend_into - (optional) Whether to descend into components on child Blocks (default=True)
```
shermanjasonaf, line 814: Change argument name/description separators from `-` to `:`.

idaes/core/util/model_diagnostics.py
```diff
@@ -694,10 +713,152 @@ def find_irreducible_degenerate_sets(self, verbose=True, tee=False):
...
+def ipopt_solve_halt_on_error(model, options=None):
+    """
+    Run IPOPT to solve model with debugging output enabled.
+
+    This function calls IPOPT to solve the model provided with settings
+    to halt on AMPL evaluation errors and report these with symbolic names.
+
+    Args:
+        model - Pyomo model to be solved.
+        options - solver options ot be passed to IPOPT
```
shermanjasonaf, line 851: Change arg name/description separators from `-` to `:`

shermanjasonaf (commented):
Functionality looks good. See below for comments on documentation.

idaes/core/util/model_diagnostics.py
```diff
@@ -694,10 +714,152 @@ def find_irreducible_degenerate_sets(self, verbose=True, tee=False):
...
+        for i in component.component_data_objects(
+            ctype=[pyo.Var, pyo.Param], descend_into=descend_into
+        ):
+            comp_list.extend(list_components_with_values_outside_valid_range(i))
+    else:
+        valid_range = get_valid_range_of_component(component)
+
+        if valid_range is not None:
+            cval = pyo.value(component)
+            if cval is not None and (cval < valid_range[0] or cval > valid_range[1]):
```
eslickj, line 837: Do you not consider ``None`` to be outside the valid range?

eslickj (approved):

idaes/core/util/model_diagnostics.py
andrewlee94: I think I did this in case of stale variables (i.e. a case where there was a variable with no initial value that did not appear in any constraints). I would also technically consider `None` to be a different issue (no value versus out of range), but we could include it here if we think it is better.

idaes/core/base/property_set.py
```diff
@@ -102,6 +108,19 @@ def required(self):
         """
         return self._required  # pylint: disable=E1101
 
+    @property
+    def valid_range(self):
+        """
+        Tuple indicating valid range of values for this property based on the data and method used
+        to calculate it. This should take the form of (lower range, upper range) and can be used
+        set and verify bounds on properties before and after solving in order to inform users of any
```
aostace01, line 116: Typo: "to" is missing from "can be used set and verify"

idaes/core/util/model_diagnostics.py
```diff
@@ -11,18 +11,19 @@
 # Please see the files COPYRIGHT.md and LICENSE.md for full copyright and
 # license information.
 #################################################################################
-"""Degeneracy Hunter is a collection of utility functions to assist in mathematical
-modeling in Pyomo.
+"""
+This module contains a collection of tools for diagnosing modelling issues.
```
aostace01, line 15: Consider replacing "modelling" with "modeling" for consistency :)

aostace01 (commented):
Looks good to me; I only have a few minor comments,

idaes/core/base/property_set.py
```diff
@@ -102,6 +108,19 @@ def required(self):
         """
         return self._required  # pylint: disable=E1101
 
+    @property
+    def valid_range(self):
+        """
+        Tuple indicating valid range of values for this property based on the data and method used
```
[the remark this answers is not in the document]
andrewlee94: State variables are properties - they just happen to be the set we chose to define the state.

aostace01 (approved):
Looks good.

idaes/core/util/model_diagnostics.py
eslickj: I guess it depends on the use case.  Maybe there is also another check for not set variables.  Intuitively I'd say None is outside the valid range.  I don't think I'm catching the point about stale.  Do you mean if a variable has a value of None, that you want to ignore it because if you've run the problem and it is still None then it is not used and doesn't matter?

andrewlee94: Yes - I didn't want unused variables cluttering up the output. I was also suggesting there should be a separate check for variables with no value.

PR #1085, opened 2023-02-09
Rework control volume code to remove unneeded terms.

andrewlee94:
## Summary/Motivation:
Pyomo is working on some changes to how expressions are generated, and in the process is changing some behaviours that IDAES has unconsciously relied upon. The most significant change is that `0*` terms will remain within the expression during unit consistency checks whereas before they were ignored. Control volumes (written before units) used this behaviour to drop unnecessary terms, but this will now cause issues when asserting unit consistency.

- Rewrite methods in control volumes which generate balance expressions to only include terms if required (rather than adding a 0 term).
- Comment out two placeholder terms in modular CEOS code for potential future functionality.
- Updated `large_residual_set` method to catch `TypeErrors` that now occur when constraints have a body containing a `None` variable.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/base/control_volume0d.py
+++ b/idaes/core/base/control_volume0d.py
@@ -562,64 +562,52 @@ class ControlVolume0DBlockData(ControlVolumeBlockData):
         if balance_type == MaterialBalanceType.componentPhase:
 
             def user_term_mol(b, t, p, j):
-                if custom_molar_term is not None:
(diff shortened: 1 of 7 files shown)


eslickj (approved):

dallan-keylogic (approved):
Looks good overall.

PR #1086, opened 2023-02-09
Minor updates to install docs

### LINKED ISSUE #1078: Installing Multiple Versions of IDAES
It is sometimes nice to install multiple versions of IDAES.  For example, some models may only work with a particular older version of IDAES.  IDAES provided an environment variable ``IDAES_DATA`` that lets you set a location for the IDAES data directory. Using directories other than the default can allow you to have multiple versions of IDAES with different configurations and different copies of the binary extensions.

If users use conda or venv to manage environments, the ``IDAES_DATA`` environment variable could probably be set by the environment. Users could probably have other setups for this too.  This may be less of an issue if we can install binaries with pip, but I'm not optimistic that we'l have that soon.

I propose we add documentation to the installation instructions about managing multiple model installations.

eslickj:
- Add a note about installing multiple versions of IDAES with  the IDAES_DATA env var
- Update macOS install instructions to indicate that ARM64 binaries are in-fact available
- Bump the binary version to 3.2.0.  The main point of that release is to add property functions, but it also fixes a bug in the Mac binaries.  It should be perfectly backward compatible with 3.1.0, so it should have no effect for most people. 
- Include the PETSc in the install instructions.  I didn't want to complicate things by providing branching install options, and I think it's clear enough that it's optional.  It's probably better people install PETSc when they don't need it that that they complain the things needing PETSc don't work.  It an extra ~ 25 Mb.

### Legal Acknowledgement


docs/tutorials/getting_started/install_templates/idaes_data.md
```diff
@@ -0,0 +1,5 @@
+.. note::
+    If you want to install multiple versions of IDAES in different environments, you can set the ``IDAES_DATA`` environment variable before installing
+    IDAES or using IDAES.  The easiest way to do this is to set the environment variable as part of activating your environment before
```
eslickj, line 3: redundant IDAES

andrewlee94 (approved):

docs/tutorials/getting_started/install_templates/idaes_data.md
```diff
@@ -0,0 +1,5 @@
+.. note::
+    If you want to install multiple versions of IDAES in different environments, you can set the ``IDAES_DATA`` environment variable before installing
+    or using IDAES.  The easiest way to do this is to set the environment variable as part of activating your environment before installing IDAES.  
+    Setting this environment variable will keep binary extensions and configuration files separate for the different IDAES versions.  For more 
+    information about setting environment variables as part of a conda or virtual environment, see the conda or venv documentation.
```
lbianchi-lbl, line 5: I think it'd be useful to add links to the respective pages. I've found the one for Conda on the documentation [here](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#saving-environment-variables), but I wasn't able to find the same for "vanilla" Python venvs.

lbianchi-lbl (commented):
Thanks @eslickj for updating the docs. I've left a comment for a possible enhancement if we think it's worth to add links to how to permanently set environment variables when activating virtual environments.

More in general, I've noticed that the instructions for the two versions of macOS have the same post-installation steps (using a screenshot because the content is merged together from different rst/md files):

image

However, I'm concerned that running the test commands for Intel Macs will result in a number of tests failing. Not sure what's the best way to address this (ideally, we'd know exactly which tests fail and we'd be able to mark and consequently exclude them from running), but we could at least add a note to the tune of:

```rst
.. note:: When running the test command above on Intel Macs, a number of failures will likely occur. This is to be expected as this architecture is **officially unsupported**.
```

eslickj:
@lbianchi-lbl, I'm not sure about the tests.  They were in the docs before.  I think most tests requiring external functions should skip if they are not available.  I know some solver tests will fail on purpose to tell you you don't have certain solvers or capabilities.  My worry is that test may be set up to require the HSL.  I guess I can try to run the tests without the extensions and see what happens.  Unless the solution is just to remove the line that says to run tests, I guess we can follow up with a new issue about testing without extensions.

As for setting environment variables with venv, I guess that since you activate an environment like this ``source <venv>/bin/activate`` you could probably just add a line to the activate script to set environment variables.  I actually typically use a script to active my conda environments, and just add environment variables to that.  I can probably try that out.

eslickj:
I just did a quick test to see what would happen if I ran the tests with no solvers.

```
(e1) john@Johns-Mac-mini idaes-pse % pytest --pyargs idaes -W ignore
================================================================== test session starts ==================================================================
platform darwin -- Python 3.9.6, pytest-7.2.1, pluggy-1.0.0
rootdir: /Users/john/git/idaes-pse, configfile: pytest.ini
plugins: anyio-3.6.2
collected 6612 items / 1 error / 5 deselected / 5 skipped / 6607 selected     
======================================================================== ERRORS =========================================================================
____________________________________ ERROR collecting idaes/apps/uncertainty_propagation/tests/test_uncertainties.py ____________________________________
idaes/apps/uncertainty_propagation/tests/test_uncertainties.py:33: in <module>
    kaug_available = SolverFactory("k_aug").available()
../../envs/e1/lib/python3.9/site-packages/pyomo/opt/base/solvers.py:93: in available
    raise ApplicationError("Solver (%s) not available" % str(self.name))
E   pyomo.common.errors.ApplicationError: Solver (k_aug) not available
================================================================ short test summary info ================================================================
ERROR idaes/apps/uncertainty_propagation/tests/test_uncertainties.py - pyomo.common.errors.ApplicationError: Solver (k_aug) not available
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
====================================================== 5 skipped, 5 deselected, 1 error in 23.32s =======================================================
```

It looks like the missing k_aug caused an error that just stopped the tests from running.

andrewlee94:
@eslickj That test/code looks like it needs to be rewritten to be skipped in that case. Especially seeing as that is code in `apps` and is thus "extension code".

eslickj:
I'm working on adding skips to tests that require external functions right now.  I'll put a separate PR in for that.  Some tests will also fail if you install Ipopt with conda just because some linear solvers aren't available and some problems don't solve with MUMPS.  I think the best idea is just to let those fail and we'll add a note to the docs about it.

eslickj:
I did some more documentation editing.  I also added binaries for macOS on intel, although they do not have the HSL, it is still a big improvement over just getting ipopt with conda, and is much closer to feature complete.

docs/tutorials/getting_started/binaries.rst
```diff
@@ -17,10 +17,13 @@ license for HSL linear solvers that allows inclusion in IDAES's compiled version
 of Ipopt.
 
 All technical papers, sales and publicity material resulting from use of
-the HSL codes within IPOPT must contain the following acknowledgement: HSL, a
+the HSL codes within Ipopt must contain the following acknowledgement: HSL, a
 collection of Fortran codes for large-scale scientific computation. See
 http://www.hsl.rl.ac.uk.
 
+.. note:: Currently the macOS x86_64 binaries to not include HSL, so HSL linear solvers
```
blnicho, line 24: Typo: "**do** not include"

blnicho (approved):
I found one typo but otherwise this looks great. I even learned some new Sphinx syntax!

PR #1091, opened 2023-02-14
Better initial guesses when initializing Control Volumes

andrewlee94:
## Summary/Motivation:
As part of working towards a new initialization API, this PR adds some new methods to estimate initial guesses for state variables in Control Volumes, and deploys this to the existing initialization routines for control volumes. Overall, this makes a very small change to the current initialization schemes, but gives us a method that we can leverage as needed.

- Adds general methods to `control_volume_base` for setting values for state vars on one `StateBlock` based on another `StateBlock` and the associated control volume quantities (e.g. `deltaP`).
- Add `estimate_state` methods to CV0D and CV1D that use the general methods to generate guesses for state variables based on the inlet state.
- Update `CV0D.initialize` and `CV1D.initialize` to use the new methods.

### Legal Acknowledgement


idaes/core/base/control_volume_base.py
```diff
@@ -960,3 +961,81 @@ def _get_representative_property_block(self):
...
+            # Similarly, mass and mole fractions are hard to deal with.
+            # Energy terms also have problems, as they are defined on ain intensive
+            # basis, so we would need to convert heat and work
+            if n.startswith("pressure"):
+                # Pressure, see if there is a fixed deltaP
+                if hasattr(self, "deltaP") and self.deltaP[index].fixed:
+                    v2val = value(v1 + self.deltaP[index])
+            elif n.startswith("temperature"):
+                # Temperature, see if there is a fixed deltaT
+                if hasattr(self, "deltaT") and self.deltaT[index].fixed:
```
eslickj, line 1038: I'm not really sure if it matters, but there could be a deltaT (I think) with no temperature variable.

eslickj (approved):

idaes/core/base/tests/test_control_volume_1d.py
```diff
@@ -3682,3 +3682,129 @@ def test_report():
...
+
+    m.fs.cv = ControlVolume1DBlock(
+        property_package=m.fs.pp,
+        transformation_method="dae.finite_difference",
+        transformation_scheme="BACKWARD",
+        finite_elements=10,
+    )
+
+    m.fs.cv.add_geometry(length_domain_set=[0, 0.2, 0.5, 0.7, 1.0])
+    m.fs.cv.add_state_blocks(has_phase_equilibrium=False)
```
dallan-keylogic, line 3701: Do we want a `length_domain_set` inconsistent with the number of specified finite elements? Also, don't we want to apply the discretizing transformation before testing?

idaes/core/base/control_volume_base.py
```diff
@@ -960,3 +961,81 @@ def _get_representative_property_block(self):
...
+        Returns:
+            None
+
+        """
+        if v2.fixed:
+            # Do not touch fixed Vars
+            pass
+        elif v2.value is not None and not always_estimate:
+            # Var has a value and we are not always estimating - do nothing
+            pass
```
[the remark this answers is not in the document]
andrewlee94: This is the reason for the `always_estimate` argument - it is there to let the user tell us whether we should overwrite existing values or not. How they use it, and whether they leave variable uninitialized or not, it up to them.

idaes/core/base/tests/test_control_volume_1d.py
andrewlee94: Not for this case - this is just to add some points so we get multiple StateBlocks when we construct the model to make sure the values propagate properly. The test only calls the `estimate_state` method, so we don;t actually solve anything.

dallan-keylogic (approved):

PR #1092, opened 2023-02-15
Fixing check for kaug in uncertainty tests

andrewlee94:
## Summary/Motivation:
PR #1086 identified an issue where tests for the uncertainty propagation tool failed if `kaug` was not installed. The cause was identified as an error in how the test was checking for the presence of `kaug`.

- Add `exception_flag=False` to all checks for solver presence in `test_uncertainties.py`
- Clean up some imports at the same time

### Legal Acknowledgement

### DIFF
--- a/idaes/apps/uncertainty_propagation/tests/test_uncertainties.py
+++ b/idaes/apps/uncertainty_propagation/tests/test_uncertainties.py
@@ -78,7 +78,9 @@ class TestUncertaintyPropagation:
         np.testing.assert_array_almost_equal(
             results.cov, np.array([[6.30579403, -0.4395341], [-0.4395341, 0.04193591]])
         )
-        assert results.propagation_f == pytest.approx(5.45439337747349)
+        assert results.propagation_f == pytest.approx(
+            5.45439337747349, abs=1e-8, rel=1e-8
+        )
 
     @pytest.mark.component
     def test_quantify_propagate_uncertainty2(self):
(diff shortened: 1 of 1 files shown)


idaes/apps/uncertainty_propagation/tests/test_uncertainties.py
```diff
@@ -12,26 +12,24 @@
...
     clean_variable_name,
 )
-from pyomo.opt import SolverFactory
-from pyomo.environ import *
-import pyomo.contrib.parmest.parmest as parmest
 
-ipopt_available = SolverFactory("ipopt").available()
-kaug_available = SolverFactory("k_aug").available()
-dotsens_available = SolverFactory("dot_sens").available()
+sys.path.append(os.path.abspath(".."))  # current folder is ~/tests
```
eslickj, line 28: Why are we doing this?  I suspect it may be left from someone doing development without installing idaes?

idaes/apps/uncertainty_propagation/tests/test_uncertainties.py
```diff
@@ -12,26 +12,24 @@
...
-from unittest.mock import patch
-
-sys.path.append(os.path.abspath(".."))  # current folder is ~/tests
 import numpy as np
 import pandas as pd
-from scipy import sparse
 import pytest
-from pytest import approx
+
+from pyomo.environ import *
```
eslickj, line 19: This is bad form, I think.  Probably want to import just what you use.

eslickj (approved):

idaes/apps/uncertainty_propagation/tests/test_uncertainties.py
eslickj: Looking at the whole file now, I think this is probably not needed.

dallan-keylogic (approved):

PR #1097, opened 2023-02-17
Deprecating categorize_dae_variables

### LINKED ISSUE #945: Deprecate caprese method categorize_dae_variables?
Looking through `caprese.categorize` I see that there are two methods that do the same thing: `categorize_dae_variables_and_constraints` and `categorize_dae_variables`.  The former seems to be more general the latter, and the latter has problems like having empty lists as default arguments (rather than using `None` and creating a new empty list in the function). 

@Robbybp , can you tell us if `categorize_dae_variables` is obsolete and can be deprecated and/or removed, or whether we still need it for something?

andrewlee94:
## Summary/Motivation:
As discussed in issue #945, the `categorize_dae_variables` method has been superseded by another method and should be deprecated.

- Deprecate `categorize_dae_variables` - this method was untested so no change in tests.
-

### Legal Acknowledgement

### DIFF
--- a/idaes/apps/caprese/categorize.py
+++ b/idaes/apps/caprese/categorize.py
@@ -21,9 +21,9 @@ from pyomo.environ import (
 )
 from pyomo.dae import DerivativeVar
 from pyomo.common.collections import ComponentSet, ComponentMap
-from pyomo.util.slices import slice_component_along_sets
 from pyomo.core.expr.visitor import identify_variables
 from pyomo.contrib.incidence_analysis.interface import IncidenceGraphInterface
+from pyomo.common.deprecation import deprecated
 
 import idaes.apps.caprese.nmpc_var as nmpc_var
 from idaes.apps.caprese.common.config import (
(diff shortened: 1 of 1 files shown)


eslickj (approved):

dallan-keylogic:
Deprecating this method is good, as far as it goes, but I think I recall @Robbybp saying at the IDAES Technical meeting that even `categorize_dae_variables_and_constraints` was *also* getting replaced in some capacity by something in Pyomo. However, I don't remember the exact details, so it would be nice if he could chime in as well.

Robbybp (approved):

Robbybp:
@dallan-keylogic Yes but this "something else", basically just more compact and modular methods for identifying "differential" and "algebraic" variables/equations, is sitting on my hard drive waiting for me to (a) write tests and make sure it has all the functionality it needs and (b) figure out where it should go.

If anybody needs this functionality, I'm happy to pass along the code, but I probably won't get to cleaning it up for release any time in the (very) near future.

PR #1098, opened 2023-02-17
Bypass race condition in solver logging

dallan-keylogic:
## Summary/Motivation:
Apply @jsiirola 's fix from #887 on a fresh branch.

### Legal Acknowledgement

### DIFF
--- a/idaes/logger.py
+++ b/idaes/logger.py
@@ -16,6 +16,7 @@ import threading
 
 from contextlib import contextmanager
 from pyomo.common.tee import capture_output
+from pyomo.common.log import LogStream
 
 
 # Throw the standard levels in here, just let you access it all in one place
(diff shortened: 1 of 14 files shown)


idaes/logger.py
```diff
@@ -265,7 +266,7 @@ def add_valid_log_tag(tag):
     idaes.cfg.valid_logger_tags.add(tag)
 
 
-class IOToLogTread(threading.Thread):
+class IOToLogThread(threading.Thread):
```
andrewlee94, line 269: This is the reason the backward compatibility tests are failing, but I don't think we care at this point.

If we do, the fix is to register the name change in the IDAES `dev-tools` repo.

dallan-keylogic: Do we even still need this class? The only place it was used in the main repository was for the solver log, which has been removed by John Siirola's code.

eslickj: I don't think this is intended to be used directly by normal users.  It should be safe to delete without affecting backward compatibility.

andrewlee94 (approved):

PR #1100, opened 2023-02-22
Set default ipopt iteration limit to 200

### LINKED ISSUE #1099: Add max_iter to default ipopt config?
I was running the ideas-pse tests without HSL stuff and some solves failed after like 3000 iterations.  I think most initialization routines and other things that use a solver in the idaes-pse repo use ``get_sover()``, so we could probably just globally restrict the number of iterations by adding "max_iter" to the defaults IDAES config.  For things in the core idaes repo we may also want to use ``get_solver()`` everywhere.

eslickj:
## Fixes #1099 

This sets the ipopt default config to have a max_iter of 200.  This can be overridden by explicitly setting max_iter if you are using get_solver() or have the IDAES solver config system turned on. I do not think this breaks any tests in the main repo, but it could break an example.  If you set the iteration limit too low pretty much all tests fail, so I know this is working, but I don't really know if 200 is too high.  At some point we should take a harder look at fragile tests and possibly tune this in the process.

The purpose of this is to let people now if they set something up that requires a suspiciously large number of iterations and to limit the amount of time a failing test can take.

### Legal Acknowledgement

### DIFF
--- a/data/config.yaml
+++ b/data/config.yaml
@@ -1,8 +1,8 @@
 _id: d779730c7cc94155877a24e2e8be2ad1
-created: '2023-02-22T12:23:34.478902'
+created: '2022-04-18T14:36:30.754421'
 description: Built-in IDAES workspace
 htmldocs:
 - https://idaes-pse.readthedocs.io/en/stable
 - '{dmf_root}/docs/build/html'
-modified: '2023-02-22T12:23:34.478902'
+modified: '2022-04-18T14:36:30.754421'
 name: IDAES
--- a/data/config.yaml
+++ b/data/config.yaml
@@ -1,8 +1,8 @@
 _id: d779730c7cc94155877a24e2e8be2ad1
-created: '2022-04-18T14:36:30.754421'
+created: '2023-02-22T12:23:34.478902'
 description: Built-in IDAES workspace
 htmldocs:
 - https://idaes-pse.readthedocs.io/en/stable
 - '{dmf_root}/docs/build/html'
-modified: '2022-04-18T14:36:30.754421'
+modified: '2023-02-22T12:23:34.478902'
 name: IDAES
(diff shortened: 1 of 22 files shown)


data/config.yaml
```diff
@@ -1,8 +1,8 @@
 _id: d779730c7cc94155877a24e2e8be2ad1
-created: '2022-04-18T14:36:30.754421'
+created: '2023-02-22T12:23:34.478902'
 description: Built-in IDAES workspace
 htmldocs:
 - https://idaes-pse.readthedocs.io/en/stable
 - '{dmf_root}/docs/build/html'
-modified: '2022-04-18T14:36:30.754421'
+modified: '2023-02-22T12:23:34.478902'
```
andrewlee94, line 7: This file has slipped in again.

idaes/apps/caprese/examples/cstr_model.py
```diff
@@ -42,8 +43,10 @@
 
 # See if ipopt is available and set up solver
 if SolverFactory("ipopt").available():
-    solver = SolverFactory("ipopt")
-    solver.options = {"tol": 1e-6, "mu_init": 1e-8, "bound_push": 1e-8}
+    solver = get_solver(
+        solver="ipopt",
+        options={"tol": 1e-6, "mu_init": 1e-8, "bound_push": 1e-8},
```
andrewlee94, line 48: Just to be sure - this won't affect the options in other tests will it?

andrewlee94 (approved):
Only one minor comment about the `data/config.yaml` file that keeps getting changed by the tests and slipping into PRs.

idaes/apps/caprese/examples/cstr_model.py
eslickj: No, the options just override whatever is in the config if anything.  They don't change the config.

dallan-keylogic (approved):

PR #1103, opened 2023-02-23
Fixing mutable default arguments

### LINKED ISSUE #330: Mutable default arguments

andrewlee94:
## Summary/Motivation:
Fixing remaining instances of mutable default arguments in functions and methods, and enable pylint check for this.

-
-

### Legal Acknowledgement

### DIFF
--- a/idaes/models_extra/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/hetero_reactions.py
+++ b/idaes/models_extra/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/hetero_reactions.py
@@ -245,7 +245,7 @@ class _ReactionBlock(ReactionBlockBase):
     whole, rather than individual elements of indexed Reaction Blocks.
     """
 
-    def initialize(blk, outlvl=idaeslog.NOTSET, optarg={"tol": 1e-8}, solver="ipopt"):
+    def initialize(blk, outlvl=idaeslog.NOTSET, optarg=None, solver="ipopt"):
         """
         Initialisation routine for reaction package.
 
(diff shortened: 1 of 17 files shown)


idaes/models_extra/gas_solid_contactors/properties/oxygen_iron_OC_oxidation/hetero_reactions.py
```diff
@@ -245,7 +245,7 @@ class _ReactionBlock(ReactionBlockBase):
     whole, rather than individual elements of indexed Reaction Blocks.
     """
 
-    def initialize(blk, outlvl=idaeslog.NOTSET, optarg={"tol": 1e-8}, solver="ipopt"):
+    def initialize(blk, outlvl=idaeslog.NOTSET, optarg=None, solver="ipopt"):
         """
         Initialisation routine for reaction package.
```
dallan-keylogic, line 250: Initialisation -> Initialization

dallan-keylogic (approved):

eslickj (approved):

PR #1107, opened 2023-02-24
Verify potentially unsafe tarfiles before extraction

jsiirola:
## Summary/Motivation:
This PR addresses CVE-2007-4559 in a more robust manner than what was proposed in #1028.  While that PR would catch tarfile members with either absolute or relative paths that point outside of the target directory, it did not address issues where the tarfile could contain symbolic (hard or soft) links that (in aggregate) could walk up the directory hierarchy and out of the target directory (which was one of the first vulnerabilities pointed out in CVE-2007-4559).

This PR addresses the issue by *manually* walking the filesystem for each member name resolving both symbolic links that already exist on the filesystem, as well as any links that would be created by the tarfile payload itself.  The final "resolved path" is then checked to ensure it is within the target directory tree.

- verify each member of a tarfile before extraction to ensure all members will be created within the target directory tree
- test the verifier using two small examples

### Legal Acknowledgement


blnicho (approved):

lbianchi-lbl (approved):
I don't understand enough of the details to have an opinion on the implementation, but I trust the explanation and the tests.

PR #1061, opened 2023-01-10
Updating copyright for 2023

ksbeattie:
## Fixes

The copyright notice to include 2023

## Summary/Motivation:
Updated Copyright and License files, a few others and all .py files (by running addheader) to update the copyright years.

## Changes proposed in this PR:
-
-

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1109, opened 2023-02-25
Updating copyright for 2023

ksbeattie:
## Fixes

The copyright notice to include 2023.  This PR replaces #1061 

## Summary/Motivation:

Updated Copyright and License files, a few others and all .py files (by running addheader) to update the copyright years.

This uses the new header format as requested in https://github.com/IDAES/examples/issues/7

-
-

### Legal Acknowledgement


ksbeattie:
Special notes (for when I do this again for other repos) are:

A command to look for files that are only whitespace:

    find . -not -path '.*/.git/*' -type f -name '*.py' -not -empty | while read f; do if ! grep -qz '[^[:space:]]' "$f"; then echo $f is only spaces; fi; done

How to truncate the final newline from the header file used by `addheader` so you don't get an extra line in the header in each file.

    truncate -s -1 file_header.txt

lbianchi-lbl (approved):
Probably ripping off the band-aid is the right thing to do, although I'm pretty sure this will be "fun" to merge into the `2.0_rel` branch...

ksbeattie:
> Probably ripping off the band-aid is the right thing to do, although I'm pretty sure this will be "fun" to merge into the `2.0_rel` branch...

I could try a different PR to/from that branch, replaying the same there, maybe that would be easier for the release.

andrewlee94 (approved):

PR #1111, opened 2023-03-01
Adding more pylint checks

andrewlee94:
## Changes proposed in this PR:
- W404
- W0622
- W1505

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1112, opened 2023-03-02
Even more pylint checks

andrewlee94:
## Summary/Motivation:
Fixes a lot more pylint issues and activates checks.

### Legal Acknowledgement


dallan-keylogic:
I see a lot of disabling W0611 for "convenience imports", whatever that means.  We should either enforce this rule on ourselves or not enforce it on others.

andrewlee94:
@dallan-keylogic Unfortunately that ends up breaking backward compatibility. Ideally we would fix them, but for now I want to make sure we don't start getting any more.

dallan-keylogic (approved):

lbianchi-lbl (changes requested):
Just leaving this as a "Request changes" so that it doesn't automerge:

- We should remove the `v1-compat.yml` workflow file since that is already failing on `main` after #1111 (and causing unrelated PRs such as #1107 to fails as well)
- I'm starting to lose track of which checks are what using only the numeric code, so we should think of what would be the best way of using names instead. Probably just splitting the command in `.github/workflows/core.yml` using `\` would be the easiest

andrewlee94:
@lbianchi-lbl Whilst I am using numeric codes for now, I how to eventually get this down to a short list where we can use names. However, numeric codes are a lot easier for me to work with when testing things, so I would prefer to stay with them for now (as an interim solution).

lbianchi-lbl (approved):

PR #1126, opened 2023-03-10
Spring cleaning for dependencies, Pt. 1: The Tree-Shaking

lbianchi-lbl:
## Summary/Motivation:

- Remove dependencies towards our ALARA (As Low As Reasonably Achievable) goal

Remove dependencies belonging to the following categories:

- [x] Dependencies without any Ctrl-F-able imports
- [x] Dependencies that are only used in "leaf" modules (i.e. where adding import guards only affects that module, as opposed to other areas in the codebase)

### Legal Acknowledgement

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -85,7 +85,6 @@ kwargs = dict(
         "pyyaml",
         "requests",  # for ui/fsvis
         "scipy",
-        "sympy",
         "tinydb",
         "xlrd",  # for DMF read of old .xls Excel files
         "openpyxl",  # for DMF read of new .xls Excel files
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -68,37 +68,26 @@ def dmf_data_files(root: str = DMF_DATA_ROOT) -> List[Tuple[str, List[str]]]:
...
-        "distro",  # help identify linux distros for binary downloads
-        "flask",  # for ui/fsvis
-        "flask-cors",
         "jupyter",
         "lxml",
         "matplotlib",
         "nbconvert",
         "nbformat",
         "numpy",
-        "networkx",
```
andrewlee94, line 84: I don;t think we should necessarily have this as a dependency, but this is something we might rely on implicitly - it gets used in the Pyomo SequentialDecomposition tool (but the dependency probably should be handled there).

andrewlee94 (approved):
Looks good.

eslickj (approved):

PR #1127, opened 2023-03-10
Removing imports of six and from __future__

andrewlee94:
## Summary/Motivation:
This PR cleans up so old Python 2 compatibility imports.

-
-

### Legal Acknowledgement

### DIFF
--- a/idaes/core/dmf/tests/test_getver.py
+++ b/idaes/core/dmf/tests/test_getver.py
@@ -73,12 +73,9 @@ class fake_module:
 
 @pytest.mark.unit
 def test_more_func():
-    import six
-
-    # doesn't have __path__ but has __loader__ with path
-    dmf.get_version_info(six)
     # __loader__ without path
     dmf.get_version_info(fake_module)
+    # TODO: No assert!
 
 
 @pytest.mark.unit
--- a/idaes/core/surrogate/pysmo/polynomial_regression.py
+++ b/idaes/core/surrogate/pysmo/polynomial_regression.py
@@ -14,10 +14,6 @@
 # pylint: disable=missing-module-docstring
 # pylint: disable=missing-function-docstring
 
-# Imports from the python standard library
-from __future__ import division
-
-# from builtins import int, str
 import os.path
 import warnings
 import pickle
@@ -28,7 +24,6 @@ import numpy as np
 import pandas as pd
 import scipy.optimize as opt
 from scipy import stats
-from six import string_types
 
 from pyomo.environ import (
     ConcreteModel,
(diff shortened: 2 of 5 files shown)


bpaul4 (commented):
As I understand it, removing these imports should be fine if the tests pass and results match. In the case of `__future__.division`, is there a way to check that `polynomial_regression` is returning the expected output? It may be that `__future__.division` is identical to the corresponding Python3 call, in that case we wouldn't need further checks.

andrewlee94:
@bpaul4 `__future__.division` was a Python 2-3 compatibility feature so that systems running Python 2 would get Python 3 behaviour - i.e. it ensured Python 3 division. Now that we only support Python3 we do not need this.

bpaul4 (approved):
Thanks for clarifying, the changes look good to me.

idaes/core/dmf/tests/test_getver.py
```diff
@@ -73,12 +73,9 @@ class fake_loader:
 
 @pytest.mark.unit
 def test_more_func():
-    import six
-
-    # doesn't have __path__ but has __loader__ with path
-    dmf.get_version_info(six)
```
lbianchi-lbl, line 79: From what I can understand from the context, `six` is being used here as an example of a module that has an empty `__path__` attribute, which in turn I imagine is used to test whether the DMF is able to deal with such cases. Pinging @dangunter in case he wants to chime in on this.

andrewlee94: That was my interpretation too, but it runs into the issue of either:

a) `six` is not a direct dependency of IDAES so it might not be there, or
b) adding `six` as a dependency just to test this functionality.

Given the lack of use of the DMF, I took the easy route and just removed this part of the test. Also, I will note this test does not assert anything (and I almost deleted the whole test for that reason).

lbianchi-lbl: I agree that taking the test out seems the most reasonable thing as having `six` as a dependency just for this test seems unnecessary. Specifically about the lack of `assert` - I think it's just "asserting it doesn't fail", i.e. that calling the function doesn't cause an exception to be raised.

lbianchi-lbl (approved):

idaes/core/dmf/tests/test_getver.py
dangunter: Right. At one point six was a direct dependency of IDAES, but of course I didn't remember that this test existed when six was removed (probably by me, but ¯\_(ツ)_/¯ ). Removing the test is fine. The DMF is at the back of the refrigerator right now waiting for the next cleaning.

PR #1133, opened 2023-03-15
Spring cleaning for dependencies, Pt. 2: The Tree-Chopping

lbianchi-lbl:
## Motivation

- Running `pip install "."` from this branch in a fresh environment results in the following being installed:

```log
  Package         Version
--------------- ----------
certifi         2022.12.7
click           8.1.3
contourpy       1.0.7
cycler          0.11.0
fonttools       4.39.0
idaes-pse       2.0.0.dev3
kiwisolver      1.4.4
matplotlib      3.7.1
numpy           1.24.2
packaging       23.0
pandas          1.5.3
Pillow          9.4.0
pip             23.0.1
ply             3.11
Pyomo           6.5.0
pyparsing       3.0.9
python-dateutil 2.8.2
pytz            2022.7.1
scipy           1.10.1
setuptools      65.6.3
six             1.16.0
wheel           0.38.4
```

- The current set of dependencies is most similar to the combined `ui,dmf,grid,omlt` targets:

```sh
pip install ".[ui,dmf,grid,omlt]"
```

```log
Package                  Version
------------------------ ----------
addheader                0.3.2
anyio                    3.6.2
argon2-cffi              21.3.0
argon2-cffi-bindings     21.2.0
arrow                    1.2.3
asttokens                2.2.1
attrs                    22.2.0
backcall                 0.2.0
beautifulsoup4           4.11.2
bleach                   6.0.0
certifi                  2022.12.7
cffi                     1.15.1
charset-normalizer       3.1.0
click                    8.1.3
colorama                 0.4.6
comm                     0.1.2
contourpy                1.0.7
coramin                  0.1.1
cycler                   0.11.0
debugpy                  1.6.6
decorator                5.1.1
defusedxml               0.7.1
et-xmlfile               1.1.0
exceptiongroup           1.1.1
executing                1.2.0
fastjsonschema           2.16.3
fonttools                4.39.0
fqdn                     1.5.1
gridx-egret              0.5.3
gridx-prescient          2.2.1
idaes-pse                2.0.0.dev3
idna                     3.4
iniconfig                2.0.0
ipykernel                6.21.3
ipython                  8.11.0
ipython-genutils         0.2.0
ipywidgets               8.0.4
isoduration              20.11.0
jedi                     0.18.2
Jinja2                   3.1.2
jsonpointer              2.3
jsonschema               4.17.3
jupyter                  1.0.0
jupyter_client           8.0.3
jupyter-console          6.6.3
jupyter_core             5.2.0
jupyter-events           0.6.3
jupyter_server           2.4.0
jupyter_server_terminals 0.4.4
jupyterlab-pygments      0.2.2
jupyterlab-widgets       3.0.5
kiwisolver               1.4.4
lxml                     4.9.2
MarkupSafe               2.1.2
matplotlib               3.7.1
matplotlib-inline        0.1.6
mistune                  2.0.5
nbclassic                0.5.3
nbclient                 0.7.2
nbconvert                7.2.10
nbformat                 5.7.3
nest-asyncio             1.5.6
networkx                 3.0
notebook                 6.5.3
notebook_shim            0.2.2
numpy                    1.24.2
omlt                     1.1
openpyxl                 3.1.2
packaging                23.0
pandas                   1.5.3
pandocfilters            1.5.0
parso                    0.8.3
pexpect                  4.8.0
pickleshare              0.7.5
Pillow                   9.4.0
Pint                     0.20.1
pip                      23.0.1
platformdirs             3.1.1
pluggy                   1.0.0
ply                      3.11
prometheus-client        0.16.0
prompt-toolkit           3.0.38
psutil                   5.9.4
ptyprocess               0.7.0
pure-eval                0.2.2
pycparser                2.21
Pygments                 2.14.0
Pyomo                    6.5.0
pyparsing                3.0.9
PyPDF2                   3.0.1
pyrsistent               0.19.3
pytest                   7.2.2
python-dateutil          2.8.2
python-json-logger       2.0.7
pytz                     2022.7.1
PyYAML                   6.0
pyzmq                    25.0.1
qtconsole                5.4.1
QtPy                     2.3.0
requests                 2.28.2
rfc3339-validator        0.1.4
rfc3986-validator        0.1.1
scipy                    1.10.1
seaborn                  0.12.2
Send2Trash               1.8.0
setuptools               65.6.3
six                      1.16.0
sniffio                  1.3.0
soupsieve                2.4
stack-data               0.6.2
terminado                0.17.1
tinycss2                 1.2.1
tinydb                   4.7.1
tomli                    2.0.1
tornado                  6.2
traitlets                5.9.0
uri-template             1.2.0
urllib3                  1.26.15
wcwidth                  0.2.6
webcolors                1.12
webencodings             0.5.1
websocket-client         1.5.1
wheel                    0.38.4
widgetsnbextension       4.0.5
xlrd                     2.0.1

```

- Split non-core dependencies in categories selectable as `extras_require` targets

## Caveats

- [x] ~Most likely causes widespread test failures~ Tests are now passing
- [x] ~Would still need to update the documentation~ Documentation has been updated (I might have missed something)

## Decisions

- After some deliberation, it was decided *not* to exposes a catch-all `extras_require` target (e.g. `complete`, `all`). This follows the "explicit is better than implicit" principle and requires downstream clients to pro-actively "opt in" to the sets of optional dependencies they need

## For reviewers

- [ ] Check documentation, especially installation steps
  - I've done a first pass based on what I could find, but it's entirely possible that I've missed something

### Legal Acknowledgement


setup.py
```diff
@@ -59,6 +59,90 @@ def dmf_data_files(root: str = DMF_DATA_ROOT) -> List[Tuple[str, List[str]]]:
...
+            "tinydb",  # resourcedb
+            "xlrd",  # tables (implicitly by pandas.read_excel())
+            "openpyxl",  # tables (implicitly by pandas.read_excel())
+        ],
+    )
+    surrogate = [
+        "omlt==1.1",  # fix the version for now as package evolves
+    ]
+    mlai = [
+        "tensorflow",
```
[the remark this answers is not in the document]
lbianchi-lbl: I agree this makes sense from a conceptual standpoint and that's how I originally grouped it. However, if we want to have one `extras_require` target to match the current set of dependencies, the "topical" targets must be such that `omlt` (which is installed by default now) and `tensorflow` (which is not) are separate.

More concretely, there's also the factor of `tensorflow` being of eating-into-my-ISP-data-caps size (not to mention the multiple widespread side effects caused by merely importing it), which IMO are all good reasons to limit as much as possible the circumstances in which it does get installed.

Note that the other "solution" for these constraints would be that *both* OMLT and Tensorflow are not part of the `complete` set of dependencies, which, as far as I'm concerned, I think we should consider as a possible choice ("installation backward-compatibility" considerations aside).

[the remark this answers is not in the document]
andrewlee94: We should check with the OMLT developers and see if it can be used without tensorflow or not. If there is a use case where you can do something without tensorflow then they can probably remain separate, but if you really need both then we should make OMLT optional as well.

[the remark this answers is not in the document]
lbianchi-lbl: I've reworked the grouping a bit taking this comment into account. There's now an `omlt` target comprising `omlt` and `tensorflow`.

setup.py
```diff
@@ -57,6 +55,81 @@ def dmf_data_files(root: str = DMF_DATA_ROOT) -> List[Tuple[str, List[str]]]:
...
+    ['requests', 'pint']
+    >>> list(extras)
+    ['ui', 'examples', 'dmf', 'surrogate', 'mlai', 'grid', 'optional']
+    >>> dict(extras)
+    {'ui': ['requests', 'pint'], 'examples': ['jupyter', 'nbconvert', 'nbformat'], ...}
+    """
+
+    ui = [
+        "requests",
+        "pint",
```
andrewlee94, line 73: This might fall out else where, but `pint` is something we use fairly broadly (if maybe indirectly through Pyomo). I do believe that we use `pint` directly in a few places.

lbianchi-lbl: > This might fall out else where, but `pint` is something we use fairly broadly (if maybe indirectly through Pyomo). I do believe that we use `pint` directly in a few places.

That's a good point. `pint` is already part of the default dependencies so having it repeated here is not strictly needed. The reason why I added is an attempt to be explicit, i.e. "if you're importing it, it should be in your dependencies", and `pint` is imported and used directly in the `idaes.core.ui` code.

Note that this is anyway somewhat of a moot point, as this dependency will become `idaes-ui` once we complete the migration of the UI codebase to IDAES/idaes-ui (#1146).

.github/workflows/integration.yml
```diff
@@ -223,8 +224,10 @@ jobs:
             runner-image: ubuntu-20.04
           - install-mode: pip-default
             pip-install-target: '.'
-          - install-mode: pip-optional
-            pip-install-target: .[optional]
+          - install-mode: pip-complete
+            pip-install-target: .[complete]
```
andrewlee94, line 228: Are these still valid options? I seem to recall you deciding to drop the "complete" and "all" options, and I do not see them later in setup.py.

docs/how_to_guides/versioned_idaes_install.rst
```diff
@@ -37,16 +37,16 @@ Installing IDAES
 ----------------
 .. _updating_install:
 
-* To get a previous `IDAES release <https://github.com/IDAES/idaes-pse/releases>`_, for example v1.13 ::
+* To get a previous `IDAES release <https://github.com/IDAES/idaes-pse/releases>`_, for example 2.0 ::
 
-   pip install idaes-pse==1.13
+   pip install idaes-pse==2.0
 
 * To get the latest version from the GitHub main branch ::
```
andrewlee94, line 44: One thing I do not see here is instructions on how to get the optional dependencies; this is one of our user facing sets of install instructions so we probably need to cover it here (or point to the README for more details and to avoid the possibility of conflicting instrucitons).

andrewlee94 (commented):
A couple of comments that I wanted to check on, but otherwise this is looking good to me.

.github/workflows/integration.yml
lbianchi-lbl: You're absolutely correct, thanks for spotting that. I'll modify the workflow files to account for this.

docs/how_to_guides/versioned_idaes_install.rst
lbianchi-lbl: I've added a How-to page on optional dependencies: https://idaes-pse--1133.org.readthedocs.build/en/1133/how_to_guides/opt_dependencies.html

andrewlee94 (approved):
Looks good to me.

ksbeattie (approved):
LGTM

PR #1143, opened 2023-03-21
Fixing model statistics to work for indexed blocks

andrewlee94:
## Summary/Motivation:
The current model statistics functions do not work when provided with an `IndexedBlock` as these do not have a `component_data_objects` method. This PR updates these methods to catch cases where the block is indexed and to first iterate over the `BlockData` objects.

- Add new private generator method that checks for indexed blocks and iterates over BlockData objects if required.
- Update tests to cover indexed blocks.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/model_statistics.py
+++ b/idaes/core/util/model_statistics.py
@@ -27,6 +27,24 @@ from pyomo.common.collections import ComponentSet
 
 
 # -------------------------------------------------------------------------
+# Generator to handle cases where the input is an indexed Block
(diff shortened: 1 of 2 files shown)


dallan-keylogic:
Is there a reason why indexed blocks in Pyomo don't have a component data objects method @jsiirola ? This new method might live more naturally there.

eslickj (approved):

dallan-keylogic (approved):

PR #1148, opened 2023-03-27
Update approach for documenting Config Blocks

andrewlee94:
## Summary/Motivation:
With Pyomo v6.5.0, some of the methods we use to document the Config Blocks in our models have been deprecated. This PR updates the relevant lines of code to use newer methods.

- Update`process_block.py` to better format Config block and suppress deprecation warnings.
- Update how Config blocks are included in Initializers.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/initialization/initializer_base.py
+++ b/idaes/core/initialization/initializer_base.py
@@ -23,7 +23,7 @@ from pyomo.environ import (
     Var,
 )
 from pyomo.core.base.var import _VarData
-from pyomo.common.config import ConfigBlock, ConfigValue, add_docstring_list
+from pyomo.common.config import ConfigBlock, ConfigValue, String_ConfigFormatter
 
 from idaes.core.util.model_serializer import to_json, from_json, StoreSpec, _only_fixed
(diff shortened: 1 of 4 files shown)


eslickj (approved):
Looks good to me.  I assume the generated documentation looks right.

idaes/core/initialization/initializer_base.py
```diff
@@ -110,7 +108,17 @@ def __init__(self, **kwargs):
 
     def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
-        cls.__doc__ = add_docstring_list(cls.__doc__ if cls.__doc__ else "", cls.CONFIG)
+        cls.__doc__ = cls.__doc__ + cls.CONFIG.generate_documentation(
```
jsiirola, line 111: This is missing the "Args:" header for google-style documentation.  Alternatively, you could switch to using numpydoc and the built-in `format='numpydoc'` formatter.

jsiirola (approved):
Approving, as the new code should be as correct (or incorrect) as the existing implementation.

PR #1157, opened 2023-03-30
Improve `ipopt_solve_halt_no_error` model diagnostic function for NL_v2

bknueven:
## Summary/Motivation:
See https://github.com/Pyomo/pyomo/issues/2793

- Add `export_defined_variables=False` to call to solve method in `ipopt_solve_halt_on_error`
- Modify test so it fails on `main`.

### Legal Acknowledgement


jsiirola:
FWIW, there are pretty significant performance implications for turning off the use of defined variables.  Would a better solution be to add a log processor that looks for the error string and maps it back to the named expression?

andrewlee94:
@jsiirola I am not really sure what that would entail, but this function is also intended to be a debugging tool so I am less concerned about performance than if this were something we called all the time.

eslickj (approved):

bknueven:
This would be a more useful debugging tool if it could point to the problematic Expression as opposed to the problematic Constraint. But that seems like a nontrivial extension of the existing function.

andrewlee94 (approved):

jsiirola (approved):

PR #1158, opened 2023-03-31
Update Helmholtz EOS

eslickj:
- Move transport properties to external functions
- More accurate thermal conductivity for water
- Switch to direct function calls with state variables rather than calculating intermediate delta and tau. 

### Legal Acknowledgement


andrewlee94 (commented):
I have a few questions.

1. I see the component files have been removed - what are these being replaced with? Are the json files now in `idaes-ext`?
2. Related to this, how does the component registration work now? I see that the registration functions take a string for the component name, but I am not sure how this links to the parameter definitions.
3. How do we document which components we support? There needs to be some way to let users know which components we support (and it would be really good if could automate this).

eslickj:
@andrewlee94,

1) Those files were transport property models. Due to some work with Carrier I moved those to the external functions.

2) The component name serving is how it finds the parameter files.  There is a setting for the director to look for parameter files the looks for {comp}_parameters.json ... For IDAES, we assume the parameter files are located in a directory where the binaries are. An alternative would be to include parameter files with IDAES-PSE.  It's a question for the future probably.  

3) There is a lot we need to document.  I'll try to do another PR, and see if I can automate the documentation of supported components and if the transport models are available.  It would be nice to automate citations too.

andrewlee94 (approved):

lbianchi-lbl:
We don't know to what extent this could affect backward compatibility, so we should test it.

eslickj:
@lbianchi-lbl, the changes should be internal to the property model.  So far, I haven't had to update any models to make things work with the changes.  I don't expect it to affect backward compatibility, but we should test.  I don't know how backward compatibility testing works.

andrewlee94:
@eslickj The backward compatibility tests are currently in a separate repo: https://github.com/IDAES/idaes-compatibility

We need to get a nightly workflow setup to run these (and maybe an on-PR run too).

agarciadiego (approved):

PR #1159, opened 2023-03-31
Model statistics scaling

dallan-keylogic:
## Summary/Motivation:
To find variables near their lower/upper bounds, judge relative proximity to bounds based on scaling factors rather than creating a surrogate scaling factor from the lower/upper bounds themselves.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/model_statistics.py
+++ b/idaes/core/util/model_statistics.py
@@ -709,7 +709,7 @@ def variables_near_bounds_generator(
             continue
 
         if relative:
-            sf = get_scaling_factor(v)
+            sf = get_scaling_factor(v, default=1)
         else:
             sf = 1
 
--- a/idaes/core/util/model_statistics.py
+++ b/idaes/core/util/model_statistics.py
@@ -25,6 +25,8 @@ from pyomo.dae import DerivativeVar
 from pyomo.core.expr.current import identify_variables
 from pyomo.common.collections import ComponentSet
 
+from idaes.core.util.scaling import get_scaling_factor
+
 
 # -------------------------------------------------------------------------
 # Generator to handle cases where the input is an indexed Block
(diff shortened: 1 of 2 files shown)


andrewlee94 (changes requested):
I am not sure this PR is doing things the right way - it seems to be conflating the issue of scaling with that of absolute and relative tolerance.

For one, I must ask if scaling is even an issue when checking if a variable is close to its bounds? For the relative case this is definitely not an issue, although one could argue that in the absolute case you might want to look at the variable in scaled space.

My initial reaction is that scaling should be a new argument to the function, and that it should only be applied if absolute tolerances are used.

andrewlee94:
@dallan-keylogic That means we have two separate issues here - the calculation of the relative tolerance and the application of the scaling (if applicable), and I think this PR is still conflating the two of them. Notably, with `relative=True`, I do not think we are checking a relative tolerance with the new approach; rather it is using the provided tolerance as an absolute tolerance in the scaled space.

I think we need to treat each of these separately:

1. We should add a  `scaled` argument to the function (default `True`) which determines whether we look up a scaling factor or set `sf=1`.
2. I think we also need to rethink what it means to have relative tolerances in this case; notably you cannot have a relative tolerance with a value of 0. Perhaps we should do what `pytest.approx` and similar functions do and have separate absolute and relative tolerances for this reason (for backward compatibility, we maintain the existing arguments but add new `abs_tol` and `rel_tol` argument and map the old arguments to those if provided).

bknueven:
> So long as IPOPT (or whatever solver) is going to be looking at the scaled version of the variable in order to determine proximity to constraints (for backing off, for size of the dual variables, etc), the scaled version is what we actually care about more than the unscaled version. There is an argument we might want to be using this tool on the scaled model rather than the unscaled model (once the new scaling tools are created), but I'm living in the present.

Note that for some things, (e.g., `bounds_push`), Ipopt does not consider the user-provided scaling factor. See: https://github.com/coin-or/Ipopt/discussions/555. This can sometimes cause trouble with our various sequential initialization routines.

andrewlee94:
@dallan-keylogic I've implemented separate absolute and relative tolerances as part of the diagnostics toolbox PR: https://github.com/IDAES/idaes-pse/pull/1228

For the issue of scaling, given the way we are going regarding scaling in Pyomo, I think the best way ti s to not consider scaling at all within the method and only look at the model as it stands. If users want to check bounds in scaled space, then they should use this method on the scaled model directly, rather than us trying to repeat the scaling here.

PR #1166, opened 2023-04-06
RBF patch for basis function display.

### LINKED ISSUE #1164: "gaussian" rbf type always displayed when loading surrogate using `PysmoSurrogate.load_from_file()`
I performed independent surrogate training for all radial basis functions using the config argument in  `PysmoRBFTrainer()` When training, the terminal display is for example,
```
Default parameter estimation method is used.

Parameter estimation method:  algebraic
Basis function:  imq
Regularization done:  True
```
The main thing is that it correctly reports the basis function.

However, when reloading ANY of the surrogates with various rbf's using `PysmoSurrogate.load_from_file()`, the terminal output ALWAYS shows
```
Default parameter estimation method is used.

Parameter estimation method:  algebraic
Gaussian basis function is used.
Basis function:  gaussian
Regularization done:  True
```
 This may be a problem in line 318 of idaes\core\surrogate\pysmo\radial_basis_function.py, but not certain

**EDIT** @lbianchi-lbl : this is the line mentioned: https://github.com/IDAES/idaes-pse/blob/f9c4fd319a590a839007db2605543b523904fa86/idaes/core/surrogate/pysmo/radial_basis_function.py#L318

OOAmusat:
## Fixes
#1164 

## Summary/Motivation:
Fix basis function reported for RBF when using PysmoSurrogate.load_from_file()

- Passes basis function type into dummy RBF call, ensuring right basis function is printed to screen.

### Legal Acknowledgement


OOAmusat:
@andrewlee94 could you please check that this fixes the problem on your end as well?

andrewlee94:
@OOAmusat You would need to tell me what to do to test/replicate this issue. However, I think a better solution would be to have tests which checks this behaviour which will also help prevent regressions.

OOAmusat:
@andrewlee94 It wasn't an error. 

The way the loading function works is that we create a dummy class (with the default RBF settings) to initialize the object with the necessary attributes,  and then populate the values of those attibutes from the saved file. What is printed to screen is from the dummy class initialization, not the final object returned to the user.

What I have done now is to pass in the actual basis function when creating the dummy attribute so that the correct basis function is displayed.

We already have tests that check that the right attributes are actually loaded into the model.

andrewlee94:
The end result might have been right, but the printed output was unexpected and that is a bug/issue. You should add a test for this use case that checks the output to the user and verifies that it is what is expected.

andrewlee94 (approved):

bpaul4 (approved):
The change to the main PySMO file makes sense to me. I'm less familiar with the dictionary you check against in the new test, but if it's just a normal output assertion then it looks good to me as well.

PR #1170, opened 2023-04-18
More Initializer Testing

andrewlee94:
## Summary/Motivation:
This PR adds more testing and refinement of Initializers for core unit models.

-
-

### Legal Acknowledgement


idaes/core/initialization/block_triangularization.py
```diff
@@ -119,9 +119,21 @@ def _solve_block_data(self, block_data, solver):
...
+            # ExternalFunctions do not play nice with sympy differentiation
+            # (the default used in calculate_variable_from_constraint).
+            # If we see a TypeError, suggest users try using reverse_numeric
+            # differentiation.
+            raise TypeError(
+                "A TypeError was encountered in the scc solver. This can occur "
+                "in models that involve ExternalFunctions. We suggest you try "
+                "setting calculate_variable_options=\u007b'diff_mode': "
+                "pyomo.core.expr.calculus.differentiate.Modes.reverse_numeric\u007d. "
+            )
```
eslickj, line 139: Not for this, but is this because you are using simpy for derivatives?  Could you use pynumero?

andrewlee94: This is a Pyomo tool so that is a question for them, but yes it is because they are using `sympy` by default for the derivatives.

eslickj (approved):

idaes/core/initialization/block_triangularization.py
jsiirola: @eslickj, this is orthogonal to pynumero.  But yes, it is possible to automatically fall back on numeric differentiation (or just always use numeric differentiation).  I just pushed up Pyomo/pyomo#2812 to have `calculate_variable_from_constraint()` automatically fall back on numeric differentiation if symbolic fails (and the user didn't care what differentiation engine to use).

agarciadiego (approved):

PR #1173, opened 2023-04-19
Last set of core model Initializers

andrewlee94:
## Summary/Motivation:
This PR adds Initializers and tests for the remaining core unit models. With this, I think we can say the initial deployment of the tools is ready and start working on documentation so that more users can test the API.

-
-

### Legal Acknowledgement


eslickj (approved):

eslickj:
Not really for this PR, but maybe a suggestion for the future.  Should we modify the old model initialization methods to use the initializer objects to eliminate code duplication, and avoid problems in the future when the old method goes away.?

andrewlee94:
@eslickj I was considering that - I just didn't want to go to the effort of doing it right now.

agarciadiego (approved):

PR #1197, opened 2023-06-07
Reformulate solubility_products constraints

lxhowl:
## Fixes
pH-dependent solubility tests from WaterTAP

## Summary/Motivation:
This solubility equilibrium constraint used a smoothing strategy to compare precipitate amount _s_ and the solution state _Q_. Only one of _s_ and _Q_ can be greater than zero at any time. When _s_ and _Q_ differ in several orders of magnitude, the problem becomes hard to scale, especially when _Q_ is in log form, the original formulation will sacrifice the accuracy of _s_, the precipitate amount, which could be very small (i.e., 1e-10) and important in water-related applications.

Fix outside of bounds warning in the initialization of pH-dependent solubility tests.

- Normalize the precipitate amount _s_ to be 0-1 with respect to its solubility product constant _Ksp_.

- Update mole_frac_phase_comp bounds in FpcTP to be consistent with log_mole_frac_phase_comp bounds.

### Legal Acknowledgement


bknueven:
For a bit more context, Xinhong has been working to make some of chemistry models in WaterTAP more stable. She found this issue with the formulation and came up with this fix to keep these smooth_max constraints scaled consistently.

idaes/models/properties/modular_properties/state_definitions/FpcTP.py
```diff
@@ -142,7 +142,7 @@ def mole_frac_comp(b, j):
 
     b.mole_frac_phase_comp = Var(
         b.phase_component_set,
-        bounds=(1e-20, 1.001),
+        bounds=(1e-50, 1.001),
```
andrewlee94, line 145: I am not sure about this change, at least at in the general code. Firstly, for most applications 1e-50 is indistinguishable from 1e-20 (which I suppose is an argument that the change doesn't matter). However, it does suggest that we really should allow users to set the bounds on mole fractions in the configuration which I think we currently do not.

Secondly, if we change the value here that would suggest we should be changing it in all the other state definitions too for consistency there.

idaes/models/properties/modular_properties/reactions/equilibrium_forms.py
```diff
@@ -189,7 +189,7 @@ def return_expression(b, rblock, r_idx, T):
         if Qunits is not None:
             Q = Q / Qunits
 
-        return s - smooth_max(0, s - Q, rblock.eps) == 0
+        return Q - smooth_max(0, Q - s / (s + b.k_eq[r_idx]), rblock.eps) == 0
```
andrewlee94, line 192: I have a few questions about this formulation:

1. Firstly, this looks like something that can and should be handled by scaling the variables for the solid flow so they are not so close to zero.
2. Secondly, why did you use the equilibrium constant in the normalizing term?

andrewlee94 (changes requested):
I have a few questions about why you chose the formulation you did, as well as about consitency with other state definitions.

idaes/models/properties/modular_properties/reactions/equilibrium_forms.py
```diff
@@ -287,7 +287,8 @@ def return_expression(b, rblock, r_idx, T):
         Q = b.log_k_eq[r_idx] - e
         # Q should be unitless due to log form
 
-        return s - smooth_max(0, s - Q, rblock.eps) == 0
+        s = s * 10 / (s + b.k_eq[r_idx])
```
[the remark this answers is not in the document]
adowling2: What are typical values for k_eq. I am thinking about the scaling of s with the proposed reformulation.

When k_eq = 0, s <-- 10

When k_eq = 10,000, s <-- s / 1,000 (approximately)

Is this the intended behavior?

Comments could help explain on of @andrewlee94's questions better.

idaes/models/properties/modular_properties/state_definitions/FpcTP.py
lxhowl: In [idaes/models/properties/modular_properties/base/generic_property.py](https://github.com/IDAES/idaes-pse/blob/main/idaes/models/properties/modular_properties/base/generic_property.py) line 4688-4696, the bounds of `log_mole_frac_phase_comp` was set to be `(-100, log(1.001))`. This is a declaration before `mole_frac_phase_comp` in `FpcTP`. The inconsistency in bounds caused a warning that the value of `mole_frac_phase_comp` is out of bounds when we do the initialization of low concentration solubility problem (some inlet component fraction was set at 1e-20). 

Since e^(-100) is approximately 4e-44, I put 1e-50 there.

And yes, I think all bounds in state definitions should be changed consistently, or we can remove one between the original var and the log-transformed var.

idaes/models/properties/modular_properties/reactions/equilibrium_forms.py
lxhowl: `s` here is the solid molar flow rate `flow_mol_phase_comp['Sol']` and it has no clear upper bound. Depending on the feed flow rate and composition, it could end up in any order of magnitude. It's hard to properly scale it before solving the equilibrium. 

While `Q` = Ksp-f(c) has clear bounds, i.e., [0,Ksp], Ksp is typically 1e-37~1e-4 for solubility products before nondimensionlization (expect smaller Ksp). 

The original form easily ends up a positive and large `s` while `Q` is not a perfect zero considering the relative difference between f(c) and Ksp. 

So I first considered `s/(s+a)` to make the precipitate term in [0,1]. Then I chose a = Ksp to bring up similar accuracy in s and Q (or f(c)). Now the constraint can be easily scaled as other equilibrium reactions where usually something divided by their Keq.

While this `solubility_product` is still not perfect in dealing with small concentrations. We have the `log_solubility_product`, where `Q` = ln(Ksp)-ln(f(c)), with bounds [0,ln(Ksp)-ln(0)]. The upper bound ln(Ksp)-ln(1e-50) is ~1e2 considering the lower bounds of `mole_frac_phase_comp` and for common low concentrations, Q is ~1e1. This is why I consider a factor of 10 of the precipitate term in `log_solubility_product`.

To compare `s` and `Q` (decide which one is the zero term), we need similar

idaes/models/properties/modular_properties/reactions/equilibrium_forms.py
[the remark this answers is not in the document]
lxhowl: Besides the comments above, to compare `s` and `Q` (decide which one is the zero term), we also need similar sensitivity of both terms at low concentrations. `Q` in log form is very sensitive to low concentrations and the first derivative of Q, dQ=1/f(c).

The first derivative of the updated `s` term is dS=Ksp/(s+Ksp)^2, for s << Ksp (low precipitate), dS=1/Ksp, Ksp>=f(c), so I consider a factor that also matches the sensitivity and 10 works best.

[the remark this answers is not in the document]
andrewlee94: OK, I think I understand where the 10 is coming from, but that means I think it should be a mutable param as well rather than a hard-coded value so that users can tune this parameter if required.

idaes/models/properties/modular_properties/reactions/equilibrium_forms.py
lxhowl: See the derivatives in the following comment, a=Ksp is a very robust setup considering the sensitivity of f(c) in `Q`, and also Ksp is a constant property. Though Ksp varies with temperature (similar to Keq), it is always in a similar order of magnitude as f(c).

In general, this formulation is to refine the zero value comparison between `s` and `Q` and guarantee a similar accuracy for both terms regardless of any scaling provided.

andrewlee94: But `Ksp` is not constant (unless you make a simplifying assumption). My question is whether `a` should be a constant of similar magnitude to `Ksp` instead; this would help with sparsity, scaling and initialization as well as it would be `a` and `Ksp` separate.

lxhowl: Maybe `k_eq_ref`? The Ksp at the reference temperature. Ksp will be larger when the temperature increases, and it will still keep accuracy in small `s`. But when the temperature goes below the reference temperature, Ksp will decrease, and we may just lose 1-2 orders of magnitude accuracy.

andrewlee94: I would not couple it explicitly to any physical variable - just make it a separate smoothing/scaling parameter that the user can set. `K_eq_ref` would be a good default value, but the issue again is that that is not always known a priori. Also, by uncoupling it it lets the (experienced) user pick a value that is good for their application; e.g. if they know they are notably above the conditions for the reference state, then they can estimate `K_sp` for that state and use that as the smoothing parameter.

idaes/models_extra/column_models/tests/test_plate_heat_exchanger.py
```diff
@@ -163,7 +162,8 @@ def test_dof(self, phe):
     @pytest.mark.component
     def test_initialize(self, phe):
         initialization_tester(
-            phe, duty=(245000, pyunits.W), optarg={"bound_push": 1e-8, "mu_init": 1e-8}
+            phe,
+            duty=(245000, pyunits.W),  # , optarg={"bound_push": 1e-8, "mu_init": 1e-8}
```
andrewlee94, line 166: @lxhowl Your PR should not have affected these tests - we need to look into the reason why these tests were failing a bit more to see why they were failing before doing anything to these (which might be in a separate PR).

idaes/models/properties/modular_properties/state_definitions/electrolyte_states.py
```diff
@@ -85,7 +85,7 @@ def _apparent_species_state(b):
     b.mole_frac_phase_comp_true = Var(
         b.params.true_phase_component_set,
         initialize=1 / len(b.params.true_species_set),
-        bounds=(1e-20, 1.001),
+        bounds=(0, 1.001),
```
andrewlee94, line 88: There is a good reason why these were set to a small number instead of 0 - if mole fraction goes to zero many of the equation in the properties can become singular thus it is necessary to set the bound away from zero. In the rare cases where someone is trying to track mole fractions with values of less than 1e-20 they should change this bound in the model after creation (at which point I have to ask if that is even meaningful).

andrewlee94 (changes requested):
@lxhowl Changing the bounds on the mole fractions is not the correct way to address the variables outside of bounds warning - that is something that needs to be fixed in either the model or the tool that is setting the value outside the bounds. The value of 1w-20 was deliberately chose to avoid singularities, and this change is likely what caused the testing in the plate heat exchanger model to start failing.

agarciadiego:
@lxhowl if the user does not specify `k_eq_ref` how robust is the `s_norm` default value? Have we tested this with another example where the `k_eq_ref` is unknown? @andrewlee94 would it make more sense to have an if statement to use this formulation in case the user has an a priori knowledge of `k_eq_ref`

andrewlee94:
@agarciadiego I am leaning towards just having the one version - the old one already had a smoothing factor that the user should be setting anyway so this just adds more flexibility (and better scaling overall).

If we think it is necessary to keep both formulations, then I think they should be separate methods the user chooses from rather than an if statement in the code. An if buried in the code is far harder for the user to find and understand.

@lxhowl This discussion did remind me however that you need to update the documentation to reflect the new formulation and explain the new parameters (and how the user should adjust them).

lxhowl:
@agarciadiego The tests I updated contains cases without `k_eq_ref` specified. I will add assertions for `s_norm` and `s_scale` default values checking. 

The formulation of the updated `s` term is meant to keep it in the same level of accuracy as the `Q` term. As `s_norm` decreases, low concentration `s` can be captured. The default s_norm value 1e-4 without `k_eq_ref` specified is set in considerations of common Ksp range 1e-37~1e-4.

idaes/models/properties/modular_properties/reactions/equilibrium_forms.py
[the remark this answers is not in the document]
dallan-keylogic: > Q in log form is very sensitive to low concentrations and the first derivative of Q, dQ=1/f(c).

Could you go more into the sensitivity to low concentrations? It looks like it's using the log concentration forms like it should.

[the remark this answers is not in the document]
lxhowl: > > Q in log form is very sensitive to low concentrations and the first derivative of Q, dQ=1/f(c).
> 
> Could you go more into the sensitivity to low concentrations? It looks like it's using the log concentration forms like it should.

Yes, this is an improved form proposed for `Q` in log form at the beginning. As you may know, the log transform of the equilibrium reaction would perform better at low concentrations. It also improves the sensitivity of Q over component concentrations for solubility products. 

The constraint here is to decide which is the "exact" zero term between `Q` and `S`. So for those low-concentration cases (e.g., Ksp ~1e-20), the reformated precipitate term S=s/(s+Ksp) is going 1 when `s` is relatively large compared to Ksp or 0 when `s` is relatively small compared to Ksp. `s` in this way is able to keep a similar precision as f(c). This idea also works for the non-log-transformed form.

Besides a similar precision, we also prefer a similar sensitivity of `Q` and `S` over state variables in this constraint. The first derivatives in previous comments are evaluated for dQ/df(c) and dS/ds for log-transformed form. This is the form we found better performance in some low concentration cases from WaterTAP.

lbianchi-lbl:
The test failures are likely due to #1215

idaes/models/properties/modular_properties/reactions/equilibrium_forms.py
```diff
@@ -284,10 +314,12 @@ def return_expression(b, rblock, r_idx, T):
             if sunits is not None:
                 s = s / sunits
 
+        s = rblock.s_scale * s / (s + rblock.s_norm)
+
         Q = b.log_k_eq[r_idx] - e
         # Q should be unitless due to log form
 
-        return s - smooth_max(0, s - Q, rblock.eps) == 0
+        return Q - smooth_max(0, Q - s, rblock.eps) == 0
```
eslickj, line 322: If one or both Q and s need to be zero, could this just bee smooth min(Q, s) == 0?

lxhowl: Current expression goes `Q-0.5(Q-s+|Q-s|) == 0` while the smooth min gives `0.5(Q+s-|Q-s|)==0`. The complexity remains the same. The current expression keeps Q as a leading term, which might be easy for users to scale the solubility product with other equilibrium reactions.

andrewlee94: @eslickj I think this formulation is correct - it is a bit more complicated than just one of Q or s needing to be zero.

lxhowl:
> @lxhowl Is this ready for re-review yet?
YES, it is ready.

idaes/models/properties/modular_properties/reactions/equilibrium_forms.py
```diff
@@ -131,6 +131,20 @@ def build_parameters(rblock, config):
             mutable=True, initialize=1e-4, doc="Smoothing parameter for smooth maximum"
         )
 
+        rblock.s_norm = Param(
+            mutable=True,
+            initialize=1e-4,
+            doc="Normalizing factor for solid precipitation term",
+        )
+        if hasattr(rblock, "k_eq_ref"):
+            rblock.s_norm.value = rblock.k_eq_ref.value
```
andrewlee94, line 140: I think it would be better to do this as `rblock.s_norm.set_value(rblock.k_eq_ref)`; this will let the set value method handle any necessary unit conversions or edge-cases that might appear.

idaes/models/properties/modular_properties/reactions/equilibrium_forms.py
```diff
@@ -234,6 +250,20 @@ def build_parameters(rblock, config):
...
```
andrewlee94, line 259: As above.

idaes/models/properties/modular_properties/reactions/tests/test_solubility_product_verification.py
```diff
@@ -52,6 +52,7 @@
 
 
 solver = get_solver()
+solver.options["tol"] = 1e-8
```
andrewlee94, line 55: Is this necessary? I am generally hesitant when a test requires changing a solver option as there can be side effects of this.

idaes/models/properties/modular_properties/reactions/equilibrium_forms.py
eslickj: @andrewlee94, I'm sure you are probably right.  Can someone explain the difference to me?  Why is this one preferred?  I think both should work, but I may be missing something. To me, the the one in the code currently just seems convoluted.

idaes/models/properties/modular_properties/reactions/equilibrium_forms.py
lxhowl: `rblock.k_eq_ref` is not a Param but a Var, it does not allow me to use `set_value` here.

idaes/models/properties/modular_properties/reactions/tests/test_solubility_product_verification.py
bknueven: It seems to be necessary for these models for reliable convergence.

I was confused about this at first, as I thought `tol` just affected the convergence criterion. But it turns out it also controls how small the barrier parameter `\mu` can get, based on equation (7) in the [implementation paper](http://cepac.cheme.cmu.edu/pasilectures/biegler/ipopt.pdf).

Occasionally these are converging to a point of local infeasibility with the larger lower bound on `\mu`.

andrewlee94: So it is probably something that could be fixed with scaling then. For now I think it is probably Ok to do this then.

idaes/models/properties/modular_properties/reactions/equilibrium_forms.py
lxhowl: @andrewlee94 
image

andrewlee94: I thought that should have worked - either way as you noted it works if you use `value()` so we should go with that.

idaes/models/properties/modular_properties/reactions/tests/test_solubility_product_verification.py
eslickj: I've been looking at something like this elsewhere, and I may have a suggestion.  I'm having trouble quickly tracing the variables and their bounds, but I speculate that the solid amount and saturation index (or whatever is used here) both have bounds of 0.  The complementarity constraint enforces that one is 0 and the other is more than 0, so the bounds are not necessary.  In my case, when I remove the bounds of the saturation index and solid amount, the problem solves much better.  The problem is hard enough without IPOPT pushing it away from the solution.

dallan-keylogic: The problem is having both an active bound for one variable at zero while complementarity simultaneously forces it to be zero. (I think) it causes linearly dependent rows in the KKT system, which causes IPOPT problems and causes dual variables to blow up (because they're no longer unique). I'm afraid of just removing both bounds, though, because then the complementarity might find a solution where both are negative.

eslickj: I'm not sure there is a solution where both are negative.  The constraint is roughly min(SI, Csolid) = 0.  I think that means they both have to be >= 0.

idaes/models/properties/modular_properties/reactions/equilibrium_forms.py
andrewlee94: @eslickj I cannot remember the details - I asked Larry Biegler to help with this is and this is the form he suggested.

andrewlee94 (approved):
Looks good to me.

agarciadiego (approved):

PR #1209, opened 2023-06-15
Fix bug related to cloning modular properties methods

### LINKED ISSUE #1206: Modular property methods do not work for cloned models
@radhakrishnatg identified an issue where cloning a model using the modular properties and then running `initialize` and failed with degrees of freedom issues. Further investigation revealed that the `define_state_vars` method on the clone state block was point to variables in the original model rather than the clone and thus running `initialize` was changing degrees of freedom on the original model instead of the clone.

The reason for this is because the modular properties creates the `define_state_vars` method dynamically for each instance of the state block as a static method with no arguments, and thus it does not update when the model is cloned. This also applies to all other methods dynamically created by the state definition functions.

The easiest solution is to make the object an argument to these methods in all of the state definition modules and assign them as bound methods rather than a static methods. We should also add a test for cloning a modular state block and check that the cloned methods point to components on the right model.

radhakrishnatg:
## Fixes
#1206 

- Changed staticmethods to classmethods to fix the bug
- Added a test to check methods point to variables on the clone 

### Legal Acknowledgement


radhakrishnatg:
@andrewlee94 Please let me know if I missed any methods.

andrewlee94 (approved):
@radhakrishnatg I think you caught everything. Thank you.

eslickj (approved):

PR #1210, opened 2023-06-20
Show platform name when binaries not found with bin-platform command

eslickj:
## Summary/Motivation:

The ``idaes bin-platform`` command tells you which set of binaries will be downloaded, but if there is no binary set mapped for your specific platform, it would just say binaries are not found.  This modifies the output so that it tells you how your platform is identified if binaries are not found.  This will make it a little easier to identify and map platforms to binaries that work for them, or to identify additional platforms where binary builds may be needed.

### Legal Acknowledgement


andrewlee94 (approved):

OOAmusat (approved):
LGTM

PR #1199, opened 2023-06-08
Update IDAES-EXT version

eslickj:
New version of IDAES-EXT with these changes

- A function was added to calculate x/(exp(x) - 1), which deals with x = 0, by using a Taylor series expansion near x = 0.
- Isothermal compressibility and fugacity coefficient functions were added to the Helmholtz EoS functions
- PETSc (https://petsc.org/release/) was updated to 3.19.2

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1211, opened 2023-06-20
Add option to change the initial condition solver.

eslickj:
## Summary/Motivation:

This adds an option to use a different nonlinear solver for the initial condition solving step.  This is motivated by the failures in https://github.com/IDAES/idaes-pse/pull/1199, but it doesn't solve the issue.  It provides more options for potentially addressing or debugging it. 

### Legal Acknowledgement


andrewlee94 (approved):
Looks OK to me - my only question is if there is something that should be updated in the docs about this (beyond just the doc string).

eslickj:
@andrewlee94, I looked at the docs, and I think the only update needed is the docstring.

bpaul4 (approved):

eslickj:
@lbianchi-lbl or @andrewlee94, it looks like the example tests are failing, but as far as I can tell it is unrelated to this PR.

bpaul4:
@eslickj yes, the failures appear unrelated. I opened an issue in #1215.

PR #1224, opened 2023-07-18
Fix for decorators in StateBlocks

eslickj:
## Fixes

This should fix decorators for Expressions and Constraints not working in state blocks.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/base/property_base.py
+++ b/idaes/core/base/property_base.py
@@ -782,6 +782,9 @@ should be constructed in this state block,
                   component.
         """
         try:
+            # try the Pyomo Block's __getattr__ method first which will return 
+            # decorators for creating components on the block (e.g. Expression, 
+            # Constraint, ...).
             return super().__getattr__(attr)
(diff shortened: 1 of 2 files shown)


idaes/core/base/property_base.py
```diff
@@ -781,7 +781,10 @@ def __getattr__(self, attr):
             attr: an attribute to create and return. Should be a property
                   component.
         """
-        return build_on_demand(self, attr)
+        try:
+            return super().__getattr__(attr)
```
andrewlee94, line 785: A comment might be useful here to remind us why this is needed in future.

andrewlee94 (changes requested):
This looks good, and I think I can see how and why it is necessary. My main request is if you could also add this to the reaction base class, as I believe there is a separate but equivalent method there.

dallan-keylogic (approved):

andrewlee94 (approved):

PR #1228, opened 2023-07-25
DiagnosticsToolbox

andrewlee94:
## Summary/Motivation:
This PR adds the first iteration of the new `DiagnosticsToolbox` class. Documentation and tests are coming, but we need input and feedback on the API before it gets set in stone.

A draft tutorial can be found here: https://app.box.com/folder/218647852677 (for those with access to the IDAES Box folder).

- New `DiagnosticsToolbox` class
-

### Legal Acknowledgement


idaes/core/util/model_diagnostics.py
```diff
@@ -17,29 +17,851 @@
...
+      but these will not appear in the Next Steps section.
+    3. Next Steps - these are recommended methods to call from the DiagnosticsToolbox to
+      get further information on warnings. If no warnings are found, this will suggest
+      the next report method to call.
+
+    """
+
+    def __init__(self, model: Block):
+        if not isinstance(model, Block):
+            raise ValueError("model argument must be an instance of a Pyomo Block.")
```
Robbybp, line 131: Should this be `TypeError`?

idaes/core/util/model_diagnostics.py
```diff
@@ -17,29 +17,851 @@
...
+        self.zero_tolerance = 1e-6
+        # TODO: Add scaling tolerance parameters
+
+    def _vars_fixed_to_zero(self):
+        # Set of variables fixed to 0
+        zero_vars = ComponentSet()
+        for v in self.model.component_data_objects(Var, descend_into=True):
+            if v.fixed and value(v) == 0:
+                zero_vars.add(v)
+        return zero_vars
```
Robbybp, line 144: Is there any reason this needs to be a method on the class? Rather than just a function, e.g. `_vars_fixed_to_zero(model, tolerance=0.0)`?

idaes/core/util/model_diagnostics.py
```diff
@@ -17,29 +17,851 @@
...
+        warnings, _ = self._collect_numerical_warnings()
+        if len(warnings) > 0:
+            raise AssertionError(f"Numerical issues found ({len(warnings)}).")
+
+    def _write_report_section(
+        self,
+        stream,
+        lines_list,
+        title=None,
+        else_line=None,
```
Robbybp, line 679: Should `else_line` have a more descriptive name? E.g. `line_if_empty`?

idaes/core/util/model_diagnostics.py
```diff
@@ -17,29 +17,851 @@
...
+        """
+        self._write_report_section(
+            stream=stream,
+            lines_list=large_residuals_set(self.model, tol=self.residual_tolerance),
+            title=f"The following constraint(s) have large residuals:",
+            header="=",
+            footer="=",
+        )
+
+    def check_dulmage_mendelsohn_partition(self):
```
Robbybp, line 402: As we are not really "checking" anything here, should this method be named something like `get_dulmage_mendelsohn_partition`?

idaes/core/util/model_diagnostics.py
```diff
@@ -17,29 +17,851 @@
...
+        igraph = IncidenceGraphInterface(self.model)
+        var_dm_partition, con_dm_partition = igraph.dulmage_mendelsohn()
+
+        # Collect under- and order-constrained sub-system
+        uc_var = var_dm_partition.unmatched + var_dm_partition.underconstrained
+        uc_con = con_dm_partition.underconstrained
+        oc_var = var_dm_partition.overconstrained
+        oc_con = con_dm_partition.overconstrained + con_dm_partition.unmatched
+
+        return uc_var, uc_con, oc_var, oc_con
```
Robbybp, line 423: For large under/over constrained subsystems, decomposing into independent blocks (connected components) is very useful. This can be done with:
```python
uc_vblocks, uc_cblocks = igraph.get_connected_components(uc_var, uc_con)
oc_vblocks, oc_cblocks = igraph.get_connected_components(oc_var, oc_con)
```
The partition of the systems are lists-of-lists, so the overconstrained variables can be recovered with `oc_con = sum(oc_vblocks, [])`, or these independent blocks can be printed individually (which I would recommend).

idaes/core/util/model_diagnostics.py
Robbybp: As an aside, is there any reason that this needs to be a method on the class, rather than a function that takes `model` as an argument? I guess the reason is that, if we wanted to optimize things later, we could have the incidence graph cached?

idaes/core/util/model_diagnostics.py
```diff
@@ -17,29 +17,851 @@
...
+                free_vars_in_constraints.add(v)
+                if not _var_in_block(v, self.model):
+                    ext_free_vars_in_constraints.add(v)
+                if v.lb is not None:
+                    if v.ub is not None:
+                        free_vars_lbub.add(v)
+                    else:
+                        free_vars_lb.add(v)
+                elif v.ub is not None:
+                    free_vars_ub.add(v)
```
Robbybp, line 756: This seems like a lot of logic for a method that is presumably just doing some reporting. I guess it is not used elsewhere?

idaes/core/util/model_diagnostics.py
```diff
@@ -17,29 +17,851 @@
...
+
+    """
+
+    def __init__(self, model: Block):
+        if not isinstance(model, Block):
+            raise ValueError("model argument must be an instance of a Pyomo Block.")
+        self.model = model
+        # TODO: Work out how to manage and document these
+        self.residual_tolerance = 1e-5
+        self.zero_tolerance = 1e-6
```
Robbybp, line 135: These should probably be arguments in the constructor, right?

Robbybp (commented):
Overall, this looks really good. Several minor comments, then I'd recommend printing out the independent diagonal blocks of the overconstrained/underconstrained subsystems individually, as this is often much easier to read if these sets decompose.

idaes/core/util/model_diagnostics.py
andrewlee94: The intention was to do that eventually - I just wasn't sure of the best way to do it yet (separate arguments versus passing to a Config Block). Suggestions are welcome.

idaes/core/util/model_diagnostics.py
andrewlee94: Thanks for the example - this was on my list of things to do.

idaes/core/util/model_diagnostics.py
andrewlee94: Caching was one reason, but also bringing everything into a single class so it was easy to find for the user (most of the issues we got from the interviews could be summed up as "I didn't know it existed").

idaes/core/util/model_diagnostics.py
andrewlee94: This probably should be moved out to a function so that it doesn't clutter up the doc string.

idaes/core/util/model_diagnostics.py
Robbybp: I think a config block makes sense. That way it's pretty easy to also pass these arguments into e.g. the `report_structural_issues` method without having to make sure the call signatures stay in sync.

andrewlee94:
@dallan-keylogic The poorly scaled variables check currently uses the `badly_scaled_variables` method out of the scaling toolbox, so it is a rather simple check. We used this for the prototype in order to have a basic demonstration using existing tools that were robust - I hope to either supplement or replace this with an svd tool once we have time to address the current issues and limitation with that tool.

andrewlee94:
@Robbybp @adowling2 @dallan-keylogic @jsiirola @mrmundt The initial version of the Diagnostics Toolbox is now ready for review so that we can get it into the upcoming August IDAES release.

I have added lots of testing and some basic documentation.

andrewlee94:
@Robbybp Would you have time to review this? We want to get this into the August release in two weeks and I am struggling to find volunteers.

Robbybp:
Yes, I can review this, when do you need a review by?

andrewlee94:
@Robbybp Some time next week should be fine. I think the release date is the 31st so we should try to get this merged before then (plus you and I are both at Sandia the week of the release).

idaes/core/util/model_diagnostics.py
```diff
@@ -17,29 +17,987 @@
...
+
+    To get started:
+
+      1. Create an instance of your model (this does not need to be initialized yet).
+      2. Fix variables until you have 0 degrees of freedom. Many of these tools presume
+         a square model, and a square model should always be the foundation of any more
+         advanced model.
+      3. Create an instance of the DiagnosticsToolbox and provide the model to debug as
+         the model argument.
+      4. Call the report_structural_issues() method.
```
Robbybp, line 180: Can you format "report_structural_issues()" as code? I.e. with double back-ticks

docs/explanations/model_diagnostics/index.rst
```diff
@@ -0,0 +1,45 @@
...
+
+Once you have at least a partial solution to your model, the next step is to check for numerical issues in your model, such as possible bounds violations and poor scaling. Due to their nature, numerical issues depend on having a solution to the model, and they can often be limited to certain model states (i.e., it is possible to have a model which behaves well at one model state only to fail badly if you change your model state). As such, numerical checks should be performed at a number of points across the full range of expected states to try to ensure that the model is well-posed across the full modeling range. Any issues that are identified here should be addressed before moving on, and remember that after any changes you should always start by checking for structural issues again.
```
Robbybp, line 28: I think this explanation (and the accompanying diagram) is really good. However, the explanation is a large amount of text that I think people (myself included) are likely to skip over. It might be nice to separate the paragraphs into titled subsections so people can easily skip to the subsection they are interested in. So that these paragraphs can be sub-sections of "General Workflow", maybe the document should use the format
```rst
==========================
Model Diagnostics Workflow
==========================
...

General Workflow
================
...

Choose a model to debug
-----------------------

Start with a square model
-------------------------

Check for structural issues
---------------------------

Check for numeric issues
------------------------

etc.
```

Robbybp (commented):
A few comments on documentation before I review the Python code

idaes/core/util/model_diagnostics.py
```diff
@@ -17,29 +17,987 @@
...
+
+        stream.write("=" * MAX_STR_LENGTH + "\n")
+
+    def display_variables_with_extreme_jacobians(self, stream=stdout):
+        """
+        Prints the variables associated with columns in the Jacobian with extreme
+        L2 norms. This often indicates poorly scaled variables.
+
+        Tolerances can be set via the DiagnosticsToolbox config.
+
```
Robbybp, line 542: What do you think about allowing tolerances to be set in this function? E.g.
```python
dt = DiagnosticsToolbox(m, jacobian_large_value_caution=1e8)
# Display with 1e8 threshold
dt.display_variables_with_extreme_jacobians(m)
# If the output is too noisy, we may want to display with a higher threshold
# to see only the worst offenders
dt.display_variables_with_extreme_jacobians(m, jacobian_large_value_caution=1e10)
```
This would require something like:
```python
def display_variables_with_extreme_jacobians(self, stream=stdout, **kwds):
    config = self.config(kwds)
    ...
```

I think this is more user-friendly (allows setting tolerances closer to where you actually use them), although the advantage is relatively minor as long as constructing another `DiagnosticsToolbox` is cheap.

Robbybp: This goes for all the similar "numerical issues" functions. Leveraging `self.config` would allow these methods to all have consistent call signatures.

idaes/core/util/model_diagnostics.py
```diff
@@ -17,29 +17,987 @@
...
+        the over- and under-constraint sub-problems.
+
+        Returns:
+            list-of-lists variables in each independent block of the under-constrained set
+            list-of-lists constraints in each independent block of the under-constrained set
+            list-of-lists variables in each independent block of the over-constrained set
+            list-of-lists constraints in each independent block of the over-constrained set
+
+        """
+        igraph = IncidenceGraphInterface(self.config.model)
```
Robbybp, line 456: We probably want `IncidenceGraphInterface(model, include_inequality=False)`

idaes/core/util/model_diagnostics.py
```diff
@@ -17,29 +17,987 @@
...
+    ),
+)
+CONFIG.declare(
+    "relative_tolerance",
+    ConfigValue(
+        default=1e-4,
+        domain=float,
+        description="Relative tolerance to for variables close to bounds.",
+    ),
+)
```
Robbybp, line 100: These should have more descriptive names, so we know they only apply to variable-bound differences.

idaes/core/util/model_diagnostics.py
```diff
@@ -17,29 +17,987 @@
...
 
 _log = idaeslog.getLogger(__name__)
 
 
+MAX_STR_LENGTH = 84
+TAB = " " * 4
+
+# TODO: Add suggested steps to cautions - how?
+
+CONFIG = ConfigDict()
```
Robbybp, line 83: I'm a little concerned that there are too many config options, which will be hard to remember. Especially since some of these could be easily confused (e.g. one might misuse `zero_value_tolerance` thinking that it applies to residuals).

What is the minimum set of tolerances that any single method needs? It seems like just a "zero tolerance" and possibly a "small tolerance". Then large value thresholds can just default to the inverse of these tolerances (or `float("inf")` if either is exactly zero). 

What do you think about a ConfigDict like this:
```python
from pyomo.common.config import ConfigValue, ConfigDict

CONFIG = ConfigDict()
CONFIG.declare("tolerance", ConfigValue(default=1e-8, domain=float))
CONFIG.declare("caution_tolerance", ConfigValue(default=1e-4, domain=float))
CONFIG.declare("large_value_threshold_warning", ConfigValue(default=1e8, domain=float))
CONFIG.declare("large_value_threshold_caution", ConfigValue(default=1e4, domain=float))
```
Then individual methods can just use e.g. `tolerance` for whatever is relevant for the particular method. This would work better if the individual methods can accept `**kwds`, so that one can call two methods with different tolerances, without having to know exactly what the right config arg for each method is:
```python
dt.display_variable_at_or_outside_bounds(tolerance=1e-6)
dt.display_constraints_with_extreme_jacobians(tolerance=1e-10)
```

I guess the downside of this is that we can't fine-tune each individual tolerance then call `report_numerical_issues`. There is also a potential backward compatibility issue if we implement a general `tolerance` arg then later decide we need more fine-grained options.

Robbybp: Maybe the way to go is to just stick with the fine-grained tolerance options, but add some way to print all the config options so they can be easily discovered.

Robbybp (commented):
Some minor change requests.

I left some ideas on making config arguments more user-friendly (i.e. reduce number of arguments and allow them to be passed to individual methods), but I'm not sure the benefits outweight the drawbacks on these. My only concrete request on config arguments is to add a method to print all argument names and their descriptions.

idaes/core/util/model_diagnostics.py
andrewlee94: Firstly, the config block is automatically included in the class doc string (via the Pyomo decorator magic), so user can easily see all the config arguments. This also extends to the documentation generated by Sphinx so it is all easy to see.

Secondly, I did start with fewer tolerances, but came to found that I wanted most of them as I generally wanted different tolerances for the Jacobian versus the residuals, etc.

Robbybp: > Firstly, the config block is automatically included in the class doc string (via the Pyomo decorator magic), so user can easily see all the config arguments. This also extends to the documentation generated by Sphinx so it is all easy to see.

Great, I didn't know (or forgot) that. I see now that what I'm asking for is a subset of what `help(DiagnosticsToolbox)` already does.

idaes/core/util/model_diagnostics.py
Robbybp: > This is one case where I intended config blocks to be changed dynamically as required - the methods all look to the config block for the tolerances and there is nothing preventing users from changing when they want to.

If we expect users to change tolerances in the config, I would rather they do it with function arguments than by dynamically altering `self.config`.

> That said, this does run counter to how config blocks are used elsewhere in IDAES where they are only checked once at build time, and thus we discourage changing them after the fact.

But even in IDAES, IIRC initialize methods accept solver arguments at call-time rather than (only) as config arguments. I think allowing this is more intuitive, but I also don't think it is necessary (yet) so have no problem with leaving it out.

idaes/core/util/model_diagnostics.py
```diff
@@ -847,10 +1808,178 @@ def ipopt_solve_halt_on_error(model, options=None):
...
+
+
+def _vars_violating_bounds(model):
+    violated_bounds = ComponentSet()
+    for v in model.component_data_objects(Var, descend_into=True):
+        if v.value is not None:
+            if v.lb is not None and v.value <= v.lb:
+                violated_bounds.add(v)
+            elif v.ub is not None and v.value >= v.ub:
+                violated_bounds.add(v)
```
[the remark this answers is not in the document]
Robbybp: I typically wouldn't regard a variable being `1e-8` above its upper bound as a violation. So to see the "real violations", the user should take the difference between variables that violate their bounds and variables that are at their bounds?

idaes/core/util/model_diagnostics.py
```diff
@@ -17,29 +17,987 @@
...
+       but these will not appear in the Next Steps section.
+    3. Next Steps - these are recommended methods to call from the DiagnosticsToolbox to
+       get further information on warnings. If no warnings are found, this will suggest
+       the next report method to call.
+
+    """
+
+    def __init__(self, **kwargs):
+        self.config = CONFIG(kwargs)
+        if not isinstance(self.config.model, Block):
```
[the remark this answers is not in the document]
Robbybp: `IncidenceGraphInterface` and `PyomoNLP` do not support IndexedBlocks, for instance. And I suspect many other tools, like `large_residuals_set` do not as well. Conceptually, we could update all the tools to support `IndexedBlock`, but I don't think this is a good idea.

idaes/core/util/model_diagnostics.py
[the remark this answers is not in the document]
andrewlee94: I added the tolerance, and a new config argument for it with a default tolerance of 0 (mostly to start with the most strict definition and allow users to relax it as required depending on their problem and behaviour of their preferred solver).

idaes/core/util/model_diagnostics.py
andrewlee94: The old initialization approach used arguments at call time, but the new `Initializer` objects actually work in a similar fashion to this (a config block which users can dynamically update - for better or worse).

idaes/core/util/model_diagnostics.py
```diff
@@ -17,29 +17,987 @@
...
 _log = idaeslog.getLogger(__name__)
 
 
+MAX_STR_LENGTH = 84
+TAB = " " * 4
+
+# TODO: Add suggested steps to cautions - how?
+
+CONFIG = ConfigDict()
+CONFIG.declare("model", ConfigValue(description="Pyomo model object to be diagnosed."))
```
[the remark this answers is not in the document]
andrewlee94: I've changed `model` to be a positional argument which is used to set a private `self._model` attribute (and updated the docs accordingly). I've also added `self.model` as a `@property` so that users can access (but not change) it easily as well as an explicit `self.set_model(model)` method to allow users to change the model they are looking at.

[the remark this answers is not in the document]
Robbybp: Do we expect users to need to (a) retrieve and (b) change the model that their `DiagnosticsToolbox` instance is currently debugging?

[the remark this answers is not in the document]
andrewlee94: Maybe - being able to retrieve it lets you confirm what model (or Block) you are looking at.

As for changing the model, I could see it being useful for a few reasons, such as switching between scaled and unscaled version of the same model, or when trying to debug more complex hierarchical models whilst keeping your current config options (e.g. moving from `m.b` to `m`). That said, creating a new instance of the toolbox is also cheap so you could argue it is better to just make users create a new instance for clarity (although then you have multiple instances of the toolbox around which could be just as confusing).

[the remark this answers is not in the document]
Robbybp: Changing to a "sub-model" with the same config settings is the most compelling reason to me. Although you could just create a new instance and use `**old_dtbox.config` as keyword args.

I like the `model` property, but at the moment I'm against letting users change the model. Letting the model change adds more mutable state that the user has to keep track of, and there is a pretty easy workaround (create a new instance). It will also make `DiagnosticsToolbox` more difficult to extend, because if we ever do any caching, we will have to invalidate all the caches.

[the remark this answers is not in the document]
andrewlee94: I've removed the `set_model` method.

docs/explanations/model_diagnostics/index.rst
```diff
@@ -0,0 +1,63 @@
...
+
+Start with a Square Model
+"""""""""""""""""""""""""
+
+Next, you should ensure you model has zero degrees of freedom (as best you can); whilst your ultimate goal may be to some an optimization problem with degrees of freedom, you should always start from a square model first. Firstly, this is because many of the model diagnosis tools work best with square models. Secondly, all models are based on the foundation of a square model; an optimization problem is just a square model with some degrees of freedom added. If your underlying square model is not well-posed then any more advanced problem you solve based on it is fundamentally flawed (even if it solves).
```
Robbybp, line 28: Typo: "**your** model"

lbianchi-lbl, line 28:
```suggestion
Next, you should ensure you model has zero degrees of freedom (as best you can); whilst your ultimate goal may be to solve an optimization problem with degrees of freedom, you should always start from a square model first. Firstly, this is because many of the model diagnosis tools work best with square models. Secondly, all models are based on the foundation of a square model; an optimization problem is just a square model with some degrees of freedom added. If your underlying square model is not well-posed then any more advanced problem you solve based on it is fundamentally flawed (even if it solves).
```

Here I'm not sure if "to some" was intended or a typo, as both "to solve" and "to some [users]" could make sense in the context.

idaes/core/util/model_diagnostics.py
```diff
@@ -17,29 +17,1022 @@
...
+       solver issues. Not all cautions need to be addressed, but users should investigate
+       each one to ensure that the behavior is correct and that they will not be the source
+       of difficulties later. Methods exist to provide more information on all cautions,
+       but these will not appear in the Next Steps section.
+    3. Next Steps - these are recommended methods to call from the DiagnosticsToolbox to
+       get further information on warnings. If no warnings are found, this will suggest
+       the next report method to call.
+
+    Args:
+        model - model to be diagnosed. The DiagnosticsToolbox does not support indexed Blocks.
```
Robbybp, line 218: Screenshot 2023-08-24 at 5 27 37 PM

This is what I see in the docs. Not sure exactly how to get this to format properly. Maybe `model: Model to be diagnosed...`?

Robbybp (approved):
Other than the docstring formatting, I think this is ready.

docs/explanations/model_diagnostics/index.rst
```diff
@@ -0,0 +1,63 @@
...
+==========================
+
+.. contents::
+    :depth: 3
+    :local:
+
+Introduction
+------------
+
+Writing well-posed equation-oriented models is a significant challenge, and even the most experienced developers often have to spend a lot of time diagnosing and resolving issues before a model is able to solve reliably. This documentation is intended to assist users with this process by outlining a general workflow for model development and debugging in order to more easily identify and resolve modeling issues.
```
Robbybp, line 11: Currently the link to the DiagnosticsToolbox API reference is all the way down at the bottom of this page. Maybe it would be helpful to add the sentence
```rst
A detailed description of the ``DiagnosticsToolbox`` API can be found :ref:`here<reference_guides/core/util/diagnostics/diagnostics_toolbox:Diagnostics Toolbox>`
```
to this paragraph.

idaes/core/util/tests/test_model_diagnostics.py
```diff
@@ -82,6 +1261,22 @@ def v_exp():
...
+@pytest.mark.unit
+def test_deprecate_degeneracy_hunter(caplog):
+    m = ConcreteModel()
+    m.v = Var()
+    m.o = Objective(expr=m.v)
+    dh = DegeneracyHunter(m)
+
+    msg = (
+        "DEPRECATED: DegeneracyHunter is being deprecated in favor of the new "
+        "DiagnosticsToolbox.  (deprecated in 2.0.0, will be removed in (or after) 3.0.0)"
```
lbianchi-lbl, line 1273:
```suggestion
        "DiagnosticsToolbox.  (deprecated in 2.2.0, will be removed in (or after) 3.0.0)"
```

I think this is what is currently causing the tests to fail.

idaes/core/util/model_diagnostics.py
```diff
@@ -17,29 +17,1023 @@
...
+            dstring = "Degrees"
+            if abs(dof) == 1:
+                dstring = "Degree"
+            warnings.append(f"WARNING: {dof} {dstring} of Freedom")
+        if len(uc) > 0:
+            cstring = "Components"
+            if len(uc) == 1:
+                cstring = "Component"
+            warnings.append(f"WARNING: {len(uc)} {cstring} with inconsistent units")
+            next_steps.append("display_components_with_inconsistent_units()")
```
blnicho, line 689:
```suggestion
            next_steps.append(self.display_components_with_inconsistent_units.__name__+"()")
```

I think we should programmatically generate these method names so that we will get an error here if the method name ever changes. With the current implementation a developer could miss this line when changing method names and wouldn't get an error or test failure.

This comment applies to every place where we are doing `next_steps.append`.

idaes/core/util/model_diagnostics.py
```diff
@@ -17,29 +17,1023 @@
...
+
+        """
+        warnings, _ = self._collect_numerical_warnings()
+        if len(warnings) > 0:
+            raise AssertionError(f"Numerical issues found ({len(warnings)}).")
+
+    def report_structural_issues(self, stream=stdout):
+        """
+        Generates a summary report of any structural issues identified in the model provided
+        and suggest next steps for debugging model.
```
blnicho, line 946:
```suggestion
        and suggests next steps for debugging the model.
```

blnicho (approved):

PR #1230, opened 2023-07-27
Add properties and minor fixes for property sets

eslickj:
## Summary/Motivation:

Add some default properties and minor fixes for property_set

### Legal Acknowledgement


idaes/core/base/property_set.py
```diff
@@ -755,6 +756,11 @@ class StandardPropertySet(PropertySetBase):
         doc="Specific Gibbs Energy (Molar Basis)",
         units="ENERGY_MOLE",
     )
+    ionic_strength = PropertyMetadata(
```
andrewlee94, line 761: Should this be defined in the electrolyte properties instead?

idaes/core/base/property_set.py
```diff
@@ -947,6 +958,80 @@ class StandardPropertySet(PropertySetBase):
         units=pyunits.dimensionless,
     )
 
+    # Log10 terms
```
andrewlee94, line 963: We should probably think about whether we want to support log10 and ln separately, pick one of these as the standard, or just have "log" but let the user decide which base to use (which is probably a bad idea).

If we start supporting both log10 and ln then this list is going to get very long, plus we might want to ask if "log" should really be "ln". We should probably update the doc string for all the base-e terms to explicitly indicate the base (unless we go to a floating base).

idaes/core/base/property_set.py
```diff
@@ -947,6 +958,80 @@ class StandardPropertySet(PropertySetBase):
...
+    log10_pressure = PropertyMetadata(
+        name="log10_pressure",
+        doc="Log of Pressure",
+        units=pyunits.dimensionless,
+    )
+
+    # p*
+    pH = PropertyMetadata(
+        name="pH",
+        doc="pH, -log10([H+])",
```
andrewlee94, line 1023: This should be `-log10(a[H+])` - pH is defined based on activity.

idaes/core/base/property_set.py
eslickj: I think most people would expect log10, at least I would for these.  I could be wrong though, or it may depend on the property and use case.  At least I'd like to use log10 for what I'm working on now.  Not sure of the best approach though.

idaes/core/base/property_set.py
eslickj: Yes, I think that makes sense.  I'll move them.

idaes/core/base/property_set.py
```diff
@@ -40,25 +40,27 @@ def __init__(
...
         )  # This method does some basic validation of input
-
+        self._lock_setattr = True
         # TODO: For future, this would be the place to store default scaling information, etc.
         # TODO: Could also define default bounds, nominal values, etc.
 
     def __setattr__(self, key, value):
-        raise TypeError("Property metadata does not support assignment.")
+        try:
+            assert self._lock_setattr == True
```
andrewlee94, line 58: Should this be an `if/raise`? At the moment you will get an `AssertionError` and the `TypeError` will never trigger.

idaes/core/base/property_set.py
```diff
@@ -364,9 +370,13 @@ def __init__(self, parent):
                     )
 
     def __setattr__(self, key, value):
-        raise TypeError(
-            "PropertySets do not support direct assignment. Please use define_property"
-        )
+        try:
+            assert self._lock_setattr == True
```
andrewlee94, line 374: As above.

pytest.ini
```diff
@@ -1,7 +1,7 @@
 [pytest]
-addopts = --pyargs idaes
-          --durations=100
-          -W ignore
+#addopts = --pyargs idaes
```
andrewlee94, line 2: Should this be here?

idaes/core/base/property_set.py
eslickj: For now I just put a few log10 properties in the ElectrolytePropertySet class, so we can figure this out later.

pytest.ini
eslickj: No, I forgot to revert it back.

idaes/core/base/property_set.py
eslickj: It could be `if hasattr(...) and self._lock_setattr == True:`.  I think this works right though since the tests didn't break. Since the lock `self._lock_setattr = True` is added at the end of `__init__`, the assert will pass and the TypeError will be raised, once you pass the `self._lock_setattr = True` line in `__init__`.  At least I'm pretty sure this is tested, that you can't set new attributes. Maybe the if is easier to understand though.

andrewlee94: It think that would be better, otherwise the user could get a bare `AssertionError`.

andrewlee94 (approved):

agarciadiego (approved):

PR #1232, opened 2023-07-27
Add fixed bed 0D TSA unit model

Daison2102:
## Summary/Motivation:
This PR will add a 0D equilibrium based fixed bed TSA (fixed_bed_tsa0d) unit model to the IDAES library. The fixed_bed_tsa0d unit model consists of a fixed bed four-step temperature swing adsorption (TSA) cycle. The model is implemented as a 0-D equilibrium-based shortcut model without axial variation of states and with a semi-analytical solution of the cyclic steady state. Unit testing is coming.

- Unit model for an equilibrium based fixed bed Temperature Swing Adsorption (TSA) process 
- Associated documentation for fixed bed TSA model

### Legal Acknowledgement


idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4522 @@
...
+        self._add_pressurization_step()
+
+        # build up variables and constraints for adsorption step
+        self._add_adsorption_step()
+
+        # discretization of time domain
+        self._apply_transformation()
+
+        # add performance equations
+        self. _make_performance()
```
andrewlee94, line 480: Unnecessary space

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4522 @@
...
+                        else:
+                            kname = str(i[1:])
+                        var = port_obj.vars[k][time_point, i[1:]]
+                        stream_attributes[n][k+" "+kname] = value(var)
+                        stream_attributes["Units"][k+" "+kname] = \
+                            units.get_units(var)
+
+        return DataFrame.from_dict(stream_attributes, orient="columns")
+
+    def _get_costing_electric_boiler(self):
```
andrewlee94, line 3486: Should this be part of the unit model file? Costing is generally something separate from the unit model. This would also help cut down the size of this file (which is huge).

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4522 @@
...
+
+        var_dict["Amount of CO2 to atmosphere [mol/s]"] = self.emissions_co2
+
+        var_dict[
+            "Concentration of CO2 emitted to atmosphere [ppm]"
+        ] = self.emissions_co2_ppm
+
+        return var_dict
+
+    def summary(self, export=False):
```
andrewlee94, line 3411: This looks like it might be something that should go into a utility function file, rather than part of the unit model itself.

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4522 @@
...
+                ):
+                    iscale.set_scaling_factor(
+                        self.cooling.energy_balance_ode[t], 1e-6)
+
+        # scaling for inlet states
+        iscale.set_scaling_factor(self.flow_mol_in, 1e-3)
+        iscale.set_scaling_factor(self.temperature_in, 1e-2)
+        iscale.set_scaling_factor(self.pressure_in, 1e-5)
+
+    def plots(self):
```
andrewlee94, line 3238: I would suggest moving this out to a separate utility method file, in part to help keep the size of the files down.

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4522 @@
...
+            if f_x0 * f_x2 < 0:
+                x1 = x2
+            else:
+                x0 = x2
+
+            # update counter and set up new condition |f(x_2)| > error
+            iter += 1
+            condition = (f_x2**2)**0.5 > 1
+
+    def initialize_build(blk,
```
andrewlee94, line 2540: We are starting to move to a new API for initialization, so I would recommend we update this to use the new API so that we don;t have to come back later.

andrewlee94 (commented):
As a first comment, the unit model file is huge, which makes it hard to review. I suggest breaking it up into smaller pieces (e.g. moving plotting and utility functions out to a separate file).

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4522 @@
...
+shortcut model. Industrial & Engineering Chemistry Research, 2015, 54(11),
+3027-3038
+
+"""
+# Import Python libraries
+import matplotlib.pyplot as plt
+from pandas import DataFrame
+from sys import stdout
+import textwrap
+import json
```
andrewlee94, line 53: Are all of these imports required?

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4522 @@
...
+from idaes.core.util.model_statistics import degrees_of_freedom
+
+from idaes.core.solvers import get_solver
+from idaes.core.util.tables import stream_table_dataframe_to_string
+from idaes.models.unit_models import (
+    SkeletonUnitModel,
+    Heater
+)
+from idaes.models_extra.power_generation.properties import (
+    FlueGasParameterBlock
```
andrewlee94, line 94: Unit models should generally not import property packages - for one this runs the risks of adding circular imports, and also breaks the encapsulation/isolation of properties from unit models.

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
AlexNoring: moved costing to separate file

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
AlexNoring: moved to a util file

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
AlexNoring: also moved plots to the util file

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
AlexNoring: removed unnecessary imports

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
AlexNoring: @Daison2102 How do you think we should address this? Can the isotherm calculations handle a component list that is different from the one specified within the model?

Daison2102: @andrewlee94 @AlexNoring This model represents more of a flowsheet than a single unit model. The flowsheet is composed of a 4-step TSA cycle + compressor + heater. The two imported property packages are used in the compressor (FlueGasParameterBlock) and the heater (iapws95).
I believe the best way to address Andrew's comment is to add config arguments, allowing the user to pass those property packages if they select the "compressor" or "steam_calculation" config arguments.

AlexNoring: The property packages are now passed as config arguments

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
AlexNoring: Added support for the new initialization API

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4563 @@
...
+            self._add_parameters_mmen_Mg_MOF_74()
+        elif self.config.adsorbent == "polystyrene-amine":
+            self._add_parameters_polystyrene_amine()
+
+        # add design and operating variables
+        self.flow_mol_in_total = Var(
+            initialize=0.014,
+            domain=PositiveReals,
+            units=units.mol / units.s,
+            doc="Total mole flow rate at inlet [mol/s]",
```
andrewlee94, line 1242: I would suggest not including the units in the doc string as this just duplicates the units attribute (and opens up the possibility of them conflicting).

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4563 @@
...
+        """
+        Method to add general required parameters to run fixed bed TSA model.
+        This include fluid and wall column properties, component set,
+        etc.
+
+        """
+        self.component_list = Set(
+            initialize=["N2", "CO2", "H2O", "O2"], doc="List of components"
+        )
+        self.component_list_subset = Set(
```
andrewlee94, line 1504: It would help to have a more descriptive name for these subsets - what is the reason for the two subsets?

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4563 @@
...
+            units=units.kg / units.mol,
+            doc="Molar mass [kg/mol]",
+        )
+        self.visc_d = Param(
+            initialize=1.9e-5,
+            units=units.Pa * units.s,
+            doc="Gas phase viscosity [Pa s]",
+        )
+
+    def _add_parameters_zeolite_13x(self):
```
andrewlee94, line 1531: Related to a previous comment, I suggest breaking these out to one or more separate files so that these can become arguments to the model. This will make it easier to add new sorbents and also help reduce the size of this file.

andrewlee94 (commented):
Here are some comments of the first part of the code.

idaes/models_extra/temperature_swing_adsorption/util.py
```diff
@@ -0,0 +1,155 @@
...
+
+    df = DataFrame.from_dict(summary_dir, orient="columns")
+    del df["pos"]
+    if export:
+        df.to_csv(f"{tsa.local_name}_summary.csv")
+
+    print("\n" + "=" * 84)
+    print(f"summary {tsa.local_name}")
+    print("-" * 84)
+    stdout.write(textwrap.indent(stream_table_dataframe_to_string(df), " " * 4))
```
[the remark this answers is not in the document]
AlexNoring: Switched to the `stream=stdout` approach

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4563 @@
...
+                Fin = 100  # mol/s
+                Q = -500000  # W
+                blk.steam_heater.unit.inlet.flow_mol[0].fix(Fin)
+                blk.steam_heater.unit.heat_duty.fix(Q)
+
+                # 7.3) check degrees of freedom and solve
+                if degrees_of_freedom(blk.steam_heater) == 0:
+
+                    # initialize steam heater
+                    heater_initializer = SingleControlVolumeUnitInitializer()
```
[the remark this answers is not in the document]
AlexNoring: Updated to use the `get_submodel_initializer()` method

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4563 @@
...
+from idaes.core.util.exceptions import ConfigurationError
+from idaes.core.util.model_statistics import degrees_of_freedom
+
+from idaes.core.solvers import get_solver
+from idaes.core.initialization import (
+    ModularInitializerBase,
+    SingleControlVolumeUnitInitializer,
+)
+from idaes.models.unit_models import SkeletonUnitModel, Heater
+from idaes.models.properties.iapws95 import htpx
```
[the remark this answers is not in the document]
AlexNoring: Removed import

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4563 @@
...
+
+import idaes.logger as idaeslog
+
+__author__ = "Daison Yancy Caballero"
+
+# Set up logger
+_log = idaeslog.getLogger(__name__)
+
+
+class FixedBedTSA0DInitializer(ModularInitializerBase):
```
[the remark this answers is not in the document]
AlexNoring: Moved to separate file

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4563 @@
...
+
+    """
+
+    def initialize(
+        self,
+        model: Block,
+        initial_guesses: dict = None,
+        json_file: str = None,
+        output_level=None,
+        exclude_unused_vars: bool = False,
```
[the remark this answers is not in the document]
AlexNoring: I set the default to `True`. If `False` is passed I added a warning but still overwrite to `True`.

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4563 @@
...
+            blk.name, self.get_output_level(), tag="unit"
+        )
+
+        # create solver
+        opt = get_solver(self.config.solver, self.config.solver_options)
+
+        # initialization of fixed bed TSA model unit
+        init_log.info("Starting fixed bed TSA initialization")
+
+        # fix states at inlet if they aren't fixed yet
```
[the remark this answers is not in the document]
AlexNoring: Removed this code and the code that reverts the inlet state

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4563 @@
...
+
+        blk._calculate_variable_from_constraint(
+            variable_list=vars_lst_heating, constraint_list=cons_lst_heating, obj=blk
+        )
+
+        # 1.2) initial solution using false position method
+
+        # deactivate final condition constraint and fix time
+        blk.heating.fc_temperature_eq.deactivate()
+        blk.heating.time.fix(1e3)
```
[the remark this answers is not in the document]
AlexNoring: Added units

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4563 @@
...
+        else:
+            raise ConfigurationError(
+                "Degrees of freedom is not zero during initialization of "
+                "adsorption step. Fix/unfix appropriate number of variables "
+                "to result in zero degrees of freedom for initialization."
+            )
+
+        # 5 - solve entire fixed bed TSA model
+
+        # 5.1) unfix variables and activate constraints that were fixed and
```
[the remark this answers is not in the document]
AlexNoring: Replaced this code with a save to and load from json

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4563 @@
...
+        # set up logger
+        init_log = idaeslog.getInitLogger(
+            cycle_step.name, self.get_output_level(), tag="unit"
+        )
+        solve_log = idaeslog.getSolveLogger(
+            cycle_step.name, self.get_output_level(), tag="unit"
+        )
+
+        # create solver
+        opt = get_solver(self.config.solver, self.config.solver_options)
```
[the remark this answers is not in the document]
AlexNoring: Added solver as class attribute

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4563 @@
...
+
+        # save solution in initial guess, f(x0)
+        if str(cycle_step).split(".")[-1] == "heating":
+            f_x0 = value(blk.temperature_desorption - cycle_step.temperature[1])
+        else:
+            f_x0 = value(cycle_step.temperature[1] - blk.temperature_adsorption)
+
+        # iterate to find an interval with the root
+        condition = True
+        if f_x0 >= 0:
```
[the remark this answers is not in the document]
AlexNoring: I added a variable to keep track of the initial `f_x0` so the two while loops could be combined.

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4563 @@
...
+    )
+    CONFIG.declare(
+        "calculate_beds",
+        ConfigValue(
+            default=False,
+            domain=In([True, False]),
+            description="Construction flag for calculation of number of beds",
+            doc="""Construction flag for calculation of number of beds
+        - False (default): number of beds is fixed and defined by user
+        - True: number of beds is calculated for an specific
```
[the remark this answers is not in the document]
AlexNoring: Fixed

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4563 @@
...
+        pressure drop in the column""",
+        ),
+    )
+    CONFIG.declare(
```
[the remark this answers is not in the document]
AlexNoring: @Daison2102 Can you comment on why it was done this way? For now I've combined the `number_of_beds` and `calculate_beds` arguments following Andrew's suggestion.

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4563 @@
...
+            and self.config.transformation_scheme is None
+        ):
+            self.config.transformation_scheme = "LAGRANGE-RADAU"
+        elif (
+            self.config.transformation_method == "dae.finite_difference"
+            and self.config.transformation_scheme != "BACKWARD"
+            and self.config.transformation_scheme != "FORWARD"
+        ):
+            raise ConfigurationError(
+                "{} invalid value for "
```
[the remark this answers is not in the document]
AlexNoring: Fixed

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4563 @@
...
+                " "
+                "dae.finite_difference"
+                ".".format(self.name)
+            )
+        elif (
+            self.config.transformation_method == "dae.collocation"
+            and self.config.transformation_scheme != "LAGRANGE-RADAU"
+        ):
+            raise ConfigurationError(
+                "{} invalid value for "
```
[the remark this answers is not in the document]
AlexNoring: Fixed

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,4563 @@
...
+            raise ConfigurationError(
+                "Steam calculations are set to rigorous "
+                "but no property package was provided"
+            )
+
+        # add required parameters
+        self._add_general_parameters()
+
+        # add adsorbent parameters
+        if self.config.adsorbent == "Zeolite-13X":
```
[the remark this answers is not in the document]
AlexNoring: I think this will be good to do in the future if/when we need to add a new sorbent.

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
AlexNoring: Removed units from doc strings

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
AlexNoring: `component_list_subset` is for components involved in the isotherm calculations while components in `component_list_subset_2` are all sent to one outlet steam without detailed calculations. I renamed to `isotherm_components` and `other_components`.

andrewlee94: For safely, it might be better to have only one of these subsets and for everything else do `component_list-subset`. This way, you can never have a case where there is a component that falls throguh the cracks.

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,3677 @@
...
+            doc="Pressure at inlet [Pa]",
+        )
+
+        # create empty port
+        p = Port(noruleinit=True, doc="Inlet port for fixed bed TSA")
+        # add port object as an attribute to model
+        setattr(self, "inlet", p)
+        # dictionary containing state of inlet stream
+        inlet_dict = {
+            "flow_mol_comp": self.flow_mol_in,
```
andrewlee94, line 992: I think you can do this as part of the `Port` declaration as well (probably using the `initialize` argument).

Again, this approach is only needed when you have to procedurally generate the port contents.

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,3677 @@
...
+            doc="Temperature at CO2 rich stream",
+        )
+        self.pressure_co2_rich_stream = Var(
+            self.flowsheet().time,
+            units=units.Pa,
+            doc="Pressure at CO2 rich stream",
+        )
+
+        # create empty port
+        p = Port(noruleinit=True, doc="Outlet port for CO2 rich stream")
```
andrewlee94, line 1053: As above.

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,3677 @@
...
+            return sum(b.mole_frac[t, j] for j in self.isotherm_components) == 1.0
+
+        # component mass balance ODE
+        @self.heating.Constraint(
+            self.heating.time_domain,
+            self.isotherm_components,
+            doc="Component mass balance - only for CO2",
+        )
+        def component_mass_balance_ode(b, t, i):
+            if t == b.time_domain.first() or i == "N2":
```
andrewlee94, line 1317: A minor word of caution, but this does presume the presence of N2 in the system. For a DAC system this is to be expected, and is probably a good approach to improve robustness (my intuition tells me an overall mass balance will be easier to solve that having to solve it implicitly via the full set of component balances).

However, it does result in a loss of generality (which probably doesn't matter for now), in that the model cannot be used for systems without N2.

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,3677 @@
...
+                    / units.mol
+                )
+
+            if self.config.steam_calculation == SteamCalculationType.rigorous:
+                return (
+                    self.flow_mass_steam
+                    == self.steam_heater.unit.control_volume.properties_out[0].flow_mass
+                )
+
+    def _step_initialize(
```
andrewlee94, line 2593: Should this method (and the one that follows) be moved to be with the `Initializer`?

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,3677 @@
...
+            if f_x0 * f_x2 < 0:
+                x1 = x2
+            else:
+                x0 = x2
+
+            # update counter and set up new condition |f(x_2)| > error
+            count += 1
+            condition = (f_x2**2) ** 0.5 > 1
+
+    def initialize_build(blk, outlvl=idaeslog.NOTSET, solver=None, optarg=None):
```
andrewlee94, line 2824: We are deprecating this approach, so I think you can safely remove this (maybe leave an exception that tells the user this is deprecated and to use the `Initializer` instead. Otherwise, you are duplicating a lot of code.

andrewlee94 (commented):
Here are some more comments - I've gone throguh all of the model itself now (although I focused on the code rather than the model equations).

idaes/models_extra/temperature_swing_adsorption/initializer.py
```diff
@@ -0,0 +1,911 @@
...
+        Initialization routine for fixed bed TSA unit.
+
+        Args:
+            blk: model to be initialized
+
+        Returns:
+            None
+
+        Raises:
+            ConfigurationError: If degrees of freedom is not zero at the start
```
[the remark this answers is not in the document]
AlexNoring: Switched instances of `ConfigurationError` to `InitializationError`

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,3677 @@
...
+                / particle_sphericity
+                / bed_voidage_mf**3
+                * b.Reynolds_mf**2
+                + 10
+                * 150
+                * (1 - bed_voidage_mf)
+                / particle_sphericity**2
+                / bed_voidage_mf**3
+                * b.Reynolds_mf
+                == 10
```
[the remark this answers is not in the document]
AlexNoring: Replaced hard-coded scaling factor with one set in the `calculate_scaling_factors` method

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,3677 @@
...
+            initialize={"CO2": -37000, "N2": -18510},
+            units=units.J / units.mol,
+            doc="Heat of adsorption",
+        )
+        self.temperature_ref = Param(
+            initialize=298.15,
+            units=units.K,
+            doc="Reference temperature",
+        )
+        self.n_ref = Param(
```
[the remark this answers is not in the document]
AlexNoring: Renamed all isotherm parameters and intermediate calculation variables with more descriptive names

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,3677 @@
...
+        self.dens_mass_sol = Param(
+            initialize=3200,
+            units=units.kg / units.m**3,
+            doc="Density of adsorbent",
+        )
+        self.particle_dia = Param(
+            initialize=2e-3, units=units.m, doc="Particle diameter"
+        )
+        # isotherm parameters
+        self.dh_ads = Param(
```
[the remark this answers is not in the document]
AlexNoring: The isotherm equations are used in each TSA step. They are built by a separate method so they can be called multiple times.

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,3677 @@
...
+        rates of N2 and CO2 only:
+        flow_mol_in_total = flow_mol_in["CO2"] + flow_mol_in["N2"]
+
+        """
+        # set up inlet port for fixed bed TSA (exhaust gas stream to TSA inlet)
+        self.flow_mol_in = Var(
+            self.flowsheet().time,
+            self.component_list,
+            initialize=10,
+            # domain=PositiveReals,
```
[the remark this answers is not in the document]
AlexNoring: Removed

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,3677 @@
...
+            initialize=101325,
+            domain=PositiveReals,
+            units=units.Pa,
+            doc="Pressure at inlet [Pa]",
+        )
+
+        # create empty port
+        p = Port(noruleinit=True, doc="Inlet port for fixed bed TSA")
+        # add port object as an attribute to model
+        setattr(self, "inlet", p)
```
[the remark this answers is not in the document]
AlexNoring: Switched to your code

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
AlexNoring: Now passing `inlet_dict` as `initialize` argument

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,3677 @@
...
+                b.flow_mol_n2_rich_stream[0, j] for j in self.isotherm_components
+            )
+
+        @self.Expression(
+            doc="Concentration of CO2 in stream emitted to atmosphere [ppm]"
+        )
+        def emissions_co2_ppm(b):
+            return b.mole_frac_n2_rich_stream["CO2"] * 100 * 1e4
+
+    def _calculate_variable_from_constraint(
```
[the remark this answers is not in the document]
AlexNoring: Renamed to `_calculate_and_fix_variable_from_constraint`

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
AlexNoring: Removed along with the old `initialize_build` method

idaes/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.py
```diff
@@ -0,0 +1,3677 @@
...
+            for k, i in flags.items():
+                if v.name == k and i:
+                    v.unfix()
+
+    def fix_initialization_states(self):
+        self.inlet.flow_mol_comp.fix()
+        self.inlet.temperature.fix()
+        self.inlet.pressure.fix()
+
+    def calculate_scaling_factors(self):
```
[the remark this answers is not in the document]
AlexNoring: Moved the hard-coded scaling factors here

docs/reference_guides/model_libraries/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.rst
```diff
@@ -0,0 +1,154 @@
+Fixed Bed Temperature Swing Adsorption (TSA)
+=============================================
+
+The fixed bed TSA unit model consists of a fixed bed four-step temperature swing adsorption (TSA) cycle. 
+The model is implemented as a 0-D equilibrium-based shortcut model without axial variation of states and 
+with a semi-analytical solution of the cyclic steady state. The following assumptions were considered in 
+the model implementation:
+
+* The gas phase behaves ideally.
```
andrewlee94, line 9: I think you need to add an assumption here about the presence of N2 as an inert (as the set up of the material balances assumes this).

docs/reference_guides/model_libraries/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.rst
```diff
@@ -0,0 +1,154 @@
...
+
+IDAES Implementation
+--------------------
+Figure 2 shows a schematic representation of the implementation of the fixed bed TSA model in IDAES:
+
+* The fixed bed TSA model was implemented in IDAES as a Unit Model Class with several configuration arguments to provide flexibility to the model. 
+* Custom methods were implemented to build each cycle step by using the generic skeleton model unit available in IDAES.
+* The model includes one inlet port (feed stream to the system) and two outlet ports (CO2 rich stream and N2-rich stream).
+* Expressions are included to calculate pressure drop in the column and minimum fluidization velocity to avoid unrealistic pressure drop in the column and also constrain the feed velocity to be lower than the minimum fluidization velocity.
+* Mass balances are indexed by components that will facilitate the addition of more components to the model.
```
andrewlee94, line 83: A caveat here is that N2 must be present (maybe CO2 as well)?

andrewlee94 (approved):
I had a couple of minor comments about the documentation, but otherwise I think this looks pretty good. Somebody with more knowledge of the equations needs to do a double check of the implementation however, as well as take another look over all the code.

MAZamarripa (approved):
lgtm

PR #1235, opened 2023-08-03
PETSc Tweaks

dallan-keylogic:
-Only include active blocks when constraints/variables are collected for PETSc solve
-Rename `snes_solver` option to `initial_solver` because the point of the option is to permit use of initial solvers besides snes
-Add symbolic solver labels to initial condition problem
-Get rid of unused list `fix_derivs`

### Legal Acknowledgement


idaes/core/solvers/petsc.py
```diff
@@ -420,8 +420,8 @@ def petsc_dae_by_time_element(
     initial_variables=None,
     detect_initial=True,
     skip_initial=False,
-    snes_solver="petsc_snes",
-    snes_options=None,
+    initial_solver="petsc_snes",
```
andrewlee94, line 423: Seeing as we are pushing to get backward compatibility tests into place, this look like a place where we are going to need backward compatibility options and warnings. I.e. we need to keep the old options around, and have them raise deprecation warnings if they are used before sending the value to the new argument (and raise an exception if both old and new are provided).

andrewlee94 (changes requested):

idaes/core/solvers/petsc.py
andrewlee94: If it is that new then we are probably OK, but we need to double check that is the case.

eslickj (approved):
I wanted to make this change when I added the snes_solver option, since it's not necessarily and SNES solver anymore, but I didn't want to break anything.  Once @andrewlee94 gets his deprecation warning, I'm good with this.

idaes/core/solvers/petsc.py
```diff
@@ -500,8 +508,17 @@ def petsc_dae_by_time_element(
         between.construct()
 
     solve_log = idaeslog.getSolveLogger("petsc-dae")
-    regular_vars, time_vars = flatten_dae_components(m, time, pyo.Var)
-    regular_cons, time_cons = flatten_dae_components(m, time, pyo.Constraint)
+    if snes_options is not None:
+        logger = idaeslog.getIdaesLogger("idaes")
```
andrewlee94, line 512: We generally use `logger = idaeslog.getLogger(__name__)` for this - using `__name__` gives more granularity and avoids hard-coded string constants. Also, Pyomo has a deprecation warning function for things like this:

```
from pyomo.common.deprecation import deprecation_warning

deprecation_warning(
    msg=msg, logger=_log, version="2.0.0", remove_in="3.0.0"
)
```

idaes/core/solvers/petsc.py
```diff
@@ -477,10 +478,17 @@ def petsc_dae_by_time_element(
...
+        snes_options (dict): [DEPRECATED in favor of initial_solver_options] nonlinear equation solver options
 
     Returns (PetscDAEResults):
         See PetscDAEResults documentation for more information.
     """
+    if snes_options is not None and initial_solver_options is not None:
+        raise RuntimeError(
+            "Both (deprecated) snes_options and initial_solver_options keyword arguments were specified. "
+            "Please specify your initial solver options using only the initial_solver_options argument."
+        )
```
andrewlee94, line 490: For clarity to the reader, I would be inclined to move the deprecation warning to immediately after this. This makes it clear why you can;t have both options, and also means that from this point on you don;t have to worry about the deprecation option at all (at the moment, there is the possibility of something happening to the two options between here and when the deprecation is checked).

andrewlee94 (commented):
Looking good - just a couple of minor comments to avoid some edge cases and be more inline with other deprecations.

andrewlee94 (approved):
LGTM

PR #1237, opened 2023-08-09
Contributes script to install IDAES and COIN solvers on Google Colab

adowling2:
Contributes script to install IDAES and COIN solvers on Google Colab

## Summary/Motivation:
- Installing IDAES (and solvers via get-extensions) is a little quirky on Colab
- We've used (a prior version of) this script for teaching with Google Colab for over a year with >40 student testers

- Contributes one file

### Legal Acknowledgement


adowling2:
@jsiirola is another good reviewer.

I'd like feedback on:
- How should this get tested?
- Are we okay with the API / level of commenting?
- Are we okay with leaving the `install_CBC`, etc. backup methods? The default usage is just to run `install_idaes()` and then `install_ipopt()`

jsiirola (changes requested):
In addition to the specific comments below, there are several general changes that should be made:
- fix the identified spelling errors
- remove the logic for direct downloads from `ampl.com`
- I would suggest replacing (all?) uses of `os.system` with `subprocess.run`

scripts/colab_helper.py
```diff
@@ -0,0 +1,218 @@
...
+            command_with_output('./ipopt_l1 -v')
+            
+
+        # Check again if Ipopt is available
+        if not package_available("ipopt"):
+
+            # As a backup, try copying the executable from AMPL
+            if on_colab():
+                print("Installing Ipopt via zip file...")
+                os.system('wget -N -q "https://ampl.com/dl/open/ipopt/ipopt-linux64.zip"')
```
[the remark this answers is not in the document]
adowling2: I just removed these.

scripts/colab_helper.py
```diff
@@ -0,0 +1,218 @@
...
+    print(r)
+
+
+def install_idaes():
+    ''' Installs latest version of IDAES-PSE via pip
+    
+    '''
+
+    # Check if idaes is available. If not, install it
+    if not package_available("idaes"):
```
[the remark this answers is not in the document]
adowling2: Thanks for the great suggestion. I implemented it.

scripts/colab_helper.py
```diff
@@ -0,0 +1,218 @@
...
+            # Add symbolic link for idaes solvers
+            # These are needed for Colab to find the solvers
+            os.system("ln -s /root/.idaes/bin/ipopt ipopt")
+            os.system("ln -s /root/.idaes/bin/k_aug k_aug")
+            os.system("ln -s /root/.idaes/bin/couenne couenne")
+            os.system("ln -s /root/.idaes/bin/bonmin bonmin")
+            os.system("ln -s /root/.idaes/bin/cbc cbc")
+            os.system("ln -s /root/.idaes/bin/clp clp")
+            os.system("ln -s /root/.idaes/bin/ipopt_l1 ipopt_l1")
+            os.system("ln -s /root/.idaes/bin/dot_sens dot_sens")
```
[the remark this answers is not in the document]
adowling2: Also, I want to make these solvers accessible without `import idaes` for two reasons:
* Have the script print the solver versions, which is a good check and helpful for debugging.
* Reuse this script in workshops/classes/tutorials that introduce Pyomo. In my experience, skipping `import idaes` is easier with brand-new Pyomo users.

adowling2:
@jsiirola This is ready for you to review.

@dangunter You might also find this helpful/interesting. I've been using a script like this to install Pyomo and IDAES in Google Colab for my graduate elective.

adowling2:
@ksbeattie @lbianchi-lbl I have black on the most updated version of this one file on my computer; it said no changes. I also confirmed I do not have any uncommitted or unpushed changes. Yet the black test is still failing.

For everyone: I tested this on Google Colab (hence several commits this morning.) I am not sure how to automate testing in Google Colab or if that is really even needed.

andrewlee94:
@adowling2 `black` is very temperamental - you need to have the correct version of `black` installed as the formatting changes slightly from version to version.

adowling2:
@andrewlee94 Now that I have your attention, do you want to review this quickly?

andrewlee94 (approved):
This looks OK to me, but I am no expert on this sort of stuff.

scripts/colab_helper.py
```diff
@@ -0,0 +1,142 @@
...
+            check=True,
+        )
+        print("Checking ipopt version:")
+        _print_single_solver_version("ipopt")
+
+
+def _update_path():
+    """Add idaes executables to PATH"""
+    if not re.search(re.escape(":/root/.idaes/bin/"), os.environ["PATH"]):
+        os.environ["PATH"] += ":/root/.idaes/bin/"
```
jsiirola, line 120: This is a minor nit, but we probably want to put `idaes/bin` first and not last in the path (the idea being that we would be overriding anything that was previously installed into the system)

scripts/colab_helper.py
```diff
@@ -0,0 +1,142 @@
...
+import os.path
+import os
+import re
+
+import subprocess
+
+
+def _check_available(executable_name):
+    """Utility to check in an executable is available"""
+    return shutil.which(executable_name) or os.path.isfile(executable_name)
```
jsiirola, line 49: (More of an FYI than an action) This is find for collab (where we now it is Linux-based).  However, a more portable approach would be to use something akin to Pyomo's `pyomo.common.fileutils.Executable` (which is what Pyomo uses for locating executables)

scripts/colab_helper.py
```diff
@@ -0,0 +1,142 @@
...
+        print("Installing idaes via pip...")
+        subprocess.run(
+            [sys.executable, "-m", "pip", "install", "-q", "idaes_pse"], check=True
+        )
+        print("idaes was successfully installed")
+        v = subprocess.run(
+            ["idaes", "--version"], check=True, capture_output=True, text=True
+        )
+        print(v.stdout)
+        print(v.stderr)
```
jsiirola, line 86: If you are just going to print out the output anyway, why bother with `capture=True`?

jsiirola (commented):
This is getting pretty close.

scripts/colab_helper.py
adowling2: I fixed this and tested it works on Colab.

scripts/colab_helper.py
jsiirola: On the terminal, the output comes to the terminal if you don't specify `capture_output`.  I wonder if this is different for the collab environment?  If that's the case, we want to at least be consistent and capture/print for *all* subprocess calls.

scripts/colab_helper.py
```diff
@@ -0,0 +1,142 @@
...
+    import colab_helper
+    colab_helper.install_idaes()
+    colab_helper.install_ipopt()
+```
+
+For testing purposes, you may need to use:
+"https://raw.githubusercontent.com/adowling2/idaes-pse/colab-install-script/scripts/colab_helper.py"
+
+"""
+
```
lbianchi-lbl, line 37: I'd recommend to add some versioning information, e.g. by adding a line here with `__version__ = "0.23.8"` (or any other date-based versioning schema).

lbianchi-lbl (changes requested):
I think having this script here can be valuable for users of IDAES on Google Colab in addition to and beyond the original audience. A few general thoughts:

- I don't think there's a way to test this automatically (without an unreasonable amount of effort)
- Since the script is tested and will continue to be used in the future, I think we can pragmatically rely on this "outside testing" to detect possible breaking changes
- However, since we're adding code to the repository that's not being tested by the CI, before merging, we should:
  -  Identify who the maintainer(s) of this script will be, e.g. by adding an entry to the `CODEOWNERS` file
  - Add some kind of simple versioning information to the script itself (since the script won't have access to either the Git history or the `idaes` version at runtime)

scripts/colab_helper.py
adowling2: I expected that behavior on Colab, but no luck when testing in a cell.

adowling2: @jsiirola, I like your suggestion. I added a verbose flag to display the captured pip, idaes get-extensions, etc. output

scripts/colab_helper.py
adowling2: Great suggestion

adowling2:
@jsiirola @lbianchi-lbl Thanks for the suggested changes. This should be ready.

Changes:
- Capture all subprocess output. Added verbose flag to toggle on output for pip install, idaes get-extensions, etc.
- Added a version
- Added my name in `CODEOWNERS`

adowling2:
@jsiirola @lbianchi-lbl This is ready for your re-review.

jsiirola (approved):

PR #1238, opened 2023-08-09
Initial REE Costing Framework

bpaul4:
## Summary/Motivation:
Initial framework for REE costing as part of PrOMMiS cost estimation efforts.

- Adds first group of cost accounts for REE plants and support to load/execute them in the IDAES Costing Framework
- Required modifications to IDAES Power Plant Costing Framework, including new methods for REE account handling and an expansion of the report call
- Example flowsheet and test file with the first two accounts, which will be expanded to include all accounts later
- Some updates to the power plant methods and documentation, notably for a new option to support multiple trains of a unit operation
- REE costing documentation and bound estimation methods coming soon

### Legal Acknowledgement


idaes/models_extra/power_generation/costing/costing_dictionaries.py
```diff
@@ -64,3 +64,9 @@ def load_sCO2_costing_dictionary():
     with open(os.path.join(directory, "sCO2_costing_parameters.json"), "r") as file:
         sCO2_costing_params = json.load(file)
     return sCO2_costing_params
+
+
+def load_REE_costing_dictionary():
```
andrewlee94, line 69: Is this the right place for this? It seems really odd that something to do with REEs is ending up in the power plant sub-folder. This might indicate that the `power_generation/costing` folder really needs to be moved out to a `models_extra/costing` as it has spread beyond just power generation.

idaes/models_extra/rare_earth_recovery/ree_initial_flowsheet.py
```diff
@@ -0,0 +1,106 @@
...
+# National Technology & Engineering Solutions of Sandia, LLC, Carnegie Mellon
+# University, West Virginia University Research Corporation, et al.
+# All rights reserved.  Please see the files COPYRIGHT.md and LICENSE.md
+# for full copyright and license information.
+#################################################################################
+
+__author__ = "Costing Team (B. Paul and M. Zamarripa)"
+__version__ = "1.0.0"
+
+import pyomo.environ as pyo
```
andrewlee94, line 17: Again, is this the right place for this? For one, it is rather isolated at the moment, and it appears to be more of a minimal test case for costing than an actual flowsheet (thus the name would probably be misleading to users). Would this be better placed in the PrOMMiS workspace, which is where the first process flowsheet will go as well (at least until we have something more comprehensive to show to the world)?

andrewlee94: https://github.com/prommis/workspace

idaes/models_extra/power_generation/costing/costing_dictionaries.py
andrewlee94: I think it really should be moved before this is merged - otherwise it is likely to linger. Seeing as this is our public repo, we should make an effort to ensure that it is user friendly.

bpaul4:
Closing this in favor of https://github.com/prommis/workspace/pull/13

PR #1239, opened 2023-08-10
Remove some logger config stuff

### LINKED ISSUE #1140: Failures on Python 3.11 for logging configuration tests
image

The offending test (which will be skipped on 3.11 starting with #1137) is:

image

(Using a screenshot since permalinks with the commit hash might not work if the PR head branch is deleted or rebased)

eslickj:
## Fixes #1140

This removes some ill-conceived configuration tools that affect the logger configuration.  This could alter logging setting when importing idaes, which is bad.  This also removes a not-very-useful tool to configure IDAES from the command line, which I'm 100% sure has never been used. 

What's left in the IDAES logging module is just a couple extra log levels, a way to tag loggers for filtering purposes, and a context manager to redirect solver output to a logger. 

### Legal Acknowledgement


idaes/commands/config.py
```diff
@@ -1,137 +0,0 @@
...
-import json
-import click
-import pyomo.common.config
-
-import idaes
-from idaes.commands import cb
-import idaes.config as cfg
-
-
-@cb.command(name="config-write", help="Write the IDAES config")
```
andrewlee94, line 31: Whilst removing this might be a good idea, in the spirit of backward compatibility we probably shouldn't just remove it without warning. If there is an alternative way to set the config then we should probably leave the options but have them return some sort of exception that tells users how to do this another way. Otherwise, we really should keep these but log a deprecation warning for the user.

andrewlee94 (commented):
Whilst I don't mind removing stuff that sees little to no use, we really should try to do a proper deprecation.

idaes/commands/config.py
eslickj: I hear you, and I can do that, but I'm about 100% sure that no one ever used this.

eslickj: On second thought, it would be better if possible to just remove this, since we don't want to configure the logger.  Alternatively we could just make all the commands still exist but print a message that they don't do anything anymore.

andrewlee94: That was my thought.

andrewlee94 (approved):

bpaul4 (approved):

andrewlee94:
@eslickj It looks like these changes changed how some logger messages are captured in two tests when running in non-editable mode. Could you take a look ot work out what happened (hopefully it is an easy fix).

eslickj:
@andrewlee94, I'm looking into it.

eslickj:
@andrewlee94, I couldn't find it, but something must change the logging level during testing.  The test passed when I ran it locally without running the rest of the tests.  I fixed it by setting the log level with caplog.  After the test, when you set the logger level that way it goes back to whatever it was.  I guess this means some test is setting a logger level and it may interfere with other logging tests.  I guess best practice for tests that involve logging may be to set the level with caplog.

Issue #1241, opened 2023-08-11
Add Pylint checks forcing Pyomo rules to use only local variables


dallan-keylogic:
While debugging issues with collocation discretization for dynamics, we found that [nonintuitive behavior of Python closures when defining functions with nonlocal variables](https://code.activestate.com/recipes/502271/) can cause major problems. The behavior at issue is this:

```
In [1]: string = "foo"

In [2]: def print_foo():
   ...:     print(string)
   ...:

In [3]: string = "bar"

In [4]: print_foo()
bar
```

The behavior I expected was that when `print_foo` is defined, it would freeze the value of  `string`  to be what it was when `print_foo` was defined. Instead, it is still updated from the outer context. This can cause a ton of trouble when Pyomo rules are created using nonlocal variables. For example, in the `Separator`, a rule is defined using multiple variables from outside the local scope:

![image](https://github.com/IDAES/idaes-pse/assets/88728506/b170aed2-e32d-4d7f-aa92-f34c88ab80cd)

The same problem occurs: the nonlocal variables change in a loop, the equations are generated correctly the first time, but then are incorrectly generated at additional timepoints by PyomoDAE.

To prevent this problem in the future, I propose that checks be added (via Pylint?) that any function with `rule` in the name, and any functions with the `@Constraint` or `@Expression` decorators be forced to use only local variables.

dallan-keylogic:
This issue is related to #1240 , arguably the first example belongs there as a demonstration of the perfidy of `lambdas`, but the second issue persists even without `lambdas`.

lbianchi-lbl:
I've searched for "closures" in the Pylint 2.12.2 docs and `cell-var-from-loop` seems to be the only match: https://pylint.readthedocs.io/en/v2.12.2/technical_reference/features.html#variables-checker-messages

![image](https://github.com/IDAES/idaes-pse/assets/48035537/91f12556-2876-47cb-8e8a-9538e0fb8169)

Unfortunately, I'm not sure if that check can be used as-is for this purpose. I'll try to look into (a) searching for similar checks that might have been added in later versions of Pylint, and (b) how the check is implemented in Pylint, to get a sense of how hard it would be to adapt.

lbianchi-lbl:
I agree with the general principle that "something is better than nothing" when it comes to these checks.

[...]

Issue #1240, opened 2023-08-11
Add Pylint Check for Lambda Functions


dallan-keylogic:
In the course of debugging an issue with collocation, @jsiirola pointed out that `lambda` functions do not behave well with `copy.deepcopy`, which can prevent Pyomo from cloning models correctly.

```
>>> a = [1,2,3]
>>> def foo(x):
...    print(id(x), x)
...
>>> xx = lambda: foo(a)
>>> yy = copy.deepcopy(xx)
>>> xx()
140447310398160 [1, 2, 3, 5]
>>> yy()
140447310398160 [1, 2, 3, 5]
>>> a.append(6)
>>> xx()
140447310398160 [1, 2, 3, 5, 6]
>>> yy()
140447310398160 [1, 2, 3, 5, 6]
```

It is unclear to the user that `a` is not going to be deep copied as well (which [probably relates to closure rules](https://code.activestate.com/recipes/502271/)). The preferred way of accomplishing something like this is using `functools.partial`.:

```
>>> import functools
>>> x = functools.partial(foo, a)
>>> import copy
>>> y = copy.deepcopy(x)
>>> x()
140447310398160 [1, 2, 3]
>>> y()
140447178509520 [1, 2, 3]
>>> a.append(5)
>>> x()
140447310398160 [1, 2, 3, 5]
>>> y()
140447178509520 [1, 2, 3]
```

Another example occurs when trying to use `lambda` functions to reduce the function's signature as part of a loop:

```
def rule_flow_mol_comp(blk, t, j, props, port):
    return (
        props[t].flow_mol_comp[j]
        == blk.number_cells * port.flow_mol[t] * port.mole_frac_comp[t, j]
    )
for side in ["fuel", "oxygen"]:
    side_comps = getattr(self.solid_oxide_cell, f"{side}_component_list")
    for direction in ["in", "out"]:
        props = getattr(self, f"{side}_properties_{direction}")
        port_name = f"{side}_{direction}let"  # e.g., fuel_inlet, oxygen_outlet
        port = getattr(self.solid_oxide_cell, port_name)
        setattr(
            self,
            f"{port_name}_flow_mol_comp_eqn",
            pyo.Constraint(
                tset,
                side_comps,
                rule=lambda blk, t, j: rule_flow_mol_comp(
                    blk, t, j, props, port
                ),
        )
```

[...]

PR #1242, opened 2023-08-16
Function Scoping Fixes

dallan-keylogic:
## Fixes
-- Fixes function closure issues outlined in #1240 and #1241 for the Separator and SOC.

### Legal Acknowledgement

### DIFF
--- a/idaes/models/unit_models/mixer.py
+++ b/idaes/models/unit_models/mixer.py
@@ -1180,11 +1180,10 @@ objects linked to all inlet states and the mixed state,
                     iscale.constraint_scaling_transform(c, s, overwrite=False)
 
         if hasattr(self, "enthalpy_mixing_equations"):
+            def scale_gen(t):
+                for p in self.mixed_state[t].phase_list:
+                    yield self.mixed_state[t].get_enthalpy_flow_terms(p)
+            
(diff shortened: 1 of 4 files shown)


andrewlee94:
@jsiirola Would you have time to take a quick look at this to double check that there aren't any other unexpected issues related to this (either ones that were already there or something added by this change)?

andrewlee94 (approved):
This looks good to me, pending the tests passing.

dallan-keylogic:
@andrewlee94 , I renamed some variables, which seems to have triggered CodeCov. Is there a way to just force the rest of the checks to run anyway?

andrewlee94:
@dallan-keylogic It looks like the CodeCov failure is due to coverage creep in the repo - our threshold for coverage is slowly creeping up as we force better coverage in the codebase which means that older sections can fall below the new threshold. In this case I am happy to bypass that check.

bpaul4 (approved):
This looks good to me. Interesting idea placing repeated exceptions in callable methods, too.

dallan-keylogic:
> This looks good to me. Interesting idea placing repeated exceptions in callable methods, too.

If I have to copy and paste once I cringe, if I have to do it twice it becomes a method. That way if you want to change something, you only need to change it one place rather than ten different places scattered throughout the code.

PR #1254, opened 2023-08-30
Add version constraint to exclude pandas 2.1 as a workaround for Pysmo test failures

### LINKED ISSUE #1253: Pysmo tests np.testing.assert_array_equal() fail because of ordering
- Current best guess is that something is affecting the ordering and therefore causes the assertions to fail
- Possibly related to pandas 2.1 being released

lbianchi-lbl:
-
-

### Legal Acknowledgement

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -131,7 +131,7 @@ kwargs = dict(
         "pint",  # required to use Pyomo units
         "networkx",  # required to use Pyomo network
         "numpy",
-        "pandas<2.1",
+        "pandas==2.1",
         "scipy",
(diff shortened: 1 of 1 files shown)


andrewlee94 (approved):
Only comment is that we really should have a comment as to why it was needed for future reference/reminder.

lbianchi-lbl:
> Only comment is that we really should have a comment as to why it was needed for future reference/reminder.

OK thanks, good point. I've added the comment and pushed again to re-trigger the checks.

dallan-keylogic (approved):
Approving so integration checks run

PR #1256, opened 2023-08-30
SVD Toolbox for diagnostics

andrewlee94:
## Summary/Motivation:
This reworks the current SVD code to be part of the new diagnostics toolbox, and adds a few new methods.

- New `SVDToolbox` class
- Copy and refactor existing SVD code into `SVDToolbox`
- New methods for displaying SVD results in standard form, as well as displaying links between variables and constraints
- Link to `SVDToolbox` from `DiagnosticsToolbox`
- Basic documentation, including placeholder for further discussion of how to interpret SVD results.

### Legal Acknowledgement


andrewlee94:
@adowling2 If you (or a student) have time to take a look at this, I have done a first pass at cleaning up Degeneracy Hunter and getting it ready to integrate with the new Diagnostics Toolbox. As part of this, I also switched the default solver to SCIP and gave the user the option to select a solver of their choice.

adowling2 (commented):
Here are some quick comments. Would you like help writing documentation?

idaes/core/util/model_diagnostics.py
```diff
@@ -1041,11 +1169,649 @@ def report_numerical_issues(self, stream=stdout):
...
+        m_dh.abs_nu = Var(m_dh.C, bounds=(0, M + m_small))
+
+        m_dh.pos_xor_neg = Constraint(m_dh.C)
+
+        # Constraint to enforce set is degenerate
+        if issparse(self.jacobian):
+            J = self.jacobian.tocsc()
+
+            def eq_degenerate(m_dh, v):
+                if np.sum(np.abs(J[:, v])) > 1e-6:
```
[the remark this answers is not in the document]
andrewlee94: I've moved this into the config block (which can be set as an argument to __init__) - do you have a suggestion of a good name for it that explains what it does.

Also, whilst we are at it, do you have more descriptiveness suggestions for M, small_m and nu that will explain to a user what they do? I am guessing that they are an explicit big-M transformation, but a lot of users probably aren't going to understand what that means.

andrewlee94:
@adowling2 Any help on writing documentation of model diagnostics would be appreciated (both specifically for Degeneracy Hunter and in general). For Degeneracy Hunter, we will definitely need some documentation on how to interpret the output and understanding what it is telling the user.

I also had a general question about the code to make sure I got the implementation right. Using your simple test problem with a duplicated equality constraint (con2 and con5), plus three inequality constraints, I get the following output:

```
Irreducible Degenerate Sets

    Irreducible Degenerate Set 0
        nu	Constraint Name
        -1.0	con2

    Irreducible Degenerate Set 1
        nu	Constraint Name
        1.0	con5

====================================================================================```

I just wanted to check to make sure that was the expected result in this case - notably the fact that `con2` and `con5` show up as separate irreducible degenerate sets surprised me initially. My initial assumption (not having used the tool before) was that I would get a single IDS that contained both `con2` and con5` (indicating that the combination of these two constraints was degenerate).

**Review by bknueven — COMMENTED** [review]
I appreciate the code reorganization as it has helped me understand the degeneracy hunter and the SVD toolbox better.

For the degeneracy hunter, I would like to explore at some point the formulation of the `candidates_milp`, as I think you can get away with fewer parameters while also generalizing the problem a bit to pick up on near degeneracies. The formulation I have in mind though would not be equivalent and therefore would need to be benchmarked and compared to the current one, so the change would not be appropriate for this PR.

File: idaes/core/util/model_diagnostics.py [hunk 2]
```diff
@@ -1041,10 +1219,680 @@ def report_numerical_issues(self, stream=stdout):
...
+        self.ids_milp.nu[cons_idx].unfix()
+
+        # Create empty dictionary
+        ids_ = {}
+
+        if check_optimal_termination(results):
+            # We found an irreducible degenerate set
+            for i in self.ids_milp.C:
+                # Check if constraint is included
+                if self.ids_milp.y[i]() > 0.9:
```
jsiirola, line 1814: As a general rule, it would be better to make the embedded constant (0.9) a named attribute (at the module scope?)

idaes/core/util/model_diagnostics.py
```diff
@@ -1041,10 +1219,680 @@ def report_numerical_issues(self, stream=stdout):
...
+
+        results = self._get_solver().solve(self.candidates_milp, tee=tee)
+
+        self.degenerate_set = {}
+
+        if check_optimal_termination(results):
+            # We found a degenerate set
+            for i in self.candidates_milp.C:
+                # Check if constraint is included
+                if self.candidates_milp.abs_nu[i]() > self.config.m_small * 0.99:
```
jsiirola, line 1715: As a general rule, it would be better to make the embedded constant (0.99) a named attribute (at the module scope?)

jsiirola (commented):
Overall, this looks pretty good.  Just some minor nits you might consider addressing.

jsiirola (approved):

adam-a-a (approved):
LGTM

PR #1260, opened 2023-08-31
Active flag for PETSc

dallan-keylogic:
## Summary/Motivation:
Adds `active` flag for `flatten_dae_components` to avoid active equations from deactivated blocks from being collected.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/solvers/petsc.py
+++ b/idaes/core/solvers/petsc.py
@@ -517,10 +517,20 @@ def petsc_dae_by_time_element(
         between = pyo.Set(initialize=sorted(between))
         between.construct()
 
+    # TODO add checks for whether between has more than one element and whether time is discretized
+
     solve_log = idaeslog.getSolveLogger("petsc-dae")
 
-    regular_vars, time_vars = flatten_dae_components(m, time, pyo.Var, active=True)
-    regular_cons, time_cons = flatten_dae_components(m, time, pyo.Constraint, active=True)
+    # Need a reprenative time for flatten_dae_components to work, because otherwise things like
+    # sums of mole fraction constraints being deactivated at t0 (because all the mole fractions are fixed),
+    # etc., cause flatten_dae_components to miss a constraint that is active at t>t0
+    t_represenative = between.at(2) # Two because Pyomo sets start at one
+
+    # TODO document fact that if user wants to change active equations at different times, they need to
+    # call this function multiple times
+
+    regular_vars, time_vars = flatten_dae_components(m, time, pyo.Var, active=True, indices=(t_represenative,))
+    regular_cons, time_cons = flatten_dae_components(m, time, pyo.Constraint, active=True, indices=(t_represenative,))
     tdisc = find_discretization_equations(m, time)
 
     solver_dae = pyo.SolverFactory("petsc_ts", options=ts_options)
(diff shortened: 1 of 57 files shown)


idaes/core/solvers/petsc.py
```diff
@@ -517,10 +518,28 @@ def petsc_dae_by_time_element(
...
+
     solve_log = idaeslog.getSolveLogger("petsc-dae")
 
-    regular_vars, time_vars = flatten_dae_components(m, time, pyo.Var)
-    regular_cons, time_cons = flatten_dae_components(m, time, pyo.Constraint)
+    # Need a representative time for flatten_dae_components to work, because otherwise things like
+    # sums of mole fraction constraints being deactivated at t0 (because all the mole fractions are fixed),
+    # etc., cause flatten_dae_components to miss a constraint that is active at t>t0
+    if representative_time is None:
+        representative_time = between.at(2)  # Two because Pyomo sets start at one
```
Robbybp, line 535: If `representative_time` is provided, can you make sure it is an element of `between`?

idaes/core/solvers/petsc.py
```diff
@@ -517,10 +518,28 @@ def petsc_dae_by_time_element(
...
         between.construct()
 
+    # Make sure that time is a ContinuousSet that has been discretized
+    try:
+        time.get_discretization_info()["nfe"]
+    except AttributeError:
+        raise RuntimeError("Argument time is not a Pyomo ContinuousSet")
+    except KeyError:
+        raise RuntimeError("The ContinuousSet time has not been discretized")
+
```
Robbybp, line 528: Is this the best way to check for a discretized continuous set? I would probably have used `isinstance(time, ContinuousSet)` and `"nfe" in time.get_discretization_info()` to avoid catching arbitrary `AttributeError`/`KeyError`.

Robbybp (commented):
Some minor comments, but this looks fine.

idaes/core/solvers/petsc.py
dallan-keylogic: I wanted to avoid using `isinstance` because it's a bit looked-down upon in Python programming---typically you want to just try doing what you want to do and ask for forgiveness later. In this case, however, it's probably better to just use it.

It would be better if `ContinuousSet` had a function `has_been_discretized` or similar to check directly rather than checking dictionary members.

bpaul4 (approved):

idaes/core/solvers/petsc.py
```diff
@@ -517,10 +519,31 @@ def petsc_dae_by_time_element(
         between = pyo.Set(initialize=sorted(between))
         between.construct()
 
+    # Make sure that time is a ContinuousSet that has been discretized
+    if isinstance(time, pyodae.ContinuousSet):
+        try:
+            time.get_discretization_info()["nfe"]
+        except KeyError:
+            raise RuntimeError("The ContinuousSet time has not been discretized")
```
Robbybp, line 527: Would this work here?
```python
if not time.get_discretization_info():
    raise RuntimeError(f"ContinuousSet {time.name} is not discretized")
```
@blnicho How would you recommend checking whether a ContinuousSet is discretized?

dallan-keylogic: I think before the set is discretized, the dictionary already exists, but is empty.

blnicho: I do this check in `pyomo.dae` using the following:

    if 'scheme' not in time.get_discretization_info():
        print("time is not discretized")

Robbybp (approved):

PR #1268, opened 2023-09-12
Diagnostic tool to look for potential evaluation errors

michaelbynum:
## Summary/Motivation:
This PR adds a method to the `DiagnosticToolBox` that looks for potential evaluation errors such as the log of a negative number. I'm marking it as draft while I work on tests/docs, but I wanted to create the PR to get some initial feedback.

Example:

```
import idaes
import pyomo.environ as pe


m = pe.ConcreteModel()
m.x = pe.Var()
m.y = pe.Var()
m.obj = pe.Objective(expr=m.x**2 + m.y**2)
m.c1 = pe.Constraint(expr=m.y >= pe.log(m.x))
m.c2 = pe.Constraint(expr=m.y >= (m.x - 1)**2.5)
dtb = DiagnosticsToolbox(m)
dtb.report_potential_evaluation_errors()
```

gives

```
Found 2 potential evaluation errors

    c1: Potential log of a negative number in log(x); Argument bounds are (-inf, inf)
    c2: Potential evaluation error in (x - 1)**2.5; base bounds are (-inf, inf); exponent bounds are (2.5, 2.5)

```

Note that adding the constraint `m.c3 = pe.Constraint(expr=m.x - 1 >= 0)` does not change this report because there could still be evaluation errors in infeasible path solvers (like Ipopt). However, if this constraint is added and FBBT is used before checking for evaluation errors, then no warnings will be issued.

### Legal Acknowledgement


andrewlee94:
@michaelbynum Just to put down in writing the comments I made at the working group meeting, suggest renaming `report_potential_evaluation_errors` to `display_potential_evaluation_errors` to be more consistent with our naming elsewhere, and then adding a check in `report_structural_issues` to collect the potential evaluation errors and report the number of issues found in the summary (if greater than zero).

Robbybp (commented):
Thanks for putting this together, @michaelbynum. A few questions on the detection of exponent evaluation errors. Also, do you think that the `_EvalErrorWalker` should be part of, say, `pyomo.util`? I guess this PR does not preclude the possibility of "promoting" it eventually if there is a need or use case.

idaes/core/util/model_diagnostics.py
```diff
@@ -1045,6 +1061,202 @@ def report_numerical_issues(self, stream=stdout):
...
+
+def _check_eval_error_pow(node: NumericExpression, warn_list: List[str], caution_list: List[str]):
+    arg1, arg2 = node.args
+
+    integer_domains = ComponentSet([Binary, Integers])
+
+    # if the exponent is an integer, there should not be any evaluation errors
+    if isinstance(arg2, _GeneralVarData) and arg2.domain in integer_domains:
+        # life is good. The exponent is an integer variable
+        return None
```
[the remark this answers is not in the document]
michaelbynum: If the base is 0 and the exponent is a negative integer, then there could be an evaluation error. Good catch. I'll fix that.

idaes/core/util/model_diagnostics.py
```diff
@@ -1045,6 +1061,202 @@ def report_numerical_issues(self, stream=stdout):
...
+    lb2, ub2 = _get_bounds_with_inf(arg2)
+    if lb2 == ub2 and lb2 == round(lb2):
+        # life is good. The exponent is fixed to an integer
```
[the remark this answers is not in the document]
michaelbynum: `lb2 == round(lb2)` will only be `True` if the exponent is an integer.

idaes/core/util/model_diagnostics.py
```diff
@@ -1045,6 +1061,202 @@ def report_numerical_issues(self, stream=stdout):
...
+            and repn.constant == round(repn.constant)
+            and all(i.domain in integer_domains for i in repn.linear_vars)
+            and all(i[0].domain in integer_domains for i in repn.quadratic_vars)
+            and all(i[1].domain in integer_domains for i in repn.quadratic_vars)
+            and all(i == round(i) for i in repn.linear_coefs)
+            and all(i == round(i) for i in repn.quadratic_coefs)
+    ):
+        # Life is good. The exponent is a linear or quadratic expression containing
+        # only integer variables with integer coefficients
+        return None
```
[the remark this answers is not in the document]
michaelbynum: Negative is okay unless the base is 0. I'll add handling for the base being 0.

idaes/core/util/model_diagnostics.py
```diff
@@ -1045,6 +1061,202 @@ def report_numerical_issues(self, stream=stdout):
...
+    msg += f'base bounds are ({lb1}, {ub1}); '
+    msg += f'exponent bounds are ({lb2}, {ub2})'
+    warn_list.append(msg)
+
+
+def _check_eval_error_log(node: NumericExpression, warn_list: List[str], caution_list: List[str]):
+    _caution_expression_argument(node, node.args, caution_list)
+    lb, ub = _get_bounds_with_inf(node.args[0])
+    if lb <= 0:
+        msg = f'Potential log of a negative number in {node}; Argument bounds are ({lb}, {ub})'
```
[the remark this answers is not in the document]
michaelbynum: Will do

michaelbynum:
> @michaelbynum Just to put down in writing the comments I made at the working group meeting, suggest renaming `report_potential_evaluation_errors` to `display_potential_evaluation_errors` to be more consistent with our naming elsewhere, and then adding a check in `report_structural_issues` to collect the potential evaluation errors and report the number of issues found in the summary (if greater than zero).

Done.

idaes/core/util/model_diagnostics.py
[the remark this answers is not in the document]
michaelbynum: Fixed.

idaes/core/util/model_diagnostics.py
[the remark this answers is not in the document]
michaelbynum: Fixed.

idaes/core/util/model_diagnostics.py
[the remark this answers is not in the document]
michaelbynum: Fixed

michaelbynum:
Working on tests now.

Robbybp:
Would it be possible or make sense to include an option where bounds are treated as strict inequalities? I.e. I think some modelers would think that `log(x); x >= 0` is acceptable if their solver is (strictly) feasible path with respect to this bound.

michaelbynum:
> Would it be possible or make sense to include an option where bounds are treated as strict inequalities? I.e. I think some modelers would think that `log(x); x >= 0` is acceptable if their solver is (strictly) feasible path with respect to this bound.

Good point. I agree. Should it be an option, or should that just be the behavior?

michaelbynum:
I guess it is solver dependent, so it should probably be an option.

idaes/core/util/model_diagnostics.py
```diff
@@ -1229,6 +1244,10 @@ def report_structural_issues(self, stream=None):
         warnings, next_steps = self._collect_structural_warnings()
         cautions = self._collect_structural_cautions()
 
+        eval_error_warnings, eval_error_cautions = self._collect_potential_eval_errors()
```
andrewlee94, line 1247: 1. Is there an easy way this could be moved inside the `_collect_structural_warnings` and `_collect_structural_cautions` methods? It looks like you did this for efficiency so you only need to walk the model once. However, there is an `assert_no_structural_issues` method as well (that only checks for warnings) which could not include this as written (although the solution might be to just add this code there as well). I am also working on adding diagnostics to the convergence tester (parameter sweep) tool, and I am looking at using these methods there too (which would mean they need to become public).
2. Could you add an entry to `next_steps` as well pointing users to the associated `display` method.

andrewlee94 (commented):
A couple of (hopefully) small requests from me. There are a few ways we could address the first comment, and we might want to discuss different options to work out the best way to do this.

idaes/core/util/model_diagnostics.py
michaelbynum: I think I have addressed both of these now. I realized I was being silly with the cautions. They were completely unnecessary/wrong.

michaelbynum:
@Robbybp, I added an option as you suggested.

michaelbynum:
If tests pass, then this will hopefully be ready to go.

idaes/core/util/model_diagnostics.py
```diff
@@ -251,6 +266,14 @@ def svd_sparse(jacobian, number_singular_values):
...
     ),
 )
+CONFIG.declare(
+    "strict_evaluation_error_detection",
+    ConfigValue(
+        default=True,
+        domain=bool,
+        description="If False, warnings will not be generated for things like log(x) with x >= 0",
+    ),
+)
```
Robbybp, line 276: I'm not sure what the default should be here. For Ipopt, with default options (i.e. `bound_relax_factor=1e-8`), we are strictly feasible path with respect to bounds of zero, but not other bounds (at least according to this option's documentation). So something like `log(1 - x); x <= 1` could still hit the evaluation error. I think I lean towards treating this as an advanced option that people can turn on if they know what they're doing, although I think we will have to test this in the wild to really know what the "right" default is.

michaelbynum: I tested the example you mentioned with x initialized to 2, and I did not get an evaluation error (if I set the nlp_scaling_method to none).

michaelbynum: Which is consistent with the way we do fraction to the boundary in parapint.

michaelbynum: That said, I agree with your overall point.

Robbybp: > I tested the example you mentioned with x initialized to 2, and I did not get an evaluation error (if I set the nlp_scaling_method to none).

What if you set `bound_push` to zero (or anything less than `1*bound_relax_factor`, I think)?

The point is just that the evaluation error *could* happen.

michaelbynum: I think we are agreeing? The default will produce a warning for the example you gave.

Robbybp: Oh I interpreted the name of the option as "treat bounds as if they are strict". But instead, it is "be more strict about the region where evaluation errors occur"?

michaelbynum: Exactly. Naming...

Robbybp: I would prefer the name to describe how the bounds are treated, rather than how the evaluation error region is treated (as this is mathematically constant). What about `"eval_error_strict_bounds"`? I'm also open to other arguments/proposals.

andrewlee94 (approved):

idaes/core/util/model_diagnostics.py
michaelbynum: Just to be clear, are you suggesting renaming and reversing the meaning?

Robbybp: Yes, I'm suggesting renaming the option, although I'd prefer to get another opinion, as naming preferences can be fairly idiosyncratic.

I would be fine with a name that describes the "region of evaluation error", but would prefer to call it something other than "strict" as this makes me think of how the bounds are handled (unless others think the name makes sense).

andrewlee94: Personally I read "strict_evaluation_error_detection" in the sense that @michaelbynum used it, but equally when I read the doc string I asked "what exactly happens when it is set to `False` and why?" This does have me leaning towards @Robbybp's suggestion as would focus the documentation on "why you might want to set this to relax the definition of bounds".

Robbybp: Since this is a potentially niche use case, maybe we should defer including this option for now (and keep the default behavior of treating bounds as if they are non-strict), then add this later if we find that the output is often dominated by bounds of zero that we don't actually expect to hit.

ksbeattie:
Ping.  We're looking at cutting the RC just after Thanksgiving, will this make it?

michaelbynum:
> @Robbybp & @michaelbynum: ping. We're looking at cutting the RC just after Thanksgiving, will this make it?

Should make it.

idaes/core/util/model_diagnostics.py
michaelbynum: @andrewlee94, are you okay with removing the option for now and adding later?

andrewlee94: That is fine by me.

andrewlee94:
@michaelbynum Note that you have a few lingering pylint issues too:

 idaes/core/util/model_diagnostics.py:59:0: E0611: No name 'generate_standard_repn' in module 'pyomo.repn.standard_repn' (no-name-in-module)
idaes/core/util/model_diagnostics.py:1337:8: W0612: Unused variable 'res' (unused-variable)
idaes/core/util/model_diagnostics.py:1358:4: C0116: Missing function or method docstring (missing-function-docstring)
idaes/core/util/model_diagnostics.py:1840:4: C0116: Missing function or method docstring (missing-function-docstring)
idaes/core/util/model_diagnostics.py:25:0: W0611: Unused Sequence imported from typing (unused-import)
idaes/core/util/model_diagnostics.py:32:0: W0611: Unused is_fixed imported from pyomo.environ (unused-import)

idaes/core/util/model_diagnostics.py
```diff
@@ -251,6 +267,14 @@ def svd_sparse(jacobian, number_singular_values):
         description="Tolerance for raising a warning for small Jacobian values.",
     ),
 )
+CONFIG.declare(
+    "warn_for_evaluation_error_at_bounds",
```
michaelbynum, line 271: @Robbybp, @andrewlee94 - what do you think of this name?

andrewlee94: Sounds good to me.

andrewlee94 (approved):

idaes/core/util/model_diagnostics.py
Robbybp: Looks good to me

michaelbynum: Awesome. Thank you both.

michaelbynum:
I'm not sure why tests are getting skipped...

andrewlee94:
@michaelbynum I think it is OK - it is just GitHub getting confused.

Robbybp (approved):

PR #1276, opened 2023-10-02
[WIP] Unpack idaes-local tar file in get-extensions command

Robbybp:
## Fixes
Part of https://github.com/IDAES/idaes-ext/issues/261

## Summary/Motivation:
Make CyIpopt as easy to use as possible.

- After unpacking `idaes-solvers-<os>.tar.gz` into `.idaes/bin`, unpack `idaes-local-<os>.tar.gz` (which is included in idaes-solvers) into `.idaes`

Opening as a WIP as there are some unresolved details:
- Current implementation assumes that the parent directory of the "install-to directory" is writable (and the correct place to unpack idaes-local). I'm not sure what the best way to specify this directory is.
- Implementation could probably be streamlined if we distributed `idaes-local-<os>.tar.gz` on the downloads page of idaes-ext (instead of inside the idaes-solvers tar file)
- We currently distribute idaes-local for MacOS as `idaes-local-darwin-arm64.tar.gz`, which is inconsistent with all the other tar files distributed for MacOS (which are `*-darwin-aarch64.tar.gz`). I would recommend changing this in idaes-ext

### Legal Acknowledgement

### DIFF
--- a/idaes/commands/util/download_bin.py
+++ b/idaes/commands/util/download_bin.py
@@ -150,7 +150,15 @@ def _get_checksums(fd, to_path, release, nochecksum):
     return _read_checksum_file(check_to)
 
 
-def _create_download_package(platform, to_path, url, extra, extras_only, library_only):
+def _create_download_package(
+    platform,
+    to_path,
+    url,
+    extra,
+    extras_only,
+    library_only,
+    local_only=False,
+):
     pname = []
     ptar = []
     ftar = []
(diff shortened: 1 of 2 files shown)


ksbeattie:
@Robbybp is this expected for the Nov release?

Robbybp:
I'm going to close this for now. Once https://github.com/IDAES/idaes-ext/pull/262 is merged, we will need to cut an idaes-ext release. At approximately the same time, I will re-open this PR (and probably combine it with an `idaes install cyipopt` command). While this PR could go forward as-is, it will need to be rewritten once https://github.com/IDAES/idaes-ext/pull/262 is merged, and the benefit of unpacking the idaes-local tar file will not be realized until then either.

@ksbeattie As to whether this is expected for the Nov release, it depends on how long it takes us to get https://github.com/IDAES/idaes-ext/pull/262 merged. My preference would be to get this in, but I will be at a conference next week and probably not able to work on https://github.com/IDAES/idaes-ext/pull/262 until I'm back, so I'd say it's a toss-up.

PR #1282, opened 2023-11-01
Add SCIP solver for testing

### LINKED ISSUE #1281: Enabling SCIP as an available solver for testing
- This would be needed to fully test some functionality, e.g. #1256
- SCIP can be installed using Conda (requires an activated Conda environment): `conda install -c conda-forge scip`
- However, this also causes a different (in the sense of non-`idaes-ext`) version of IPOPT to be installed in the Conda environment:
  ```sh
  $ python -c "from pyomo.environ import SolverFactory as F; s = F('ipopt'); print(s.executable())"
  /opt/conda/envs/test-idaes-scip/bin/ipopt
  $ python -c "import idaes; from pyomo.environ import SolverFactory as F; s = F('ipopt'); print(s.executable())"
  /home/ludo/.idaes/bin/ipopt
  ```

My vague concern is that having two versions of IPOPT might lead to hard(er)-to-debug environment issues. However, I don't know enough about the Pyomo solver interface and/or the internals of `idaes-ext` to understand whether this can be problematic, so I've opened this issue as a spin-off of an ongoing conversation with @andrewlee94 to discuss it with a wider group.

lbianchi-lbl:
- Add non-public util function to enable SCIP solver (using `ampl_module_scip`)
- Add `scip_solver` pytest fixture for model diagnostics tests
  - If a test uses the `scip_solver` fixture (i.e. requires SCIP to run), it will be automatically skipped without the need of adding individual `pytest.mark.skipif()` markers if SCIP is not available
- Add step in GHA installation action to install `ampl_module_scip` from https://pypi.ampl.com

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/testing.py
+++ b/idaes/core/util/testing.py
@@ -21,8 +21,13 @@ This module contains utility functions for use in testing IDAES models.
 __author__ = "Andrew Lee"
 
 
+import os
+from typing import Callable, Union
+
 from pyomo.environ import Constraint, Set, units, Var
 from pyomo.common.config import ConfigBlock
+from pyomo.common import Executable
+from pyomo.common.dependencies import attempt_import
 
 from idaes.core import (
     declare_process_block_class,
@@ -409,3 +414,31 @@ class ReactionBlockData(ReactionBlockDataBase):
             return MaterialFlowBasis.mass
         else:
             return MaterialFlowBasis.other
+
+
+def _enable_scip_solver_for_testing(
+    name: str = "scip",
+) -> Union[Callable[[], None], None]:
+    ampl_module_scip, is_available = attempt_import("ampl_module_scip")
+    if not is_available:
+        _log.warning(
+            "ampl_module_scip must be installed to enable SCIP solver for testing"
+        )
+        return None
+
+    new_path_entry = ampl_module_scip.bin_dir
+    # TODO prepending new_path_entry to PATH means that the "testing" SCIP will "shadow"
+    # existing executable directories already on PATH
+    # this behavior would match the (implied) semantics of this function, i.e.
+    # "enable SCIP solver used for testing (at the expense of other non-testing SCIP executables)"
+    os.environ["PATH"] = os.pathsep.join([new_path_entry, os.environ["PATH"]])
+    Executable.rehash()
+
+    def undo_changes():
+        try:
+            os.environ["PATH"].remove(new_path_entry)
+            Executable.rehash()
+        except ValueError:
+            pass
+
+    return undo_changes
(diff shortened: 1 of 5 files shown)


idaes/core/util/testing.py
```diff
@@ -409,3 +414,31 @@ def get_reaction_rate_basis(b):
...
+    # TODO prepending new_path_entry to PATH means that the "testing" SCIP will "shadow"
+    # existing executable directories already on PATH
+    # this behavior would match the (implied) semantics of this function, i.e.
+    # "enable SCIP solver used for testing (at the expense of other non-testing SCIP executables)"
+    os.environ["PATH"] = os.pathsep.join([new_path_entry, os.environ["PATH"]])
+    Executable.rehash()
+
+    def undo_changes():
+        try:
+            os.environ["PATH"].remove(new_path_entry)
```
lbianchi-lbl, line 439: Ughhh, I just realized that this won't work since `os.environ["PATH"]` is a `str` rather than a `list`. Serves me right for trying to be smart and implement a way to revert global changes to the environment...

lbianchi-lbl, line 439: Ughhh, I just realized that this won't work since os.environ["PATH"] is a str rather than a list. Serves me right for trying to be smart and implement a way to revert global changes to the environment...

andrewlee94 (approved):

idaes/core/util/testing.py
```diff
@@ -409,3 +414,36 @@ def get_reaction_rate_basis(b):
...
+    # "enable SCIP solver used for testing (at the expense of other non-testing SCIP executables)"
+    os.environ["PATH"] = os.pathsep.join([new_path_entry, os.environ["PATH"]])
+    Executable.rehash()
+
+    def remove_from_path():
+        path_as_list = os.environ["PATH"].split(os.pathsep)
+
+        try:
+            path_as_list.remove(new_path_entry)
+        except ValueError:
```
mrmundt, line 442: It's not a bad idea to add a test to cover this case to future-proof against any changes in how AMPL and/or Python handle their `PATH` entries / modifications in the future.

lbianchi-lbl: I agree it'd be nice to have the "undo" functionality tested. It's a bit tricky but I'll see what I can do

lbianchi-lbl: @mrmundt I've added a basic test for the PATH manipulation: https://github.com/IDAES/idaes-pse/pull/1282/commits/986ba84bbc6b51640fa3a590083dde10523f2685

idaes/core/util/tests/test_scip_for_testing.py
mrmundt, line 1: Thanks, @lbianchi-lbl ! Can I note how funny this name is, though? `test for testing` ;)

mrmundt (approved):

idaes/core/util/tests/test_scip_for_testing.py
lbianchi-lbl: Thanks for noting, that was very much on purpose! I like to think of it as "Barber paradox for software engineers" :sweat_smile:

PR #1288, opened 2023-11-16
Fix methods in SVD Toolbox

### LINKED ISSUE #1285: SVD Toolbox display_constraints_including_variable error
When using `display_constraints_including_variable`, I got an error:

```
tbx.display_constraints_including_variable(m.fs.stripper_section.reflux_mixer.rich_solvent_state[0.0].temperature)
Traceback (most recent call last):
  File "c:\users\dallan\repos\idaes-pse\idaes\core\util\model_diagnostics.py", line 1553, in display_constraints_including_variable
    var_idx = var_list.index(variable)
  File "c:\users\dallan\repos\pyomo\pyomo\core\expr\relational_expr.py", line 330, in __bool__
    return super().__bool__()
  File "c:\users\dallan\repos\pyomo\pyomo\core\expr\relational_expr.py", line 47, in __bool__
    raise PyomoException(
pyomo.common.errors.PyomoException: Cannot convert non-constant Pyomo expression (fs.stripper_section.reflux_mixer.rich_solvent_state[0.0].temperature  ==  fs.absorber_section.makeup_mixer.mixed_state[0.0].mole_frac_comp[MEA]) to bool.
This error is usually caused by using a Var, unit, or mutable Param in a
Boolean context such as an "if" statement, or when checking container
membership or equality. For example,
    >>> m.x = Var()
    >>> if m.x >= 1:
    ...     pass
and
    >>> m.y = Var()
    >>> if m.y in [m.x, m.y]:
    ...     pass
would both cause this exception.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "c:\users\dallan\repos\idaes-pse\idaes\core\util\model_diagnostics.py", line 1555, in display_constraints_including_variable
    raise AttributeError(f"Could not find {variable.name} in model.")
AttributeError: Could not find fs.stripper_section.reflux_mixer.rich_solvent_state[0.0].temperature in model.
```

That variable is in fact in the model. I'm pretty sure that this is the result of using `list().index` in order to perform the variable lookup. I think that method uses `==` internally, which leads to the Pyomo error. I'm not sure why it failed this time while it succeeded previously in the same debugging session. You can perform the same lookup by using `nlp.get_primal_indices` (beware, it takes a list of constraints), or `nlp.get_constraint_indices` for constraints.

I'll probably end up writing the fix myself.

dallan-keylogic:
## Summary/Motivation:
Use PyNumero methods to get constraint/variable indices rather than `list.index` because (I think) `list.index` uses `==` to compare elements, which is not what should be done for Pyomo objects. Fixes #1285 .

Also add in string formatting for magnitude of Jacobian elements. Now it uses scientific notation with three digits past the decimal place. It's a lot better than having to deal with 16 digits.

### Legal Acknowledgement


idaes/core/util/model_diagnostics.py
```diff
@@ -1450,6 +1450,10 @@ def __init__(self, model: _BlockData, **kwargs):
             self._model, scaled=False, equality_constraints_only=True
         )
 
+        # Get list of equality constraint and variable names
+        self.eq_con_list = self.nlp.get_pyomo_equality_constraints()
```
andrewlee94, line 1454: I think you should make these private. There is no reason the user should be changing these.

andrewlee94 (changes requested):
Only one minor request.

idaes/core/util/model_diagnostics.py
dallan-keylogic: I don't think they should be changing them, but they might have the desire to look at them. We can still make them semiprivate, though.

andrewlee94 (approved):

Robbybp (approved):
Good fix. `list.index` is indeed not safe here.

idaes/core/util/model_diagnostics.py
```diff
@@ -1620,23 +1620,19 @@ def display_constraints_including_variable(self, variable, stream=None):
...
         # Get index of variable in Jacobian
         try:
-            var_idx = var_list.index(variable)
-        except (ValueError, PyomoException):
+            var_idx = self.nlp.get_primal_indices([variable])[0]
+        except (KeyError, PyomoException):
             raise AttributeError(f"Could not find {variable.name} in model.")
 
-        nonzeroes = self.jacobian[:, var_idx].nonzero()
+        nonzeros = self.jacobian.getcol(var_idx).nonzero()
```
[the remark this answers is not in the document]
Robbybp: `getcol` and indexing do the same thing AFAIK. So one is slow and one is fast depending on whether the matrix CSC or CSR. If the matrix is COO, both are slow.

PR #1291, opened 2023-11-20
Add support for dynamics to MS Contactor

andrewlee94:
## Summary/Motivation:
This add support for dynamics to the MS Contactor model.

- Add method to construct volume and volume fraction terms.
- Add phase fractions to handle multiphase streams
- Added code to construct holdup and accumulation terms as required
- Added accumulation terms to balance equations
- Added units to 0 term in balances for steady-state cases

### Legal Acknowledgement

### DIFF
--- a/idaes/models/unit_models/mscontactor.py
+++ b/idaes/models/unit_models/mscontactor.py
@@ -18,7 +18,16 @@ from functools import partial
 from pandas import DataFrame
 
 # Import Pyomo libraries
-from pyomo.environ import Block, Constraint, RangeSet, Reals, Set, units, Var
+from pyomo.environ import (
+    Block,
+    Constraint,
+    Expression,
+    RangeSet,
+    Reals,
+    Set,
+    units,
+    Var,
+)
 from pyomo.common.config import ConfigDict, ConfigValue, Bool, In
 from pyomo.contrib.incidence_analysis import solve_strongly_connected_components
 from pyomo.dae import DerivativeVar
@@ -586,6 +595,10 @@ class MSContactorData(UnitModelBlockData):
             )
             def sum_volume_frac(b, t, e):
                 return 1 == sum(b.volume_frac_stream[t, e, s] for s in b.streams)
+
+            for stream in self.config.streams.keys():
+                phase_list = getattr(self, stream).phase_list
+                _add_phase_fractions(self, stream, phase_list)
 
     def _build_material_balance_constraints(self, flow_basis, uom):
         # Get units for transfer terms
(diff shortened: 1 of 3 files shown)


agarciadiego (approved):
LGTM

ArkoprabhoDasgupta (approved):

idaes/models/unit_models/mscontactor.py
```diff
@@ -1246,11 +1394,78 @@ def _pressure_balance_rule(blk, t, s, stream, uom):
...
+        sum_of_phase_fractions = Constraint(
+            b.flowsheet().time,
+            b.elements,
+            rule=partial(
+                _sum_phase_frac_rule, phase_frac=phase_fraction, phase_list=phase_list
+            ),
+            doc=f"Sum of phase fractions constraint for stream {stream}",
+        )
+        b.add_component(stream + "_sum_phase_fractions", sum_of_phase_fractions)
+
```
dallan-keylogic, line 1453: Phase volume fraction needs to be given by the property package, and appropriate constraints need to be written here. 

```
def phase_volume_frac_rule(b, p):
    return b.flow_mol_phase[p] * b.dens_mol_phase[p] / sum( b.flow_mol_phase[q] * b.dens_mol_phase[q] for q in b.params.phase_set)
```

idaes/models/unit_models/mscontactor.py
```diff
@@ -558,15 +569,49 @@ def _build_heterogeneous_reaction_blocks(self):
...
+    def _add_geometry(self, uom):
+        if self.config.has_holdup:
+            # Add volume for each element
+            # TODO: Assuming constant volume for now
+            self.volume = Var(
+                self.elements,
+                initialize=1,
+                units=uom.VOLUME,
+                doc="Volume of element",
+            )
```
dallan-keylogic, line 581: Consider if you want to index `self.volume` by time now if you have plans to make it time-varying eventually. Having an index that a variable is constant along is easier than trying to add the index after the fact.

dallan-keylogic (changes requested):
I think partitioning volume between different streams is fine, with the proviso that specific submodels need to provide the constraints that perform that partitioning. However, the phase volume fraction needs to be provided by property packages, and that constraint should be added as part of this model.

idaes/models/unit_models/mscontactor.py
dallan-keylogic: Actually, I got mixed up, it should be 
````
def phase_volume_frac_rule(b, p):
    return b.flow_mol_phase[p] * b.vol_mol_phase[p] / sum( b.flow_mol_phase[q] * b.vol_mol_phase[q] for q in b.params.phase_set)
```

dallan-keylogic (approved):
Upon further reflection, phase fraction is handled appropriately now, its calculation is outsourced to child models of this class, like other unit models involving the `ControlVolume`.

PR #1293, opened 2023-11-28
Fixing errors due to bound checking in nl writer

andrewlee94:
## Summary/Motivation:
Pyomo 6.7.0 added checks for fixed variables which violate bounds which caused a number of our tests to fail (about half were deliberate violations in the diagnostics tools for testing, but the rest were real).

- Updated tests to remove bound violations
-

### Legal Acknowledgement


jsiirola (approved):
Looks reasonable, but is there a plan to restore the tests that trigger the initialization error?

andrewlee94:
@jsiirola Not at this point, as the `model.initialize()` API is deprecated.

PR #1295, opened 2023-12-01
Fixing bug in TSA tests

### LINKED ISSUE #1290: Bug in TSA tests
There are a few bugs/issues in the tests for the TSA column, specifically `test_report` (line 166) in `test_fixed_bed_tsa_0d.py`

1. The test fails when run with only `unit` tests (`pytest -m unit`) due to the model not having been solved. This is because the tests involving a solver are `component` tests which do not get run in this case resulting in uninitialized values. This test should be either flagged as a `component` test, or better, the underlying model reworked so that this method will not fail.
2. This test does not assert anything, and thus only catches exceptions. Rather than test the `report` method, this should test the underlying methods for collecting the stream table and performance components.
3. Tests for the summary method should be broken out into a separate test.
4. The plotting functions should also be separated into a different test, although it is hard to assert anything with these.

AlexNoring:
## Fixes #1290

Fixes #1290 by adding initial values to all model variables. It also adds assert statements to the reporting method tests and moves the tests for the summary and plotting methods into separate tests.

### Legal Acknowledgement


andrewlee94 (approved):
Looks good.

bpaul4 (approved):

PR #1297, opened 2023-12-06
Rename SVDToolbox to JacobianAnalysisToolbox

andrewlee94:
## Summary/Motivation:
In preparation for adding some additional diagnostics tools, this PR renamed the SVDToolbox to the more generic JacobianAnalysisToolbox. Seeing as the `SVDToolbox` was added after the v2.1 release, this should not cause any backward compatibility issues.

- Rename SVDToolbox
- Update docs accordingly
- Fix a minor bug in Degeneracy Hunter related to avoid an edge case with trivial constraints.

### Legal Acknowledgement


MarcusHolly (approved):

bpaul4 (approved):

Robbybp:
I think this name might be general enough to be somewhat confusing. `DiagnosticsToolbox` already does a significant amount of Jacobian analysis. Also, my interpretation of the purpose of `SVDToolbox` is to cache the result of the SVD, then call useful methods on it. I worry that if we turn it into a general Jacobian analysis toolbox, we will end up with too much state to easily keep track of. Additionally, we may run into conflicting method definitions. For instance, if we add a QR factorization method, then `display_rank_of_equality_constraints` will need to decide whether to use the result of SVD or the result of QR.

If we go the route of a general Jacobian analysis toolbox, I would like to start with a well-defined idea of what we are caching and how we specify the method to use for a particular analysis.

dallan-keylogic:
@Robbybp It'd be my preference that the Jacobian analysis tools in the normal diagnostics toolbox get merged over here, rather than the other way around. Some example workflow ended up in my FOCAPD paper, where you determine that something is a problematic Var using the DiagnosticsToolbox then you have to use the SVDToolbox to figure out which Constraints it appears in. The other option is adding those methods to the normal diagnostics toolbox.

Robbybp:
> @Robbybp Would your suggestion be to leave this as-is and put the new ill-conditioning tools into `DiagnosticsToolbox` then?

That is my current recommendation, at least for the parallel row/col identifier. There is nothing that stops us from adding `JacobianAnalysisToolbox` later as a pure addition if we want a single class that is in charge of caching expensive linear algebra operations.

I am less confident about what to do with the optimization-based analysis tool. I still think this is very similar to Degeneracy Hunter:
- Both solve optimization problems where the result is a vector of weights that prove (almost) linear dependence
- Degeneracy Hunter minimizes the 0-norm of the weight vector. Whereas the ModelAnalyzer tool minimizes the parameter that relaxes the linear dependence constraint, with an option to add a 1-norm penalty to the objective as regularization.
- Degeneracy Hunter normalizes with `y[i].fix(1)`, while ModelAnalyzer normalizes with `sum(y[:]) == 1` (or maybe it enforces that the 1-norm is 1, I forget).

I think the right option medium-term is probably to combine these into a more general formulation, so we don't have multiple LP/MIP-based ill-conditioning/degeneracy analyzers lying around. For now I think I advocating putting the analyzer into a stand-alone function.

bknueven:
> I am less confident about what to do with the optimization-based analysis tool. I still think this is very similar to Degeneracy Hunter:
> 
>     * Both solve optimization problems where the result is a vector of weights that prove (almost) linear dependence
> 
>     * Degeneracy Hunter minimizes the 0-norm of the weight vector. Whereas the ModelAnalyzer tool minimizes the parameter that relaxes the linear dependence constraint, with an option to add a 1-norm penalty to the objective as regularization.
> 
>     * Degeneracy Hunter normalizes with `y[i].fix(1)`, while ModelAnalyzer normalizes with `sum(y[:]) == 1` (or maybe it enforces that the 1-norm is 1, I forget).
> 
> 
> I think the right option medium-term is probably to combine these into a more general formulation, so we don't have multiple LP/MIP-based ill-conditioning/degeneracy analyzers lying around. For now I think I advocating putting the analyzer into a stand-alone function.

I agree - I think the ModelAnalyzer should eventually be rolled into Degeneracy Hunter. ModelAnalyzer uses a LASSO approach as compared to Degeneracy Hunter's best subset selection approach.

For very degenerate problems I think the ModelAnalzyer could benefit from an implementation which is more integrated with a specific LP code. In either case the LP or MIP ends up sharing some of the same issues as the original NLP.

PR #1298, opened 2023-12-08
Add custom sampling option to PySMO

OOAmusat:
## Summary/Motivation:
This PR adds a sampling method that allows users to specify the sampling distributions to be used for each input variable/column. Current interface provides random, uniform and normal distribution sampling options.

- Add ``CustomSampling`` class
- Add tests

### Legal Acknowledgement


bpaul4 (approved):
I think this looks very good, and the coverage is very high as well. Nicely done!

idaes/core/surrogate/pysmo/sampling.py
```diff
@@ -1536,3 +1535,199 @@ def sample_points(self):
...
+        elif (sampling_type.lower() == "creation") or (
+            sampling_type.lower() == "selection"
+        ):
+            sampling_type = sampling_type.lower()
+            self.sampling_type = sampling_type
+        else:
+            raise Exception(
+                'Invalid sampling type requirement entered. Enter "creation" for sampling from a range or "selection" for selecting samples from a dataset.'
+            )
+        print("Sampling type: ", self.sampling_type, "\n")
```
andrewlee94, line 1613: This is probably a comment for the future, but we should probably switch to using loggers for all this output so that we can control how much gets seen and where.

idaes/core/surrogate/pysmo/sampling.py
```diff
@@ -1536,3 +1535,199 @@ def sample_points(self):
...
+            elif len(data_input) != 2:
+                raise Exception("data_input must contain two lists of equal lengths.")
+            elif not isinstance(data_input[0], list) or not isinstance(
+                data_input[1], list
+            ):
+                raise Exception("data_input must contain two lists of equal lengths.")
+            elif len(data_input[0]) != len(data_input[1]):
+                raise Exception("data_input must contain two lists of equal lengths.")
+            elif data_input[0] == data_input[1]:
+                raise Exception("Invalid entry: both lists are equal.")
```
andrewlee94, line 1653: I think this misses the case where any two matching pairs are equal. I.e. it only checks if ALL dimensions have lower bounds equal to upper bounds, not that any dimension has matching bounds.

idaes/core/surrogate/pysmo/sampling.py
```diff
@@ -1536,3 +1535,199 @@ def sample_points(self):
...
+            elif number_of_samples <= 0:
+                raise Exception("number_of_samples must a positive, non-zero integer.")
+            self.number_of_samples = number_of_samples
+
+        elif self.sampling_type == "creation":
+            if not isinstance(data_input, list):
+                raise ValueError(
+                    'List entry of two elements expected for sampling_type "creation."'
+                )
+            elif len(data_input) != 2:
```
andrewlee94, line 1644: These three checks all produce the same exception - this would seem to me to be a case of a single `elif` with  a set of `or` statements.

idaes/core/surrogate/pysmo/sampling.py
```diff
@@ -1536,3 +1535,199 @@ def sample_points(self):
...
+                raise Exception("data_input must contain two lists of equal lengths.")
```
andrewlee94, line 1645: Is a base `Exception` the right error here? I would think a `ValueError` would be more appropriate.

idaes/core/surrogate/pysmo/sampling.py
```diff
@@ -1536,3 +1535,199 @@ def sample_points(self):
...
+                )
+
+            # Catch potential errors in number_of_samples
+            if number_of_samples is None:
+                print(
+                    "\nNo entry for number of samples to be generated. The default value of 5 will be used."
+                )
+                number_of_samples = 5
+            elif number_of_samples > self.data.shape[0]:
+                raise Exception(
```
andrewlee94, line 1630: Base `Exception` again.

idaes/core/surrogate/pysmo/sampling.py
```diff
@@ -1536,3 +1535,199 @@ def sample_points(self):
...
+            self.data_headers = data_headers
+
+            # Catch potential errors in number_of_samples
+            if number_of_samples is None:
+                print(
+                    "\nNo entry for number of samples to be generated. The default value of 5 will be used."
+                )
+                number_of_samples = 5
+            elif not isinstance(number_of_samples, int):
+                raise Exception("number_of_samples must be an integer.")
```
andrewlee94, line 1674: As above.

idaes/core/surrogate/pysmo/sampling.py
```diff
@@ -1536,3 +1535,199 @@ def sample_points(self):
...
+        self.dist_vector = list_of_distributions
+
+    def generate_from_dist(self, dist_name):
+        if dist_name.lower() in ["uniform", "random"]:
+            dist = getattr(np.random.default_rng(), dist_name.lower())
+            var_values = np.array(dist(size=self.number_of_samples))
+            return dist, var_values
+        elif dist_name.lower() == "normal":
+            dist = getattr(np.random.default_rng(), "normal")
+            var_values = dist(loc=0.5, scale=1 / 6, size=self.number_of_samples)
```
andrewlee94, line 1706: How are you determining the mean and standard deviation here? From the code, I suspect you are taking the mean to be the average of the upper and lower bound and the standard deviation to be the difference in bounds divided by 6.

I think most users would expect to provide the mean and standard deviation explicitly, with the possibility of providing upper and lower bounds to prevent extreme values.

idaes/core/surrogate/pysmo/sampling.py
```diff
@@ -1536,3 +1535,199 @@ def sample_points(self):
...
+            if sum(
+                [1 for i in range(0, var_values.shape[0]) if var_values[i] > 1]
+            ) + sum([1 for i in range(0, var_values.shape[0]) if var_values[i] < 0]):
+                warnings.warn(
+                    "Points adjusted to remain within specified Gaussian bounds."
+                )
+            var_values_truncated = np.array(
+                [1.0 if j > 1.0 else 0.0 if j < 0.0 else j for j in var_values]
```
andrewlee94, line 1714: What happens if you have multiple values that exceed the bound? Wouldn't that create the possibility of multiple samples with the same value (even if the chances are very small)?

idaes/core/surrogate/pysmo/tests/test_sampling.py
```diff
@@ -2240,5 +2241,631 @@ def test_sample_points_02(self, array_type):
...
+        np.testing.assert_array_equal(CSClass.data, input_array)
+        np.testing.assert_array_equal(CSClass.number_of_samples, 6)
+        np.testing.assert_array_equal(CSClass.x_data, np.array(input_array)[:, :-1])
+        assert CSClass.dist_vector == ["uniform", "normal"]
+
+    @pytest.mark.unit
+    @pytest.mark.parametrize("array_type", [np.array, pd.DataFrame])
+    def test__init__selection_03(self, array_type):
+        input_array = array_type(self.input_array)
+        with pytest.raises(Exception):
```
andrewlee94, line 2290: I would recommend testing that the right `Exception` is raised here by using the `match` argument. This also helps make it clear what each test is looking for.

idaes/core/surrogate/pysmo/tests/test_sampling.py
```diff
@@ -2240,5 +2241,631 @@ def test_sample_points_02(self, array_type):
...
+            [i, j, ((i + 1) ** 2) + ((j + 1) ** 2)]
+            for i in np.linspace(0, 10, 21)
+            for j in np.linspace(0, 10, 21)
+        ]
+    )
+    full_data = {"x1": y[:, 0], "x2": y[:, 1], "y": y[:, 2]}
+
+    @pytest.mark.unit
+    @pytest.mark.parametrize("array_type", [np.array, pd.DataFrame])
+    def test__init__selection_01(self, array_type):
```
andrewlee94, line 2258: Could you use more meaningful test names here? E.g. `test_selection_no_samples` instead of `test__init__selection_03`.

This makes it much easier to debug in future if something goes wrong as the name of the test tells you something about what should have happened.

idaes/core/surrogate/pysmo/tests/test_sampling.py
```diff
@@ -2240,5 +2241,631 @@ def test_sample_points_02(self, array_type):
...
+            )
+            for dist_type in ["uniform", "random", "uniform"]:
+                dist_res, scaled_samples = CSClass.generate_from_dist(dist_type)
+                assert type(scaled_samples) == np.ndarray
+                assert scaled_samples.shape == (CSClass.number_of_samples,)
+                assert dist_res.__name__ == dist_type
+
+    @pytest.mark.unit
+    @pytest.mark.parametrize("array_type", [np.array])
+    def test_sample_points_01(self, array_type):
```
andrewlee94, line 2750: These look like tests that could (should) be done using fixed inputs. I.e. do not use the sample generation and mock up a set of known distributions that we can then use to ensure we get the right answer out.

As far as I can tell, `sample_points` is a case where we can test the method in isolation used manufactured data - i.e. provide  the input specification, a mock up of a random dataset and a concrete answer we expect to get back out.

andrewlee94 (changes requested):
A few requests, mostly for more specific exception types and descriptive test names.

idaes/core/surrogate/pysmo/sampling.py
OOAmusat: Fixed, with tests.

idaes/core/surrogate/pysmo/sampling.py
OOAmusat: Fixed.

idaes/core/surrogate/pysmo/sampling.py
OOAmusat: Resolved - we just need to document this properly.

idaes/core/surrogate/pysmo/sampling.py
OOAmusat: Fixed.

OOAmusat: Fixed.

idaes/core/surrogate/pysmo/sampling.py
OOAmusat: Fixed.

idaes/core/surrogate/pysmo/tests/test_sampling.py
OOAmusat: Done

idaes/core/surrogate/pysmo/tests/test_sampling.py
OOAmusat: Done.

idaes/core/surrogate/pysmo/sampling.py
OOAmusat: Yes, we can always change that later.

idaes/core/surrogate/pysmo/sampling.py
OOAmusat: @andrewlee94 yes it would, with enough samples.

The other options are:

1. To return the values as is, but make it clear to the user in the documentation that this is a possibility.
2. Try to re-generate new random numbers from the distribution to replace any values exceeding the bounds. This was my first thought actually, but I'm worried that it could lead to the distribution not being maintained.

So maybe option 1? And make it clear that the bounds are for 99.7%, and that there may be a few values outside of the bounds.

andrewlee94: I think documenting option 1 is the safest for now - another possibility would be to do 1 with the option of 2 (i.e. give an option to strictly enforce the bounds, noting that this means that the distributions will not be strictly maintained).

Either way, this is hopefully mostly theoretical.

OOAmusat: Fixed. Doing 1 with the option of 2. Also added tests that check this (by generating generating 10000 samples from the 0-1 distribution)

idaes/core/surrogate/pysmo/sampling.py
OOAmusat: Done now.

andrewlee94 (approved):

ksbeattie:
@OOAmusat looks like this did not make the Dec 2.3.0 release...

OOAmusat:
> @OOAmusat looks like this did not make the Dec 2.3.0 release...

Yes, I don't know why as it had the required approvals and tests passing...

PR #1299, opened 2023-12-08
Adding diagnostics tools for ill conditioning

andrewlee94:
## Summary/Motivation:
This PR adds new diagnostics capabilities for detecting near-parallel variables and constraints and for finding sources of ill conditioning in models (similar to DegeneracyHunter).

- Adds method to check for near parallel rows and columns in Jacobians
- Add hook for checking for parallel rows and columns in DiagnosticsToolbox
- Adds beta method to analyze Jacobian for near singularity based on rows or columns (with logger warning about future changes)

### Legal Acknowledgement


idaes/core/util/model_diagnostics.py
```diff
@@ -15,7 +15,7 @@
 This module contains a collection of tools for diagnosing modeling issues.
 """
 
-__author__ = "Alexander Dowling, Douglas Allan, Andrew Lee"
+__author__ = "Alexander Dowling, Douglas Allan, Andrew Lee, Robby Parker, Ben Kneuven"
```
bknueven, line 18:
```suggestion
__author__ = "Alexander Dowling, Douglas Allan, Andrew Lee, Robby Parker, Ben Knueven"
```

idaes/core/util/model_diagnostics.py
```diff
@@ -3051,6 +3128,236 @@ def ipopt_solve_halt_on_error(model, options=None):
...
+                jidx = var_list.index(j)
+                iidx = con_list.index(i)
+                jac_entry = jac[iidx, jidx]
+
+            if jac_entry != 0:
+                lhs += jac_entry * b.y[j]
+
+        return lhs == b.res[i]
+
+    inf_prob.by = Constraint(set2, rule=b_rule)
```
bknueven, line 3303: After using this on some larger flowsheets, I found the model build time to be fairly long. We could probably significantly speed it up by iterating over the non-zeros in the Jacobian instead.

We'd need to re-work the logic around selecting the row or column to do so.

idaes/core/util/model_diagnostics.py
```diff
@@ -3051,6 +3128,236 @@ def ipopt_solve_halt_on_error(model, options=None):
...
+    # Objective -- minimize residual
+    inf_prob.min_res = Objective(expr=inf_prob.res_norm)
+
+    solver = SolverFactory("cbc")  # TODO: Consider making this an option
+
+    # tighten tolerances  # TODO: If solver is an option, need to allow user options
+    solver.options["primalT"] = inverse_target_kappa
+    solver.options["dualT"] = inverse_target_kappa
+
+    solver.solve(inf_prob, tee=False)
```
bknueven, line 3331:
```suggestion
    results = solver.solve(inf_prob, tee=False)
    if check_optimal_termination(results):
        # TODO: maybe we should tighten tolerances first?
        raise RuntimeError("Ill conditioning diagnostic problem infeasible") 
```
The problem by construction should not be infeasible, but I am sure it the solver could report it as such for a sufficiently ill-conditioned system.

idaes/core/util/model_diagnostics.py
```diff
@@ -3051,6 +3128,236 @@ def ipopt_solve_halt_on_error(model, options=None):
...
+
+    result_norm = inf_prob.res_norm.value
```
bknueven, line 3333: I have seen cases where this number is slightly negative. In this case we should probably bail out, or tighten the tolerances and try again (and if that fails then bail out).
```suggestion
    result_norm = inf_prob.res_norm.value
    if result_norm < 0.0:
        # TODO: try again with tighter tolerances?
        raise RuntimeError("Ill conditioning diagnostic problem has numerically troublesome solution")
```

idaes/core/util/model_diagnostics.py
```diff
@@ -3051,6 +3128,236 @@ def ipopt_solve_halt_on_error(model, options=None):
...
+    if result_norm >= inverse_target_kappa:
+        return []
+
+    # find an equivalent solution which minimizes y_norm
+    inf_prob.min_res.deactivate()
+    inf_prob.res_norm.fix()
+
+    inf_prob.min_y = Objective(expr=inf_prob.y_norm)
+
+    solver.solve(inf_prob, tee=False)
```
bknueven, line 3343:
```suggestion
    # if the this problem is numerically infeasible, we can still report something to the user
    results = solver.solve(inf_prob, tee=True, load_solutions=False)
    if check_optimal_termination(results):
        inf_prob.solutions.load_from(results)
```

idaes/core/util/model_diagnostics.py
```diff
@@ -3051,6 +3128,247 @@ def ipopt_solve_halt_on_error(model, options=None):
...
+            "Must be 'row' or 'column'."
+        )
+
+    jac, nlp = get_jacobian(model, scaled=False)
+
+    inverse_target_kappa = 1e-16 / target_feasibility_tol
+
+    # Get list of row/column names
+    con_list = [i.name for i in nlp.get_pyomo_constraints()]
+    var_list = [i.name for i in nlp.get_pyomo_variables()]
```
bknueven, line 3263:
```suggestion
    # Get a mapping of row/column names to index in jac
    con_dict = {i.name: idx for idx, i in enumerate(nlp.get_pyomo_constraints())}
    var_dict = {i.name: idx for idx, i in enumerate(nlp.get_pyomo_variables())}
```

idaes/core/util/model_diagnostics.py
```diff
@@ -3051,6 +3128,247 @@ def ipopt_solve_halt_on_error(model, options=None):
...
+    inf_prob.res_neg = Var(set2, initialize=1e-2, bounds=(0, None))
+    inf_prob.res = Expression(set2, rule=lambda m, i: m.res_pos[i] - m.res_neg[i])
+
+    def b_rule(b, i):
+        lhs = 0.0
+
+        for j in set1:
+            if direction == "row":
+                iidx = var_list.index(i)
+                jidx = con_list.index(j)
```
bknueven, line 3291:
```suggestion
                iidx = var_dict[i]
                jidx = con_dict[j]
```

idaes/core/util/model_diagnostics.py
```diff
@@ -3051,6 +3128,247 @@ def ipopt_solve_halt_on_error(model, options=None):
...
+                jac_entry = jac[jidx, iidx]
+            else:
+                jidx = var_list.index(j)
+                iidx = con_list.index(i)
```
bknueven, line 3295:
```suggestion
                jidx = var_dict[j]
                iidx = con_dict[i]
```

idaes/core/util/model_diagnostics.py
```diff
@@ -3051,6 +3128,247 @@ def ipopt_solve_halt_on_error(model, options=None):
...
+    inverse_target_kappa = 1e-16 / target_feasibility_tol
+
+    # Get list of row/column names
+    con_list = [i.name for i in nlp.get_pyomo_constraints()]
+    var_list = [i.name for i in nlp.get_pyomo_variables()]
+
+    # Build test problem
+    inf_prob = ConcreteModel()
+    inf_prob.con_set = Set(initialize=con_list)
+    inf_prob.var_set = Set(initialize=var_list)
```
bknueven, line 3268:
```suggestion
    inf_prob.con_set = Set(initialize=con_dict.keys())
    inf_prob.var_set = Set(initialize=var_dict.keys())
```

bknueven (commented):
The below set of changes avoids cubic time complexity in the `check_ill_conditioning` model construction.

idaes/core/util/tests/test_model_diagnostics.py
```diff
@@ -2883,3 +3037,224 @@ def test_display(self):
...
+
+        m.v1 = Var(initialize=1e-8)
+        m.v2 = Var()
+        m.v3 = Var()
+        m.v4 = Var()
+
+        m.c1 = Constraint(expr=m.v1 == m.v2 - 0.99999 * m.v4)
+        m.c2 = Constraint(expr=m.v1 + 1.00001 * m.v4 == 1e-8 * m.v3)
+        m.c3 = Constraint(expr=1e8 * (m.v1 + m.v4) + 1e10 * m.v2 == 1e-6 * m.v3)
+        m.c4 = Constraint(expr=-m.v1 == -0.99999 * (m.v2 - m.v4))
```
Robbybp, line 3106: Columns of the Jacobian corresponding to each variable:
```
v1: (1, 1, 1e8, -1)
v2: (-1, 0, 1e10, 1-1e5)
v3: (0, -1e8, -1e6, 0)
v4: (1-1e5, 1+1e5, 1e8, -(1-1e5))
```
`v1` and `v4` are parallel in an absolute sense, but we may also want to consider `(v1, v2)` and `(v2, v4)` to be parallel pairs in a relative sense, since 1 is very small relative to the large coefficients of `1e8` and `1e10`.

See my PR into this branch, where I've implemented this relative tolerance check.

Robbybp (commented):
A few small comments, then I've opened a PR into this branch where I do the following:
- Clean up variable names in some of the code I wrote for the parallel vector checker
- Implement a relative tolerance check for parallel vectors.

The PR is here: https://github.com/andrewlee94/idaes-pse/pull/19. Let me know what you think about the changes.

andrewlee94:
@Robbybp @bknueven Do you think this PR is ready to merge now, or should we hold off? I will be out of the country from the end of the week until basically the beginning of February.

idaes/core/util/model_diagnostics.py
```diff
@@ -3051,6 +3129,238 @@ def ipopt_solve_halt_on_error(model, options=None):
...
+                absprod = abs(prod[0, 0])
+                diff = abs(absprod - unorm * vnorm)
+                if diff <= tolerance or diff <= tolerance * max(unorm, vnorm):
+                    parallel.append((uidx, vidx))
+
+    parallel = [(components[uidx], components[vidx]) for uidx, vidx in parallel]
+    return parallel
+
+
+def check_ill_conditioning(
```
Robbybp, line 3219: I think this name is a too vague, as there are many different ways of checking for ill-conditioning. Maybe `compute_ill_conditioning_certificate` or `compute_worst_conditioned_subset`?

idaes/core/util/model_diagnostics.py
```diff
@@ -3051,6 +3129,238 @@ def ipopt_solve_halt_on_error(model, options=None):
...
+        results_set = RangeSet(0, nlp.n_primals() - 1)
+        jac = jac.transpose().tocsr()
+    elif direction == "column":
+        components = nlp.get_pyomo_variables()
+        components_set = RangeSet(0, len(components) - 1)
+        results_set = RangeSet(0, nlp.n_constraints() - 1)
+        jac = jac.tocsr()
+
+    # Build test problem
+    inf_prob = ConcreteModel()
```
Robbybp, line 3274: Why is this called `inf_prob`?

idaes/core/util/model_diagnostics.py
```diff
@@ -3051,6 +3129,238 @@ def ipopt_solve_halt_on_error(model, options=None):
...
+    # Objective -- minimize residual
+    inf_prob.min_res = Objective(expr=inf_prob.res_norm)
+
+    solver = SolverFactory("cbc")  # TODO: Consider making this an option
+
+    # tighten tolerances  # TODO: If solver is an option, need to allow user options
+    solver.options["primalT"] = target_feasibility_tol * 1e-1
+    solver.options["dualT"] = target_feasibility_tol * 1e-1
+
+    results = solver.solve(inf_prob, tee=False)
```
Robbybp, line 3323: This does not have to happen now, but I generally like splitting these types of functions into two functions: One that simply builds the model, and one that just builds the model (by calling the first function), then solves the model. That way an advanced user can build the model and inspect it, apply a transformation, or use a custom solver. It also makes it easier to see what the default solver and options are.

I'm open to arguments that this is a bad idea, and again, this does not need to happen now.

Robbybp (commented):
A few minor comments. These do not necessarily need to happen now, as this is beta functionality, but I would like them to be considered before this is made more permanent. In addition, I think a more permanent implementation should be more configurable to incorporate a DegeneracyHunter-like option.

idaes/core/util/model_diagnostics.py
```diff
@@ -3051,6 +3129,238 @@ def ipopt_solve_halt_on_error(model, options=None):
...
+    inf_prob.res_norm = Var()
+    inf_prob.res_norm_constr = Constraint(
+        expr=inf_prob.res_norm
+        == sum(inf_prob.res_pos.values()) + sum(inf_prob.res_neg.values())
+    )
+
+    # Objective -- minimize residual
+    inf_prob.min_res = Objective(expr=inf_prob.res_norm)
+
+    solver = SolverFactory("cbc")  # TODO: Consider making this an option
```
Robbybp, line 3317: It would be nice to default to HiGHS or SCIP, although I understand if that can't happen for the time being.

idaes/core/util/model_diagnostics.py
bknueven: `compute_ill_conditioning_certificate` is better.

`compute_worst_conditioned_subset` would be a bit inaccurate since it might involve solving a MIP version of this problem.

idaes/core/util/model_diagnostics.py
bknueven: We actually solve two problems at the moment -- we minimize the norm of the RHS, then we fix the norm of the RHS and minimize the 1-norm of the vector. That makes a cleaner workflow more difficult.

I also imagine eventually trying to re-solve with tighter tolerances if the answer is nonsensical, e.g., the norm of the RHS is negative. These things can be done automatically much more straightforwardly if we pick a single solver.

idaes/core/util/model_diagnostics.py
bknueven: The best solver to pick would probably be SoPlex as we can set arbitrary precision and utilize its iterative refinement capability. But I went with `cbc` because it ships with `idaes get-extensions`.

idaes/core/util/model_diagnostics.py
bknueven: probably should just go with `m`?

Robbybp: I like `m`. I'm fine with another name too, I just don't know what "inf" is.

idaes/core/util/model_diagnostics.py
Robbybp: Since we're solving the minimum-residual problem globally, we have the "worst conditioned" combination of rows/columns, right? Then you're saying that there is no reason to believe that the result will be a subset?

idaes/core/util/model_diagnostics.py
Robbybp: I did not realize that we solve two problems. I see the difficulty in splitting this up in that case.

idaes/core/util/model_diagnostics.py
Robbybp: "Infeasibility" of the linear dependence constraint?

idaes/core/util/model_diagnostics.py
bknueven: I don't think so, but it will tend to be a subset because we minimize the 1-norm of the multiplier alongside requiring an optimal basic feasible solution to the LP, i.e., it's just LASSO.

Put another way, we could try to minimize the number of non-zeros in the multiplier, which would give a minimal subset demonstrating the ill conditioning.

idaes/core/util/model_diagnostics.py
andrewlee94: FYI, for testing at least we do have the AMPL version of SCIP available now. For DegeneracyHunter, I made the solver a user argument with SCIP as the default.

idaes/core/util/model_diagnostics.py
Robbybp, line 3219: I think `check_ill_conditioning` is too vague of a name here. I suggest `compute_ill_conditioning_certificate`

Robbybp (changes requested):
Other than the name of `check_ill_conditioning`, I think this is ready. 

I'd still like to see this merged with DegeneracyHunter eventually, because I think having two different optimization-based degeneracy/ill-conditioning detectors will be confusing for users. This does not have to hold up this PR.

idaes/core/util/model_diagnostics.py
bknueven: I like this name better.

If we were honest it might be `try_to_compute_ill_conditioning_certificate`, as this LP itself will also be ill-conditioned. SoPlex has the capability to solve LPs to arbitrary precision, but it's not available currently as a (direct) Pyomo solver. And you cannot access the same options through SCIP, unfortunately.

andrewlee94: I've updated the name to `compute_ill_conditioning_certificate`

Robbybp (approved):
Looks good

andrewlee94:
@bknueven Would you have time to review this? I think it is just waiting on a second reviewer.

bknueven (approved):

PR #1307, opened 2023-12-18
Add seed fixing option to PySMO's sampling methods to enhance reproducibility

OOAmusat:
## Summary/Motivation:
This PR adds the ability for users to specify a seed with ``rand_seed`` when using the ``LatinHypercubeSampling``, ``CVTSampling`` and ``CustomSampling`` classes in PySMO. Fixing ``rand_seed`` ensures that the sampling methods return the same results every time the methods are run, allowing for reproducibility.

- Add ``rand_seed`` argument to CVT, LHS and custom sampling methods
- Add tests

### Legal Acknowledgement


idaes/core/surrogate/pysmo/sampling.py
```diff
@@ -594,6 +596,12 @@ def __init__(
             self.number_of_samples = number_of_samples
             self.x_data = bounds_array  # Only x data will be present in this case
 
+        if rand_seed is not None:
+            if not isinstance(rand_seed, int):
```
andrewlee94, line 600: Would it be better to do `int(rand_seed)` instead to catch cases where `rand_seed` is an integer in `float` form? You could either do this as a `try/except` or even just rely on the native exception from `int()`.

idaes/core/surrogate/pysmo/sampling.py
```diff
@@ -1412,6 +1422,12 @@ def __init__(
             raise Exception("Invalid tolerance input")
         self.eps = tolerance
 
+        if rand_seed is not None:
```
andrewlee94, line 1425: Seeing as this is in `__init__`, should this code be moved to the base class to avoid duplicating the code in each subclass?

idaes/core/surrogate/pysmo/sampling.py
OOAmusat: Done.

idaes/core/surrogate/pysmo/sampling.py
OOAmusat: No, I'd rather leave it here because it is an argument specific to a subset of the sampling methods.

idaes/core/surrogate/pysmo/tests/test_sampling.py
```diff
@@ -492,6 +492,42 @@ def test__init__selection_right_behaviour_with_specified_no_samples(
...
+            number_of_samples=6,
+            sampling_type="selection",
+            rand_seed=rand_seed,
+        )
+        np.testing.assert_array_equal(LHSClass.data, input_array)
+        np.testing.assert_array_equal(LHSClass.number_of_samples, 6)
+        np.testing.assert_array_equal(LHSClass.x_data, np.array(input_array)[:, :-1])
+        assert LHSClass.seed_value == rand_seed
+
+    @pytest.mark.unit
```
bpaul4, line 513: Should this test pass with a decimal, non-integer random seed value? See my comment on line 600 of the sampling module.

idaes/core/surrogate/pysmo/sampling.py
```diff
@@ -594,6 +596,13 @@ def __init__(
             self.number_of_samples = number_of_samples
             self.x_data = bounds_array  # Only x data will be present in this case
 
+        if rand_seed is not None:
+            try:
```
bpaul4, line 600: Running a decimal seed such as `int(15.1)` returns the floored value `15`, while running a non-float seed such as `int("15.1")` raises a `TypeError` - so, neither case will be caught as written here. Should line 602 be `np.random.seed(rand_seed)` and lines 603 nad 604 except and raise a `TypeError`?

idaes/core/surrogate/pysmo/sampling.py
```diff
@@ -1412,6 +1423,13 @@ def __init__(
             raise Exception("Invalid tolerance input")
         self.eps = tolerance
 
+        if rand_seed is not None:
+            try:
```
bpaul4, line 1427: See comment on line 600.

idaes/core/surrogate/pysmo/sampling.py
```diff
@@ -1732,13 +1752,21 @@ def __init__(
             )
         self.normal_bounds_enforced = strictly_enforce_gaussian_bounds
 
+        if rand_seed is not None:
+            try:
+                self.seed_value = int(rand_seed)
```
bpaul4, line 1757: See comment on line 600.

idaes/core/surrogate/pysmo/sampling.py
OOAmusat: @bpaul4 One of Andrew's comments (see above) was that the test should pass for floats (1in case someone enters something like 15.0), so it is currently implemented to pass to decimal, non-integer seed values. 

On the error: ``int("15.1")`` raises a ``ValueError`` for me, not a ``TypeError``:
image

bpaul4: Thanks for clarifying, in that case I think it's OK.

bpaul4 (approved):

idaes/core/surrogate/pysmo/sampling.py
OOAmusat: Is it raising a TypeError on your end? It would be good to know if it is, before I break the whole of IDAES :)

bpaul4: No, you're correct. `np.random.seed("15.1")` throws a `TypeError`.

dangunter: According to the docstring random.seed is a legacy function. Good opportunity to use the best practice interface instead I think.

OOAmusat: The new best practice interface ``np.random.default_rng()`` requires some code refactoring for it to work in this case. Because the new interface requires the creation of a local random number generator (unlike ``np.random.seed`` which is global), it breaks several of the static methods (e.g.,  in the sampling classes. 

The new method is already used in the case of the ``CustomSampling`` class. Further changes may be more suitable in the future if/when the sampling class is refactored.

dangunter (commented):
LGTM

dangunter (approved):
LGTM

PR #1315, opened 2024-01-11
Enable unnecessary-lambda Pylint check

lbianchi-lbl:
## Summary/Motivation:

- `unnecessary-lambda` is a built-in Pylint check, that _might_ help addressing some of the issues with `lambda`s described in #1240 
- However, this only looks as cases where `my_func = lambda x, y: my_other_func(x, y)`, i.e where one could just simply set `my_func = my_other_func`
- More to the point, it does nothing to detect and prevent other uses of `lambda` that cause most of the issues in #1240 (specifically, by definition, the closure scope gotchas only arise when the `lambda` has fewer arguments that the function being used)

- Enable `unnecessary-lambda` Pylint check

### Legal Acknowledgement

### DIFF
--- a/.pylint/pylintrc
+++ b/.pylint/pylintrc
@@ -13,7 +13,6 @@ disable=no-self-argument,
     fixme,
     raise-missing-from,
     logging-fstring-interpolation,  # not sure why pylint doesn't like f-strings
-    unnecessary-lambda,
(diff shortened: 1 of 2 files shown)


ksbeattie (approved):
LGTM

PR #1324, opened 2024-02-01
Python 3.12 fix for idaes command

eslickj:
## Fixes

Python 3.12 compatibility issue with the idaes command.

- Fix minor issue due to pkgutil API change
- Catch exception if a command module fails to import a dependency

### Legal Acknowledgement

### DIFF
--- a/idaes/commands/__init__.py
+++ b/idaes/commands/__init__.py
@@ -20,10 +20,11 @@ import time
 _command_import_start_time = time.time()
 
 import pkgutil
+import click
 from idaes.commands.base import command_base as cb
 
 # import all the commands
-for loader, dotted_module_name, is_pkg in pkgutil.walk_packages(__path__):
+for finder, dotted_module_name, is_pkg in pkgutil.walk_packages(__path__):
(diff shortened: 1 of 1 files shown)


andrewlee94 (approved):
Looks simple enough.

lbianchi-lbl (changes requested):
- I think the fix itself looks fine (even though most of the dynamic import functionality has been in the stdlib's `importlib` for a while)
- The fact that this code is run at import time (as opposed to e.g. being wrapped in a function and called on-demand) is what's giving me a bit more pause
  - This applies to the existing code in general, but more specifically to using `click.echo()` for module-level code (maybe it's OK, but I don't know enough of click to understand the possible implications)
- Independently from the above, since this is a Python 3.12 issue, I suggest we wait until 3.12 is added to the CI job matrix to make sure it works (this is being tracked separately in #1330). I'll try to open a PR shortly

lbianchi-lbl:
I've applied these changes as part of #1346 to be able to run the tests at all, so I think this can be closed.

PR #1334, opened 2024-02-09
Extract Codecov upload steps in CI workflow to dedicated dependent job

lbianchi-lbl:
## Summary/Motivation:

- Codecov upload failures are frustrating, as they currently require re-running the entire test suite that generates the report (20-40+ minutes)
- The proper fix is to use a Codecov upload token, saved as a repository secret
  - However, this require a major restructuring of the CI workflow because `pull_request` triggers cannot access `secrets`, requiring the creation of a new, non-`pull_request` workflow
  - Synchronizing triggers  and passing data between different workflows is considerably more complicated compared to doing the same between different jobs in the same workflow
- As an intermediate fix, we can separate the Codecov upload so that it runs in a separate job within the `pull_request`-triggered workflow, so that when it fails, it can be re-run separately (which should only take a few seconds)

- In `pytest` job: replace step uploading report to Codecov with `actions/upload-artifact`
- Create new `upload-coverage` job dependent on `pytest` that just

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -135,24 +135,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ["linux", "win64"]
-        include:
-          - os: linux
-            codecov-os: linux
-          - os: win64
-            codecov-os: windows
+        report-variant: ["linux", "win64"]
     steps:
       # the checkout step is needed to have access to codecov.yml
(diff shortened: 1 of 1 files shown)


lbianchi-lbl:
@ksbeattie @andrewlee94 FYI - I'm merging this in now to restore Codecov upload functionality for `main` and other PRs.

PR #1336, opened 2024-02-13
Support calcuation of critical properties of mixtures in Modular Properties

andrewlee94:
## Summary/Motivation:
This PR adds support for calculating the critical properties of mixtures in the Modular Properties framework, which is necessary to support the new Smooth VLE formulation. This is done by adding  a new method to EoS classes which contains instructions for how to calculate the mixture critical properties appropriate to the EoS.

When dealing with systems with multiple phases/EoS, only one EoS can be used thus it is necessary to determine which EoS to use. This PR applies the following steps:

1. Iterate over all phases and take the EoS associated with the first liquid phase encountered.
2. If no liquid phases are found, use the EoS associated with the vapor phase (assumes only 1 vapor phase).
3. If no liquid or vapor phases are found, raise an exception.

- Add new unit definitions for specific volumes, and deprecate non-standard `MOLAR_VOLUME` unit.
- Add `compress_fact_crit` and `dens_mol_crit` to the standard property set.
- Add methods and metadata to core modular properties code to support critical properties as build-on-demand.
- Add placeholder `build_critical_properties` method to `EoSBase`.
- Add `build_critical_properties` method to `IdealEoS`. Cubic equivalent WIP.

### Legal Acknowledgement

### DIFF
--- a/docs/explanations/components/property_package/general/phase_def.rst
+++ b/docs/explanations/components/property_package/general/phase_def.rst
@@ -34,7 +34,7 @@ Equations of state (or equivalent methods) describe the relationship between dif
 
 A wide range of equations of states are available in literature for different applications and levels of rigor, and the IDAES Generic Property Package Framework provides a number of prebuilt modules for users, which are listed below.
 
-Equation of state packages may allow for user options (e.g. choosing a specific type of cubic equation of state). The options are set using the `equation_of_state_options` argument, and the options available are described in the documentation of each equation of state module.
+Equation of state packages may allow for user options (e.g., choosing a specific type of cubic equation of state). The options are set using the `equation_of_state_options` argument, and the options available are described in the documentation of each equation of state module.
 
 Equation of State Libraries
 """""""""""""""""""""""""""
@@ -44,6 +44,20 @@ Equation of State Libraries
 
     eos/ideal
     eos/cubic
+    
+Critical Properties of Mixtures
+"""""""""""""""""""""""""""""""
+
+Calculation of the critical properties of mixtures depends on the equation of state being used. As the Modular Property Package framework allows users to specify different equations of state for each phase, the following logic is used to determine which equation of state to use for calculating critical properties.
+
+1. If a vapor-liquid equilibrium pair is defined in the `"phases_in_equilibrium"` configuration argument, then the liquid phase from this pair is used (IDAES generally assumed supercritical fluids are liquid like).
+2. If no vapor-liquid equilibrium pair is defined, then the first liquid phase define is used.
+3. If no liquid phases are defined then the vapor phase is used (it is assumed there will only be one vapor phase).
+4. If no vapor phase is defined, then a `PropertyPackageError` is returned as there is no suitable phase for calculating critical properties.
+
+Note that not all Equations of State are suitable for calculating critical properties as many cannot represent the critical conditions. The Ideal equation of State is one common example of an equation of state that DOES NOT support critical properties.
+
+During initialization, the critical properties of the mixture are approximated using the mole fraction weighted sum of the component critical properties (note that users must provide values for all component critical properties (i.e., `compress_fact_crit`, `dens_mol_crit`, `pressure_crit` and `temperature_crit`) for initialization if critical properties are to be calculated.
 
 Phase-Specific Parameter
 ^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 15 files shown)


dallan-keylogic:
Most of this PR is good, but I think that adding critical properties to the ideal EoS is fundamentally misguided.  There's no point at which the vapor and liquid phases become identical for the ideal EoS. I looked up "Kay's Rule" for calculating T_crit and P_crit for mixtures, and it appears to be used for estimating mixture critical properties to use for calculating Z in a corresponding states chart. For the ideal EoS Z=1 if we have a vapor and Z=0 if we have a liquid, there is no middle ground.

andrewlee94:
@dallan-keylogic This should be ready for review. I've removed the ideal critical properties and added the cubics.

idaes/models/properties/modular_properties/eos/ceos.py
```diff
@@ -901,6 +902,125 @@ def vol_mol_phase_comp(b, p, j):
...
+
+        m.add_component("A_crit", Constraint(rule=rule_A_crit))
+
+        def rule_B_crit(m):
+            bm_crit = getattr(m, "bm_crit")
+            return EoS_param[ctype]["coeff_b"] * Cubic.gas_constant(
+                m
+            ) * m.temperature_crit == (bm_crit * m.pressure_crit)
+
+        m.add_component("B_crit", Constraint(rule=rule_B_crit))
```
dallan-keylogic, line 988: Can't these be `Expressions`? Or is the idea to avoid divisions.

idaes/models/properties/modular_properties/eos/tests/test_ceos_PR.py
```diff
@@ -1065,3 +1074,310 @@ def test_vol_mol_phase_comp(m):
...
+                "c": {
+                    "parameter_data": {
+                        "omega": 0.3,
+                        "compress_fact_crit": 1,
+                        "dens_mol_crit": 1,
+                        "pressure_crit": 3e5,
+                        "temperature_crit": 300,
+                    }
+                },
+            },
```
dallan-keylogic, line 1114: Here we might start to run into problems with the critical properties not being consistent. Knowing how the cubic EoS works, I imagine what happens is that T_crit and P_crit influence the mixture critical properties while rho_mol_crit and Z_crit do not.

dallan-keylogic (changes requested):
The implementation looks largely fine, though I'm wondering why we're maintaining two separate initialization routines for the modular properties.  It probably has to do with the old method of initialization vs the new method of initialization, but having one method call the other or extracting the common code into a function would be much cleaner. That impacts this PR only incidentally, though (where the same typos/stylistic choices were made in both parallel initialization methods). 

Potentially more critically, we have four critical properties the user can provide, but they're related through an equation. Thus we have only three degrees of freedom. The user could, potentially, provide inconsistent data. The answer is probably "the user is responsible for understanding which properties are used by their EoS and for providing consistent data", but I thought I should raise the issue.

docs/explanations/components/property_package/general/phase_def.rst
```diff
@@ -44,6 +44,20 @@ Equation of State Libraries
...
     eos/ideal
     eos/cubic
+    
+Critical Properties of Mixtures
+"""""""""""""""""""""""""""""""
+
+Calculation of the critical properties of mixtures depends on the equation of state being used. As the Modular Property Package framework allows users to specify different equations of state for each phase, the following logic is used to determine which equation of state to use for calculating critical properties.
+
+1. If a vapor-liquid equilibrium pair is defined in the `"phases_in_equilibrium"` configuration argument, then the liquid phase from this pair is used (IDAES generally assumed supercritical fluids are liquid like).
+2. If no vapor-liquid equilibrium pair is defined, then the first liquid phase define is used.
```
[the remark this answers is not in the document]
andrewlee94: Dicts are now guaranteed to be insertion ordered, so they will be deterministic.

docs/explanations/components/property_package/general/phase_def.rst
```diff
@@ -44,6 +44,20 @@ Equation of State Libraries
...
+3. If no liquid phases are defined then the vapor phase is used (it is assumed there will only be one vapor phase).
+4. If no vapor phase is defined, then a `PropertyPackageError` is returned as there is no suitable phase for calculating critical properties.
+
+Note that not all Equations of State are suitable for calculating critical properties as many cannot represent the critical conditions. The Ideal equation of State is one common example of an equation of state that DOES NOT support critical properties.
+
+During initialization, the critical properties of the mixture are approximated using the mole fraction weighted sum of the component critical properties (note that users must provide values for all component critical properties (i.e., `compress_fact_crit`, `dens_mol_crit`, `pressure_crit` and `temperature_crit`) for initialization if critical properties are to be calculated.
```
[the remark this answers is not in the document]
andrewlee94: User beware. The alternative is to allow the user to provide 3 or the 4 option, but then we need to have 4 different methods for calculating the missing one (and the logic to determine which one to use).

idaes/models/properties/modular_properties/eos/ceos.py
andrewlee94: No, these are the actual constraints we need to solve to find T_c and P_c.

andrewlee94:
Regarding having the two initialization routines, it is backward compatibility for the older approach. As they are in separate classes and the new approach also has a few things the old one does not, I did not want one to call the other. It would be possible to have them use a single, common method for the main initialization routine, but at the time I wanted the new version to package everything it would need neatly.

dallan-keylogic (approved):
LGTM

docs/explanations/components/property_package/general/phase_def.rst
```diff
@@ -44,6 +44,20 @@ Equation of State Libraries
 
     eos/ideal
     eos/cubic
+    
+Critical Properties of Mixtures
+"""""""""""""""""""""""""""""""
+
+Calculation of the critical properties of mixtures depends on the equation of state being used. As the Modular Property Package framework allows users to specify different equations of state for each phase, the following logic is used to determine which equation of state to use for calculating critical properties.
+
+1. If a vapor-liquid equilibrium pair is defined in the `"phases_in_equilibrium"` configuration argument, then the liquid phase from this pair is used (IDAES generally assumed supercritical fluids are liquid-like).
```
agarciadiego, line 53: assumed like in the past? I would change it assumes

agarciadiego (approved):
LGTM

Issue #1327, opened 2024-02-07
PETSc Solver Derivative Estimation Bug?


eslickj:
I think the block to estimate time derivatives was mistakenly indented extra.  https://github.com/IDAES/idaes-pse/blob/190451a457fcc69aa73c7253f09bec17cd870112/idaes/core/solvers/petsc.py#L714C1-L720C64

The calculate derivatives block is in the temporary model context, so I think the discretization constraints are always deactivated.

PR #1342, opened 2024-02-19
Fix calculate_derivatives for PETSc

dallan-keylogic:
## Fixes
Addresses #1327

### Legal Acknowledgement


idaes/core/solvers/tests/test_petsc.py
```diff
@@ -988,3 +988,161 @@ def diff_eq_rule(m, t):
...
+        },
+        skip_initial=True,  # With u and x fixed, no variables to solve for at t0
+        calculate_derivatives=True,
+    )
+    # It should backfill values for interpolated points like t=1
+    assert pyo.value(m.dxdt[0]) == pytest.approx(-0.7580125427537873, rel=1e-3)
+    assert pyo.value(m.dxdt[1]) == pytest.approx(-0.2034733131369807, rel=1e-3)
+    # No value assigned at 2 because there's no corresponding discretization equation
+    # TODO I don't understand how a value is getting assigned here
+    # assert m.dxdt[2].value is None
```
dallan-keylogic, line 1148: @eslickj , could you help me understand why we're getting derivative values populated here? It's definitely not through the `calculate_derivatives` function, but something being returned by the PETSc solver interface.

ksbeattie:
@eslickj, any chance you could review this?

idaes/core/solvers/tests/test_petsc.py
blnicho: I think the problem is that with the forward finite difference discretization you should technically be deactivating `m.diffeq[2]` however this runs afoul of the `representative_time` infrastructure in the petsc solver. Seems like there is an assumption baked into the petsc solver logic that the model will be discretized using an implicit approach. Which makes sense since our petsc docs explicitly mention that only "implicit TS solvers are supported".

blnicho (changes requested):

idaes/core/solvers/tests/test_petsc.py
dallan-keylogic: I've tried deactivating `m.diffeq[2]`, and it still gets populated. It isn't populated by the `calculate_derivatives` function, but directly from the solver.

idaes/core/solvers/petsc.py
```diff
@@ -428,7 +428,7 @@ def petsc_dae_by_time_element(
     symbolic_solver_labels=True,
     between=None,
     interpolate=True,
-    calculate_derivatives=True,
+    calculate_derivatives=False,
```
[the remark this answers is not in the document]
dallan-keylogic: Because 1) Due to the indent issue, derivatives weren't being calculated by default, so this is consistent with the behavior before this fix and 2) I wanted the user to understand how `calculate_derivatives` worked (by solving the Pyomo discretization equations for derivative values) before having derivative values populated (lest they think the derivative values were, for example, taken straight from PETSc).

idaes/core/solvers/petsc.py
```diff
@@ -731,19 +731,71 @@ def calculate_time_derivatives(m, time):
...
+                    disc_eq = disc_dict[key]
+                    for t in time:
+                        if t < between.first() or t > between.last():
+                            # Outside of integration range, skip calculation
+                            continue
+                        # Check time index to decide what constraints to scale
+                        if t == time.first() or t == time.last():
+                            try:
+                                if disc_eq[t].active and not deriv[t].fixed:
+                                    old_value = deriv[t].value
```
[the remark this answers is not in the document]
dallan-keylogic: No, not needed for this conditional.

idaes/core/solvers/petsc.py
```diff
@@ -731,19 +731,71 @@ def calculate_time_derivatives(m, time):
...
+                                    )
+                            except ValueError:
+                                # Reset deriv value to old value
+                                if disc_eq[t].active and not deriv[t].fixed:
+                                    deriv[t].value = old_value
+
+                        else:
+                            if disc_eq[t].active and not deriv[t].fixed:
+                                deriv[t].value = 0  # Make sure there is a value
+                                calculate_variable_from_constraint(deriv[t], disc_eq[t])
```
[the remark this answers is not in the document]
dallan-keylogic: One thing the old code does that this code does not do is to catch exceptions only on the conditions `between.first() or t == between.last()` or `t == time.first() or t == time.last()`. If they get hit with a `KeyError` or `ValueError` some other time (for whatever reason), I'd like that exception to be raised. However, the way to handle it is probably to check those conditionals inside the `Except` blocks and reraise the errors if they're not met.

idaes/core/solvers/petsc.py
```diff
@@ -731,19 +731,58 @@ def calculate_time_derivatives(m, time):
...
+                            else:
+                                raise err
+                        except ValueError as err:
+                            # At edges of between, it's unclear which adjacent
+                            # values of state variables have been populated.
+                            # Therefore we might get hit with value errors.
+                            if t == between.first() or t == between.last():
+                                # Reset deriv value to old value
+                                if disc_eq[t].active and not deriv[t].fixed:
+                                    deriv[t].value = old_value
```
dallan-keylogic, line 783: @lbianchi-lbl Do you know what Pylint is complaining about here? We can only get to this logic if old_value has already been assigned in the `try` block.

lbianchi-lbl: From a quick look I'd guess it might be because of `old_value` is defined inside the `if` block on L764?

dallan-keylogic: Moved it outside the `if` block, still getting the pylint error. I'm guessing it's because Pylint doesn't know that accessing `deriv[t].value` is only going to raise a `KeyError` in that context, and will never raise a `ValueError`. I assume this is a case for some Pylint ignore statement.

The test failures appear to be unrelated to this PR.

dallan-keylogic:
@andrewlee94  @lbianchi-lbl Could you rubber stamp this so integration tests can run?

andrewlee94:
@dallan-keylogic You can use the `CI:run_integration` tag to manually trigger the integration tests.

eslickj (approved):
Looks good as far as I can tell.  Not sure if my approval matters anymore, but I approved anyway.

ksbeattie:
> Looks good as far as I can tell. Not sure if my approval matters anymore, but I approved anyway.

@eslickj your approval always matters to us!

PR #1346, opened 2024-02-21
Add support for Python 3.12

lbianchi-lbl:
## Summary/Motivation:

Python 3.12 is now the most recent stable version and we should ensure it's supported and tested.

- Add Python 3.12 to CI job matrices and docs
- Update default Python version in CI to 3.11

### Legal Acknowledgement


jsiirola (approved):

blnicho (approved):

idaes/core/dmf/commands.py
```diff
@@ -391,7 +391,9 @@ def cat_resources(path, objects=(), color=True):
     # get all resources,
     # display any that match an object as a prefix
     for r in d.find():
-        for oid in unmatched:
+        # iterate on a copy of the set to avoid
+        # mutating an object being iterated on
+        for oid in set(unmatched):
```
jsiirola, line 396: I would make this a `list` and not a `set`: list is faster, doesn't require hashability, and most importantly, is deterministic.  I think I would have gone with:
```python
unmached = list(objects)
for r in d.find():
    for i, oid in enumerate(unmached):
        if not oid:
            continue
        if r.uuid.startswith(oid):
            unmached[i] = None
            # ...
```

jsiirola (approved):
This is fine, although there is a potential nondeterminism in the `dmf/commands.py` that may (or may not) be an issue.

lbianchi-lbl:
Definitely orthogonal to the main purpose of this PR, but after updating Pylint (which was needed for 3.12 compatibility), I figured out a way to format its log so that it's nicely rendered by GitHub:

image

idaes/core/dmf/commands.py
lbianchi-lbl: As a note for posterity (?): I've moved the Pylint-related changes (only incidentally due to Python 3.12), including the one this comment refers to, to their own PR (#1351, now merged as 57cf0a7f31f3b5aafabfe103c5af8d028bb69b37)

PR #1351, opened 2024-02-23
Update Pylint to 3.0.3

lbianchi-lbl:
## Summary/Motivation:

This was originally part of #1346, but I thought it would be better to move the Pylint-specific changes to a separate PR.

- Update Pylint to 3.0.3
- Apply the necessary changes to resolve all outstanding messages

### Legal Acknowledgement


idaes/core/surrogate/pysmo/polynomial_regression.py
```diff
@@ -356,6 +368,8 @@ def __init__(
             print("The default training/cross-validation split of 0.75 is used.")
             training_split = 0.75
         elif training_split >= 1 or training_split <= 0:
+            # PYLINT-TODO
+            # pylint: disable-next=broad-exception-raised
```
andrewlee94, line 372: Not for this PR, but this looks like we should create an issue about using more specific error types here.

andrewlee94 (approved):
This looks good to me.

agarciadiego (approved):
LGTM

idaes/core/surrogate/pysmo/polynomial_regression.py
lbianchi-lbl: I definitely agree, I almost thought I could do it as part of this PR since it's straightforward from the technical point of view, but then I thought we might want to think it through a bit more. I've just created #1352 to track this.

PR #1354, opened 2024-02-24
Pin ampl-module-scip version to 20240121

### LINKED ISSUE #1353: CI failures in test_model_diagnostics.py::TestDegeneracyHunter::test_solve_candidates_milp
- Affects the most likely unrelated PRs #1342 and #1350
- Current guess is that it's due to a new version of `ampl-module-scip` being released (see https://portal.ampl.com/dl/pypi/ampl-module-scip/)
- Will try to open a PR to pin the `ampl-module-scip` version to the last available one known to be working
  - Judging from https://github.com/IDAES/idaes-pse/actions/runs/8021757576/job/21914589132#step:5:1362, this seems to be 20240121

lbianchi-lbl:
### Legal Acknowledgement

### DIFF
--- a/.github/actions/setup-idaes/action.yml
+++ b/.github/actions/setup-idaes/action.yml
@@ -8,6 +8,10 @@ inputs:
     description: 'Command to use to install `install-target`'
     required: false
     default: pip --no-cache-dir install --progress-bar off
+  ampl-scip-version:
+    description: Version of AMPL Scip solver to install
+    required: false
+    default: '20240121'
 runs:
   using: "composite"
   steps:
@@ -46,5 +50,5 @@ runs:
       shell: bash -l {0}
       run: |
         echo '::group::Output of "pip install ampl_module_scip" command'
-        ${{ inputs.install-command }} --index-url https://pypi.ampl.com ampl_module_scip
+        ${{ inputs.install-command }} --index-url https://pypi.ampl.com ampl_module_scip==${{ inputs.ampl-scip-version }}
         echo '::endgroup::'


andrewlee94:
@lbianchi-lbl We probably need to investigate why the new version of ampl-scip is causing issues, as this could be a sign of something we need to deal with.

dangunter:
@andrewlee94, since this seems to block everything else I suggest we merge the fix and open an issue to investigate after that.

dangunter (approved):
This version seems to fix the immediate problem.

.github/actions/setup-idaes/action.yml
```diff
@@ -8,6 +8,10 @@ inputs:
     description: 'Command to use to install `install-target`'
     required: false
     default: pip --no-cache-dir install --progress-bar off
+  ampl-scip-version:
+    description: Version of AMPL Scip solver to install
+    required: false
+    default: '20240121'
```
andrewlee94, line 14: We should not be fixing solver versions wherever possible. In this case, from the issue we are seeing it is likely the failure is because the value being checked can equally be positive or negative (same magnitude) and that the different version of SCIP pick differently. Hopefully, this can be fixed by making the assert check the absolute value of the Pyomo variable instead (both values are equally correct).

andrewlee94 (changes requested):

ksbeattie (approved):
For the sake of getting an RC done on time, I'd like to merge this.  @andrewlee94 could you try your suggested fix testing against both the 20240121 and 20240221 in a different PR?

PR #1201, opened 2023-06-08
[WIP] Price-taker framework for DISPATCHES

adam-a-a:
## Changes proposed in this PR:
-
-

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1358, opened 2024-02-29
Price taker model for DISPATCHES, Rehashed

djlaky:
## Fixes
Compared to #1201, operational constraints mathematical form was corrected. Unnecessary functions were removed/merged. Additional user flexibility was added for constructing cost objectives.

## Summary/Motivation:
Resurrecting #1201 to finish price taker framework in accordance with project milestones. 

Framework allows the user to construct price-taker models for design and/or operational optimization considering time-varying market price data.

### Legal Acknowledgement


lbianchi-lbl:
It looks like one of the tests it's taking too long to run on Windows:

image

I'm going to cancel the workflow to be able to proceed with the release.

lbianchi-lbl:
image

This test failure is caused by the object being a `pathlib.Path` instance rather than a `str`. To check for the file extension `.xls` in this case, you can use [`Path.suffix`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.suffix).

pytest.ini
lbianchi-lbl, line 1: Minor thing, but this change can (and probably should, to avoid conflicts) be reverted since it's being applied by #1362.

ksbeattie:
@dlakes94 any chance you could come to our dev call to talk about this?

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
```diff
@@ -0,0 +1,138 @@
...
+    NonNegativeReals,
+    Constraint,
+)
+from functools import reduce
+from idaes.core.base.process_base import declare_process_block_class
+from idaes.models.unit_models import SkeletonUnitModelData
+from pyomo.common.config import ConfigValue, In
+
+
+def deepgetattr(obj, attr):
```
radhakrishnatg, line 28: @lbianchi-lbl I borrowed this function from internet. Is it okay to have it in the repository? If not, is there an alternative for this function?

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
```diff
@@ -0,0 +1,138 @@
...
+            )
+
+        if self.config.append_lmp_data:
+            self.LMP = Param(
+                initialize=1,
+                mutable=True,
+                doc="Parameter: Will be updated to LMP value at given time",
+            )
+
+        self.config.model_func(self, **self.config.model_args)
```
radhakrishnatg, line 138: @dlakes94 Need to pass `design_blk` as an argument if it is specified. Alternatively, we can delete lines 86-92 and let the user pass it in `model_args`.

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
lbianchi-lbl: @radhakrishnatg the Python standard library has [operator.attrgetter()](https://docs.python.org/3/library/operator.html#operator.attrgetter) that you can use instead:

```
import operator

m.fs.unit = MyUnitModel()

assert m.fs.unit is operator.attrgetter("fs.unit")(m)
```

andrewlee94: @radhakrishnatg What exactly are you trying to test with this anyway? I suspect there are better ways to do whatever you are trying to do using native Pyomo functionality.

Also, on borrowing from the internet, it depends on where you got it and what license they place on the code. E.g., you cannot copy code from StackOverflow into IDAES as the licenses are incompatible.

radhakrishnatg: Thank you @lbianchi-lbl and @andrewlee94 ! 
@andrewlee94 I want a method similar to `getattr` which also works for attribute of an attribute e.g., `modified_getattr(m, "fs.compressor[1].input")`. Ludovico's suggestion works for me. Please let me know if there are better methods available in Pyomo.

idaes/apps/grid_integration/multiperiod/price_taker_model.py
```diff
@@ -0,0 +1,965 @@
...
+
+        """
+        if os.path.exists(file_path):
+            path_to_file = file_path
+        else:
+            raise ValueError(
+                f"The file path {file_path} does not exist. Please check your file path."
+            )
+
+        if ".xls" in path_to_file.suffix:
```
radhakrishnatg, line 296: The extension should be `.xlsx`

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
lbianchi-lbl: I'm aware of Pyomo's [find_component()](https://pyomo.readthedocs.io/en/stable/library_reference/aml/index.html#pyomo.environ.ConcreteModel.find_component) method, but the Pyomo folks will have the proper answer.

idaes/apps/grid_integration/multiperiod/price_taker_model.py
```diff
@@ -0,0 +1,965 @@
...
+
+        Returns:
+            lmp_data:   dict of representative day LMP data, indices are indexed
+                        by integers starting at 1 (ex: {1: {1: 4, 2: 3, 3: 5},
+                                                        2: {1: 1, 2: 7, 3: 3}})
+                                               Format: {day: {time_period: LMP}}
+            weights:    dict of weights for each representative day, indexed the
+                        same way as lmp_data      (ex: {1: 45, 2: 56})
+                                               Format: {day: weight}
+        """
```
lbianchi-lbl, line 221: @dlakes94 this seems to resolve the Sphinx warnings:

```suggestion
        """
        Clusters the given price signal in n_clusters. This method supports k-means,
        k-medoids, etc., techniques for clustering.

        Args:
            raw_data: Columnar data for a given LMP signal.
            n_clusters: Number of clusters desired for the data (representative days).

        Returns:
            lmp_data: A dictionary of representative day LMP data, indices are indexed
                    by integers starting at 1. Example: ::

                        {1: {1: 4, 2: 3, 3: 5},
                        2: {1: 1, 2: 7, 3: 3}}

            weights: A dictionary of weights for each representative day, indexed the
                    same way as lmp_data. Example: ::

                        {1: 45, 2: 56}
        """
```

idaes/apps/grid_integration/multiperiod/price_taker_model.py
dlakes94: I agree, but I think this allows for users to read both '.xls' and '.xlsx' files. I can change if we don't want '.xls' extensions to be compatible.

idaes/apps/grid_integration/multiperiod/price_taker_model.py
dlakes94: Thank you! I will incorporate this.

lbianchi-lbl (commented):
I have a general question about the test files: would it be possible to use a text format (e.g. JSON or CSV) instead of xlsx? Binary files don't play nice with version control, so the motivation for using them should be strong enough to justify the significant tradeoffs.

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
andrewlee94: @radhakrishnatg What is your actual use case for this? I would hesitate to write an entirely new function to do this, as there is likely things that will already do that for you. For one, as @lbianchi-lbl noted there is the `find_component()`.

djlaky: I've updated the files to use Pyomo's native `find_component()` function instead of the borrowed `deepgetattr()` function.

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
djlaky: Pushed the responsibility to the user so design_blk is now expected in model_args if the user's model depends on the design_blk.

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
```diff
@@ -0,0 +1,122 @@
...
+    Constraint,
+)
+from functools import reduce
+from idaes.core.base.process_base import declare_process_block_class
+from idaes.models.unit_models import SkeletonUnitModelData
+from pyomo.common.config import ConfigValue, In
+
+
+@declare_process_block_class("DesignModel")
+class DesignModelData(SkeletonUnitModelData):
```
andrewlee94, line 29: Out of interest, why are you inheriting `SkeletonUnitModelData` here and not `UnitModelData` (or another base class)?

setup.py
```diff
@@ -97,6 +97,8 @@ class ExtraDependencies:
     ]
     grid = [
         "gridx-prescient>=2.2.1",  # idaes.tests.prescient
+        "scikit-learn",
```
andrewlee94, line 100: Just to check, what are these dependencies required for and are they absolutely necessary?

idaes/apps/grid_integration/multiperiod/price_taker_model.py
```diff
@@ -0,0 +1,968 @@
...
+            k_values, inertia_values, curve="convex", direction="decreasing"
+        )
+        n_clusters = elbow_point.knee
+
+        if n_clusters is None:
+            raise ValueError(
+                f"Could not find elbow point for given kmin, kmax. Consider increasing the range of kmin, kmax."
+            )
+
+        print(f"Optimal # of clusters is: {n_clusters}")
```
andrewlee94, line 185: We generally discourage `print` statements inside code; using a logger is generally better as it gives control over where the output goes.

idaes/apps/grid_integration/multiperiod/price_taker_model.py
```diff
@@ -0,0 +1,968 @@
...
+
+        if int(n_clusters) + 2 >= kmax:
+            _logger.warning(
+                f"Optimal number of clusters is close to kmax: {kmax}. Consider increasing kmax."
+            )
+
+        if plot == True:
+            plt.plot(k_values, inertia_values)
```
andrewlee94, line 193: Should this be moved into a separate method so that it can be called by the user separately?

idaes/apps/grid_integration/multiperiod/tests/test_price_taker.py
```diff
@@ -0,0 +1,1376 @@
...
+@pytest.mark.unit
+def test_elbow_plot(excel_data):
+    m = PriceTakerModel()
+
+    daily_data = m.generate_daily_data(excel_data["BaseCaseTax"])
+    m.get_optimal_n_clusters(daily_data, plot=True)
+    test_fig = plt.gcf()
+    plt.close("all")
+
+    assert test_fig is not None
```
andrewlee94, line 100: Does this really test much of the actual code? It runs the plotting code, but all this assert does is make sure the resulting plot got closed which doesn't really test the plot. Note that testing plotting code is generally very hard - we generally only test the data used to generate the plot and leave the actual plotting untested.

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
djlaky: I believe this was the first thing that came to mind when the original development was happening. @radhakrishnatg can probably comment more on this.

idaes/apps/grid_integration/multiperiod/tests/test_price_taker.py
djlaky: You're correct this just tests if the plot was successfully called; it does not check if the "correct" plot or "correct" data was displayed.

I'm inclined to _remove_ the plotting test. Please see my other review comment for further information on the optimal clusters code snippet ([link](https://github.com/IDAES/idaes-pse/pull/1358/files#r1591054731)).

idaes/apps/grid_integration/multiperiod/price_taker_model.py
```diff
@@ -0,0 +1,968 @@
...
+    RangeSet,
+    Objective,
+    Constraint,
+    NonNegativeReals,
+    Expression,
+    maximize,
+    Param,
+)
+
+from idaes.apps.grid_integration import MultiPeriodModel
```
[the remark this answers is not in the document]
djlaky: We will note this for our continued development, thank you.

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
radhakrishnatg: @andrewlee94 From what I remember, `UnitModelData` class has several additional methods that I think are not needed. I just need a simple Pyomo block. If `UnitModelData` is a better alternative, then @djlaky, please make the change.

idaes/apps/grid_integration/multiperiod/price_taker_model.py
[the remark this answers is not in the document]
radhakrishnatg: @andrewlee94 Thank you for the note! I was not aware of time indexing. In DISPATCHES, we create one instance of the flowsheet, initialize it and `clone` it. Also, in most cases, models tend be linear/quadratic, so initialization is not needed. If we want to use the time index, then I believe we have to make a lot of changes to the `MultiPeriodModel` class. Maybe for another PR.

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
andrewlee94: It might be OK, I was mostly curious (and thought that SkeletonUnitModel might have inherited from UnitModelBlock anyway, but I just checked and it does not). However, do you use the methods defined on SkeletonUnitModel at all (add_ports, fix_initialization_states, initialize (deprecated)), or would ProcessBlock be just as good (the parent of SkeletonUnitModel)?

idaes/apps/grid_integration/multiperiod/price_taker_model.py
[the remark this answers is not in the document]
Robbybp: Yes, I have some tools in [`pyomo-mpc`](https://github.com/pyomo/pyomo/tree/main/pyomo/contrib/mpc) for working with dynamic models. These should be useful for working with time-indexed steady state flowsheets, as this is very close to the use case they were designed for. To my basic understanding, `MultiPeriodModel` and the related tooling in Dispatches seems like they are solving a similar problem from a different angle.

idaes/apps/grid_integration/multiperiod/price_taker_model.py
radhakrishnatg: Makes sense. No need to change.

idaes/apps/grid_integration/multiperiod/price_taker_model.py
```diff
@@ -0,0 +1,1008 @@
...
+        blk_name = op_blk.split(".")[-1] + "_capacity_limits"
+        setattr(self.mp_model, blk_name, Block())
+        blk = getattr(self.mp_model, blk_name)
+
+        # Constraint rules for avoiding overlap for multiple-commodity naming
+        def capacity_low_limit_rule(b, t):
+            return (
+                min_capacity * op_mode[self.mp_model.set_period.at(t)]
+                <= var_commodity[self.mp_model.set_period.at(t)]
+            )
```
radhakrishnatg, line 473: You don't have to change it, but you can write it this way:
`min_capacity * op_mode[t] <= var_commodity[t]`

andrewlee94: I think you should make that change - for one it simplifies the code and takes it easier to read, and secondly it removes one unnecessary method call that could potentially break in the future.

radhakrishnatg (approved):
Looks good to me.

idaes/apps/grid_integration/multiperiod/price_taker_model.py
```diff
@@ -0,0 +1,1008 @@
...
+        if not have_skl:
+            raise ImportError(
+                f"Clustering feature requires optional import 'scikit-learn'."
+            )
+        kmeans = KMeans(n_clusters=n_clusters).fit(daily_data.transpose())
+        centroids = kmeans.cluster_centers_
+        labels = kmeans.labels_
+
+        # Set any centroid values that are < 1e-4 to 0 to avoid noise
+        centroids = centroids * (centroids >= 1e-4)
```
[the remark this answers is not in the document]
djlaky: I made the change. Good catch

idaes/apps/grid_integration/multiperiod/price_taker_model.py
andrewlee94: Where and how is this rule used? My inclination would be handle this at the component generation step, and only used the indexing elements you want. That way, this rule will only ever get a single value for `t`.

idaes/apps/grid_integration/multiperiod/price_taker_model.py
```diff
@@ -0,0 +1,1008 @@
...
+        var_commodity = {
+            t: self.mp_model.period[t].find_component(op_blk + "." + commodity_var)
+            for t in self.mp_model.period
+        }
+        max_capacity = self.find_component(design_blk + "." + capacity_max)
+        min_capacity = self.find_component(design_blk + "." + capacity_min)
+
+        # Importing in the necessary variables
+        if not hasattr(self, "range_time_steps"):
+            self.range_time_steps = RangeSet(len(self.mp_model.set_period))
```
andrewlee94, line 462: Can you comment on why this might not exist at this point? It seems to me that this is something that should exist before you get here.

idaes/apps/grid_integration/multiperiod/price_taker_model.py
andrewlee94: Also, why not have two separate indexing sets in this case, one for period and one for year? That way they would be clearly distinguished (and year can be a single element if you don't need it).

Whilst dealing with arbitrary length indexing elements is possible, it does add some complexity and edge cases you need to be able to handle. This brings me to another question, do you have tests for both cases (and any edge cases you can envision occurring)?

radhakrishnatg: @djlaky We have declared:
```python
op_mode = {
            t: self.mp_model.period[t].find_component(op_blk + ".op_mode")
            for t in self.mp_model.period
        }
```
so, `t = self.mp_model.set_period.at(t)`. If elements of `set_period` are tuples, then `t` is also a tuple. Right now, we only plan to support representative days. The multiyear case can be handled by creating an instance of the `PriceTakerModel` for each year. If single index vs. tuple is the issue, then you can construct the constraint in this manner.
```
m.capacity_low_lim = Constraint(m.set_period)
for t in m.set_period:
    m.capacity_low_lim.body(min_capacity * op_mode[t] <= var_commodity[t])
```
This can handle both single index and double index. Double check the method name once again. It has to be either `body` or `expr` or something like that.
Alternatively, you can separate it out depending on full year vs. representative days.

radhakrishnatg: Apparently, there is another approach:
```python
@m.Constraint(m.set_period)
def capacity_low_lim(blk, *t):
    return m.capacity_low_lim.body(min_capacity * op_mode[t] <= var_commodity[t])
```
This works for both single index and tuple.

djlaky: For constraints that have no skipping, sure, but for the ramping constraints and startup/shutdown there are issues:

https://github.com/djlaky/idaes-pse/blob/price-taker-model/idaes/apps/grid_integration/multiperiod/price_taker_model.py#L644-L656

https://github.com/djlaky/idaes-pse/blob/f10ea59ad3f2a430f749748b42be8419e34982a3/idaes/apps/grid_integration/multiperiod/price_taker_model.py#L751-L759

idaes/apps/grid_integration/multiperiod/price_taker_model.py
andrewlee94: Looking more closely, I would recommend breaking this into a separate method. This will make testing easier as well.

idaes/apps/grid_integration/multiperiod/price_taker_model.py
```diff
@@ -0,0 +1,1008 @@
...
+                    f"Excel file was provided but no sheet was specified. Using the first sheet of the excel file."
+                )
+                sheet = 0
+            full_data = pd.read_excel(path_to_file, sheet_name=[sheet])[sheet]
+        elif ".csv" in path_to_file.suffix:
+            full_data = pd.read_csv(
+                path_to_file,
+            )
+
+        if column_name is None:
```
andrewlee94, line 324: A minor comment, but I would generally put the simple checks first before trying to open the file. These checks are a lot cheaper than opening a file, and I prefer to fail sooner rather than later.

idaes/apps/grid_integration/multiperiod/tests/test_price_taker.py
```diff
@@ -0,0 +1,1406 @@
...
+    # data = pd.read_excel(file_path, sheet_name=1)
+    DATA_DIR = Path(__file__).parent
+    file_path = DATA_DIR / "FLECCS_princeton.csv"
+    data = pd.read_csv(file_path)
+    return data
+
+
+@pytest.mark.unit
+def test_seed_value():
+    m = PriceTakerModel()
```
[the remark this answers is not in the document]
andrewlee94: Seeing as `seed` is a property with a custom `settr`, I would suggest testing for the presence of `_seed` and the failure cases for `seed` here.

idaes/apps/grid_integration/multiperiod/tests/test_price_taker.py
```diff
@@ -0,0 +1,1406 @@
...
+
+    # Older versions get n_clusters = 10, Newer versions n_clusters = 11
+    m = PriceTakerModel()
+
+    daily_data = m.generate_daily_data(excel_data["BaseCaseTax"])
+    n_clusters, inertia_values = m.get_optimal_n_clusters(daily_data, plot=True)
+    test_fig = plt.gcf()
+    plt.close("all")
+
+    assert 9 <= n_clusters <= 15
```
andrewlee94, line 105: This is not a great test (I assume this is what you were referring to with different platforms/versions giving different results). This also does not test the inertia values which would be good to include (if possible). Would it be possible to use a much simpler dummy data set to test this which gives a well defined solution for testing?

Also - this is where break the plotting function out as a separate method would help as you could have a separate test for it.

andrewlee94 (changes requested):
1. We will need documentation for all the new classes - at a minimum some technical references for the API (autodocs), but and explanation with a simple demonstration of how to use the tools would be nice if possible (looking at the code, this will be very important).
2. I do not see a test file for design_and_operation_model.py. This might be rolled into another tests, but I recommend having a separate test file for this to help with future maintenance.
3. In general, I think you could (and should) do a lot more unit testing of methods in isolation. From what I can see in the tests, there are not a lot of tests that target individual methods. E.g. I see no test that targets `add_startup_shutdown` and actually tests that it does what it is supposed to do (I see one test for a failure case, but that is it).
4. Most of the tests I see are based on what I assume is an actual case study. I would suggest thinking about whether you can construct simpler dummy cases for unit testing all the methods in isolation, and where possible mock up any necessary starting information so that you have fewer moving parts and simpler cases to test with. The full case study is good for an end-to-end component/integration test of the system, but often make it harder to do unit testing.

ksbeattie:
@djlaky we're hoping to cut a release at the end of the month - next week.  Do you think this will be done in time for that?

adowling2:
We will not have time to incorporate the feedback from Andrew. If we are okay with merging without those suggestions (maybe open an issue to track proposed changes), then yes, this can probably get merged.

ksbeattie:
Pushing this to the Aug release

ksbeattie:
@djlaky any update on this?

idaes/apps/grid_integration/multiperiod/price_taker_model.py
```diff
@@ -0,0 +1,968 @@
...
+
+        while j <= len(raw_data):
+            daily_data[day] = raw_data[i:j].reset_index(drop=True)
+            i = j
+            j = j + self._horizon_length
+            day = day + 1
+
+        return daily_data
+
+    def get_optimal_n_clusters(
```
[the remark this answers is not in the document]
adam-a-a: @lbianchi-lbl I was wondering if you noticed this thread and had any thoughts about bounds on scikitlearn version supported in IDAES. For example, @djlaky points out that the optimal number of clusters in his test case can be either 10 or 11, depending on whether older or newer versions of scikit-learn are used. I am unsure which versions @djlaky is referring to when he describes them as "older" or "newer" versions, but it might be good to consider a solution to this issue to help merge this Pricetaker model.

ksbeattie:
I'm moving this to the Nov release, since it doesn't look like it will make the Aug release.

ksbeattie:
@MarcusHolly @djlaky, we need an update on this.

MarcusHolly:
We're going back to working on this. Hopefully it can be merged this month

ksbeattie:
@djlaky @MarcusHolly what's the news on this?

MarcusHolly:
It's being looked at (and hopefully finished) for this month

MarcusHolly:
> 1. We will need documentation for all the new classes - at a minimum some technical references for the API (autodocs), but and explanation with a simple demonstration of how to use the tools would be nice if possible (looking at the code, this will be very important).
> 2. I do not see a test file for design_and_operation_model.py. This might be rolled into another test, but I recommend having a separate test file for this to help with future maintenance.
> 3. In general, I think you could (and should) do a lot more unit testing of methods in isolation. From what I can see in the tests, there are not a lot of tests that target individual methods. E.g. I see no test that targets `add_startup_shutdown` and actually tests that it does what it is supposed to do (I see one test for a failure case, but that is it).
> 4. Most of the tests I see are based on what I assume is an actual case study. I would suggest thinking about whether you can construct simpler dummy cases for unit testing all the methods in isolation, and where possible mock up any necessary starting information so that you have fewer moving parts and simpler cases to test with. The full case study is good for an end-to-end component/integration test of the system, but often make it harder to do unit testing.

@andrewlee94 Would you be able to give this PR another review. I've addressed most of your specific comments, and some of your above feedback has already been addressed. The rest can be covered in subsequent PR(s). Namely, the remaining action items are to add a demonstration/how-to, add a distinct test file for design_and_operation_model.py, and add more test(s) for methods in isolation

docs/reference_guides/apps/grid_integration/multiperiod/Price_Taker.rst
```diff
@@ -0,0 +1,55 @@
...
+       g(d,u_{s,t},δ_{s,t}) = 0; ∀_{s} ∈ S, t ∈ T
+
+    .. math::
+
+       h(d,u_{s,t},δ_{s,t},u_{s,t+1},δ_{s,t+1}) = 0; ∀_{s} ∈ S, t ∈ T
+
+
+The price taker multi-period modeling workflow involves the integration of multiple software platforms into the IDAES optimization model
+and can be broken down into two distinct functions, as shown in the figure below. In part 1, simulated or historical
+ISO (International Organization for Standardization) data is used to generate locational marginal price (LMP)
```
andrewlee94, line 37: I am fairly that ISO stands for Independent System Operator or something similar.

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
```diff
@@ -0,0 +1,121 @@
...
+)
+from pyomo.common.config import ConfigValue, In
+from idaes.core.base.process_base import declare_process_block_class
+from idaes.core.base.process_base import ProcessBlockData
+
+
+@declare_process_block_class("DesignModel")
+class DesignModelData(ProcessBlockData):
+    """
+    Class for containing design model ...
```
andrewlee94, line 30: I still do not know enough about what this class actually does to approve this. The docs do not explain what this is or how to use it, and I am not sure what this code actually adds (i.e. is it even necessary?).

For one, this needs a descriptive doc string to explain what it does.

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
```diff
@@ -0,0 +1,121 @@
...
+    )
+    CONFIG.declare(
+        "model_args",
+        ConfigValue(
+            default={},
+            doc="Dictionary containing arguments needed for model_func",
+        ),
+    )
+
+    # noinspection PyAttributeOutsideInit
```
andrewlee94, line 48: Why is this needed?

idaes/apps/grid_integration/multiperiod/price_taker_model.py
```diff
@@ -0,0 +1,1053 @@
...
+          but are never used during cashflow construction
+"""
+import pandas as pd
+import numpy as np
+
+have_skl = True
+have_kn = True
+try:
+    from sklearn.cluster import KMeans
+except:
```
andrewlee94, line 32: This should use a more specific Exception type (ImportError?)

idaes/apps/grid_integration/multiperiod/price_taker_model.py
```diff
@@ -0,0 +1,1053 @@
...
+    have_skl = False
+try:
+    from kneed import KneeLocator
```
andrewlee94, line 35: Are these existing dependencies, or are they adding new dependencies we need to be able to handle? If these are new dependencies, then this needs to be discussed in more detail (i.e. a meeting).

idaes/apps/grid_integration/multiperiod/price_taker_model.py
```diff
@@ -0,0 +1,1053 @@
...
+        self.seed = seed
+        self.horizon_length = horizon_length
+
+    @property
+    def seed(self):
+        return self._seed
+
+    @seed.setter
+    def seed(self, value):
+        if not isinstance(value, int):
```
andrewlee94, line 80: Minor comment, but would it be more efficient to just cast value to an int? I.e.:

```
try:
    self._seed = int(value)
except TypeError as e:
   raise TypeError("My message")
```

Also, I think `TypeError` might be more applicable here.

andrewlee94 (changes requested):
I am still not sure what some of the code here is for and why it is needed, and the existing docs do not really explain this. Thus, I am hesitant to approve this as-is in part because this suggests there might need to be more design discussion and secondly because with the existing documentation I am not confident of the maintainability of the code.

docs/reference_guides/apps/grid_integration/multiperiod/Price_Taker.rst
adam-a-a: That's correct- "Independent System Operator".

idaes/apps/grid_integration/multiperiod/price_taker_model.py
MarcusHolly: Import errors are raised in specific methods since not all the methods require one or both of these imports (See L173 and L276)

andrewlee94: That is not the point I was getting at - it is generally bad practice to use a bare `except` as it can mask unexpected exceptions. Always specify the type of exception you want to handle, so that anything you did not expect gets raised as normal.

andrewlee94: Also, if different parts of the code have different dependencies, then that is a suggestion that they should be different modules to isolate the dependencies.

idaes/apps/grid_integration/multiperiod/price_taker_model.py
adam-a-a: @lbianchi-lbl FYI

MarcusHolly: Linking to a previous discussion that explains the changes in setup.py: https://github.com/IDAES/idaes-pse/pull/1201#issuecomment-1585113725. If there is a better way to go about this, let us know.

setup.py
```diff
@@ -47,6 +47,8 @@ class ExtraDependencies:
     ]
     grid = [
         "gridx-prescient>=2.2.1",  # idaes.tests.prescient
+        "scikit-learn>=1.1",
```
andrewlee94, line 50: We need to have some discussion of this at least before we go ahead. Adding dependencies makes maintenance harder and adds more potential points of conflict. Could you provide some discussion here of why these are needed and cannot be provided by existing dependencies (@lbianchi-lbl will want to hear this too)?

One possible solution we should discuss is whether this could and should live in a separate repository where it has greater control over its environment (with out impacting the rest of IDAES) and also means that ongoing maintenance of this code has less effect on IDAES overall.

idaes/apps/grid_integration/multiperiod/price_taker_model.py
lbianchi-lbl: For optional dependencies, you can look into `attempt_import()` from `pyomo.common.dependencies` that streamlines the syntax a bit. You should be able to find usage  examples in the IDAES codebase and/or Pyomo docs.

setup.py
MarcusHolly: Scikit-learn (sklearn) is needed to import Kmeans, which is a [clustering technique](https://scikit-learn.org/1.5/modules/clustering.html) used to determine the sum of squares error for a given number of clusters in a given dataset (price signals in this case). Kneed is needed to import KneeLocator and uses the output of Kmeans to determine what the optimal number of clusters is that a dataset should be broken down into

As far as I know, no existing dependencies cover clustering techniques

adam-a-a: @andrewlee94 @lbianchi-lbl - here is a reminder of the old comment thread related to this: https://github.com/IDAES/idaes-pse/pull/1358/files#r1653251726

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
adam-a-a: @radhakrishnatg would be the best person to respond to this.

setup.py
andrewlee94: From a quick glance, `scipy` appears to have some clustering capability (https://docs.scipy.org/doc/scipy/reference/cluster.html).

My main concern here is that we have had issues in the past with dependency conflicts creating issues with testing the tools (in that a failure in one optional dependency, even if only used in one small part of the code, can break the testing for the entire codebase).

MarcusHolly: I'll look into this...

MarcusHolly: I've implemented the tools from scipy

idaes/apps/grid_integration/multiperiod/price_taker_model.py
```diff
@@ -91,11 +80,30 @@ def horizon_length(self, value):
         if value <= 0:
             raise ValueError(f"horizon_length must be > 0, but {value} is provided.")
         if not isinstance(value, int):
-            raise ValueError(
+            raise TypeError(
                 f"horizon_length must be an integer, but {value} is not an integer"
             )
         self._horizon_length = value
 
+    @staticmethod
```
andrewlee94, line 88: Could you comment on why this is a `staticmethod`?

Also, a minor technical comment but for the doc string this should be called a method, not a function. E.g. https://www.codecademy.com/article/fwd-js-methods-functions

MarcusHolly: If I don't add the `staticmethod` label, I see the following message `Method 'compute_sse' may be 'static'`, so I added it just to be safe.

andrewlee94: That is not a good answer - the better answer is because it does not make use of the class itself (i.e., the `self` argument), but that leads into a deeper design question of what this method is intended to be. One question that came to mind as I read this again was whether this is intended to be something a user will interact with or not.

If a user is not expected to directly interact with this, it might be worth considering making this a private method as well (so as not to clutter the docs).

MarcusHolly: I see no reason why a user would need to interaction with this. 

Does a private method essentially just mean the docstring is restructured as done in the following example:
`    def _set_scaling_factor(
        self, component, component_type, scaling_factor, overwrite=None
    ):
        """
        PRIVATE METHOD

        The purpose of this method is to apply the correct max and min limits to scaling factors
        before setting them. Which set of limits to apply is determined solely by the component_type
        argument, and no type checking is performed.
        """`

idaes/apps/grid_integration/multiperiod/price_taker_model.py
```diff
@@ -55,6 +55,27 @@
 _logger = idaeslog.getLogger(__name__)
 
 
+def _compute_sse(data, centroids, idx):
+    """
+    PRIVATE METHOD
```
andrewlee94, line 60: Now that you have moved this outside the class, it is actually a function rather than a method. The distinction is that a method is a member of (defined inside) a class (and thus normally has the `self` argument), where as a function is not owned by a class (e.g. defined at the module level like this, or defined inside a method or function).

idaes/apps/grid_integration/multiperiod/price_taker_model.py
MarcusHolly: @lbianchi-lbl Can you provide a bit more context about optional dependencies? For this model, having access to a k-means clustering function is necessary, so what would be the purpose of an `attempt_import()`. In other words, if the import attempt failed, we would just have to display an error message saying something along the lines of "Install dependency XYZ". In which case, would there really be any difference between optional dependencies and actual dependencies?

andrewlee94: The purpose at `attempt_import` is for testing purposes to let us know if the tools we need are installed, and if not that we should skip these tests. The goal is to let us isolate potentially problematic dependencies and be able ot set up environments that do not have all dependencies without causing tests to fail.

MarcusHolly: Alright, then scipy should suffice for now since the only clustering technique we've implemented so far is k-means clustering. But if we intend to add options for more clustering techniques in the future, then we'd likely need to reconsider using `attempt_import` or adding in the dependencies for sklearn and kneed

lbianchi-lbl:
Summarizing the conversation from the dev call:

- Authors and reviewers seem to agree that the clustering methods available through SciPy are sufficient for the current needs
- Therefore, this PR should not add any new dependencies; authors can go ahead and remove/undo the corresponding code changes (`attempt_import`, etc) from this PR
- Different clustering strategies and/or implementations can be added in a future PR once this has been merged

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
adam-a-a: Based on discussions, @radhakrishnatg explained the need for these classes, particularly with regard to cost calculations later on. @andrewlee94 are there other particulars that you want addressed, aside from further documentation?

andrewlee94: @radhakrishnatg and I have discussed this. The first big thing is some basic documentation so that users and maintainers can understand what they are looking at (this can just be better doc strings). I also pointed out that the same basic interface could be done using a more standard approach of having the users write derived classes instead of the current approach of providing a method as an argument (it adds 2 or 3 lines of boilerplate code, but aligns this tool with all the rest of IDAES).

I would also like to make sure there is a maintenance plan for this and an assigned maintainer.

adam-a-a: @andrewlee94 Do you mean that we would have Design and Operation model classes as is, with the model build config options (and associated code) removed, then build model specific classes? For example, NGCC design model would inherit from DesignModel class, and it would merely add the required code under the build method to construct the model; similarly, NGCC operation model would inherit from OperationModel and add required code under build method. Is that along the lines of what you were thinking?

In the interim, we would be maintaining this code for flex desal work, but I imagine that there are other activities that would make use of this code as well, now and in the future. I guess the assigned maintainer(s) would be up for discussion. Arguably, it could be @radhakrishnatg , @MarcusHolly , and myself.

docs/reference_guides/apps/grid_integration/multiperiod/Price_Taker.rst
```diff
@@ -0,0 +1,55 @@
+Price Taker
+===========
+
+Price takers are companies or individuals who must accept market prices since they lack the market share
```
adam-a-a, line 4:
```suggestion
Price takers are entities that must accept market prices since they lack the market share
```

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
andrewlee94: Yes, that was my suggestion as it is almost identical to writing the method that is passed in as an argument.

My main reason is because this is how the other IDAES models are written and so aligns this with the standard. Otherwise we have this one section of the code that does things differently which will end up confusing someone at some point.

For the transition, you could support both - someone writing the build method just does not provide the model build config. However, you would want to add a deprecation to the `model_build` pathway and not document it (which is easy as there is no documentation yet).

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
```diff
@@ -0,0 +1,122 @@
...
+        "append_lmp_data",
+        ConfigValue(
+            default=True,
+            domain=In([True, False]),
+            doc="Should LMP data automatically be appended to the model?",
+        ),
+    )
+
+    # noinspection PyAttributeOutsideInit
+    def build(self):
```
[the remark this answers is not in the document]
adam-a-a: @andrewlee94 I was just thinking about this, and on the one hand, using derived classes would be more in line with how the rest of IDAES does things. On the other hand, using this `model_func` approach gives the potential of applying either approach: I can formulate quick and dirty models on the fly by using the config option to pass in a build function, or I can go and create a child class. So in a sense, allowing for the current approach would give more flexibility, while eliminating it would be more restrictive on the user (which might be the way to go in most cases, but not necessarily here). 

In any case, I think that any refactor of this formulation should be part of a subsequent PR. I think it is more important for this overall capability to get merged in first, since there has already been published work on this Pricetaker approach (for one).

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
adam-a-a: I am pasting my response here for better visibility:

@andrewlee94 I was just thinking about this, and on the one hand, using derived classes would be more in line with how the rest of IDAES does things. On the other hand, using this model_func approach gives the potential of applying either approach: I can formulate quick and dirty models on the fly by using the config option to pass in a build function, or I can go and create a child class. So in a sense, allowing for the current approach would give more flexibility, while eliminating it would be more restrictive on the user (which might be the way to go in most cases, but not necessarily here).

In any case, I think that any refactor of this formulation should be part of a subsequent PR. I think it is more important for this overall capability to get merged in first, since there has already been published work on this Pricetaker approach (for one).

andrewlee94: @adam-a-a My counter to the argument against derived classes is that from a code perspective there are only two additional lines of boilerplate code, and thus derived classes are effectively no harder that the `model_build` approach. Consider:

```
def my_func():
    # My code
```

is equivalent to:

```
@declare_process_block_class(MyPriceTaker)
class MyPriceTakerData(PriceTakerBase):
    def build():
        # My code
```

The one difference is how arguments to the method would be handled, but even that is fairly easy. For the derived class approach, the code just needs to refer to `self.config.model_build_arguments` (which already exists) instead of expecting the mas direct arguments.

I do agree that this could be done in a later PR however. The main reason this came up was that it is not particularly well documented how this was intended to be used so I wanted clarification on useage for this PR.

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
```diff
@@ -0,0 +1,215 @@
...
+            m.capex = Expression(expr=cost["capex"] * m.power)
+            m.fom = Expression(expr=cost["fom"] * m.power)
+
+        m = ConcreteModel()
+        m.unit_1 = DesignModel(
+            model_func=my_design_model,
+            model_args={
+                "p_min": 150, "p_max": 600, "cost": {"capex": 10, "fom": 1},
+            },
+        )
```
adam-a-a, line 66: @andrewlee94 - documentation for design model

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
```diff
@@ -0,0 +1,215 @@
...
+        m.op_unit_1 = OperationModel(
+            model_func=my_operation_model, model_args={"design_blk": m.unit_1},
+        )
+    """
```
adam-a-a, line 149: @andrewlee94 - documentation for operation model

adam-a-a:
@andrewlee94 I think we have the basic documentation in here. We also have a tutorial in the pipeline which is pretty much done. Just need approval before putting it up. Thus, maybe that can come in a subsequent PR so that the approval doesn't hold things up further.

If there are other spots that are missing documentation, can you highlight any notable gaps? To me, I think this might be ready to merge now, and we can resolve gaps via an issue that tracks those gaps along with subsequent PRs.

idaes/apps/grid_integration/multiperiod/tests/test_design_and_operation_models.py
```diff
@@ -0,0 +1,112 @@
...
+
+import pytest
+from pyomo.environ import ConcreteModel, Constraint, Expression, Var
+from idaes.apps.grid_integration import (
+    DesignModel,
+    OperationModel,
+)
+
+
+@pytest.mark.unit
```
andrewlee94, line 22: It would be good to have more tests of the different edge cases, such as what happens when the user does not provide the function to build the model.

idaes/apps/grid_integration/multiperiod/tests/test_price_taker.py
```diff
@@ -401,25 +346,21 @@ def test_optimal_clusters_kmax_logger_message2(excel_data):
...
-    with pytest.raises(
-        ImportError,
-        match=(
-            "Optimal cluster feature requires optional imports 'scikit-learn' and 'kneed'."
-        ),
-    ):
-        n_clusters, inertia_values = m.get_optimal_n_clusters(
-            daily_data, kmin=kmin, kmax=kmax
-        )
+# @pytest.mark.unit
```
andrewlee94, line 349: Commented code.

andrewlee94 (dismissed):
A few more requests. The big one is the blanket disabling of pylint warnings - most of those look like things that should actually be fixed, and by doing a blanket disabling I cannot see the lines that are actually causing the issue. This makes it harder to maintain in the future if those issues become sever enough to break the code (these ones are minor enough that they should be fine, but it is a bad habit to get into).

The two other important things are better documenting what happens if the model_func argument is not provided and what this means to the user, and overall documentation of how to actually use these tools to solve a problem (this is probably a subsequent PR).

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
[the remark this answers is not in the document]
dallan-keylogic: > In any case, I think that any refactor of this formulation should be part of a subsequent PR. I think it is more important for this overall capability to get merged in first, since there has already been published work on this Pricetaker approach (for one).

This approach is backwards. If we know a refactor is coming, it's better to do it before the PR is merged. Otherwise we need to deprecate the old class, leave it in a few cycles, and create a similarly-named new class in order to maintain backwards compatibility.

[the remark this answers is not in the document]
adam-a-a: In fact, we don't _know_ a refactor is coming and it is something that is really up for discussion at this point--which is why I said "if any refactor." This PriceTaker class has been delayed for quite some time now, and I don't think we should wait until the discussion settles.

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
```diff
@@ -10,27 +10,63 @@
...
+import logging
+from pyomo.environ import Binary, Param, Var
+from pyomo.common.config import Bool, ConfigDict, ConfigValue
 from idaes.core.base.process_base import declare_process_block_class
 from idaes.core.base.process_base import ProcessBlockData
 
+_logger = logging.getLogger(__name__)
+
 
+# pylint: disable = attribute-defined-outside-init, too-many-ancestors
```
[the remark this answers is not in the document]
lbianchi-lbl: I agree with @andrewlee94. My suggestion:

- Remove module-level directives
- Look at the failures and see if it's feasible to address them individually (either by making changes to the code, or using more narrow-scope disable directives)

idaes/apps/grid_integration/multiperiod/tests/test_design_and_operation_models.py
MarcusHolly: I've added 2 edge cases for when `model_func` is not defined in the design or operation model. Perhaps that is enough for now.

idaes/apps/grid_integration/multiperiod/design_and_operation_models.py
```diff
@@ -45,20 +81,74 @@ class DesignModelData(ProcessBlockData):
...
         super().build()
 
+        self.install_unit = Var(
+            within=Binary,
+            doc="Binary: 1, if the unit is installed, 0 otherwise",
+        )
+
+        if self.config.model_func is None:
+            # Function that builds the design model is not specified
+            return
```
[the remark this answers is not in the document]
adam-a-a: I agree that we should add a logger message here.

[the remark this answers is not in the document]
MarcusHolly: I've addressed this already, but feel free to comment on the warning itself

radhakrishnatg:
@lbianchi-lbl `scipy`'s clustering methods are not very robust. For example:
```
In [2]: import numpy as np

In [3]: data = np.array([[0,0], [0,1], [1,1], [1, 0], [10, 0], [10, 1], [11, 1], [11, 0], [5,10], [5, 11], [6, 11], [6, 10]])
```
The centroids for this dummy data must be `[0.5, 0.5], [10.5, 0.5], [5.5, 10.5]`. `scipy`'s `Kmeans2` can find the centroids successfully sometimes, but fails quite often.
```
In [30]: centroids, labels = spc.kmeans2(data.astype(float), 3)

In [31]: centroids
Out[31]:
array([[10.5,  0.5],
       [ 0.5,  0.5],
       [ 5.5, 10.5]])

In [32]: centroids, labels = spc.kmeans2(data.astype(float), 3)
<ipython-input-32-f9297b3e7da0>:1: UserWarning: One of the clusters is empty. Re-run kmeans with a different initialization.        
  centroids, labels = spc.kmeans2(data.astype(float), 3)

In [33]: centroids, labels = spc.kmeans2(data.astype(float), 3)
<ipython-input-33-f9297b3e7da0>:1: UserWarning: One of the clusters is empty. Re-run kmeans with a different initialization.        
  centroids, labels = spc.kmeans2(data.astype(float), 3)

In [34]: centroids, labels = spc.kmeans2(data.astype(float), 3)
<ipython-input-34-f9297b3e7da0>:1: UserWarning: One of the clusters is empty. Re-run kmeans with a different initialization.        
  centroids, labels = spc.kmeans2(data.astype(float), 3)

In [35]: centroids, labels = spc.kmeans2(data.astype(float), 3)
<ipython-input-35-f9297b3e7da0>:1: UserWarning: One of the clusters is empty. Re-run kmeans with a different initialization.        
  centroids, labels = spc.kmeans2(data.astype(float), 3)

In [36]: centroids, labels = spc.kmeans2(data.astype(float), 3)

In [37]: centroids
Out[37]:
array([[10.5,  0.5],
       [ 0.5,  0.5],
       [ 5.5, 10.5]])
```
When it fails to compute the centroids, the centroid values are wrong.
```
In [41]: centroids, labels = spc.kmeans2(data.astype(float), 3)
<ipython-input-41-f9297b3e7da0>:1: UserWarning: One of the clusters is empty. Re-run kmeans with a different initialization.        
  centroids, labels = spc.kmeans2(data.astype(float), 3)

In [42]: centroids
Out[42]:
array([[ 3.        ,  5.5       ],
       [10.5       ,  0.5       ],
       [-3.95230113,  7.93989469]])
```
From this point of view, `scikit-learn`'s functions are much more robust. We do not observe this kind of behavior with scikit-learn. For this reason, we would like to add scikit-learn as an additional optional dependency. Could you please tell me what is the best way to handle this?

lbianchi-lbl:
@radhakrishnatg you can try to add `scikit-learn` to `setup.py` after the existing entry(es):

https://github.com/IDAES/idaes-pse/blob/327d8f3f50022f3afb9e1c7862c4f6c759db38f1/setup.py#L49-L51

Note that you'll also have to make sure that the tests that need this dependency are skipped and don't fail when that dependency is not installed. There should be examples on how to do that already (using `attempt_import()` from Pyomo and/or `pytest.importorskip()` in the existing code.

idaes/apps/grid_integration/multiperiod/clustering.py
```diff
@@ -0,0 +1,240 @@
...
+        for d in range(num_days)
+    }
+    weights = {d + 1: sum(labels == d) for d in range(num_days)}
+
+    return lmp_data_clusters, weights
+
+
+def get_optimal_num_clusters(
+    samples: Union[pd.DataFrame, np.array],
+    kmin: int = 1,
```
MarcusHolly, line 131: Default should be at least 2 since that is the minimum that silhouette requires

idaes/apps/grid_integration/multiperiod/clustering.py
```diff
@@ -0,0 +1,240 @@
...
+sklearn, sklearn_avail = attempt_import("sklearn")
+
+if sklearn_avail:
+    from sklearn.cluster import KMeans
+    from sklearn.metrics import silhouette_score
+
+_logger = idaeslog.getLogger(__name__)
+
+
+def generate_daily_data(raw_data: list, horizon_length: int):
```
MarcusHolly, line 34: @radhakrishnatg I've updated the file I am working on locally to use `generate_daily_data` as defined here rather than my older implementation. One issue that I'm running into is the presence of NaNs. In order to get around this, I've replaced all the NaNs with the mean of the corresponding column. If there's a better way to deal with NaNs, let me know. I'll attach my updated version of this function below:

```
def generate_daily_data(raw_data: list, horizon_length: int):
    """
    Function used to generate the daily data in a usable format
    from the raw data provided.
    Args:
        raw_data : list,
            Columnar data for a given LMP signal
        horizon_length : int,
            Length of each representative day/period
    Returns:
        daily_data: pd.DataFrame,
            Correctly formatted daily LMP data for later use
    """
    if horizon_length > len(raw_data):
        raise ValueError(
            f"Horizon length {horizon_length} exceeds the price signal length of {len(raw_data)}"
        )

    elements_ignored = len(raw_data) % horizon_length
    if elements_ignored:
        _logger.warning(
            f"Length of the price signal is not an integer multiple of horizon_length.\n"
            f"\tIgnoring the last {elements_ignored} elements in the price signal."
        )

    daily_data = {
        j: raw_data[((j - 1) * horizon_length) : (j * horizon_length)]
        for j in range(1, (len(raw_data) // horizon_length) + 1)
    }

    daily_data = pd.DataFrame(daily_data)
    daily_data = daily_data.fillna(daily_data.mean())
    daily_data = daily_data.transpose()

    # DataFrame arranges the data for each day as a column. Since most clustering techniques
    # require different samples as rows, we take the transpose before returning the data.
    return daily_data
```

MarcusHolly: After looking into this further, I think something is wrong with how daily_data is being set up, but I can't tell what the issue is. When running `daily_data = generate_daily_data(raw_data=data["BaseCaseTax"], horizon_length=24)`, where the BaseCaseTax is coming from `FLECCS_princeton.csv`, I get the following output for `daily_data`
image

MarcusHolly: If I use the following as `generate_daily_data`, I'm able to generate `daily_data` without any issues

```
def generate_daily_data(raw_data: list, horizon_length: int):
    """
    Function used to generate the daily data in a usable format
    from the raw data provided.

    Args:
        raw_data:   Columnar data for a given LMP signal

    Returns:
        daily_data: Correctly formatted daily LMP data for later use
    """
    day_list = list(range(1, (len(raw_data) // horizon_length) + 1))

    daily_data = pd.DataFrame(columns=day_list)

    # Extracting data to populate empty dataframe
    i = 0
    j = horizon_length
    day = 1

    if j > len(raw_data):
        raise ValueError(
            f"tried to generate daily data, but horizon length of {horizon_length} exceeds raw_data length of {len(raw_data)}"
        )

    while j <= len(raw_data):
        daily_data[day] = raw_data[i:j].reset_index(drop=True)
        i = j
        j = j + horizon_length
        day = day + 1

    daily_data = daily_data.transpose()

    return daily_data
```

image

idaes/apps/grid_integration/multiperiod/clustering.py
radhakrishnatg: Good catch! Changed the default to 2.

adam-a-a:
@adowling2 @djlaky - we think this is in a good enough place to merge, and if there are no objections, would like to merge by the next dev call coming up. Please provide any review comments that you might have before the dev call on Thursday. We also plan to follow up with subsequent PRs to make certain refinements, such as adding a tutorial example, adding a clustering method, etc., so if any suggested changes come to mind, please consider whether they should be done in this PR or could be handled in a subsequent PR. Thanks in advance!

idaes/apps/grid_integration/pricetaker/clustering.py
```diff
@@ -0,0 +1,249 @@
...
+sklearn, sklearn_avail = attempt_import("sklearn")
+
+if sklearn_avail:
+    from sklearn.cluster import KMeans
+    from sklearn.metrics import silhouette_score
+
+_logger = idaeslog.getLogger(__name__)
+
+
+def generate_daily_data(raw_data: list, horizon_length: int):
```
MarcusHolly, line 34: I think it is worth adding the following to this function:
```
if not isintance(raw_data, list):
    raw_data = list(raw_data)
```

idaes/apps/grid_integration/pricetaker/clustering.py
```diff
@@ -0,0 +1,249 @@
...
+    if kmax >= len(samples):
+        raise ValueError(f"kmax must be < len(samples). Received kmax = {kmax}")
+
+    k_values = list(range(kmin, kmax + 1))
+    inertia_values = []
+    mean_silhouette = []
+
+    for k in k_values:
+        kmeans = KMeans(n_clusters=k, random_state=seed).fit(samples)
+        inertia_values.append(kmeans.inertia_)
```
MarcusHolly, line 185: The following code has worked better for me. I don't fully understand the underlying mechanism, but the logic is that the current method forces Kmeans to use the same centroid initialization for each k whereas the alternative allows for variation in centroid initialization across different k values. In summary, Kmeans relies on good initial centroids and the alternative reduces the likelihood of repeating a poor initialization
```suggestion
    np.random.seed(seed)
    for k in k_values:
        kmeans = KMeans(n_clusters=k).fit(samples)
        inertia_values.append(kmeans.inertia_)
```

MarcusHolly (commented):
Two small things to address. Otherwise, it LGTM

idaes/apps/grid_integration/pricetaker/tests/test_clustering.py
```diff
@@ -0,0 +1,281 @@
...
+    )
+
+    assert n_clusters == 3
+
+    # Test that a figure was created
+    assert plt.gcf() is not None
+    # Test that axes were created
+    assert plt.gca() is not None
+    # Test that the plot has data
+    # assert plt.gca().has_data()
```
agarciadiego, line 175: Any reason why this test is commented out?

agarciadiego (approved):
Just a small question otherwise this looks good to me

MarcusHolly (approved):

idaes/apps/grid_integration/pricetaker/tests/test_clustering.py
radhakrishnatg: It was a simple plot. I could see the data on the plot, but the assertion was failing for some unknown reason.
There isn't anything in the code that breaks it. So, I thought it would just be sufficient to run it for code coverage. 
I commented it for now. I will see if I can fix it later.

adam-a-a (commented):
LGTM- dropping my review even though this is merged already. My intention is to potentially cover some of these comments as we move forward with subsequent PRs @radhakrishnatg @MarcusHolly.

docs/reference_guides/apps/grid_integration/multiperiod/Price_Taker.rst
```diff
@@ -0,0 +1,55 @@
...
+
+The price taker multi-period modeling workflow involves the integration of multiple software platforms into the IDAES optimization model
+and can be broken down into two distinct functions, as shown in the figure below. In part 1, simulated or historical
+ISO (Independent System Operator) data is used to generate locational marginal price (LMP)
+signals, and production cost models (PCMs) are used to compute and optimize the time-varying dispatch schedules for each
+resource based on their respective bid curves. Advanced data analytics (RAVEN) reinterpret the LMP signals and PCM
+as stochastic realizations of the LMPs in the form of representative days (or simply the full-year price signals).
+In part 2, PRESCIENT uses a variety of input parameters (design capacity, minimum power output, ramp rate, minimum up/down time, marginal cost, no load cost, and startup profile)
+to generate data for the market surrogates. Meanwhile, IDAES uses the double loop simulation to integrate detailed
+process models (b, ii) into the daily (a, c) and hourly (i, iii) grid operations workflow.
```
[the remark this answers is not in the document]
MarcusHolly: Not really - I've trimmed down the "part 2" section, but still kept the image since I think it helps to get a holistic picture of the general workflow

idaes/apps/grid_integration/pricetaker/clustering.py
```diff
@@ -0,0 +1,252 @@
...
+        opt_x : float
+            Optimal x at which curvature changes significantly
+    """
+    # The implementation is based on
+    # Ville Satopaa, Jeannie Albrecht, David Irwin, Barath Raghavan
+    # Finding a “Kneedle” in a Haystack:
+    # Detecting Knee Points in System Behavior
+    # https://raghavan.usc.edu/papers/kneedle-simplex11.pdf
+
+    return len(np.array(x)) + len(np.array(y))
```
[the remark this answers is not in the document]
MarcusHolly: I just removed this function in #1579  since I didn't think it was necessary

idaes/apps/grid_integration/pricetaker/price_taker_model.py
```diff
@@ -0,0 +1,950 @@
...
+            "Created hourly cashflow expressions at:\n\t period[d, t].total_hourly_cost, "
+            "period[d, t].total_hourly_revenue, period[d, t].net_hourly_cash_inflow."
+        )
+
+    def add_overall_cashflows(
+        self,
+        lifetime: int = 30,
+        discount_rate: float = 0.08,
+        corporate_tax_rate: float = 0.2,
+        annualization_factor: Optional[float] = None,
```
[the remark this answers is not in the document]
MarcusHolly: I made this change - we can always revert it later, if necessary

idaes/apps/grid_integration/pricetaker/price_taker_model.py
```diff
@@ -0,0 +1,950 @@
...
+    "horizon_length",
+    ConfigValue(
+        domain=PositiveInt,
+        doc="Length of each representative day/period",
+    ),
+)
+CONFIG.declare(
+    "num_clusters_range",
+    ConfigValue(
+        default=(5, 30),
```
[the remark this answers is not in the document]
MarcusHolly: I'll echo this point. I think it could just be assigned to `num_clusters`, and the lower bound should probably just be 2

idaes/apps/grid_integration/pricetaker/price_taker_model.py
```diff
@@ -0,0 +1,950 @@
...
+        cf.net_profit = Var(doc="Net profit after taxes")
+        cf.net_profit_calculation = Constraint(
+            expr=cf.net_profit == cf.net_cash_inflow - cf.fom - cf.corporate_tax
+        )
+
+        if annualization_factor is None:
+            # If the annualization factor is not specified
+            annualization_factor = discount_rate / (
+                1 - (1 + discount_rate) ** (-lifetime)
+            )
```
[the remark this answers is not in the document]
radhakrishnatg: If the user specifies `annualization_factor`, then `discount_rate` and `lifetime` will be ignored. Would it be sufficient if we just raised a warning saying that the discount rate and lifetime are being ignored?

idaes/apps/grid_integration/pricetaker/price_taker_model.py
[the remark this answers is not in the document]
radhakrishnatg: This option exists only for domain validation. Users will not be interacting with this `_config` option. Users will pass `kmin` and `kmax` to the `get_optimum_num_clusters` method. Inside the method, we use this config option to verify if the specified kmin and kmax are valid or not. To avoid this confusion, we could either add a few comments to the code, or replace this config option with two options: one for `kmin` and the other for `kmax`. I'm okay with either option.

Doesn't matter whether we choose 5 or 4, it will be an arbitrary number. We do not want it to be too low or too high. Typically, for a full-year price signal, based on my earlier work, we will need at least 10 - 12 clusters.  So, I chose 5 which is the middle value.

idaes/apps/grid_integration/pricetaker/price_taker_model.py
```diff
@@ -0,0 +1,950 @@
...
+        into representative periods before appending the data.
+
+        Args:
+            lmp_data: Union[list, tuple, pd.DataFrame]
+                List of locational marginal prices
+
+            num_representative_days: Optional[int], default=None
+                number of clusters or representative periods.
+
+            horizon_length: Optional[int], default=None
```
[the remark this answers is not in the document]
radhakrishnatg: `config` attribute in this code does not work like other IDAES models. Users will not be specifying all the config options at the time of instantiation. Depending on the need, users can pass options to individual methods, and `config` exists just to perform domain validation (verify whether the input to the method is valid or not).

idaes/apps/grid_integration/pricetaker/price_taker_model.py
```diff
@@ -0,0 +1,950 @@
...
+        """
+        self._assert_lmp_data_exists()
+        # Domain validation of kmin and kmax
+        self._config.num_clusters_range = (kmin, kmax)
+
+        daily_data = generate_daily_data(self._config.lmp_data, self.horizon_length)
+
+        return get_optimal_num_clusters(
+            daily_data, kmin, kmax, method, generate_elbow_plot, self._config.seed
+        )
```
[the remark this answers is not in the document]
radhakrishnatg: It is used in the `num_representative_days` property. The `getter` method obtains the number of representative days from `self._config.num_clusters`.

PR #1362, opened 2024-03-01
Restore ability to run pytest on individual files or directories

### LINKED ISSUE #1200: Unable to run pytest on a single test file
As originally reported by @dallan-keylogic:

## Description

- It used to be possible to run pytest on a single test file by specifying the path to the test file as a command-line argument, e.g.: `pytest idaes/models/properties/modular_properties/eos/tests/test_enrtl.py`
- However, since recently, it seems this is not possible anymore: the same command runs all tests regardless of what file is specified

## Probable cause

- Changes to `pytest.ini` in #1133 added a default command-line flag (`--pyargs idaes`) that specifies the location of the `idaes` package as the testpath
  - This was done to ensure more consistent behavior regardless of e.g. the working directory where `pytest` is invoked, as well as forcing the tests to be collected from the installed package (as opposed from the local directory, in case the two are not the same)
- However, this cause the unwanted side effect of always adding all tests inside the `idaes` package directory (and subdirectories) to the current test run, effectively rendering the option of specifying individual test files or directories non-functional

## Workarounds

- Edit `pytest.ini` locally to remove the offending `--pyargs idaes` CLI flag
- To select a subset of the tests to be run, use keywords through the `-k` CLI option https://docs.pytest.org/en/7.2.x/reference/reference.html#command-line-flags instead of passing file paths as arguments
  - Note that this has an overhead since all tests still need to be collected first, which might take several seconds because of the size of the IDAES codebase and the amount of import-time metaprogramming

## Remedies

- To reduce the possibility of these issues occurring in the future, we should consider carefully how the pytest configuration should be distributed between file(s), CLI flags, env vars, etc

## Other useful info

- To show what tests would be run with the current configuration without actually running them, add `--collect-only` to the pytest CLI options

lbianchi-lbl:
- Remove `--pyargs idaes` from `pytest.ini`
- Move markers definition from `pytest.ini` to `idaes/conftest.py` so that tests can be run without warnings without requiring `pytest.ini`, which is not available in non-editable installations without ad-hoc download

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -119,7 +119,7 @@ jobs:
           echo PYTEST_ADDOPTS="$PYTEST_ADDOPTS --cov --cov-report=xml" >> "$GITHUB_ENV"
       - name: Run pytest (not integration)
         run: |
-          pytest -m "not integration"
+          pytest --pyargs idaes -m "not integration"
       - name: Upload coverage report as GHA workflow artifact
         if: matrix.cov-report
         uses: actions/upload-artifact@v4
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -154,7 +154,7 @@ jobs:
         with:
           install-target: -r requirements-dev.txt
       - name: Run pytest (integration)
-        run: pytest -m integration
+        run: pytest --pyargs idaes -m integration
 
   examples:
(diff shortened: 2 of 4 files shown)


andrewlee94 (approved):
Looks fine to me, but I am no expert on configuring pytest.

ksbeattie (approved):
LGTM

PR #1375, opened 2024-03-14
Add diagnostics tests for all unit models

### LINKED ISSUE #1374: DiagnosticsToolbox `report_numerical_issues` raises an error when Jacobian is "exactly" singular
MWE:
```python
import pyomo.environ as pyo                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                     
m = pyo.ConcreteModel()                                                                                                                                                                                                                                              
m.x = pyo.Var([1,2], initialize=1.0)                                                                                                                                                                                                                                 
m.eq = pyo.Constraint(pyo.PositiveIntegers)                                                                                                                                                                                                                          
m.eq[1] = m.x[1] * m.x[2] == 1.5                                                                                                                                                                                                                                     
m.eq[2] = m.x[2] * m.x[1] == 1.5                                                                                                                                                                                                                                     
m.obj = pyo.Objective(expr=m.x[1]**2 + 2 * m.x[2]**2)                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                     
dt = DiagnosticsToolbox(m)                                                                                                                                                                                                                                           
dt.report_numerical_issues()                                                                                                                                                                                                                                         
```
This raises
```console
RuntimeError: Factor is exactly singular
```
from SciPy's SuperLU interface. I would expect us to catch this error and display this information gracefully.

### Version info
Python 3.9.6
SciPy 1.9.3
Recent IDAES main

andrewlee94:
## Summary/Motivation:
This PR adds tests using the diagnostics toolbox to all unit model test cases. Also adds catch for error identified in #1374 

- Patch/workaround for ASL bug in diagnostics toolbox
- Updates tests for all core unit models to include checks for structural and numerical issues.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/scaling.py
+++ b/idaes/core/util/scaling.py
@@ -28,6 +28,7 @@ variables to calculate additional scaling factors.
 __author__ = "John Eslick, Tim Bartholomew, Robert Parker, Andrew Lee"
 
 import math
+import os
 import sys
 
 import scipy.sparse.linalg as spla
@@ -49,10 +50,34 @@ from pyomo.util.calc_var_value import calculate_variable_from_constraint
 from pyomo.core import expr as EXPR
 from pyomo.common.numeric_types import native_types
 from pyomo.core.base.units_container import _PyomoUnit
+from pyomo.common.fileutils import find_library
 
 import idaes.logger as idaeslog
 
 _log = idaeslog.getLogger(__name__)
+
+
+# There appears to be a bug in the ASL which causes terminal failures
+# if you try to create multiple ASL structs with different external
+# functions in the same process. This causes pytest to crash during testing.
+# To avoid this, register all known external functions before we call
+# PyNumero.
+ext_funcs = ["cubic_roots", "general_helmholtz_external", "functions"]
+library_set = set()
+libraries = []
+
+for f in ext_funcs:
+    library = find_library(f)
+    if library not in library_set:
+        library_set.add(library)
+        libraries.append(library)
+
+if "AMPLFUNC" in os.environ:
+    env_str = "\n".join([os.environ["AMPLFUNC"], *libraries])
+else:
+    env_str = "\n".join(libraries)
+
+os.environ["AMPLFUNC"] = env_str
 
 
 def __none_left_mult(x, y):
--- a/idaes/core/util/scaling.py
+++ b/idaes/core/util/scaling.py
@@ -30,6 +30,7 @@ __author__ = "John Eslick, Tim Bartholomew, Robert Parker, Andrew Lee"
 import math
 import os
 import sys
+from typing import Optional, List
 
 import scipy.sparse.linalg as spla
 import scipy.linalg as la
@@ -63,7 +64,7 @@ _log = idaeslog.getLogger(__name__)
 # To avoid this, register all known external functions before we call
 # PyNumero.
 def _ensure_external_functions_libs_in_env(
-    ext_funcs: list[str], var_name: str = "AMPLFUNC", sep: str = "\n"
+    ext_funcs: List[str], var_name: str = "AMPLFUNC", sep: str = "\n"
 ):
     libraries_str = os.environ.get(var_name, "")
     libraries = [lib for lib in libraries_str.split(sep) if lib.strip()]
--- a/idaes/core/util/scaling.py
+++ b/idaes/core/util/scaling.py
@@ -28,9 +28,7 @@ variables to calculate additional scaling factors.
 __author__ = "John Eslick, Tim Bartholomew, Robert Parker, Andrew Lee"
 
 import math
-import os
 import sys
-from typing import Optional, List
 
 import scipy.sparse.linalg as spla
 import scipy.linalg as la
@@ -51,33 +49,10 @@ from pyomo.util.calc_var_value import calculate_variable_from_constraint
 from pyomo.core import expr as EXPR
 from pyomo.common.numeric_types import native_types
 from pyomo.core.base.units_container import _PyomoUnit
-from pyomo.common.fileutils import find_library
 
 import idaes.logger as idaeslog
 
 _log = idaeslog.getLogger(__name__)
-
-
-# There appears to be a bug in the ASL which causes terminal failures
-# if you try to create multiple ASL structs with different external
-# functions in the same process. This causes pytest to crash during testing.
-# To avoid this, register all known external functions before we call
-# PyNumero.
-def _ensure_external_functions_libs_in_env(
-    ext_funcs: List[str], var_name: str = "AMPLFUNC", sep: str = "\n"
-):
-    libraries_str = os.environ.get(var_name, "")
-    libraries = [lib for lib in libraries_str.split(sep) if lib.strip()]
-    for func_name in ext_funcs:
-        lib: Optional[str] = find_library(func_name)
-        if lib is not None and lib not in libraries:
-            libraries.append(lib)
-    os.environ[var_name] = sep.join(libraries)
-
-
-_ensure_external_functions_libs_in_env(
-    ["cubic_roots", "general_helmholtz_external", "functions"]
-)
 
 
 def __none_left_mult(x, y):
--- a/idaes/core/util/scaling.py
+++ b/idaes/core/util/scaling.py
@@ -62,22 +62,21 @@ _log = idaeslog.getLogger(__name__)
 # functions in the same process. This causes pytest to crash during testing.
 # To avoid this, register all known external functions before we call
 # PyNumero.
-ext_funcs = ["cubic_roots", "general_helmholtz_external", "functions"]
-library_set = set()
-libraries = []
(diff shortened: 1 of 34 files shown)


bpaul4 (approved):
Everything looks good to me, I just had a couple of comments.

idaes/models/properties/examples/saponification_thermo.py
```diff
@@ -83,15 +83,15 @@ def build(self):
 
         # Heat capacity of water
         self.cp_mol = Param(
-            mutable=False,
+            mutable=True,
```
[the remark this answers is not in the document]
andrewlee94: As they have units, Pyomo was implicitly making them mutable and logging a warning. I changed them to be explicitly mutable to suppress the warning.

idaes/models/unit_models/tests/test_heat_exchanger.py
```diff
@@ -548,21 +556,11 @@ def test_build(self, btx):
         assert number_total_constraints(btx) == 38
         assert number_unused_variables(btx) == 0
 
-    @pytest.mark.integration
```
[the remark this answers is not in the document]
andrewlee94: No, but I felt that was somewhat superfluous if units were consistent.

idaes/core/util/scaling.py
```diff
@@ -49,12 +50,36 @@
...
+        library_set.add(library)
+        libraries.append(library)
+
+if "AMPLFUNC" in os.environ:
+    env_str = "\n".join([os.environ["AMPLFUNC"], *libraries])
+else:
+    env_str = "\n".join(libraries)
+
+os.environ["AMPLFUNC"] = env_str
+
```
Robbybp: Why is this in the scaling module? I could see two different approaches for implementing this workaround:
1. We don't attempt to fix this for users (until we get a bug report about it), and just add the workaround to a testing module somewhere.
2. We attempt to fix this for users.

This seems to be going with option 2. However, I would argue that, if we're trying to give users access to the workaround, we should put this in a high-level module or `__init__.py`. I say this because users could hit this issue in a wide variety of situations in which `scaling` is not imported. If we just want to narrowly fix the issue with the tests, I would put this in the test module that is triggering the bug (I believe `test_model_diagnostics`).

idaes/core/util/scaling.py
```diff
@@ -49,12 +50,36 @@
...
+# if you try to create multiple ASL structs with different external
+# functions in the same process. This causes pytest to crash during testing.
+# To avoid this, register all known external functions before we call
+# PyNumero.
+ext_funcs = ["cubic_roots", "general_helmholtz_external", "functions"]
+library_set = set()
+libraries = []
+
+for f in ext_funcs:
+    library = find_library(f)
```
Robbybp: This relies on IDAES having modified the path, right? Would it make sense to be more explicit about the libraries we are adding? I.e. use `ideas.bin_directory`. The idea is just that maybe the user has other copies of these libraries on the path (or others with the same names, `functions` in particular is pretty generic).

idaes/core/util/scaling.py
```diff
@@ -49,12 +50,36 @@
...
+    if library not in library_set:
+        library_set.add(library)
+        libraries.append(library)
+
+if "AMPLFUNC" in os.environ:
+    env_str = "\n".join([os.environ["AMPLFUNC"], *libraries])
```
Robbybp: I would do something like:
```python
orig_libraries = os.environ["AMPLFUNC"].split("\n")
env_str = "\n".join(orig_libraries + libraries)
```
This is to guard against unexpected behavior if `"\n\n"` appears in `AMPLFUNC`. Hopefully this wouldn't be a problem for ASL, but I have been surprised before...

idaes/core/util/scaling.py
andrewlee94: The function that is being used for the Jacobian is here, hence why I moved it from the diagnostics. I think this is the only place we use PyNumero for now as well, so it would cover many user cases. The reason I did not put it in the top level `__init__.py`  is just because I am hesitant to have put code there (mostly learning from @lbianchi-lbl's warnings).

idaes/core/util/scaling.py
andrewlee94: I am not sure there - `find_library` is a Pyomo function so I am not sure what it really does. I am not really aware of the `idaes.bin_directory` either - I assume it is the path to where the binaries are so we could just use `path.join(idaes.bin_directory, "function_name.so")`?

idaes/core/util/scaling.py
lbianchi-lbl: @Robbybp I agree with your point that this code being run inconsistently depending on whether `idaes.core.util.scaling` is imported should be avoided as it could lead to very hard to debug behaviors.

As a clarification, my take is that nonstandard code being run as import-time side effects should be avoided if possible, for the reason above and others (including there not being an easy way to "opt out" or "undo" the changes if needed). However, if the choice is between "import-time side effects that are run always" vs "import-time side effects that only run sometimes", the first is IMO preferable. Additionally, since there's already a lot happening in `idaes/__init__.py`/`idaes/config.py`, I can see the argument for keeping all this type of code in one(-ish) place.

In summary, I agree with you in that `idaes/__init__.py` (or any other module that's unconditionally/consistently imported) would be the most reasonable place to put this code.

idaes/core/util/scaling.py
andrewlee94: @Robbybp I am not sure why, but it looks like there is a differnce in behaviour on Windows and Linux if I try this. On Linux, using `path.join(idaes.bin_directory, "function_name.so")` appears to work, but on Windows we see the same terminal failure during testing due to the ASL bug.

For now, I think I will stick with using Pyomo's `find_library` function as it appear to work for both OSes.

Robbybp: This is probably because the libraries have the `.dll` extension on Windows (and the `.dylib` extension on Mac). I would prefer to be more specific about the libraries we're adding to minimize the chance that something unexpected happens. Maybe we could pass `path.join(ideas.bin_directory, libname)` to `find_library`? Here we're only using `find_library` to deal with the OS-specific extension, but I think that's appropriate.

andrewlee94: @Robbybp It looks like that is working now.

Robbybp (approved):
One small question, but this looks good. I think this is a great step to take. (And thanks for catching the SuperLU error.)

idaes/models/unit_models/tests/test_heater.py
```diff
@@ -371,8 +370,8 @@ def sapon(self):
         m.fs.unit.inlet.conc_mol_comp[0, "H2O"].fix(55388.0)
         m.fs.unit.inlet.conc_mol_comp[0, "NaOH"].fix(100.0)
         m.fs.unit.inlet.conc_mol_comp[0, "EthylAcetate"].fix(100.0)
-        m.fs.unit.inlet.conc_mol_comp[0, "SodiumAcetate"].fix(0.0)
-        m.fs.unit.inlet.conc_mol_comp[0, "Ethanol"].fix(0.0)
+        m.fs.unit.inlet.conc_mol_comp[0, "SodiumAcetate"].fix(1e-8)
+        m.fs.unit.inlet.conc_mol_comp[0, "Ethanol"].fix(1e-8)
```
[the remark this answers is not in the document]
andrewlee94: Value fixed at exactly the bound. Whilst that is not a "real" error in the numerical sense (as it is a fixed variable), it is poor modeling practice to have 0 concentrations (and thus zero flows) so I set it to an EPS instead.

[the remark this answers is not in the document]
Robbybp: I disagree that having 0 concentrations is poor practice. As long as at least one concentration is non-zero, we don't have a zero _total_ flow. That said, I think this change is fine.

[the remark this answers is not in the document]
dallan-keylogic: You don't have to deal with zero total flow, but if you either have a power-law concentration expression or calculate entropy of mixing anywhere, you end up with an evaluation error.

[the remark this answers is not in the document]
Robbybp: For my own education/curiosity, where do concentration power-law expressions show up?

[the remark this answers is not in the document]
dallan-keylogic: Probably power-law concentration isn't the best way to describe it, but in conventional mass-action equilibrium expressions:
image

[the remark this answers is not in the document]
andrewlee94: This, and the fact that thermo properties often have logs or inverse functions in them is why I consider it bad practice to set any concentration/mass/mole fraction to absolute zero. I don't think this particular case actually has any of those, but I try to do this everywhere to err on the side of caution (and maybe get others to learn from me).

bpaul4 (approved):
The updates look good to me.

PR #1382, opened 2024-03-20
Crossflow hx

dallan-keylogic:
## Summary/Motivation:
Adds models for crossflow heat exchanger and heater used in SOC dynamic flowsheet to models_extra.

### Legal Acknowledgement


dallan-keylogic:
The test failures are from what were supposed to be minor changes to the HeatExchanger1D model. Apparently having opposite discretization methods on both sides causes problems with the energy balance.  This is a function of coarse discretization---when I switched to collocation the problem disappeared.

idaes/models/unit_models/heat_exchanger_1D.py
```diff
@@ -430,39 +430,62 @@ def build(self):
...
+                "co-current heat exchanger. "
+                "Defaulting to finite "
+                "difference method on the cold side."
+            )
+            self.config.cold_side.transformation_method = "dae.finite_difference"
+
+        if (
+            self.config.hot_side.transformation_method
+            != self.config.cold_side.transformation_method
+        ):
```
bpaul4, line 455: In one of the tests, the `transformation_scheme` is different for the hot/cold sides and expected to throw a `ConfigurationError`, but it appears that the error is only thrown if the `transformation_method` is different for the hot/cold sides.

`@pytest.mark.unit
    def test_config_validation():
        m = ConcreteModel()
        m.fs = FlowsheetBlock(dynamic=False)
    
        m.fs.properties = BTXParameterBlock(valid_phase="Liq")
    
        with pytest.raises(ConfigurationError):
>           m.fs.HX_co_current = HX1D(
                hot_side={
                    "property_package": m.fs.properties,
                    "transformation_scheme": "BACKWARD",
                },
                cold_side={
                    "property_package": m.fs.properties,
                    "transformation_scheme": "FORWARD",
                },
                flow_type=HeatExchangerFlowPattern.cocurrent,
            )
E           Failed: DID NOT RAISE <class 'idaes.core.util.exceptions.ConfigurationError'>`

idaes/models/unit_models/heat_exchanger_1D.py
```diff
@@ -430,39 +430,62 @@ def build(self):
...
+            ):
+                raise ConfigurationError(
+                    "If a collocation method is used for HeatExchanger1D, the user "
+                    "must specify the transformation scheme they want to use."
+                )
             if (
-                self.config.hot_side.transformation_method
-                != self.config.cold_side.transformation_method
-            ) or (
                 self.config.hot_side.transformation_scheme
```
bpaul4, line 474: As you mentioned, this is only caught if collocation is used. Should this check/raise be included for finite difference as well?

idaes/models/unit_models/heat_exchanger_1D.py
dallan-keylogic: The intention of the changes to the `HeatExchanger1D` was to allow the user to use an [upwind difference scheme](https://en.wikipedia.org/wiki/Upwind_scheme) for countercurrent flow and make that the default. Upwinding has some numerical benefits, but they may only apply to explicit time integration schemes. At the same time, using opposite schemes on the hot side/cold side causes energy to not be conserved. @andrewlee94  would say that's a result of the numerical approximation, and what did you expect using a first order method. Nevertheless, we need to think about what we want to do and whether introducing a change like this might cause problems for users.

idaes/models/unit_models/heat_exchanger_1D.py
dallan-keylogic: I catch it for collocation because those methods create additional points within finite elements. If the points don't match up (which they won't for different schemes), you'll get errors trying to create heat linking equations between the control volumes.

For finite differences, no additional points are created, so there's no need to check that the schemes are the same (except to perhaps warn the user that energy may not be conserved).

idaes/models_extra/power_generation/unit_models/heat_exchanger_common.py
jinliangma, line 1: The two scaling factors: sf_di_tube (Line 705) and sf_area_flow_shell_min (Line 709) are declared and not used, which is flagged by the Github diff tool.  For the shell side Re and Nu calculation inside diameter is not used anyway. We should remove those lines

dallan-keylogic: Just pushed a changeset that does that.

idaes/models_extra/power_generation/unit_models/heat_exchanger_common.py
```diff
@@ -0,0 +1,781 @@
...
+        def deltaP_shell_eqn(b, t, x):
+            # FIXME this equation doesn't have elevation change---is this right?
+            return (
+                b.deltaP_shell[t, x] * b.pitch_x
+                == -1.4
+                * b.friction_factor_shell[t, x]
+                * shell.properties[t, x].dens_mol_phase["Vap"]
+                * shell.properties[t, x].mw
+                * b.v_shell[t, x] ** 2
+            )
```
dallan-keylogic, line 444: @jinliangma , should this equation have an elevation change term?

idaes/models_extra/power_generation/unit_models/heat_exchanger_common.py
```diff
@@ -0,0 +1,781 @@
...
+        def deltaP_tube_friction_eqn(b, t, x):
+            return (
+                b.deltaP_tube_friction[t, x]
+                * pyunits.convert(b.di_tube, to_units=tube_units["length"])
+                == -0.5
+                * tube.properties[t, x].dens_mol_phase["Vap"]
+                * tube.properties[t, x].mw
+                * b.v_tube[t, x] ** 2
+                * b.friction_factor_tube[t, x]
+            )
```
dallan-keylogic, line 636: @jinliangma , should this equation have an elevation change term?

idaes/models_extra/power_generation/unit_models/heat_exchanger_common.py
jinliangma: Original cross flow heat exchanger is for boiler superheaters and economizers. The shell side has flue gas that is very low in density such that the static pressure difference due to elevation change can be neglected.  We only considered the tube side elevation effect since economizer's tube side has liquid water.  I believe this model now considers only the vapor phase so we may remove the elevation term for both sides.  It won't hurt if we add those terms for both sides though.

idaes/models_extra/power_generation/unit_models/cross_flow_heat_exchanger_1D.py
```diff
@@ -0,0 +1,815 @@
...
+        opt = get_solver(solver, optarg)
+
+        hot_side = blk.hot_side
+        cold_side = blk.cold_side
+        t0 = blk.flowsheet().time.first()
+        if not (
+            "temperature" in hot_side.properties[t0, 0].define_state_vars().keys()
+            and "temperature" in cold_side.properties[t0, 0].define_state_vars().keys()
+        ):
+            raise NotImplementedError(
```
bpaul4, line 445: @andrewlee94 do you have suggestions on the error message below, or is this fine for a user-facing message?

idaes/models_extra/power_generation/unit_models/heater_1D.py
```diff
@@ -0,0 +1,566 @@
...
+
+        # ---------------------------------------------------------------------
+        # Initialize shell block
+
+        flags = blk.control_volume.initialize(
+            outlvl=0, optarg=optarg, solver=solver, state_args=state_args
+        )
+
+        init_log.info_high("Initialization Step 1 Complete.")
+
```
bpaul4, line 460: Commented code, if it's not here for future debugging it should be cleaned up.

idaes/models_extra/power_generation/unit_models/tests/test_cross_flow_heat_exchanger_1D.py
```diff
@@ -0,0 +1,266 @@
...
+from idaes.models_extra.power_generation.properties.natural_gas_PR import (
+    get_prop,
+    EosType,
+)
+from idaes.models_extra.power_generation.unit_models import CrossFlowHeatExchanger1D
+import idaes.core.util.model_statistics as mstat
+from idaes.core.util.model_statistics import degrees_of_freedom
+from idaes.core.solvers import get_solver
+
+# Set up solver
```
bpaul4, line 31: Same as above, bound_push is commented out.

idaes/models_extra/power_generation/unit_models/tests/test_cross_flow_heat_exchanger_1D.py
```diff
@@ -0,0 +1,266 @@
...
+        "H2": 1,
+        "H2O": 1,
+        "N2": 10,
+        "Ar": 10,
+    }
+    for comp, s in _mf_scale.items():
+        pp.set_default_scaling("mole_frac_comp", s, index=comp)
+        pp.set_default_scaling("mole_frac_phase_comp", s, index=("Vap", comp))
+        pp.set_default_scaling("flow_mol_phase_comp", s * 1e-3, index=("Vap", comp))
+
```
bpaul4, line 131: Same for commented ssf lines.

idaes/models_extra/power_generation/unit_models/tests/test_heater_1D.py
```diff
@@ -0,0 +1,222 @@
...
+from idaes.models_extra.power_generation.properties.natural_gas_PR import (
+    get_prop,
+    EosType,
+)
+from idaes.models_extra.power_generation.unit_models import Heater1D
+import idaes.core.util.model_statistics as mstat
+from idaes.core.util.model_statistics import degrees_of_freedom
+from idaes.core.solvers import get_solver
+
+# Set up solver
```
bpaul4, line 30: Same here as well.

idaes/models_extra/power_generation/unit_models/cross_flow_heat_exchanger_1D.py
```diff
@@ -0,0 +1,815 @@
...
+                cst(
+                    self.heat_shell_eqn[t, z], s_Q_shell * value(self.length_flow_shell)
+                )
+                # Geometric mean is overkill for most reasonable cases, but it mitigates
+                # damage done when one stream has an unset scaling factor
+                ssf(self.temp_wall_center[t, z], (sf_T_shell * sf_T_tube) ** 0.5)
+                cst(self.temp_wall_center_eqn[t, z], (sf_Q_tube * s_Q_shell) ** 0.5)
+
+    def _get_performance_contents(self, time_point=0):
+        var_dict = {}
```
bpaul4, line 787: Commented code, is this needed for anything?

bpaul4 (approved):
Looks good, I have a few minor comments that aren't critical to merging this.

idaes/models_extra/power_generation/unit_models/cross_flow_heat_exchanger_1D.py
```diff
@@ -0,0 +1,808 @@
...
+                "whose single phase is a vapor phase. The hot side phase is not a vapor phase."
+            )
+        if not pobj_cold.is_vapor_phase():
+            raise ConfigurationError(
+                "The CrossFlowHeatExchanger1D model is valid only for property packages "
+                "whose single phase is a vapor phase. The cold side phase is not a vapor phase."
+            )
+
+        # Reference
+        add_object_reference(self, "heat_tube", tube.heat)
```
andrewlee94, line 211: This looks like a place where `References` would be better.

idaes/models_extra/power_generation/unit_models/cross_flow_heat_exchanger_1D.py
```diff
@@ -0,0 +1,808 @@
...
+        add_object_reference(self, "heat_shell", shell.heat)
+
+        shell_has_pressure_change = False
+        tube_has_pressure_change = False
+
+        if self.config.cold_side.has_pressure_change:
+            if self.config.shell_is_hot:
+                add_object_reference(self, "deltaP_tube", tube.deltaP)
```
andrewlee94, line 219: As above.

idaes/models_extra/power_generation/unit_models/cross_flow_heat_exchanger_1D.py
```diff
@@ -0,0 +1,808 @@
...
+                shell_has_pressure_change = True
+        if self.config.hot_side.has_pressure_change:
+            if self.config.shell_is_hot:
+                add_object_reference(self, "deltaP_shell", shell.deltaP)
+                shell_has_pressure_change = True
+            else:
+                add_object_reference(self, "deltaP_tube", tube.deltaP)
+                tube_has_pressure_change = True
+
+        heat_exchanger_common._make_performance_common(  # pylint: disable=W0212
```
andrewlee94, line 232: Minor comment - could you use the full name of the pylint exclusion. It makes it a lot easier to understand what is being skipped and helps for future maintenance.

idaes/models_extra/power_generation/unit_models/cross_flow_heat_exchanger_1D.py
```diff
@@ -0,0 +1,808 @@
...
+        heat_exchanger_common._make_performance_tube(  # pylint: disable=W0212
+            self,
+            tube=tube,
+            tube_units=tube_units,
+            tube_has_pressure_change=tube_has_pressure_change,
+            make_reynolds=True,
+            make_nusselt=True,
+        )
+
+        def heat_accumulation_term(b, t, x):
```
andrewlee94, line 249: It would probably be better to put this closer to where it is used so readers don;t have to jump back and forth.

idaes/models_extra/power_generation/unit_models/cross_flow_heat_exchanger_1D.py
```diff
@@ -0,0 +1,808 @@
...
+        @self.Constraint(
+            self.flowsheet().config.time,
+            shell.length_domain,
+            doc="Nusselts number equation on shell side",
+        )
+        def N_Nu_shell_eqn(b, t, x):
+            return (
+                b.N_Nu_shell[t, x]
+                == b.f_arrangement
+                * 0.33
```
andrewlee94, line 275: As above.

idaes/models_extra/power_generation/unit_models/cross_flow_heat_exchanger_1D.py
```diff
@@ -0,0 +1,808 @@
...
+                )
+                return (dT0 - dT1) / log(dT0 / dT1)
+
+            @self.Expression(self.flowsheet().config.time)
+            def overall_heat_transfer_coefficient(b, t):
+                return b.total_heat_duty[t] / (
+                    b.total_heat_transfer_area * b.log_mean_delta_temperature[t]
+                )
+
+    def initialize_build(
```
andrewlee94, line 401: General comment: I am inclined to not include this method in new models and just provide `Initializer` objects instead, as this API has been deprecated.

idaes/models_extra/power_generation/unit_models/cross_flow_heat_exchanger_1D.py
andrewlee94: I think it is fine, especially for a contrib model.

idaes/models_extra/power_generation/unit_models/heater_1D.py
```diff
@@ -0,0 +1,564 @@
...
+    useDefault,
+)
+import idaes.core.util.scaling as iscale
+from idaes.core.solvers import get_solver
+from idaes.core.util.config import is_physical_parameter_block
+from idaes.core.util.misc import add_object_reference
+import idaes.logger as idaeslog
+from idaes.core.util.tables import create_stream_table_dataframe
+from idaes.core.util.model_statistics import degrees_of_freedom
+from idaes.models_extra.power_generation.unit_models.heat_exchanger_common import (
```
andrewlee94, line 46: You should not be importing underscore functions from a helper module. If these are intended to be imported like this, then this implies they are not private functions.

idaes/models_extra/power_generation/unit_models/heater_1D.py
```diff
@@ -0,0 +1,564 @@
...
+)
+
+__author__ = "Jinliang Ma, Douglas Allan"
+
+
+@declare_process_block_class("Heater1D")
+class Heater1DData(UnitModelBlockData):
+    """Standard Trim Heater Model Class Class."""
+
+    CONFIG = ConfigBlock()
```
andrewlee94, line 59: I would recommend using the `CONFIG` from `UnitModelBlockData` as it already has the `dynamic` and `has_holdup` options, and means that if new required arguments are added in the future that these will be inherited automatically.

idaes/models_extra/power_generation/unit_models/heat_exchanger_common.py
```diff
@@ -0,0 +1,764 @@
...
+
+
+def _make_geometry_common(blk, shell_units):
+    # Number of tube columns in the cross section plane perpendicular to shell side fluid flow (y direction)
+    blk.ncol_tube = Var(
+        initialize=10.0, doc="Number of tube columns", units=pyunits.dimensionless
+    )
+
+    # Number of segments of tube bundles
+    blk.nseg_tube = Var(
```
andrewlee94, line 46: Minor comment: I would generally encourage longer names to avoid any confusion about what these are. Not that important for contrib models however.

idaes/models_extra/power_generation/unit_models/cross_flow_heat_exchanger_1D.py
```diff
@@ -0,0 +1,808 @@
...
+        # Nusselts number
+        @self.Constraint(
+            self.flowsheet().config.time,
+            tube.length_domain,
+            doc="Nusselts number equation on tube side",
+        )
+        def N_Nu_tube_eqn(b, t, x):
+            return (
+                b.N_Nu_tube[t, x]
+                == 0.023
```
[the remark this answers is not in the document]
dallan-keylogic: @jinliangma , do you know where this correlation came from?

idaes/models_extra/power_generation/unit_models/cross_flow_heat_exchanger_1D.py
dallan-keylogic: @jinliangma , do you know where this correlation came from?

idaes/models_extra/power_generation/unit_models/heater_1D.py
dallan-keylogic: The main rationale for importing underscore functions is to avoid duplicated code without making a contract with the end user to maintain it in its current form indefinitely. If they're not private functions, that implies there's also a layer of testing, etc. that should go into maintaining them.

andrewlee94: Whilst I do understand what you are trying to do, it still runs against Python coding practices. Also, given this is in `models_extra`, there is implicitly only a minimal contract for maintenance (although most users would not realize this).

One possible alternative would be to set things up so that the model inherits these methods, e.g. either throguh a base class, multiple inheritance or a mix-in class (don't ask me for details on the last one, I just know such things exist).

idaes/models_extra/power_generation/unit_models/cross_flow_heat_exchanger_1D.py
dallan-keylogic: Is there a reference for this new API?

image

andrewlee94: https://idaes-pse.readthedocs.io/en/stable/reference_guides/initialization/index.html

andrewlee94 (commented):
A few more comments but again mostly minor (and some are just suggestions you can feel free to ignore).

andrewlee94 (approved):

idaes/models_extra/power_generation/unit_models/heater_1D.py
```diff
@@ -0,0 +1,548 @@
...
+
+@declare_process_block_class("Heater1D")
+class Heater1DData(UnitModelBlockData):
+    """Standard Trim Heater Model Class Class."""
+
+    default_initializer = Heater1DInitializer
+
+    CONFIG = UnitModelBlockData.CONFIG()
+    CONFIG.declare(
+        "has_fluid_holdup",
```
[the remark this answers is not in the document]
dallan-keylogic: Yes, it is. It's a stub for potentially having mass and energy holdup in the fluid (as opposed to the energy holdup in the walls). However, due to concerns about pressure waves, its only valid value is `False` for the time being.

idaes/models_extra/power_generation/unit_models/heat_exchanger_common.py
```diff
@@ -0,0 +1,539 @@
...
+                )
+                cst(
+                    blk.conv_heat_transfer_coeff_shell_eqn[t, z],
+                    sf_conv_heat_transfer_coeff_shell * sf_do_tube,
+                )
+
+            # FIXME estimate from parameters
+            if blk.config.has_holdup:
+                s_U_holdup = gsf(blk.heat_holdup[t, z])
+                cst(blk.heat_holdup_eqn[t, z], s_U_holdup)
```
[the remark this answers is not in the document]
dallan-keylogic: Unit tests for scaling factors? Probably better to have some assertion about no numerical issues from the diagnostics toolbox.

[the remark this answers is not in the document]
andrewlee94: You really want both - the purpose of the unit test is not to check the scaling factors themselves but that the code runs smoothly (i.e. to help you narrow down where the issues is occurring rather than having to work throguh the entire unit-level method to find where it breaks). You do need a separate set of tests to make sure that the scaling factors do improve the condition of the model (for limited test cases at least).

idaes/models_extra/power_generation/unit_models/cross_flow_heat_exchanger_1D.py
```diff
@@ -0,0 +1,1041 @@
...
+            )
+            tube_units = (
+                model.config.hot_side.property_package.get_metadata().derived_units
+            )
+
+        # Trigger creation of cp for use in future initialization
+        # Important to do before initializing property packages in
+        # case it is implemented as Var-Constraint pair instead of
+        # an Expression
+        value(hot_side.properties[t0, 0].cp_mol)
```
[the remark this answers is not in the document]
dallan-keylogic: I took those out, but then Pylint complained the statements were pointless, so I put them back in.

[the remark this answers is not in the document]
dallan-keylogic: Went back and took the values out while disabling Pylint warnings for them because there's no guarantee the evaluation won't cause some error before the model is initialized.

PR #1383, opened 2024-03-22
Allow upwinding in 1D Heat exchanger

dallan-keylogic:
## Summary/Motivation:
Breaks off the changes to the `HeatExchanger1D` from #1382 into their own PR

- Allows upwind difference schemes to be used in the `HeatExchanger1D`
- Adds warnings about energy conservation when different finite difference schemes are used on either side of the heat exchanger.

### Legal Acknowledgement


idaes/models/unit_models/tests/test_heat_exchanger_1D.py
```diff
@@ -274,36 +275,197 @@ def test_config():
...
                 "transformation_scheme": "BACKWARD",
             },
             cold_side={
                 "property_package": m.fs.properties,
+                "transformation_method": "dae.finite_difference",
                 "transformation_scheme": "FORWARD",
             },
+            flow_type=HeatExchangerFlowPattern.countercurrent,
+        )
+    assert "To avoid" not in caplog.text
```
bpaul4, line 352: What is line 352 checking for?

dallan-keylogic: This is to check that the upwinding disclaimer is not logged. I'll add a note to that effect.

bpaul4 (approved):
Everything looks good, @dallan-keylogic.

idaes/models/unit_models/heat_exchanger_1D.py
```diff
@@ -430,96 +430,125 @@ def build(self):
 
         # Set flow directions for the control volume blocks and specify
         # discretization if not specified.
-        if self.config.flow_type == HeatExchangerFlowPattern.cocurrent:
-            set_direction_hot = FlowDirection.forward
-            set_direction_cold = FlowDirection.forward
+        if self.config.hot_side.transformation_method is useDefault:
+            _log.warning(
```
andrewlee94, line 434: Should this be a warning level message? I am of two minds, but I think this would be better as info.

idaes/models/unit_models/heat_exchanger_1D.py
```diff
@@ -430,96 +430,125 @@ def build(self):
...
+                "Discretization method was "
+                "not specified for the hot side of the "
+                "heat exchanger. "
+                "Defaulting to finite "
+                "difference method on the hot side."
+            )
+            self.config.hot_side.transformation_method = "dae.finite_difference"
+        if self.config.cold_side.transformation_method is useDefault:
+            _log.warning(
```
andrewlee94, line 443: As above.

idaes/models/unit_models/heat_exchanger_1D.py
```diff
@@ -430,96 +430,125 @@ def build(self):
...
             if (
-                self.config.hot_side.transformation_method
-                != self.config.cold_side.transformation_method
-            ) or (
                 self.config.hot_side.transformation_scheme
                 != self.config.cold_side.transformation_scheme
             ):
                 raise ConfigurationError(
+                    "If a collocation method is used, "
                     "HeatExchanger1D only supports similar transformation "
```
andrewlee94, line 479: Replace "similar" with "same".

idaes/models/unit_models/heat_exchanger_1D.py
```diff
@@ -430,96 +430,125 @@ def build(self):
...
+                f"Found method {self.config.hot_side.transformation_method} "
+                f"on hot side and method {self.config.cold_side.transformation_method} "
+                "on cold side."
+            )
+        if self.config.hot_side.transformation_method == "dae.collocation":
+            if (
+                self.config.hot_side.transformation_scheme is useDefault
+                or self.config.cold_side.transformation_scheme is useDefault
+            ):
+                raise ConfigurationError(
```
andrewlee94, line 469: Minor comment, but should we have a default here? I vaguely recall that one of the collocation schemes had a bug too.

idaes/models/unit_models/heat_exchanger_1D.py
```diff
@@ -430,96 +430,125 @@ def build(self):
...
-            if self.config.hot_side.transformation_scheme is useDefault:
-                _log.warning(
-                    "Discretization scheme was "
-                    "not specified for the hot side of the "
-                    "co-current heat exchanger. "
-                    "Defaulting to backward finite "
-                    "difference on the hot side."
+
+        if self.config.hot_side.transformation_scheme is useDefault:
+            _log.warning(
```
andrewlee94, line 499: Again, I think this would probably be best as an info messasge.

idaes/models/unit_models/heat_exchanger_1D.py
```diff
@@ -430,96 +430,125 @@ def build(self):
...
+                "Discretization scheme was "
+                "not specified for the hot side of the "
+                "counter-current heat exchanger. "
+                "Defaulting to backward finite "
+                "difference on the hot side."
+            )
+            self.config.hot_side.transformation_scheme = "BACKWARD"
+        if self.config.cold_side.transformation_scheme is useDefault:
+            _log.warning(
```
andrewlee94, line 508: Again.

idaes/models/unit_models/heat_exchanger_1D.py
```diff
@@ -776,6 +805,7 @@ def initialize_build(
         cold_side_units = (
             self.cold_side.config.property_package.get_metadata().get_derived_units
         )
+        # TODO What if there is more than one time point? What if t0 != 0?
```
andrewlee94, line 808: This should at least have `t0 = self.flowsheet().time.first()`.

As for multiple time points, that is on the user to determine how best to proceed. Initialization generally focuses on steady-state. Note that we would probably need to update the `Initializer` as well.

idaes/models/unit_models/heat_exchanger_1D.py
```diff
@@ -881,3 +911,13 @@ def calculate_scaling_factors(self):
...
                 overwrite=False,
             )
+
+
+def _log_upwinding_disclaimer(side, flow_config, bad_scheme, good_scheme):
+    _log.info(
+        f"For {side}, a {bad_scheme} scheme was chosen to discretize the length domain. "
+        f"However, this scheme is not an upwind scheme for {flow_config} flow, and "
+        f"as a result may run into numerical stability issues. To avoid this, "
+        f"use a {good_scheme} scheme (which may result into energy conservation issues "
```
andrewlee94, line 921: Typo: into -> in

idaes/models/unit_models/tests/test_heat_exchanger_1D.py
```diff
@@ -274,36 +275,199 @@ def test_config():
...
+        m.fs.HX_cocurrent1 = HX1D(
+            hot_side={
+                "property_package": m.fs.properties,
+            },
+            cold_side={
+                "property_package": m.fs.properties,
+            },
             flow_type=HeatExchangerFlowPattern.cocurrent,
         )
+    assert (
```
andrewlee94, line 374: Minor comment, but I wonder if this should be broken into smaller tests, one for each test case. Running them all in one test means we will only see the first failure that occurs, which might mean others are hidden until earlier failures are resolves (this makes fixing things faster if/when they break).

idaes/models/unit_models/tests/test_heat_exchanger_1D.py
```diff
@@ -764,6 +928,248 @@ def test_numerical_issues(self, btx):
...
+        assert perf_dict == {
+            "vars": {
+                "Area": btx.fs.unit.area,
+                "Length": btx.fs.unit.hot_side.length,
+            }
+        }
+
+    @pytest.mark.ui
+    @pytest.mark.unit
+    def test_get_stream_table_contents(self, btx):
```
andrewlee94, line 1035: Minor comment, but I don;t think this test is necessary (we've already tested this functionality elsewhere).

idaes/models/unit_models/tests/test_heat_exchanger_1D.py
```diff
@@ -764,6 +928,248 @@ def test_numerical_issues(self, btx):
...
+                "temperature": pytest.approx(298.15, rel=1e-4),
+                "pressure": pytest.approx(101325.0, rel=1e-4),
+            },
+        }
+
+        assert stable.to_dict() == expected
+
+    @pytest.mark.skipif(solver is None, reason="Solver not available")
+    @pytest.mark.component
+    def test_initialize(self, btx):
```
andrewlee94, line 1084: For completeness, I think we should make sure the new `Initializers` work with collocation schemes.

andrewlee94 (commented):
A lot of comments ,but most are really minor.

Regarding the level of logger messages, I can see why you put them at "warning" level, but I think we should reserve that for cases where something is going wrong, and not places where we are using default settings.

idaes/models/unit_models/heat_exchanger_1D.py
dallan-keylogic: This was originally a warning-level message before my proposed revisions. I'd be happy to downgrade it to "Caution" or "Info".

idaes/models/unit_models/heat_exchanger_1D.py
dallan-keylogic: Perhaps. Originally I wanted the user to assume full responsibility for using collocation, but since we need it for energy conservation in upwind schemes I might make `"LAGRANGE-RADAU"` the default. I should test it out on `"LAGRANGE-LEGENDRE"` as well, because that's the one that potentially has a bug in it; it creates additional continuity equations about the values of the function between finite elements, and I can easily see that it fails to do so correctly when flow is moving from 1 to 0.

idaes/models/unit_models/tests/test_heat_exchanger_1D.py
dallan-keylogic: My intention was just to copy-and-paste the existing tests for cocurrent and countercurrent configurations. Perhaps the same duplication is occuring there. I can certainly just delete this test, though.

andrewlee94: I would - the others do duplicate it, but it has been tested enough by this point.

idaes/models/unit_models/heat_exchanger_1D.py
andrewlee94: Maybe make it a caution then - it is definitely not a warning but probably warrants more than info.

idaes/models/unit_models/heat_exchanger_1D.py
andrewlee94: My vague recollection is the it was Lagrange-Legendre that had the possible bug. If so - we should make Lagrange-Radau the default.

blnicho: Is there an issue for this bug with Lagrange-Legendre?

dallan-keylogic: @blnicho @andrewlee94 It appears to work just fine for the `HeatExchanger1D` in counter-current mode, and, furthermore, both methods return the same answers for outlet temperatures.

idaes/logger.py
```diff
@@ -26,14 +26,16 @@
 CRITICAL = logging.CRITICAL  # 50
 ERROR = logging.ERROR  # 40
 WARNING = logging.WARNING  # 30
+CAUTION = 25
```
andrewlee94, line 29: I originally wondered if it would just be better to use INFO_LOW for the "cautions" rather than add additional logger levels, however the naming of the INFO levels is somewhat counter-intuitive (named by amount of output rather than importance). In that sense, I think I am actually inclined to keep CAUTION as being more intuitive.

andrewlee94 (approved):

andrewlee94:
@dallan-keylogic It looks like some tests are failing due to numerical diagnostics checks (possibly due to merging the near parallel constraint check into the main diagnostics toolbox).

dallan-keylogic:
@lbianchi-lbl @ksbeattie Punting this was unsuccessful.  Codecov demands its pound of flesh.

lbianchi-lbl:
> @lbianchi-lbl @ksbeattie Punting this was unsuccessful. Codecov demands its pound of flesh.

It looks like the jobs on our side (i.e. the only thing we have control over) are passing: 
image

The reason why the required checks don't show up is on Codecov's side. Unfortunately, there's not much we can do other than maybe trying to re-run the CI (using `git commit --allow-empty -m "Trigger CI" && git push` ain't the pretties, but it should work).

PR #1385, opened 2024-03-23
Add parallel constraint/variable check to `report_numerical_issues`

### LINKED ISSUE #1380: Add parallel variables/constraints to `report_numerical_issues`
Right now, `display_near_parallel_*` methods are considered "advanced diagnostic techniques" and only suggested if `report_numerical_issues` finds no issues. However, I would argue that they should be considered "numerical diagnostics" as they run quickly, give concise output (for an appropriately tight tolerance), and don't depend on external solvers. I propose we add them to the "numerical warnings" section of the diagnostics toolbox. I am happy to open a PR if others agree.

Robbybp:
## Summary/Motivation:
These checks are fast and precise, so they should go in the general high-level reporting method.

- Add parallel constraint/variable checks to numerical warnings section of `report_numerical_issues`
- Update expected strings in tests

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/tests/test_model_diagnostics.py
+++ b/idaes/core/util/tests/test_model_diagnostics.py
@@ -1371,8 +1371,8 @@ Model Statistics
 3 WARNINGS
 
     WARNING: 2 Constraints with large residuals (>1.0E-05)
-    WARNING: 1 pair of constraints are parallel (to tolerance 1e-08)
-    WARNING: 1 pair of variables are parallel (to tolerance 1e-08)
+    WARNING: 1 pair of constraints are parallel (to tolerance 1.0E-08)
+    WARNING: 1 pair of variables are parallel (to tolerance 1.0E-08)
 
 ------------------------------------------------------------------------------------
 0 Cautions
@@ -1456,7 +1456,7 @@ Model Statistics
     WARNING: 1 Constraint with large residuals (>1.0E-05)
     WARNING: 2 Variables with extreme Jacobian values (<1.0E-08 or >1.0E+08)
     WARNING: 1 Constraint with extreme Jacobian values (<1.0E-08 or >1.0E+08)
-    WARNING: 3 pairs of variables are parallel (to tolerance 1e-08)
+    WARNING: 3 pairs of variables are parallel (to tolerance 1.0E-08)
 
 ------------------------------------------------------------------------------------
 4 Cautions
--- a/idaes/core/util/tests/test_model_diagnostics.py
+++ b/idaes/core/util/tests/test_model_diagnostics.py
@@ -1371,8 +1371,8 @@ Model Statistics
 3 WARNINGS
 
     WARNING: 2 Constraints with large residuals (>1.0E-05)
-    WARNING: 1 pair of constraints are parallel (to tolerance 0.0001)
-    WARNING: 1 pair of variables are parallel (to tolerance 0.0001)
+    WARNING: 1 pair of constraints are parallel (to tolerance 1e-08)
+    WARNING: 1 pair of variables are parallel (to tolerance 1e-08)
 
 ------------------------------------------------------------------------------------
 0 Cautions
@@ -1436,8 +1436,8 @@ Suggested next steps:
         model.v2 = Var(initialize=0)
         model.v3 = Var(initialize=0)
 
-        model.c1 = Constraint(expr=model.v1 == model.v2)
-        model.c2 = Constraint(expr=model.v1 == 1e-8 * model.v3)
+        model.c1 = Constraint(expr=1e-2 * model.v1 == model.v2)
+        model.c2 = Constraint(expr=1e-2 * model.v1 == 1e-8 * model.v3)
         model.c3 = Constraint(expr=1e8 * model.v1 + 1e10 * model.v2 == 1e-6 * model.v3)
 
         dt = DiagnosticsToolbox(model=model)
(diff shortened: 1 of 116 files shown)


Robbybp:
Not sure why all Andrew's commits from #1375 are showing up here...

andrewlee94 (approved):
Looks good. Thanks for doing this @Robbybp

Robbybp:
Looks like this is breaking tests with new numerical warnings. I'll have to look into this.

andrewlee94:
@Robbybp Taking a quick look at the CSTR test, it is because the toolbox is now detecting near parallel components (outlet temperature and reaction rate in the case of the CSTR). I suspect some of these would fall under the category of "expected" (given some of the test models are very simple).

I did a quick check of setting the default tolerance for near-parallel components to 1e-6 which brought the number of failures down to around 14.

Robbybp:
> I did a quick check of setting the default tolerance for near-parallel components to 1e-6 which brought the number of failures down to around 14.

Tightening the tolerance to 1e-8 seems to help even more. This is backwards-incompatible, but I'm fine with tightening the tolerance here, as I think 1e-8 as a "distance from singularity" tolerance is more consistent with our other tolerances for large/small coefficients and residuals.

andrewlee94:
@Robbybp That sounds fine to me too - I am not sure where the value of 1e-4 came from anyway (it was probably arbitrary). I don't think it is a huge issue for backward compatibility either, but we will need to update the relevant test cases to be "more parallel" (which should just mean changing the coefficient in the duplicate constraint).

idaes/core/util/tests/test_model_diagnostics.py
```diff
@@ -975,10 +975,10 @@ def test_display_near_parallel_variables(self):
         model.v3 = Var()
         model.v4 = Var()
 
-        model.c1 = Constraint(expr=model.v1 == model.v2 - 0.99999 * model.v4)
-        model.c2 = Constraint(expr=model.v1 + 1.00001 * model.v4 == 1e-8 * model.v3)
+        model.c1 = Constraint(expr=1e-8 * model.v1 == 1e-8 * model.v2 - 1e-8 * model.v4)
+        model.c2 = Constraint(expr=1e-8 * model.v1 + 1e-8 * model.v4 == model.v3)
         model.c3 = Constraint(
-            expr=1e8 * (model.v1 + model.v4) + 1e10 * model.v2 == 1e-6 * model.v3
+            expr=1e3 * (model.v1 + model.v4) + 1e3 * model.v2 == model.v3
```
Robbybp, line 981: This test was pretty fragile. The coefficients on `v3` are small relative to those on other variables, so its dot product is vacuously parallel with anything. This didn't show up previously because we were relying on the absolute tolerance of 1e-4 to give it an empty sparsity pattern. Maybe this raises the question of what to do when one of the vectors we are testing is zero. I've removed the coefficients on `v3` so that this isn't an issue for now.

Also, the 1e-8 ratio between small and large coefficients for `v1` and `v2` lead to relative "distances-from-parallel" that are very close to our new tolerance of 1e-8, so I've adjusted some coefficients to give us a little more buffer.

idaes/core/util/tests/test_model_diagnostics.py
```diff
@@ -1433,8 +1436,8 @@ def test_report_numerical_issues_jacobian(self):
         model.v2 = Var(initialize=0)
         model.v3 = Var(initialize=0)
 
-        model.c1 = Constraint(expr=model.v1 == model.v2)
-        model.c2 = Constraint(expr=model.v1 == 1e-8 * model.v3)
+        model.c1 = Constraint(expr=1e-2 * model.v1 == model.v2)
+        model.c2 = Constraint(expr=1e-2 * model.v1 == 1e-8 * model.v3)
         model.c3 = Constraint(expr=1e8 * model.v1 + 1e10 * model.v2 == 1e-6 * model.v3)
```
Robbybp, line 1441: Similarly, I updated coefficients on `v1` to give some buffer around the relative tolerance. The small coefficients on `v3` are also causing it to appear as "parallel to" to the other variables, which again raises the question of how to handle zero vectors.

dallan-keylogic:
If we want to use a (in my opinion) better method for calculating parallel constraints, #1395 needs to be merged before this.

idaes/core/util/tests/test_model_diagnostics.py
dallan-keylogic: This is really because the relative tolerance should be comparing to is `diff <= tolerance * unorm *vnorm`, not `diff <= tolerance * max(unorm, vnorm)`. The absolute tolerance probably can get phased out entirely.

Robbybp: Maybe `unorm * vnorm` is the right denominator for the relative tolerance. I would still like to have an absolute tolerance, though, as I wouldn't want `(1e-9, 2e-9)` to be considered parallel to `(1, 2)`. (Unless we want to consider a zero vector as "parallel to everything", which I think is debatable.)

dallan-keylogic: I support filtering out rows/columns with norms that are effectively zero, because otherwise they'll just spam the log as being parallel to every other row. No need to use an absolute tolerance for the inner product, though, we can just choose to skip displaying the row/column if one of the norms is too small.

In the diagnostics toolbox, the tolerance for this filtering should be the `CONFIG.jacobian_small_value_warning`, the same tolerance as used in `display_variables_with_extreme_jacobians` and `display_constraints_with_extreme_jacobians` to display a warning message. That way it's displayed (once) to the user there.

PR #1393, opened 2024-04-09
Fixing spurious DoF in Gibbs reactor model

### LINKED ISSUE #1392: Handling inert species in Gibbs Reactor
The `inert_species_balance` function in gibbs reactor does not create the balance for all the inert species, but only for the ones that are linearly independent with element balance as per the logic in the function. This increases the degrees of freedom as the variables are created for all the species including inerts but there are no constraints addressing the linearly independent with element balance. 

The previous version of gibbs reactor use to create the constraints for all the inert species commit id - 119656fcfc9d249d844f680b14c74da991e6f9dd 

To give an example of this, look at the below example

```
from idaes.models.unit_models import GibbsReactor
import pyomo.environ as pyo
from idaes.models.properties.modular_properties.base.generic_property import (
    GenericParameterBlock,
)
from idaes.models.properties.modular_properties.base.generic_reaction import (
    GenericReactionParameterBlock,
)
from idaes.models_extra.power_generation.properties.natural_gas_PR import (
    get_prop,
    get_rxn,
)

m=pyo.ConcreteModel()
m.fs=FlowsheetBlock(dynamic=False)
NG_config = get_prop(
        components=[
            "H2",
            "CO",
            "H2O",
            "CO2",
            "CH4",
            "C2H6",
            "C3H8",
            "C4H10",
            "N2",
            "O2",
            "Ar",
        ]
    )
m.fs.NG_props = GenericParameterBlock(**NG_config)
m.fs.gr=GibbsReactor(has_heat_transfer=True,
        has_pressure_change=True,
        inert_species=["N2", "Ar", "O2"],
        property_package=m.fs.NG_props)

### This is the inert_species_balance function
inert_species=["N2", "Ar", "O2"]
for j in inert_species:
    cv = m.fs.gr.control_volume
    e_comp = cv.properties_out[0].config.parameters.element_comp

    # Check for linear dependence with element balances
    # If an inert species is the only source of element e,
    # the inert species balance would be linearly dependent on the
    # element balance for e.
    dependent = True

    if len(m.fs.gr.control_volume.properties_in.phase_list) > 1:
        # Multiple phases avoid linear dependency
        dependent = False
    else:
        for e in m.fs.gr.config.property_package.element_list:
            if e_comp[j][e] == 0:
                # Element e not in component j, no effect
                continue
            else:
                for i in m.fs.gr.control_volume.properties_in.component_list:
                    if i == j:
                        continue
                    else:
                        # If comp j shares element e with comp i
                        # cannot be linearly dependent
                        if e_comp[i][e] != 0:
                            print(j,e)
                            dependent = False
    print(j,dependent)
    if (
        not dependent
        and ('Vap', j) in m.fs.gr.control_volume.properties_in.phase_component_set
    ):
        print(j)
        # return 0 == (
        #     cv.properties_in[t].get_material_flow_terms(p, j)
        #     - cv.properties_out[t].get_material_flow_terms(p, j)
        # )
    else:
        print(j)
        # return Constraint.Skip
```

This gives shows that the model does not create constraints for N2 and Ar as they are linearly independent, but creates constraints for O2 as it has O. This makes the dof to increase by 2 of the flowsheet for each Gibbs Reactor.

andrewlee94:
## Summary/Motivation:
The current implementation of the Gibbs reactor model includes all elements in the Lagrange multipliers expressions, even if the elemental composition for a given component is 0. This leads to spurious degrees of freedom where the variable appears to be in active constraints but is multiplied y zero (and thus filtered out in the Pyomo solver writers).

- Add logic to catch these cases and exclude them from the expression.
-

### Legal Acknowledgement

### DIFF
--- a/idaes/models/unit_models/gibbs_reactor.py
+++ b/idaes/models/unit_models/gibbs_reactor.py
@@ -249,6 +249,10 @@ see property package for documentation.}""",
                         j
                     ][e]
                     for e in b.control_volume.config.property_package.element_list
+                    if b.control_volume.properties_out[
+                        t
+                    ].config.parameters.element_comp[j][e]
+                    != 0
                 )
             )
 
--- a/idaes/models/unit_models/gibbs_reactor.py
+++ b/idaes/models/unit_models/gibbs_reactor.py
@@ -14,7 +14,7 @@
 Standard IDAES Gibbs reactor model.
 """
 # Import Pyomo libraries
-from pyomo.environ import Constraint, Param, Reals, Reference, Var
+from pyomo.environ import Constraint, Param, Reals, Reference, Set, Var
 from pyomo.common.config import ConfigBlock, ConfigValue, In, ListOf, Bool
 
 # Import IDAES cores
@@ -212,19 +212,33 @@ see property package for documentation.}""",
 
         # Add performance equations
         # Add Lagrangian multiplier variables
+        # Only need multipliers for species which participate in reactions
+        l_set = []
+        for e in self.control_volume.config.property_package.element_list:
+            skip = True
+            for j in self.control_volume.properties_in.component_list:
+                if j in self.config.inert_species:
+                    continue
+                elif self.control_volume.properties_out.params.element_comp[j][e] != 0:
+                    skip = False
+
+            if not skip:
+                l_set.append(e)
+        self.lagrange_set = Set(initialize=l_set)
+
         e_units = self.control_volume.config.property_package.get_metadata().get_derived_units(
             "energy_mole"
         )
         self.lagrange_mult = Var(
             self.flowsheet().time,
-            self.control_volume.config.property_package.element_list,
+            self.lagrange_set,
             domain=Reals,
             initialize=100,
             doc="Lagrangian multipliers",
             units=e_units,
         )
 
-        # TODO : Remove this once sacling is properly implemented
+        # TODO : Remove this once scaling is properly implemented
         self.gibbs_scaling = Param(default=1, mutable=True)
 
         # Use Lagrangian multiple method to derive equations for Out_Fi
@@ -248,11 +262,7 @@ see property package for documentation.}""",
                     * b.control_volume.properties_out[t].config.parameters.element_comp[
                         j
                     ][e]
-                    for e in b.control_volume.config.property_package.element_list
-                    if b.control_volume.properties_out[
-                        t
-                    ].config.parameters.element_comp[j][e]
-                    != 0
+                    for e in self.lagrange_set
                 )
             )
 
--- a/idaes/models/unit_models/gibbs_reactor.py
+++ b/idaes/models/unit_models/gibbs_reactor.py
@@ -259,9 +259,7 @@ see property package for documentation.}""",
                 b.control_volume.properties_out[t].gibbs_mol_phase_comp[p, j]
                 + sum(
                     b.lagrange_mult[t, e]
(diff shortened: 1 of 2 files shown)


idaes/models/unit_models/gibbs_reactor.py
```diff
@@ -249,6 +249,10 @@ def gibbs_minimization(b, t, p, j):
                         j
                     ][e]
                     for e in b.control_volume.config.property_package.element_list
+                    if b.control_volume.properties_out[
+                        t
+                    ].config.parameters.element_comp[j][e]
+                    != 0
```
bpaul4, line 255: Should we use 0, or a very small value? Does the Pyomo solver writer specifically check for "exactly 0" values and filter those variables out?

andrewlee94: This is only for exact zero (and a fixed value at that). The issue is that the solver writer is filtering these out for us, but when we use the `degrees_of_freedom` function we still see these (and hence see spurious DoF).

his is to catch cases where the multiplier is a constant of 0 and filter them out of the expression.

andrewlee94: I also just worked out a better solution to remove the unneeded multipliers entirely (so that they don;t end up as unused variables).

idaes/models/unit_models/gibbs_reactor.py
```diff
@@ -212,19 +212,33 @@ def build(self):
...
         # Add performance equations
         # Add Lagrangian multiplier variables
+        # Only need multipliers for species which participate in reactions
+        l_set = []
+        for e in self.control_volume.config.property_package.element_list:
+            skip = True
+            for j in self.control_volume.properties_in.component_list:
+                if j in self.config.inert_species:
+                    continue
+                elif self.control_volume.properties_out.params.element_comp[j][e] != 0:
```
JavalVyas2000, line 222: AttributeError: '_IndexedGenericStateBlock' object has no attribute 'params' is thrown here. I believe a time index is missing, properties_out[0.0],params should be there instead of properties_out.params

idaes/models/unit_models/gibbs_reactor.py
```diff
@@ -248,7 +262,7 @@ def gibbs_minimization(b, t, p, j):
                     * b.control_volume.properties_out[t].config.parameters.element_comp[
```
JavalVyas2000, line 262: Would you recommend changing the code here to `properties_out[t].params` instead of `properties_out[t].config.parameters`. This is for the uniformity.

JavalVyas2000 (changes requested):

idaes/models/unit_models/gibbs_reactor.py
andrewlee94: This works for me, and I have created a test specifically for your use case to be sure.

Could you check what version of IDAES you are using? I fixed this issue a while ago and StateBlocks now support `.params`.

andrewlee94: You need to update your version of IDAES as you are at least two full releases behind.

dallan-keylogic (approved):
Looks good.

idaes/models/unit_models/gibbs_reactor.py
JavalVyas2000: This works fine with the latest version. Looks good.

JavalVyas2000 (approved):

PR #1396, opened 2024-04-12
Adding attributes to hold units used by MSContactor

andrewlee94:
## Summary/Motivation:
For unit models deriving from the MSContactor, or using it a a building block, it is often necessary to know the units of measurement used for the unit model variables. This PR gets the MSContactor to store these as attributes so that deriving models can make use of them.

- Adds two new attributes to the MSContactor to store the flow basis and units of measurement.
-

### Legal Acknowledgement

### DIFF
--- a/idaes/models/unit_models/mscontactor.py
+++ b/idaes/models/unit_models/mscontactor.py
@@ -385,17 +385,21 @@ class MSContactorData(UnitModelBlockData):
         # Call UnitModel.build
         super().build()
 
+        # Placeholders for things we will get from first StateBlock
+        self.flow_basis = None
+        self.uom = None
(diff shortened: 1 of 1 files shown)


bpaul4:
@andrewlee94 it looks like many tests directly call `_build_state_blocks`, and since `self.flow_basis = None` and `self.uom = None` are set in `build` these attributes don't exist in the tests. Should the tests call `build` instead?

andrewlee94:
@bpaul4 No - I just didn't think I'd need to update the tests.

bpaul4 (approved):

agarciadiego (approved):

PR #1397, opened 2024-04-17
Add new Complementarity formualtion for VLE with cubic EoSs

andrewlee94:
## Summary/Motivation:
This PR adds an implementation of the improved Smooth VLE formulation originally proposed in #977. This implementation preserves the old implementation for backward compatibility.

- Infrastructure work to support new formulation, and address some pylint issues in modular properties code.
- Add new class `SmoothVLE2` to implement improved formulation

### Legal Acknowledgement


idaes/models/properties/modular_properties/eos/ceos.py
```diff
@@ -1356,26 +1382,79 @@ def a(k):
...
+        expression for fw
+
+    """
     return 0.48 + 1.574 * cobj.omega - 0.176 * cobj.omega**2
 
 
 def func_alpha_soave(T, fw, cobj):
+    """
+    alpha function for SRK EoS.
+
```
dallan-keylogic, line 1415: I believe this alpha function is used by both SRK and PR.

idaes/models/properties/modular_properties/eos/ceos.py
```diff
@@ -1356,26 +1382,79 @@ def a(k):
...
+    """
     Tc = cobj.temperature_crit
     Tr = T / Tc
     return (1 + fw * (1 - sqrt(Tr))) ** 2
 
 
 def func_dalpha_dT_soave(T, fw, cobj):
+    """
+    Function to get first derivative of alpha for SRK EoS.
```
dallan-keylogic, line 1431: Same thing here.

idaes/models/properties/modular_properties/eos/ceos.py
```diff
@@ -1412,8 +1497,14 @@ def rule_am_crit_default(m, cname, a_crit):
...
 def rule_bm_default(m, b, p):
+    """
+    Standard mixing rule for b term
+    """
     return sum(m.mole_frac_phase_comp[p, i] * b[i] for i in m.components_in_phase(p))
 
 
 def rule_bm_crit_default(m, b):
+    """
+    Standard mixing rule for b term at critical point
```
dallan-keylogic, line 1508: "at critical point" -> "evaluated at critical point"

idaes/models/properties/modular_properties/examples/tests/test_BT_PR_SmoothVLE2.py
```diff
@@ -0,0 +1,891 @@
...
+        m.fs.state[1].temperature.setub(600)
+
+        for logP in [9.5, 10, 10.5, 11, 11.5, 12]:
+            m.fs.obj.deactivate()
+
+            m.fs.state[1].flow_mol.fix(100)
+            m.fs.state[1].mole_frac_comp["benzene"].fix(0.5)
+            m.fs.state[1].mole_frac_comp["toluene"].fix(0.5)
+            m.fs.state[1].temperature.fix(300)
+            m.fs.state[1].pressure.fix(10 ** (0.5 * logP))
```
dallan-keylogic, line 219: What's the point of the 0.5 in the exponent here?

idaes/models/properties/modular_properties/examples/tests/test_BT_PR_SmoothVLE2.py
```diff
@@ -0,0 +1,891 @@
...
+        assert (
+            m.fs.state[1].scaling_factor[m.fs.state[1].mole_frac_comp["toluene"]]
+            == 1000
+        )
+        assert (
+            m.fs.state[1].scaling_factor[
+                m.fs.state[1].mole_frac_phase_comp["Liq", "benzene"]
+            ]
+            == 1000
+        )
```
dallan-keylogic, line 836: For the record, I don't think 1000 is a great value for scaling factors of mole fractions. Testing mole fractions like this makes it a pain to change later. Can't we do something like test  Jacobian rows, columns, and condition number?

idaes/models/properties/modular_properties/phase_equil/smooth_VLE_2.py
```diff
@@ -0,0 +1,278 @@
...
+            henry_comps,
+            l_only_comps,
+            v_only_comps,
+        ) = identify_VL_component_list(b, phase_pair)
+
+        if l_phase is None or v_phase is None:
+            raise ConfigurationError(
+                f"{b.params.name} Generic Property Package phase pair {phase_pair[0]}-{phase_pair[1]} "
+                "was set to use Smooth VLE formulation, however this is not a vapor-liquid pair."
+            )
```
dallan-keylogic, line 61: This error should probably be thrown by `identify_VL_component_list`.

idaes/models/properties/modular_properties/phase_equil/smooth_VLE_2.py
```diff
@@ -0,0 +1,278 @@
...
+                liquid_phase = phase_pair[1]
+            else:
+                vapor_phase = phase_pair[1]
+                liquid_phase = phase_pair[0]
+            return (
+                b._teq[phase_pair]
+                - b.temperature
+                - s[vapor_phase] * t_units
+                + s[liquid_phase] * t_units
+                == 0
```
dallan-keylogic, line 90: Why make the slack variable `s` dimensionless when we're multiplying it by `t_units` here? That also implies it should be scaled.

idaes/models/properties/modular_properties/phase_equil/smooth_VLE_2.py
```diff
@@ -0,0 +1,278 @@
...
+
+        b.add_component("_tbar_constraint" + suffix, Constraint(rule=rule_teq))
+
+        eps = Param(
+            default=1e-04,
+            mutable=True,
+            doc="Smoothing parameter for complementarities",
+            units=f_units,
+        )
+        b.add_component("eps" + suffix, eps)
```
dallan-keylogic, line 101: Does `eps` need to be adjusted by the user?

idaes/models/properties/modular_properties/phase_equil/smooth_VLE_2.py
```diff
@@ -0,0 +1,278 @@
...
+            bounds=(0, None),
+            doc="Slack variable for cubic second derivative for phase p",
+            units=pyunits.dimensionless,
+        )
+        b.add_component("gn" + suffix, gn)
+
+        def rule_temperature_slack_complementarity(b, p):
+            flow_phase = b.flow_mol_phase[p]
+
+            return smooth_min(s[p] * f_units, flow_phase, eps) == 0
```
dallan-keylogic, line 124: Okay, we're treating `s` as having different units depending on the context. That's not ideal for scaling.

idaes/models/properties/modular_properties/phase_equil/smooth_VLE_2.py
```diff
@@ -0,0 +1,278 @@
...
+        T_units = blk.params.get_metadata().default_units.TEMPERATURE
+
+        liquid_phase, _, raoult_comps, henry_comps, _, _ = identify_VL_component_list(
+            blk, pp
+        )
+
+        Tbub = estimate_Tbub(blk, T_units, raoult_comps, henry_comps, liquid_phase)
+        Tdew = estimate_Tbub(blk, T_units, raoult_comps, henry_comps, liquid_phase)
+
+        assert False
```
dallan-keylogic, line 187: Debugging survival into production.

dallan-keylogic (commented):
Here are some comments, it seems like the full PR isn't finished yet.

idaes/models/properties/modular_properties/base/utility.py
```diff
@@ -288,3 +301,293 @@ def get_concentration_term(blk, r_idx, log=False):
...
+                .dT_expression(blk, liquid_phase, j, Tbub0 * T_units)
+                for j in henry_comps
+            )
+        )
+
+        # Limit temperature step to avoid excessive overshoot
+        if f / df < -50:
+            Tbub1 = Tbub0 + 50
+        elif f / df > 50:
+            Tbub1 = Tbub0 - 50
```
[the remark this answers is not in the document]
dallan-keylogic: Probably too much work, though.

idaes/models/properties/modular_properties/base/utility.py
```diff
@@ -288,3 +301,293 @@ def get_concentration_term(blk, r_idx, log=False):
...
+    Returns:
+        Lists of component names for:
+            * liquid Phase object
+            * vapor Phase object
+            * components using Raoult's Law
+            * components using Henry's Law
+            * liquid only components,
+            * vapor only components
+
+    """
```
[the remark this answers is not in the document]
andrewlee94: This is only for VLE systems, so is not expected to work in those cases. I have added an error for cases where a non-VL pair is passed as the argument.

idaes/models/properties/modular_properties/base/utility.py
```diff
@@ -288,3 +301,293 @@ def get_concentration_term(blk, r_idx, log=False):
...
+    # Use lowest component temperature_crit as starting point
+    # Starting high and moving down generally works better,
+    # as it under-predicts next step due to exponential form of
+    # Psat.
+    # Subtract 1 to avoid potential singularities at Tcrit
+    Tbub0 = (
+        min(blk.params.get_component(j).temperature_crit.value for j in raoult_comps)
+        - 1
```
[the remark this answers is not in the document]
andrewlee94: In this case, `1` is just an arbitrary value to move away from the exact critical value as some correlations are singular if `T=Tc[j]`.

idaes/models/properties/modular_properties/base/utility.py
```diff
@@ -288,3 +301,293 @@ def get_concentration_term(blk, r_idx, log=False):
...
+        Estimated dew point pressure as a float.
+
+    """
+    return value(
+        1
+        / (
+            sum(blk.mole_frac_comp[j] / blk.pressure_sat_comp[j] for j in raoult_comps)
+            + sum(
+                blk.mole_frac_comp[j] / blk.henry[liquid_phase, j] for j in henry_comps
+            )
```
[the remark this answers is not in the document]
andrewlee94: Maybe - but that would be a definite case of user stupidity.

idaes/models/properties/modular_properties/examples/tests/test_BT_PR_SmoothVLE2.py
andrewlee94: To get a half step size, as `range` does not accept non-integer arguments.

idaes/models/properties/modular_properties/examples/tests/test_BT_PR_SmoothVLE2.py
andrewlee94: Agreed, but for now I am focused on getting this working (plus the new tools will supersede this before too long).

idaes/models/properties/modular_properties/phase_equil/smooth_VLE_2.py
andrewlee94: A catch has been added to the utility method, but this can be retained as well.

idaes/models/properties/modular_properties/phase_equil/smooth_VLE_2.py
andrewlee94: Yes, users should adjust `eps` as necessary and this will be in the docs (the old formulation was the same).

idaes/models/properties/modular_properties/phase_equil/smooth_VLE_2.py
andrewlee94: Agreed - if you have any ideas on how to handle this better it would be appreciated. For now I have just focused on getting the code running.

andrewlee94:
@viiibhav @jghouse88 Not sure if either of you still receive these, but input and review on this PR would be welcome.

idaes/models/properties/modular_properties/base/utility.py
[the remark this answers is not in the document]
dallan-keylogic: Yeah, but in edge cases we could end up with a negative temperature (for example, if the user decided to work with temperature reduced by `Tc`).

idaes/models/properties/modular_properties/examples/tests/test_BT_PR_SmoothVLE2.py
dallan-keylogic: Then it's not really `logP` anymore, but `2*logP`, which isn't a great variable name. You could get around this issue with `numpy.linspace` or `numpy.logspace`, but that's probably not worth the effort changing.

idaes/models/properties/modular_properties/phase_equil/smooth_VLE_2.py
dallan-keylogic: My instinct here is to just give `s[p]` units of temperature per flowrate, then scale this equation like temperature. However, it needs to be scaled with a "temperature difference" scaling factor, not an "absolute temperature" scaling factor.

jghouse88:
> @viiibhav @jghouse88 Not sure if either of you still receive these, but input and review on this PR would be welcome.

Will take a look this week if that works. The best test would be to see if the paper examples work with this pr and see if we get similar results.

andrewlee94:
@jghouse88 I have tested this with the existing tests in IDAES, swapping the new formulation for the old one. All but two tests pass with the new formulation, but the two that fail appear to need further investigation - the temperature sweep test is failing to converge to the correct solution (it gets stuck somewhere).

idaes/models/properties/modular_properties/phase_equil/smooth_VLE_2.py
andrewlee94: Based on some testing, scaling of s, g and epsilon depends on the larger of T and flow, thus it is hard to know what to do with these. I am inclined to leave these as unit-less for that reason, and any scaling routine can handle the necessary logic behind the scenes (as I doubt users will care enough to understand what is happening).

idaes/models/properties/modular_properties/phase_equil/smooth_VLE_2.py
```diff
@@ -0,0 +1,339 @@
...
+        vl_phase_set = Set(initialize=[phase_pair[0], phase_pair[1]])
+        b.add_component("_vle_set" + suffix, vl_phase_set)
+
+        s = Var(
+            vl_phase_set,
+            initialize=EPS_INIT,
+            bounds=(0, None),
+            doc="Slack variable for equilibrium temperature",
+            units=t_units,
+        )
```
dallan-keylogic, line 115: Do we have any sort of scaling for these slack variables, or is the intention to let the autoscaler do it? This is a case where the results from the autoscaler is going to be highly dependent on initialization, and since the slack variable will frequently be near zero, you may not get any scaling at all.

idaes/models/properties/modular_properties/phase_equil/smooth_VLE_2.py
```diff
@@ -0,0 +1,339 @@
...
+        Method to calculate scaling factors for phase equilibrium
+        """
+        suffix = "_" + phase_pair[0] + "_" + phase_pair[1]
+        sf_T = iscale.get_scaling_factor(b.temperature, default=1, warning=True)
+
+        try:
+            teq_cons = getattr(b, "_teq_constraint" + suffix)
+            # pylint: disable-next=protected-access
+            iscale.set_scaling_factor(b._teq[phase_pair], sf_T)
+            iscale.constraint_scaling_transform(teq_cons, sf_T, overwrite=False)
```
dallan-keylogic, line 227: I guess we're not scaling the slacks then. If we're giving them temperature units, we might as well scale them like temperature. It would be a good first guess, at any rate, and I think the autoscalers will faceplant.

dallan-keylogic (changes requested):

docs/explanations/components/property_package/general/pe/smooth_vle2.rst
```diff
@@ -0,0 +1,67 @@
...
+
+.. math:: \min(a, b) =  0.5{\left[a + b - \sqrt{(a-b)^2 + \epsilon^2}\right]}
+
+Each complementarity requires a smoothing parameter, named :math:`\epsilon_T` and :math:`\epsilon_Z` for the temperature and cubic root constraints respectively. Within the IDAES model, these are rendered as ``eps_t_phase1_phase2`` and ``eps_z_phase1_phase2``, where ``phase1`` and ``phase2`` are the names assigned to the liquid and vapor phases in the property package (order will depend on the order these are declared).
+
+The tractability of the VLE problem depends heavily upon the values chosen for :math:`\epsilon_T` and :math:`\epsilon_Z`, with larger values resulting in smoother transitions at the phase boundaries (and thus increased tractability) at the expense of decreased accuracy near these points. It is recommended that users employ a 2-stage approach to solving these problems, starting with a larger value of :math:`\epsilon_T` and :math:`\epsilon_Z` initially to determine which region the solution lies in, followed by a second solve using smaller values to refine the solution.
```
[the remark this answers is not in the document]
andrewlee94: They are local, as they potentially need to vary unit-to-unit. Their names are show in the docs.

idaes/models/properties/modular_properties/base/generic_property.py
```diff
@@ -2424,9 +2097,13 @@ def build(self):
...
 
             t_units = self.params.get_metadata().default_units.TEMPERATURE
+            if self.temperature.value is not None:
+                t_value = value(self.temperature)
+            else:
+                t_value = None
             self._teq = Var(
                 self.params._pe_pairs,
-                initialize=value(self.temperature),
+                initialize=t_value,
```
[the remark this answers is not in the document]
andrewlee94: This is a catch for cases where T might not be a state variable. I do not expect it to happen often, but it can occur so we need to catch it.

idaes/models/properties/modular_properties/examples/tests/test_BT_PR.py
```diff
@@ -80,23 +85,34 @@ def test_T_sweep(self, m):
...
             m.fs.state[1].mole_frac_comp["benzene"].fix(0.5)
             m.fs.state[1].mole_frac_comp["toluene"].fix(0.5)
             m.fs.state[1].temperature.fix(300)
-            m.fs.state[1].pressure.fix(10 ** (0.5 * logP))
+            m.fs.state[1].pressure.fix(P)
+
+            # For optimization sweep, use a large eps to avoid getting stuck at
+            # bubble and dew points
+            m.fs.state[1].eps_t_Vap_Liq.set_value(10)
+            m.fs.state[1].eps_z_Vap_Liq.set_value(10)
```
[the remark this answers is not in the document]
andrewlee94: Whilst there is potentially more care needed with the eps values, it is not much more (and the fact that we did not need to tweak the eps values for the older implementation is likely just as much luck as anything). However, I would argue "Robust" is no the correct word in this case; both methods are sensitive to the value of eps chosen, but once a good value has been identified they will generally be robust to changes.

idaes/models/properties/modular_properties/phase_equil/smooth_VLE_2.py
dallan-keylogic: Very well, but I do expect this case will need heuristic scaling.

dallan-keylogic (approved):
LGTM

idaes/models/properties/modular_properties/base/generic_property.py
```diff
@@ -2424,9 +2097,13 @@ def build(self):
         ):
 
             t_units = self.params.get_metadata().default_units.TEMPERATURE
+            if self.temperature.value is not None:
+                t_value = value(self.temperature)
+            else:
+                t_value = None
```
bpaul4, line 2103: If this is None, does the model break or is there a catch for this case?

idaes/models/properties/modular_properties/base/tests/test_generic_property.py
```diff
@@ -1972,6 +1975,57 @@ def build_critical_properties(b, *args, **kwargs):
...
+    "state_bounds": {
+        "flow_mol": (0, 100, 1000, pyunits.mol / pyunits.s),
+        "temperature": (273.15, 300, 500, pyunits.K),
+        "pressure": (5e4, 1e5, 1e6, pyunits.Pa),
+    },
+    "pressure_ref": (101325, pyunits.Pa),
+    "temperature_ref": (298.15, pyunits.K),
+    "parameter_data": {
+        "PR_kappa": {
+            ("foo", "bar"): 0.000,
```
bpaul4, line 2023: Why "foo" and "bar"?

idaes/models/properties/modular_properties/base/tests/test_utility.py
```diff
@@ -559,3 +581,253 @@ def test_inherent_partial_pressure(self, frame2):
...
+    class DummyParameterData(GenericParameterData):
+        def configure(self):
+            self.configured = True
+
+    @pytest.fixture()
+    def frame(self):
+        m = ConcreteModel()
+        m.params = DummyParameterBlock(
+            components={
+                "a": {},
```
bpaul4, line 613: Are these supposed to be empty?

bpaul4 (approved):
Looks good, just a couple minor questions.

idaes/models/properties/modular_properties/base/generic_property.py
andrewlee94: This just affects whether we can get initial guesses for Teq. Whether the model will solve or not is a separate issue at this point; the goal here is for the code to work without encountering a terminal error.

idaes/models/properties/modular_properties/base/tests/test_generic_property.py
andrewlee94: [Standard Python terms for testing attributes.](https://www.altcademy.com/explain/foo-bar-baz)

idaes/models/properties/modular_properties/base/tests/test_utility.py
andrewlee94: Yes - this is a unit test. We need to define the components so the code runs but don't care about any actual properties.

PR #1400, opened 2024-04-24
bug fix for issue #1364: catch AttributeError in set_scaling_from_default

### LINKED ISSUE #1364: Incompatibility between `set_scaling_from_default()` and generic reaction parameter data
While experimenting with using the `set_scaling_from_default()` method to add scaling to a model that includes a reactor, I observed the following AttributeError:

```
AttributeError: 'ScalarBlock' object has no attribute 'get_default_scaling'
```

I ran `set_scaling_from_default` on my entire model, so it descended into every block in turn. Tracing the error back to its source, it appears that the problem is in how the `GenericReactionParameterData` class instantiates the individual reaction blocks, named "reaction_{idx}".
https://github.com/IDAES/idaes-pse/blob/eea1209077b75f7d940d8958362e69d4650c079d/idaes/models/properties/modular_properties/base/generic_reaction.py#L336

These reaction blocks are created as generic Pyomo Block objects, which does not have the `get_default_scaling` method defined. Instead, they should be created as ProcessBaseBlock objects, or something derived from that.

Looks like an easy 3-line fix: 2 lines to update the imports and 1 line to change the pyomo Block to an idaes ProcessBaseBlock. Happy to submit a PR if needed.

jasonmbray-p66:
## Fixes
#1364 Incompatibility between set_scaling_from_default() and generic reaction parameter data

## Summary/Motivation:
Currently set_scaling_from_default assumes all blocks have a `get_default_scaling()` method, but this is not guaranteed.

- Add try/except around call to `get_default_scaling()`
- Log warning

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/scaling.py
+++ b/idaes/core/util/scaling.py
@@ -1179,9 +1179,15 @@ def set_scaling_from_default(
                 return
 
         parent = component.parent_block()
-        dsf = parent.get_default_scaling(
-            component.parent_component().local_name, index=component.index()
-        )
+        try:
+            dsf = parent.get_default_scaling(
+                component.parent_component().local_name, index=component.index()
+            )
+        except AttributeError:
+            dsf = None
+            _log.warning(
+                f"{component.name} block missing 'get_default_scaling()' method, no scaling factor assigned."
+            )
 
         if dsf is not None:
             set_scaling_factor(component, dsf, overwrite=overwrite)


andrewlee94:
@jasonmbray-p66 Thank you for taking the time to fix this for us. Contributions from users are always appreciated.

radhakrishnatg (approved):

Issue #1369, opened 2024-03-11
Resolve Keras File Format Issues


bpaul4:
TensorFlow updated to version 2.16.1 on March 8, 2024. Python 3.9+ tests that install the latest package now throw errors similar to the message below:

```
E           ValueError: File format not supported: filepath=D:\a\idaes-pse\idaes-pse\idaes\core\surrogate\plotting\tests\keras_surrogate. Keras 3 only supports V3 `.keras` files and legacy H5 format files (`.h5` extension). Note that the legacy SavedModel format is not supported by `load_model()` in Keras 3. In order to reload a TensorFlow SavedModel as an inference-only layer in Keras 3, use `keras.layers.TFSMLayer(D:\a\idaes-pse\idaes-pse\idaes\core\surrogate\plotting\tests\keras_surrogate, call_endpoint='serving_default')` (note that your `call_endpoint` might have a different name).
```

Tests and code that use Keras should be updated accordingly.

bpaul4:
@rundxdi I briefly investigated the issue and it will require updating the KerasSurrogate "save_to_folder" and "load_from_folder" methods with serialization to a ".keras" format instead, and related front-end changes in tests that use the methods. Please work on this when you can, as it is a breaking issue in IDAES at the moment. Thank you.

rundxdi:
> @rundxdi I briefly investigated the issue and it will require updating the KerasSurrogate "save_to_folder" and "load_from_folder" methods with serialization to a ".keras" format instead, and related front-end changes in tests that use the methods. Please work on this when you can, as it is a breaking issue in IDAES at the moment. Thank you.

Thanks for identifying the needed updates.  I'll take a look tomorrow and work on updates this week.

ksbeattie:
@rundxdi any estimate on this?

ksbeattie:
@rundxdi any news on this?  Maybe join the next dev call to update us?

rundxdi:
@ksbeattie I've opened PR #1401 to address this.  It should fix keras_surrogate.py and test_keras_surrogate.py.  However, I'm having trouble getting test_keras_plotting.py to work with the changes and its existing model.  Working on that currently.

PR #1401, opened 2024-04-25
Keras file format updates

rundxdi:
## Fixes

Addresses file format changes associated with updates to tensorflow and keras mentioned in issue #1369 

## Summary/Motivation:

Tensorflow and keras version updates changed how they save/load files.  This PR updates the keras_surrogate.py file to accommodate those changes.

- Update keras_surrogate.py to save new .keras file format
- Update test_keras_surrogate.py with new versions of saved models and accounting for new save/load methods

### Legal Acknowledgement


idaes/core/surrogate/keras_surrogate.py
```diff
@@ -263,7 +263,7 @@ def save_to_folder(self, keras_folder_name):
               The name of the folder to contain the Keras model and additional
               IDAES metadata
         """
-        self._keras_model.save(keras_folder_name)
+        self._keras_model.save(os.path.join(keras_folder_name, "idaes_keras_model.keras"))
```
andrewlee94, line 266: Should the file name be hard-coded here, or should we make this an input from the user (either optional or required).

bpaul4:
@rundxdi thank you for opening this. Once this is ready to test, it would be good to undo the temporary fix implemented in https://github.com/IDAES/idaes-pse/pull/1373 to ensure the latest version of TensorFlow is used in the tests.

idaes/core/surrogate/keras_surrogate.py
bpaul4: I support letting users pass a string for the file name, so something like

```suggestion
        self._keras_model.save(os.path.join(keras_folder_name, keras_file_name + ".keras"))
```

where `keras_file_name` is an input to the save method. I don't know if we still need to specify the folder name, or if a single string for the entire path is sufficient.

ksbeattie:
@rundxdi any progress on this?

rundxdi:
> @rundxdi any progress on this?

Nothing this week -- still trying to figure out plotting tests.

idaes/core/surrogate/keras_surrogate.py
rundxdi: The folder name is used to save/load some model weights and some idaes specific information separate from the filename.  But I think the file name should be rename-able as you say -- making that change now

rundxdi:
I have two remaining issues:

1. Python 3.8 doesn't like test_keras_surrogate.py.  Tests all produce failure statuses looking like:

`Exception encountered: Unrecognized keyword arguments: ['batch_shape']`

These tests all pass on Python 3.9+.

2. The test_sofc_surrogates.py tests rely on a saved Keras surrogate that cannot be loaded with Keras v3+ (and thus cannot be loaded with Tensorflow 2.16.1).  I've found no obvious way to generate an equivalent surrogate in Keras v3 equivalent surrogate; their suggested workaround doesn't work.

lbianchi-lbl:
>     1. Python 3.8 doesn't like test_keras_surrogate.py.  Tests all produce failure statuses looking like:
> 
> `Exception encountered: Unrecognized keyword arguments: ['batch_shape']`

> 
> These tests all pass on Python 3.9+.

My guess is that this is due to NumPy (and consequently I imagine most of the downstream projects depending on it) having dropped support for Python 3.8 a while back. So, when installing on Python 3.8, an older version of the package (the latest compatible) gets installed, which I assume doesn't support the same arguments as the current version.

We're considering removing support for (i.e. stopping testing with) Python 3.8. In the meantime, you should be able to use `pytest.mark.skipif()` to skip the failing tests if they're being run: https://docs.pytest.org/en/latest/how-to/skipping.html#id1


>     2. The test_sofc_surrogates.py tests rely on a saved Keras surrogate that cannot be loaded with Keras v3+ (and thus cannot be loaded with Tensorflow 2.16.1).  I've found no obvious way to generate an equivalent surrogate in Keras v3 equivalent surrogate; their suggested workaround doesn't work.

With the disclaimer that I don't have any significant hands-on experience using Keras, I'm at least aware of there being limited or no support for cross-version compatibility (i.e. a model serialized with Keras version X won't work when you try to deserialize it using version Y). I'm not sure how much the new (current) serialization format changes things. If this is the case, I guess the solution would be to regenerate the file used in the tests using the current version. I'm not sure if this addresses the original issue, though.

andrewlee94:
As mentioned in the dev call, regarding the test failure in the power generation models you should:

1. Mark the failing tests with `pytest.mark.xfail`
2. In the failing code, add a deprecation warning indicating the tests for this code have started failing and the code will be removed no earlier than August if it is not fixed.
3. Open an issue to fix this and assign it to the code owner(s), and put this on the August release board as a High priority so we track it.

ksbeattie:
@rundxdi we expect to cut the May release next week, so if this PR is to be included, it needs to be merged very soon.

ksbeattie:
@rundxdi, this missed the May release, not on the Aug release board.

bpaul4 (commented):
Thank you @rundxdi for investigating and implementing the fix. The changes to the Keras/OMLT calls and file writing look good.

I have a couple questions. First, the models previously import from a folder containing `saved_model.pb`, `keras_metadata.pb`, and `idaes_info.json` files and a `variables/` folder containing `variables.index` and `variables.data-00000-of-00001` files. Are these still necessary, or can we remove them in favor of the `.keras` files? I've seen in my research into Keras 3 that it now doesn't support SavedModel files and includes a read-only TFSM format instead.

Second, do we want to look into the Python 3.8 failures? Since it isn't occurring in the newer Python environments, it may just be a compatibility issue that's best skipping in that one environment.

bpaul4:
@rundxdi when you have a moment, could you please provide a status update? What is required to finalize this?

rundxdi:
> @rundxdi when you have a moment, could you please provide a status update? What is required to finalize this?

Hi @bpaul4 -- we can safely remove any of the non `.keras` files.  The Python 3.8 tests should be skipped at this point.  

It should be ready to finalize/review for finalization.

bpaul4 (approved):
I think this looks good now, thanks @rundxdi!

lbianchi-lbl:
Before merging this, we should remove Python 3.8 from the CI and see if the tests pass without needing to mark them as `xfail`ing.

bpaul4:
@andrewlee94 it looks like all of your prior review comments have been addressed.

I removed the xfail markers from the keras surrogate tests and everything looks good.

idaes/models_extra/power_generation/properties/tests/test_sofc_surrogates.py
```diff
@@ -53,6 +53,7 @@ def build_rom():
 
 
 @pytest.mark.unit
+@pytest.mark.xfail
```
andrewlee94, line 56: Could you add an inline comment to explain why this is `xfailed`. I know this is mentioned in the deprecation warning in the code, but having comments here will help reinforce that this issue needs to be fixed.

andrewlee94 (commented):
@bpaul4 I was waiting for the Python 3.8 issue to be merged to make sure the tests passed. I have just one minor request.

idaes/models_extra/power_generation/properties/tests/test_sofc_surrogates.py
bpaul4: I'll add the comments now. @AlexNoring is aware of the SOFC surrogate issue and will continue the discussion here: https://github.com/IDAES/idaes-pse/issues/1431

andrewlee94 (approved):

bpaul4:
@andrewlee94 how would you suggest handling the failing example integration tests? The failures are resolved as part of https://github.com/IDAES/examples/pull/128.

andrewlee94:
@bpaul4 I would say we need to get that examples PR merged first.

idaes/core/surrogate/keras_surrogate.py
```diff
@@ -263,7 +263,9 @@ def save_to_folder(self, keras_folder_name):
               The name of the folder to contain the Keras model and additional
               IDAES metadata
         """
-        self._keras_model.save(keras_folder_name)
+        self._keras_model.save(
+            os.path.join(keras_folder_name, keras_model_name + ".keras")
+        )
```
lbianchi-lbl, line 268: Suggestion: since Keras seems to support `pathlib.Path` objects now, this could be rewritten to be slightly more concise/readable (requires replacing `import os.path` with `from pathlib import Path` at the top of the file):

```suggestion
        self._keras_model.save(
            Path(keras_folder_name, keras_model_name).with_suffix(".keras")
        )
```

idaes/core/surrogate/keras_surrogate.py
```diff
@@ -293,7 +295,11 @@ def load_from_folder(cls, keras_folder_name):
 
         Returns: an instance of KerasSurrogate
         """
-        keras_model = keras.models.load_model(keras_folder_name)
+
+        keras_model = keras.models.load_model(
+            os.path.join(keras_folder_name, keras_model_name + ".keras")
+        )
+
```
lbianchi-lbl, line 302: See previous comments for using `pathlib.Path` instead of `os.path`.

idaes/core/surrogate/keras_surrogate.py
```diff
@@ -319,12 +325,11 @@ def save_keras_json_hd5(nn, path, name):
...
+    nn.save_weights(os.path.join(path, "{}.weights.h5".format(name)))
 
 
 def load_keras_json_hd5(path, name):
-    with open(os.path.join(path, "{}.json".format(name)), "r") as json_file:
-        json_model = json_file.read()
-        nn = keras.models.model_from_json(json_model)
-    nn.load_weights(os.path.join(path, "{}.h5".format(name)))
+    nn = keras.models.load_model(os.path.join(path, "{}.keras".format(name)))
+    nn.load_weights(os.path.join(path, "{}.weights.h5".format(name)))
```
lbianchi-lbl, line 334: See comments above for using `pathlib.Path` instead of `os.path`.

lbianchi-lbl (commented):
Mostly minor comments, questions, and suggestions.

setup.py
```diff
@@ -93,7 +93,7 @@ class ExtraDependencies:
     ]
     omlt = [
         "omlt==1.1",  # fix the version for now as package evolves
-        'tensorflow < 2.16.1 ; python_version < "3.12"',
+        "tensorflow",
```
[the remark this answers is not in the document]
bpaul4: I believe that would be appropriate. although removing the tag will automatically search for the latest version. Pinning a minimum version would be a good safety net.

lbianchi-lbl:
There might be another test that needs to be skipped/xfailed as it's currently causing a failure in the "user install" integration checks: 
image

andrewlee94:
@lbianchi-lbl I think we should look into that one a bit more first, and fix it if we can. Being that is in the core code, we should at least know why it fails.

lbianchi-lbl:
> @lbianchi-lbl I think we should look into that one a bit more first, and fix it if we can. Being that is in the core code, we should at least know why it fails.

The fact that it's only failing in the user-mode check, plus the exception raised, seem to point towards the file not being included in the non-developer installation:

image

PR #1413, opened 2024-05-17
Adding USD_2022 to units library as the most recently published value

hunterbarber:
## Fixes
Append USD_YYYY list with USD_2022 value from https://toweringskills.com/financial-analysis/cost-indices/

## Summary/Motivation:
Add more up to date values for USD

- Append USD_YYYY list with USD_2022 value from https://toweringskills.com/financial-analysis/cost-indices/

### Legal Acknowledgement

### DIFF
--- a/idaes/core/base/costing_base.py
+++ b/idaes/core/base/costing_base.py
@@ -92,6 +92,7 @@ def register_idaes_currency_units():
                 "USD_2019 = 500/607.5 * USD_CE500",
                 "USD_2020 = 500/596.2 * USD_CE500",
                 "USD_2021 = 500/708.0 * USD_CE500",
+                "USD_2022 = 500/816.0 * USD_CE500",
             ]
         )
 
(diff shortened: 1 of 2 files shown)


hunterbarber:
Note the conversion value for USD_2021 in the repo differs from the one published in https://toweringskills.com/financial-analysis/cost-indices/ (708.0 compared to 708.8). I imagine changing this value in the repo would require a lot of downstream changes in value assertions for costing tests, so I'm leaving as is.

andrewlee94 (approved):

bpaul4 (approved):
@hunterbarber thank you for adding the 2022 value. We actually use 2018 as the default for most tests, so updating the 2021 value should have a small impact. Please feel free to update this value.

andrewlee94:
@hunterbarber @bpaul4 We need to be careful about changing values like this; whilst we might test it, other might be using the values and this would be an unexpected change for them. If there is good cause for the change then we can consider it, but we would also need to discuss backward compatibility.

PR #1429, opened 2024-06-04
Add new AutoScaler and CustomScalerBase classes

andrewlee94:
## Summary/Motivation:
This PR adds a draft of the new `Scaler` classes along with some general utility functions for the new scaling interface. I have generally created new functions even if older ones existed to make a clean break from the older API and assist with backward compatibility.

Draft Documentation: [Initial Outline of Documentation for Scaling Tools.docx](https://github.com/user-attachments/files/16565931/Initial.Outline.of.Documentation.for.Scaling.Tools.docx)

- New ScalerBase class
- New AutoScaler class
- New CustomScalerBase class
- Utility methods for manipulating scaling suffixes
- Demonstration tests of new methods on Gibbs reactor

### Legal Acknowledgement


idaes/core/scaling/autoscaling.py
```diff
@@ -0,0 +1,291 @@
...
+
+        self._set_scaling_factor(vardata, "variable", sf)
+
+    def _con_by_norm(self, blk, con_list=None, norm=2):
+        # Get scaled Jacobian - we want to consider any existing scaling
+        # We will account for existing scaling factors later and update them
+        jac, nlp = get_jacobian(blk, scaled=True)
+
+        if con_list is None:
+            con_list = nlp.get_pyomo_equality_constraints()
```
dallan-keylogic, line 245: In my experience, `get_jacobian` is slow because forming the Pynumero NLP is slow. If we have flowsheet-level variables, we end up getting a Jacobian for the flowsheet and every subcomponent. I'm not sure how slow forming the NLP is going to be for PropertyBlock 43 in the MEA column, but you're still at least doubling the work you're doing.

I would suggest getting the Jacobian and NLP once at the beginning of `constraints_by_jacobian_norm`, then just referring to rows and columns of the overall object as you want to scale constraints.

idaes/core/scaling/autoscaling.py
```diff
@@ -0,0 +1,291 @@
...
+
+        # Use scipy to get all the norms
+        # Should be more efficient that iterating in Python
+        axis = (
+            1  # Could make this an argument to also support variable-based norm scaling
```
adam-a-a, line 250: Based on your comment, should we consider adding another method for variable-based norm scaling?

idaes/core/scaling/autoscaling.py
```diff
@@ -0,0 +1,291 @@
...
+    ConfigValue(
+        default=False,
+        domain=Bool,
+        description="Minimum value for constraint scaling factors.",
+    ),
+)
+
+
+@document_kwargs_from_configdict(CONFIG)
+class AutoScaler:
```
adam-a-a, line 97: It'd be nice to have a method like `report_scaling` that would give you vars and constraints in one column, scaling factor values in the 2nd column, and the scaled value in the third column.

idaes/core/scaling/autoscaling.py
andrewlee94: I do not know if this makes sense or not - comments would be welcome (@dallan-keylogic ).

idaes/core/scaling/autoscaling.py
andrewlee94: The first two you can get by doing `suffix.display()`. The third part - having scaled value is easy enough for Vars, but what is the "scaled value" for a Constraint? Scaled residual is easy to get, but not always that meaningful (as the scaled Jacobian is often more meaningful).

idaes/models/unit_models/tests/test_gibbs.py
```diff
@@ -1013,583 +462,23 @@ def test_initialize_duty(self, methane):
...
+    @pytest.mark.solver
+    @pytest.mark.skipif(solver is None, reason="Solver not available")
+    @pytest.mark.integration
+    def test_verify_scaling_duty(self, methane):
+        assert jacobian_cond(methane, scaled=True) == pytest.approx(9191, abs=1)
 
-        # Check for optimal solution
-        assert check_optimal_termination(results)
+        assert len(extreme_jacobian_rows(methane, scaled=True)) == 0
+        assert len(extreme_jacobian_rows(methane, scaled=True)) == 0
```
adam-a-a, line 481: Actually, adding to my other suggestion on thinking about a reporting method for `Autoscaler`, these would also be good to show in report output (e.g., condition number, extreme jac rows/columns or at least the number of each). 

I realize that the DiagnosticsToolbox already has the capability to report on some of these things, but just thinking it'd be nice to be able to apply scaling via AutoScaler and subsequently check what you really did to model scaling with a report method.

andrewlee94: That is what the Diagnostics tools are for - putting those methods here would just be duplication of code (and the diagnostics also tell you a lot more). The eventual documentation will highlight that you really need to use diagnostics, scaling and initialization tools together to get the best results.

idaes/core/scaling/autoscaling.py
andrewlee94: I have been thinking about there best way to do this, and did realize I was probably getting the Jacobian more times than necessary. I was not aware of the issue of having higher-level components in the problem causing the entire problem to be emitted (and given how properties work that will be an issue). I am also thinking there might need to be some checks to deal with deactivated components (and that it might be possible to deactivate unnecessary components before getting the Jacobian to reduce the problem size).

andrewlee94: I think I've fixed this so this method will only be called once for any call to the scaling method.

`constraints_by_jacobian_norm` (line 156) collects all the constraints we want to scale and the top level block, and then only calls the autoscaler once.

idaes/core/scaling/autoscaling.py
dallan-keylogic: If we scale variables first based on order-of-magnitude, we can then scale constraints based on row/column norm. However, norm-based variable scaling probably won't be useful for a fully unscaled model, because the most common Jacobian entries will be O(1). 

For a partially-scaled model, though, filling in variable scaling factors based on column norms might be useful.

idaes/models/unit_models/tests/test_gibbs.py
agarciadiego: Will the diagnostics toolbox reference these new tools?

andrewlee94: @agarciadiego I am not sure there - we probably need to think about how and where we draw the line between the two. Up until now, the Diagnostics Toolbox has been about finding the issues, and hasn't said much about how to fix them (as that is not so easy to explain concisely). We can definitely mention it in the docs, and maybe we could have the `display` methods mention tools to help fix issues (where appropriate).

MarcusHolly (commented):
I spent some time trying to apply the autoscalers to the BSM2 flowsheet in WaterTAP, but I think that flowsheet is too much of a monstrosity to naively expect the autoscalers to resolve all the problems. That being said, I didn't have much time to test out the other tools extensively and I'll be out for the next two weeks.

ksbeattie:
Making this high priority so it gets the attention from others that is needed.

andrewlee94:
Some initial documentation for the code:

[Initial Scaling Routine Profiling.docx](https://github.com/user-attachments/files/16552332/Initial.Scaling.Routine.Profiling.docx)

idaes/core/scaling/autoscaling.py
```diff
@@ -0,0 +1,215 @@
...
+        else:
+            raise TypeError(f"{blk_or_cons.name} is not a block or constraint.")
+
+        # Once we have a single target block and list of constraints, call the scaler method
+        # once for all the constraints
+        self._con_by_norm(jblock, con_list=con_list, norm=norm)
+
+    def _vardata_by_magnitude(self, vardata):
+        if value(vardata) is None:
+            sf = 1.0
```
MarcusHolly, line 170: I don't think this is working as intended. I've tried replacing the scaling in the `demo_flowsheet.py` with 
```
    scaler = AutoScaler()
    scaler.variables_by_magnitude(m)
```
but I get the following error: `ValueError: No value for uninitialized NumericValue object fs.M01.minimum_pressure[0.0,1]`. I suspect this would occur anytime `value(vardata)` is None. I'm not sure if the solution is to not scale variables with a value of None or find a way to (temporarily?) fix its value

andrewlee94: Ah - the main thing is that this should be `vardata.value` and not `value(vardata)`.

MarcusHolly: That fixes the issue

idaes/core/scaling/tests/test_scaling_profiler.py
```diff
@@ -0,0 +1,554 @@
...
+    model.fs.unit.inlet.mole_frac_comp[0, "CO"].fix(1e-5)
+    model.fs.unit.inlet.mole_frac_comp[0, "H2O"].fix(1e-5)
+    model.fs.unit.inlet.mole_frac_comp[0, "NH3"].fix(1e-5)
+    model.fs.unit.inlet.temperature[0].fix(1500.0)
+    model.fs.unit.inlet.pressure[0].fix(101325.0)
+
+    model.fs.unit.outlet.temperature[0].fix(2844.38)
+    model.fs.unit.deltaP.fix(0)
+
+    from_json(model, fname=fname, wts=StoreSpec.value())
```
MarcusHolly, line 241: What is the purpose of this json file? If I wanted to use the scaling profiler on a flowsheet, would I need to generate a similar json?

andrewlee94: That is for test purposes so that we can load a solution into the model without having to spend time on initializing it first. You could replace that with a standard initialization and solve, but ti would take longer to run.

idaes/core/scaling/util.py
```diff
@@ -0,0 +1,812 @@
...
+# Design of Advanced Energy Systems (IDAES).
+#
+# Copyright (c) 2018-2023 by the software owners: The Regents of the
+# University of California, through Lawrence Berkeley National Laboratory,
+# National Technology & Engineering Solutions of Sandia, LLC, Carnegie Mellon
+# University, West Virginia University Research Corporation, et al.
+# All rights reserved.  Please see the files COPYRIGHT.md and LICENSE.md
+# for full copyright and license information.
+#################################################################################
+"""
```
dallan-keylogic, line 13: Leaving this as a marker about where to resume my review.

idaes/models/unit_models/tests/test_gibbs_scaling.py
```diff
@@ -0,0 +1,464 @@
...
+        )
+
+        scaler = GibbsReactorScaler()
+
+        scaler.variable_scaling_routine(m.fs.unit)
+
+        for v in m.fs.unit.lagrange_mult.values():
+            assert m.fs.unit.scaling_factor[v] == pytest.approx(
+                1 / (8.314 * 500), rel=1e-4
+            )
```
dallan-keylogic, line 119: Why are we expecting no heat or deltaP scaling factors here? We have both heat transfer and pressure change in the unit model. Also, if we're expecting no scaling factors, we should explicitly assert that they aren't present.

idaes/models/unit_models/tests/test_gibbs_scaling.py
```diff
@@ -0,0 +1,464 @@
...
+            submodel_scalers={
+                "control_volume.properties_in": SMScaler,
+                "control_volume.properties_out": SMScaler,
+            },
+        )
+
+        scaled = jacobian_cond(methane, scaled=True)
+
+        assert scaled < unscaled
+        assert scaled == pytest.approx(6.96238e15, rel=1e-5)
```
dallan-keylogic, line 464: Once this gets merged, I want to see what's causing the condition number to still be so badly behaved.

dallan-keylogic (approved):
Looks good besides some minor corrections and tweaks. The real test will be once we use it and once we start trying to get users to use it.

idaes/core/scaling/util.py
```diff
@@ -0,0 +1,812 @@
...
+        )
+
+    # Get suffix and assign scaling factor
+    sfx = get_scaling_suffix(component)
+
+    if not overwrite and component in sfx:
+        _log.debug(
+            f"Existing scaling factor for {component.name} found and overwrite=False. "
+            "Scaling factor unchanged."
+        )
```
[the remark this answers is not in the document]
andrewlee94: Not quite - you should be able to control the logging level by each subpackage in IDAES. I.e. you can set different levels for `idaes` and `idaes.core.scaling`.

idaes/core/scaling/util.py
```diff
@@ -0,0 +1,812 @@
...
+    if ctype not in [None, Var, Constraint]:
+        raise ValueError(
+            f"report_scaling_factors only supports None, Var or Constraint for argument ctype: "
+            f"received {ctype}."
+        )
+
+    if not isinstance(blk, (Block, BlockData)):
+        raise TypeError(
+            "report_scaling_factors: blk must be an instance of a Pyomo Block."
+        )
```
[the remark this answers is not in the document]
andrewlee94: Yes.

idaes/core/scaling/util.py
```diff
@@ -0,0 +1,812 @@
...
+        if denominator == 0:
+            # Assign a nominal value of 1 so that we can continue
+            denominator = 1
+            # Log a warning for the user
+            _log.debug(
+                "Nominal value of 0 found in denominator of division expression. "
+                "Assigning a value of 1. You should check you scaling factors and models to "
+                "ensure there are no values of 0 that can appear in these functions."
+            )
```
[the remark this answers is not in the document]
andrewlee94: A Var or Param does not need to be fixed to give a zero here, just its current value. That is probably still not good, but at least not as bad as it being fixed to zero.

As for whether this should be a warning or not - my first reaction is that the diagnostics should have caught this, but that would of course rely on the user using the diagnostics. It probably does not hurt to make this a warning however (i.e. if you looked at the diagnostics you wouldn't see this anyway).

idaes/models/unit_models/tests/test_gibbs_scaling.py
andrewlee94: My guess is the fact that the model is uninitialized, but I gave it good initial guesses for the actual final state - i.e. there is a mismatch between the current uninitialized state and the scaling factors it was given. If we went and solved the model, I suspect the value would be much lower.

idaes/core/scaling/util.py
```diff
@@ -0,0 +1,812 @@
...
+            stream.write(
+                f"\n{header}{' ' * (maxname - len(header))}{TAB}Scaling Factor\n"
+            )
+
+            for n, i in cdict.items():
+                # Pad name to length
+                stream.write(f"{n + ' ' * (maxname - len(n))}{TAB}{i}\n")
+
+
+def get_nominal_value(component):
```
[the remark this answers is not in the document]
andrewlee94: This is the underlying function that is used by the expression walker to get the nominal value of an expression/constraint. It is primarily intended to be used in the expression walker (or limited similar circumstances).

docs/explanations/scaling_toolbox/applying_scaling_tools.rst
```diff
@@ -0,0 +1,61 @@
...
+    :depth: 2
+
+Basic Usage
+-----------
+
+.. Note::
+
+  Modelers should endeavor to provide as much scaling information as possible before calling Scalers in order to provide as much information on your particular case as possible.
+
+All Scaler in the Scaling Toolbox support a ``scale_model`` method that can be called to apply scaling to a model of the appropriate type. The ``scale_model`` method can be called as shown below:
```
andrewlee94, line 14:
```suggestion
All Scalers in the Scaling Toolbox support a ``scale_model`` method that can be called to apply scaling to a model of the appropriate type. The ``scale_model`` method can be called as shown below:
```

docs/explanations/scaling_toolbox/diagnosing_scaling_issues.rst
```diff
@@ -0,0 +1,20 @@
...
+* large Jacobian Condition Numbers (>1e10),
+* variables with very large or very small values,
+* variables with values close to zero,
+* variables with values close to a bound (conditionally),
+* constraints with mismatched terms,
+* constraints with potential cancellations,
+* variables and constraints with extreme Jacobian norms,
+* extreme entries in Jacobian matrix (conditionally).
+
+If you see any of these warnings in the model diagnostics output, it is a sign that you have potential scaling issues which should be resolved in order to improve the performance, robustness and accuracy of your model.
```
andrewlee94, line 17:
```suggestion
If you see any of these warnings in the model diagnostics output, it is a sign that you have potential scaling issues which should be investigated in order to improve the performance, robustness and accuracy of your model.
```

docs/explanations/scaling_toolbox/scaling_theory.rst
```diff
@@ -0,0 +1,39 @@
...
+'''''''''''''''''''''''''''''
+
+Due to the three different aspects of scaling and the fact that they can sometimes compete with each other, it is hard (if not impossible) to define a single metric for what defines “good” scaling. However, common indicators of bad scaling are:
+
+* Variables with extremely large or small scaled magnitudes (>1e6 or <1e-6) – you should generally aim to have most variables have scaled magnitudes between 1 and 10.
+
+  * An important exception is any variable that can pass through zero, like enthalpy or bidirectional fluxes. These variables should be scaled based on “significant variation” (think something like a standard deviation) of the variable throughout different conditions of the model.
+  * Zero flux boundary conditions should be scaled like the other fluxes in the model.
+
+* Constraints with 1 or more very large terms – as each constraint must be solved to solver tolerance in order to meet the convergence criteria, it is important that change on the order of magnitude of the solver tolerance is meaningful for each constraint term.
```
andrewlee94, line 32:
```suggestion
* Constraints with 1 or more very large terms – as each constraint must be solved to solver tolerance in order to meet the convergence criteria, it is important that change on the order of magnitude of the solver tolerance is meaningful for each constraint term. Small perturbations in these large terms can causes large changes in the constraint residual.
```

docs/explanations/scaling_toolbox/scaling_theory.rst
```diff
@@ -0,0 +1,39 @@
...
+* Constraint with terms that are of a significantly different orders of magnitude. This means that a change that is significant in the small terms may be negligible for the larger terms (i.e., you can make significant changes in the small term with no effect on the larger term), whilst a change in the larger terms may result in huge changes in the smaller terms (i.e., a change in the final significant figure of the large term might result in orders of magnitude change in the smaller term, thus causing a cascade of changes through your model).
+* Any time a difference is taken between several large numbers to obtain a small number, the model is at risk of catastrophic cancellation and ill-conditioning
+* Very large Jacobian Condition Number (>1e10). The condition number represents a worst-case error amplification in the problem. For example, with a condition number of 1e10 then a round-off error due to machine precision (~1e-16) could result in a change in model state of 1e-6 (= 1e10 x 1e-16); note that most solvers use a default tolerance of 1e-6 to 1e8, thus with a condition number of 1e10 or greater, there is a risk that your solution may be dominated by round-off errors.
```
andrewlee94, line 36:
```suggestion
* Very large Jacobian Condition Number (>1e10). The condition number represents a worst-case error amplification in the problem. For example, with a condition number of 1e10 then a round-off error due to machine precision (~1e-16) could result in a change in model state of 1e-6 (= 1e10 x 1e-16); note that most solvers use a default tolerance of 1e-6 to 1e-8, thus with a condition number of 1e10 or greater, there is a risk that your solution may be dominated by round-off errors.
```

docs/explanations/scaling_toolbox/scaling_workflow.rst
```diff
@@ -0,0 +1,35 @@
...
+
+As a starting point, the following workflow is recommended when scaling any model:
+
+1. Understand as much about the model as possible, including expected input and output values and where possible the formulation of the constraints.
+2. Make use of the Diagnostics Toolbox to ensure there are no structural issues and to identify potential scaling issues that must be resolved. This also provides a reference point for checking to see that your scaling factors are improving the state of the model. Modelers are encouraged to use these tools throughout the process to monitor their progress, however note that a partially scaled model will often have more issues than a completely unscaled model (this is often expected, and not necessarily a sign that you are going the wrong way)
```
andrewlee94, line 15:
```suggestion
2. Make use of the Diagnostics Toolbox to ensure there are no structural issues and to identify potential scaling issues that must be resolved. This also provides a reference point for checking to see that your scaling factors are improving the state of the model. Modelers are encouraged to use these tools throughout the process to monitor their progress, however note that a partially scaled model will often have more issues than a completely unscaled model (this is often expected, and not necessarily a sign that you are going the wrong way). Of particular note are the ``display_variables_with_extreme_jacobians`` and ``display_constraints_with_extreme_jacobians`` methods (as well as the ``SVDToolbox`` for advanced users).
```

docs/reference_guides/scaling/scaling.rst
```diff
@@ -0,0 +1,20 @@
+Scaling Toolbox
+===============
+
+.. note::
+
+  In v2.0, IDAES is beginning to deploy a new suite of scaling tools. This documentation refers to the new tools.
+  For documentation of the older scaling tools, :ref:`see here<reference_guides/core/util/scaling:Scaling Methods>`.
```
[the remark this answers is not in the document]
andrewlee94: This is for backward compatibility until we get around to deprecating the old tools.

idaes/core/scaling/autoscaling.py
```diff
@@ -0,0 +1,232 @@
...
+from idaes.core.scaling.scaling_base import ScalerBase
+from idaes.core.scaling.util import get_scaling_factor
+from idaes.core.util.scaling import get_jacobian
+
+
+class AutoScaler(ScalerBase):
+    """
+    IDAES Autoscaling Toolbox
+
+    Contains a number of methods useful for automatically scaling models based
```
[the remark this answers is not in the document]
andrewlee94: There is, and it will be automatically included and documented via the base class config block.

docs/explanations/scaling_toolbox/scaling_toolbox.rst
```diff
@@ -0,0 +1,57 @@
...
+* fully automated, can scale an entire model with one command
+* applicable to any model
+* useful for getting initial values for scaling factors, or filling-in missing scaling factors
+
+Cons:
+
+* require an initial solution, and thus not useful for pre-initialization scaling
+* consider only the current model state, and often overscale the problem
+
+For models with an initial (ideally feasible) solution, full information on the state of the model and the Jacobian is available. A number of approaches are available that can take this information and generate scaling factors for the entire model in order to meet some overall model characteristic (e.g., minimizing the norm of the Jacobian matrix). These have the advantage of requiring minimal input from the modeler and being able to scale an entire model in one go. However, these approaches require an initial solution (and thus are not useful for pre-initialization scaling) and as they consider only a single characteristic metric calculated at the current model state, they can often over scale the model and may not provide the best performance.
```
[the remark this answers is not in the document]
andrewlee94: This one is technically correct as is - we can only target a single characteristic at a time.

idaes/core/scaling/scaling_base.py
```diff
@@ -0,0 +1,233 @@
...
+from pyomo.core.base.constraint import ConstraintData
+from pyomo.core.base.var import VarData
+
+from idaes.core.scaling.util import get_scaling_factor, set_scaling_factor
+import idaes.logger as idaeslog
+
+# Set up logger
+_log = idaeslog.getLogger(__name__)
+
+# TODO: Tests for new common methods
```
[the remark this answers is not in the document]
andrewlee94: I forgot to remove this - it was there as a reminder to me to write tests.

idaes/core/scaling/custom_scaler_base.py
```diff
@@ -0,0 +1,608 @@
...
+        self.constraint_scaling_routine(
+            model, overwrite=self.config.overwrite, submodel_scalers=submodel_scalers
+        )
+
+        # Step 4: Call constraint fill in
+        if second_stage_fill_in is not None:
+            for i in second_stage_fill_in:
+                i(model)
+
+        # TODO: Consider how to make this work
```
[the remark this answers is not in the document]
andrewlee94: @dallan-keylogic and I think we have a way to avoid the need for this which he is working on. I'll remove the TODO and let him decide if he wants to create an issue or not.

idaes/core/scaling/util.py
```diff
@@ -0,0 +1,812 @@
...
+        blk = blk_or_suffix
+    elif isinstance(blk_or_suffix, Block):
+        # Indexed block
+        blk = blk_or_suffix
+        sdict = {}
+        sdict["block_datas"] = {}
+        for bd in blk_or_suffix.values():
+            sdict["block_datas"][bd.name] = _collect_block_suffixes(
+                bd, descend_into=descend_into
+            )
```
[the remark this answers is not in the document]
andrewlee94: For all of these tools, we assume that the IDAES Controlled Suffix will be on the parent block of the component. Pyomo does support putting scaling factors elsewhere, but that would require the user to manually set those (and thus they are responsible for managing them).

idaes/core/scaling/custom_scaler_base.py
```diff
@@ -0,0 +1,608 @@
...
+    ):
+        """
+        Call scaling method for submodel.
+
+        Scaler for submodel is taken from submodel_scalers dict if present, otherwise the
+        default scaler for the submodel is used.
+
+        Args:
+            model: parent model of submodel
+            submodel: name of submodel to be scaled as str
```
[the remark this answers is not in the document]
andrewlee94: Clarified.

docs/explanations/scaling_toolbox/index.rst
```diff
@@ -0,0 +1,30 @@
...
+Introduction
+------------
+
+The numerical scaling of a model is critical to its robustness and tractability, and can mean the difference between finding a solution and a solver failure. Scaling of non-linear models is a key step in the model development and application workflow, and often requires a significant amount of time and effort. The IDAES Scaling Toolbox aims to assist users in this task.
+
+Key Points
+----------
+
+* Scaling is critical for model performance
+* Scaling is not a one-off process, but something that needs to be reassessed before each solver call
```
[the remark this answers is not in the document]
andrewlee94: This one I will argue should remain as "needs to be re-assessed" - the re-assessment might be "it is fine as it is", but the user should always ask the question.

idaes/core/scaling/custom_scaler_base.py
```diff
@@ -0,0 +1,608 @@
...
+
+# Set up logger
+_log = idaeslog.getLogger(__name__)
+
+CSCONFIG = CONFIG()
+
+DEFAULT_UNIT_SCALING = {
+    # "QuantityName: (reference units, scaling factor)
+    "Temperature": (units.K, 1e-2),
+    "Pressure": (units.Pa, 1e-5),
```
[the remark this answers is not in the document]
andrewlee94: Yes - unfortunately Pyomo has no way to distinguish those, so the user/developer needs to take care to handle those cases.

idaes/core/scaling/custom_scaler_base.py
```diff
@@ -0,0 +1,608 @@
...
+            sf = 1
+        elif scheme == "harmonic_mean":
+            sf = sum(1 / abs(i) for i in nominal)
+        elif scheme == "inverse_sum":
+            sf = 1 / sum(abs(i) for i in nominal)
+        elif scheme == "inverse_root_sum_squared":
+            sf = 1 / sum(abs(i) ** 2 for i in nominal) ** 0.5
+        elif scheme == "inverse_maximum":
+            sf = 1 / max(abs(i) for i in nominal)
+        elif scheme == "inverse_minimum":
```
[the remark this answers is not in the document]
andrewlee94: Yes - @jsiirola said the same thing.

idaes/core/scaling/custom_scaler_base.py
```diff
@@ -0,0 +1,608 @@
...
+        """
+        Call scaling method for submodel.
+
+        Scaler for submodel is taken from submodel_scalers dict if present, otherwise the
+        default scaler for the submodel is used.
+
+        Args:
+            model: parent model of submodel
+            submodel: name of submodel to be scaled as str
+            submodel_scalers: user provided dict of Scalers to use for submodels
```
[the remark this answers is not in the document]
andrewlee94: I think the reason for this was that I was thinking we did not have the objects yet, but as this is a user-facing function (and not a set of defaults) that is not the case. I will update this too.

dallan-keylogic (approved):
LGTM

PR #1442, opened 2024-06-21
Moving some slow tests to integration

andrewlee94:
## Summary/Motivation:
Run times for our non-integration tests have been creeping up again, so this PR moves a lot of slow tests (>10s) to integration. This will likely impact our apparent test coverage, as this is calculated based on the `unit` and `component` flags only, but cuts the run time for these by nearly 30% on my local machine.

### Legal Acknowledgement


idaes/models/unit_models/tests/test_heat_exchanger_1D.py
```diff
@@ -3544,7 +3544,7 @@ def model(self):
 
         return m
 
-    @pytest.mark.component
+    @pytest.mark.integration
     def test_general_hx1d_initializer(self, model):
         initializer = HX1DInitializer()
         initializer.initialize(model.fs.unit)
```
dallan-keylogic, line 3550: I wonder how this initialization time is going to be affected when we move away from block triangularization in #1436 .

dallan-keylogic (approved):
LGTM

idaes/models/unit_models/tests/test_heat_exchanger_1D.py
andrewlee94: That is a good point, but I suspect it won't have a huge effect in this case. I believe there is another equivalent test for a different configuration that runs fast enough, so I suspect this one is caused more by stiffness in the model than the property initialization. Still, would be worth waiting for #1436 to check.

andrewlee94: Update: after #1436 merged, run time for this test is still ~16 seconds, so I'll leave it in integration for now.

bpaul4 (approved):

bpaul4 (approved):

Issue #1399, opened 2024-04-23
Function to allow initial condition problem in PETSc to be debugged


dallan-keylogic:
After once again having to go into PETSc and manually edit the function to return the Pyomo block that it uses for solving the initial condition problem, I think it's time to streamline the process so that people that don't know how to edit core IDAES functions can also debug their initial condition problems.

I know that @Robbybp had a similar request a few months ago. If you have anything in particular you want out of such a helper function, please add it to this issue so that I can keep track of it.

lbianchi-lbl:
#1443 covers a portion of this PR, but does not implement the refactoring (e.g. introducing a `PETSc` class) that would be needed (or at least useful) to address this in a more general way.

sufikaur:
Will be fixed in #1824

Issue #1347, opened 2024-02-21
PETSc  DAE Solver "initial_only" only option.


eslickj:
When using the petsc.petsc_dae_by_time_element() function to solve a DAE model, I think it would be useful to have an "initial_only" option.  I may not have thought this through well enough, but the idea is that when you initialize a DAE model and you are solving with an integrator, only the initial condition matters.  Next time you solve it the trajectory is just replaced.  If I break the model up and initialize unit by unit, getting the trajectory is just a waste of time.  This isn't really a PETSc solver issue specifically, but petsc.petsc_dae_by_time_element() does conveniently split off the initial condition equations for you. I'm know there are other ways to approach this, but adding the option seems convenient. 

I think this would be easy to implement.  Just terminate the function after solving the initial conditions.  The results should be in the model.  Return the solver status and ```None``` for the trajectory.

dallan-keylogic:
Possibly what should be done is to provide a way to return the initial condition problem. I've had to do things like add break statements in the middle of the PETSc code in order to do model diagnostics because my initial condition problem was badly specified. The only trick is figuring out how to handle the context manager.

eslickj:
@dallan-keylogic, what about a new context manager that you could use like

``` python
with InitialConditionProblemManager(...) as ic_model:
    #diagnostics, solve, or whatever.
```

We could maybe just copy the first part of the petsc time stepping solver wrapper and build on TemporarySubsystemManager.  It shouldn't be too hard to put together since it's basically all there.  The only trick is you may want an option to revert any changes you make to variable values.  Like if I want to solve the initial condition problem, I want to keep the variable values.  If you are doing diagnostics, maybe you don't want to keep any changes.

Issue #1343, opened 2024-02-19
PETSc interface fails with Lagrange-Legendre collocation


dallan-keylogic:
Presently, the PETSc interface throws an error when one attempts to integrate a DAE discretized with Lagrange-Legendre discretization:

```
    m = pyo.ConcreteModel()

    m.time = pyodae.ContinuousSet(initialize=(0.0, 1.0, 2.0))
    m.x = pyo.Var(m.time)
    m.u = pyo.Var(m.time)
    m.dxdt = pyodae.DerivativeVar(m.x, wrt=m.time)

    def diff_eq_rule(m, t):
        return m.dxdt[t] == m.x[t] ** 2 - m.u[t]

    m.diff_eq = pyo.Constraint(m.time, rule=diff_eq_rule)

    discretizer = pyo.TransformationFactory("dae.collocation")
    discretizer.apply_to(m, nfe=2, scheme="LAGRANGE-LEGENDRE")

    m.u.fix(1.0)
    m.x[0].fix(0.0)

    res = petsc.petsc_dae_by_time_element(
        m,
        time=m.time,
        between=[0.0, 2.0],
        ts_options={
            "--ts_type": "beuler",
            "--ts_dt": 3e-2,
        },
        skip_initial=True,  # With u and x fixed, no variables to solve for at t0
        calculate_derivatives=True,
    )
```

produces a log with a degrees-of-freedom error:

```
2024-02-19 14:06:27 [INFO] idaes.solve.petsc-dae: Called fg_read, err: 0 (0 is good)
2024-02-19 14:06:27 [INFO] idaes.solve.petsc-dae: ---------------------------------------------------
2024-02-19 14:06:27 [INFO] idaes.solve.petsc-dae: DAE: 1
2024-02-19 14:06:27 [INFO] idaes.solve.petsc-dae: Reading nl file: C:\Users\dallan\AppData\Local\Temp\tmpwptc4we3.pyomo.nl
2024-02-19 14:06:27 [INFO] idaes.solve.petsc-dae: Number of constraints: 2
2024-02-19 14:06:27 [INFO] idaes.solve.petsc-dae: Number of nonlinear constraints: 1
2024-02-19 14:06:27 [INFO] idaes.solve.petsc-dae: Number of linear constraints: 1
2024-02-19 14:06:27 [INFO] idaes.solve.petsc-dae: Number of inequalities: 0
[...]

PR #1443, opened 2024-06-26
Add function to PETSc interface to return initial condition problem

dallan-keylogic:
## Fixes
- Addresses #1343, #1347, and #1399 
- Stops Pytest warnings from controller tests not returning `None`

## Summary/Motivation:
When using the PETSc interface, a square initial condition problem is first solved (with either PETSc-snes or IPOPT)  to make sure the initial condition is consistent before sending integrations problems to PETSc-TS. However, when you end up with degree of freedom problems or a degenerate problem, it's difficult to debug. As an expedient, I would edit the PETSc problem to return the problem in the middle of the context manager. 

I've added a function to get this problem directly and modified the PETSc interface to use this function to get the initial condition problem to solve to ensure consistency. 

### Legal Acknowledgement


Robbybp:
This looks really useful, thanks for making the change. I'll give it a review soon.

eslickj (approved):

lbianchi-lbl:
@Robbybp, just a friendly nag to review when (or if) you have some time.

dallan-keylogic:
@Robbybp , could you please review this sometime soon? I'd really like it to make the August release.

idaes/core/solvers/petsc.py
```diff
@@ -1023,3 +1063,127 @@ def from_json(self, pth):
...
+def get_initial_condition_problem(
+    model,
+    time,
+    initial_time,
+    representative_time=None,
+    initial_constraints=None,
+    initial_variables=None,
+    detect_initial=True,
+    flattened_problem=None,
+):
```
Robbybp, line 1111: Is there any reason `initial_time` is required? I would think that `time.first()` would be a reasonable default.

idaes/core/solvers/petsc.py
```diff
@@ -1023,3 +1063,127 @@ def from_json(self, pth):
...
+        initial_constraints = []
+
+    if flattened_problem is None:
+        if representative_time is None:
+            raise RuntimeError(
+                "The user must supply either the flattened problem or a representative time."
+            )
+        flattened_problem = _get_flattened_problem(
+            model=model, time=time, representative_time=representative_time
+        )
```
Robbybp, line 1155: Why is either a flattened problem or representative time required? Shouldn't this function just use the same defaults that `petsc_dae_by_time_element` uses?

idaes/core/solvers/petsc.py
```diff
@@ -1023,3 +1063,127 @@ def from_json(self, pth):
...
+            condition solve step.  Since the time-indexed constraints are picked
+            up automatically, this generally includes non-time-indexed
+            constraints.
+        initial_variables (list): This is a list of variables to fix after the
+            initial condition solve step.  If these variables were originally
+            unfixed, they will be unfixed at the end of the solve. This usually
+            includes non-time-indexed variables that are calculated along with
+            the initial conditions.
+        detect_initial (bool): If True, add non-time-indexed variables and
+            constraints to initial_variables and initial_constraints.
```
Robbybp, line 1134: I don't love the name of this option, but since it is the same as in `petsc_dae_by_time_element`, I think it should stay.

idaes/core/solvers/petsc.py
```diff
@@ -1023,3 +1063,127 @@ def from_json(self, pth):
...
+        flattened_problem (dict): Dictionary returned by get_flattened_problem.
+            If not provided, get_flattened_problem will be called at representative_time.
```
Robbybp, line 1136: I assume this option is provided so we don't have the overhead of recomputing the flattened model when calling within `petsc_dae_by_time_element`? I'm fine with this option, but would prefer for it to not be part of this function's public API. This is mostly because I don't want a user to have to construct a dict with undocumented keys (or use the private `_get_flattened_problem` function) to use this option. I recommend renaming the option to `_flattened_problem`, and adding to the docstring that its behavior is subject to change.

For future applications where a "flattened model" is necessary, the `DynamicModelInterface` from `pyomo.contrib.mpc` may be useful (see https://github.com/Pyomo/pyomo/blob/404fd6d997d24f7209e8af4c427a13d6e10f19c4/pyomo/contrib/mpc/interfaces/model_interface.py#L52 or https://pyomo.readthedocs.io/en/stable/contributed_packages/mpc/interface.html#pyomo.contrib.mpc.interfaces.model_interface.DynamicModelInterface).

idaes/core/solvers/petsc.py
```diff
@@ -1023,3 +1063,127 @@ def from_json(self, pth):
...
+    ] + initial_constraints
+    variables = [
+        var[initial_time] for var in flattened_problem["time_variables"]
+    ] + initial_variables
+
+    if len(constraints) <= 0:
+        raise RuntimeError(
+            "Zero constraints in initial condition problem, therefore "
+            "there is no block to return."
+        )
```
Robbybp, line 1181: We could just return an empty block, or None, in this case. I think I like this better than raising an error here that we then catch above. We already check the number of constraints before trying to solve the block, so an empty block should be fine, although None might be more explicit.

idaes/core/solvers/petsc.py
```diff
@@ -563,46 +594,48 @@ def petsc_dae_by_time_element(
...
+                initial_constraints=initial_constraints,
+                initial_variables=initial_variables,
+                detect_initial=detect_initial,
+                flattened_problem=flattened_problem,
+            )
+        except RuntimeError as err:
+            if "Zero constraints" in err.message:
+                init_subsystem = None
+            else:
+                raise err
```
Robbybp, line 611: See below. I'd rather this function return None, and we handle it here, than rely on parsing the error message.

Robbybp (changes requested):
Questions/comments/requests about error handling and function arguments.

idaes/core/solvers/petsc.py
dallan-keylogic: I'm trying to avoid having to flatten the problem multiple times. I haven't profiled it, but I suspect it's rather slow. I debated reimplementing `petsc_dae_by_time_element` as an `Object` so that the flattened problem can be cached. Maybe that will happen eventually.

Robbybp: I have no problem with a `flattened_problem` argument, I just think if neither it nor representative time is provided, we should default to `_get_flattened_problem(model=model, time=time, representative_time=time.at(2))`.

ksbeattie:
Progress not expected until end of Sept (per @dallan-keylogic)

PR #1454, opened 2024-07-22
Diagnostics tool for ill-posed constraints

andrewlee94:
## Summary/Motivation:
As part of working on the new scaling tools, I realised there are some simple checks we can do for detecting poorly-posed constraints that could cause scaling issues. This PR adds a new expression walker that looks for the following signs of poor scaling in constraints:

* sum expressions with terms which have significant differences in magnitude.
* sum expressions where terms cancel out (with a catch for cases of `constant == sum()`).

- New `print_compact_form` method for displaying expressions which is makes use of named `Expressions`
- New diagnostics tools for identifying potential mismatched terms and cancellations in constraints
- New expression walker for finding problematic terms in constraints to support the above
- Bumps version of Pyomo for testing

### Legal Acknowledgement


dallan-keylogic:
@Robbybp , is there any possibility of combining a tool like this with block triangularization? My intuition is that a constraint in which catastrophic cancellation occurs (taking the difference between two large numbers to get a small number) might show up differently in the block structure than one in which it does not (adding a small number to a large number to get another large number), even though individual constraint might be identical.

Robbybp:
@dallan-keylogic My only initial thought is that a "catastrophic cancellation" has different implications depending on where it appears appears in the block triangular decomposition. If it is in a diagonal block, it could be a problem (as we're relying on that term for nonsingularity). Otherwise, it is fine, as the entry could be zero and the matrix's singularity would not change.

dallan-keylogic:
Okay, a very basic example to see the issues at play. We're not concerned when we add a small number to a big number to get another big number:
image
This system has the matrix representation
image
However, the situation is different when we take the difference of two big numbers to get a small number:
image
This system has the matrix representation
image

Looking at these two systems, there doesn't seem to be any structural difference between them. The variable b is chosen to be a pivot in both cases.  However, when we define a scaled variable
image
we get a different picture.

The first system then becomes 
image
with a matrix representation
image
whose system matrix is well-conditioned with determinant -1.

The second system, on the other hand, becomes
image
with a matrix representation
image
whose system matrix is ill-conditioned with determinant -epsilon. It can be shown that the condition number blows up in the limit as epsilon goes to zero.

Robbybp:
@dallan-keylogic I would say that scaling `b` by `1/eps` is not the right choice here. A block triangularization on the scaled system could definitely identify that. We could potentially identify that `b`, as a column-singleton, should not be scaled up.

idaes/core/util/model_diagnostics.py
```diff
@@ -4122,3 +4264,269 @@ def _collect_model_statistics(model):
...
+            # Collect all warnings about mismatched terms from child nodes
+            mismatch += d[1]
+            cancelling += d[2]
+
+            # We will check for cancelling terms here, rather than the sum itself, to handle special cases
+            # We want to look for cases where a sum term results in a value much smaller
+            # than the terms of the sum
+            sums = self._sum_combinations(d[0])
+            if any(i <= self._sum_tol * max(d[0]) for i in sums):
+                cancelling.append(str(node))
```
jsiirola, line 4434: I think I understand what you are doing here, but wouldn't this complain loudly about the objective for every parameter estimation problem (MSE) if the problem actually solved?

idaes/core/util/model_diagnostics.py
```diff
@@ -4122,3 +4264,269 @@ def _collect_model_statistics(model):
...
+            if not d[3]:
+                const = False
+
+        # Return any problematic terms found
+        return mismatch, cancelling, const
+
+    def _check_equality_expression(self, node, child_data):
+        # (In)equality expressions are a special case of sum expressions
+        # We can start by just calling the method to check the sum expression
+        vals, mismatch, cancelling, const = self._check_sum_expression(node, child_data)
```
jsiirola, line 4446: Don't you need to negate the values for the second argument in the relational expression before treating it like a sum?

idaes/core/util/model_diagnostics.py
```diff
@@ -4122,3 +4264,269 @@ def _collect_model_statistics(model):
...
+        for i in chain.from_iterable(
+            combinations(values_list, r) for r in range(2, len(values_list) + 1)
+        ):
+            sums.append(abs(sum(i)))
+        return sums
+
+    node_type_method_map = {
+        EXPR.EqualityExpression: _check_equality_expression,
+        EXPR.InequalityExpression: _check_equality_expression,
+        EXPR.RangedExpression: _check_sum_expression,
```
jsiirola, line 4476: Don't you need special handling for Ranged (like Equality / Inequality)?  I would probably define a
```python
def _check_ranged(self, node, child_data):
    lhs_vals, lhs_mismatch, lhs_cancelling, lhs_const = self._check_equality(node, child_data[:2])
    rhs_vals, rhs_mismatch, rhs_cancelling, rhs_const = self._check_equality(node, child_data[1:])
    # Then merge the results and return
```

idaes/core/util/model_diagnostics.py
```diff
@@ -4122,3 +4264,269 @@ def _collect_model_statistics(model):
...
+        mismatch, cancelling, const = self._perform_checks(node, child_data)
+
+        # First, need to get value of input terms, which may be sub-expressions
+        input_mag = [self._get_value_for_sum_subexpression(i) for i in child_data]
+
+        # Next, create a copy of the external function with expected magnitudes as inputs
+        newfunc = node.create_node_with_local_data(input_mag)
+
+        # Evaluate new function and return the absolute value
+        return [value(newfunc)], mismatch, cancelling, const
```
jsiirola, line 4414: The comment doesn't match the code: missing `abs()`

idaes/core/util/model_diagnostics.py
```diff
@@ -4122,3 +4264,269 @@ def _collect_model_statistics(model):
...
+        return vals, mismatch, cancelling, const
+
+    def _check_product(self, node, child_data):
+        mismatch, cancelling, const = self._perform_checks(node, child_data)
+
+        val = self._get_value_for_sum_subexpression(
+            child_data[0]
+        ) * self._get_value_for_sum_subexpression(child_data[1])
+
+        return [val], mismatch, cancelling, const
```
jsiirola, line 4352: Almost all of your nonlinear handlers could be replaced by a single callback:
```python
def _check_general_expr(self, node, child_data):
    mismatch, cancelling, const = self._perform_checks(node, child_data)
    val = node._apply_operation(list(map(self._get_value_for_sum_subexpression, child_data)))
    return [val], mismatch, cancelling, const
```

jsiirola (commented):
Overall looks pretty reasonable.  Some questions and comments (that should probably be addressed)

idaes/core/util/model_diagnostics.py
```diff
@@ -4122,3 +4264,269 @@ def _collect_model_statistics(model):
...
+
+            # Var or Param
+            return self._check_base_type(node)
+            # might want to add other common types here
+
+        # not a leaf - check if it is a named expression
+        if (
+            hasattr(node, "is_named_expression_type")
+            and node.is_named_expression_type()
+        ):
```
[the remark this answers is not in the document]
andrewlee94: I think I'll keep the current form as it is a little easier to understand what it is doing.

idaes/core/util/model_diagnostics.py
andrewlee94: Thank you - I did not know I could do that.

idaes/core/util/model_diagnostics.py
dallan-keylogic: One of my main concerns with this tool is its inability to distinguish between true cases of catastrophic cancellation and benign situations, where, for example, we have a heat of adsorption term in a column which will be close to except near the breakthrough point.

idaes/core/util/model_diagnostics.py
```diff
@@ -4122,3 +4264,269 @@ def _collect_model_statistics(model):
...
+            vl = max(absvals)
+            vs = min(absvals)
+            if vl != vs:
+                if vs == 0:
+                    diff = log10(vl)
+                else:
+                    diff = log10(vl / vs)
+
+                if diff >= self._mm_tol:
+                    mismatch.append(str(node))
```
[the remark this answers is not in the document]
andrewlee94: Done, and made things more efficient in the process.

idaes/core/util/model_diagnostics.py
andrewlee94: Is there an easy way to test for that however? Note that this tool is only used for Cautions in the main toolbox, with the implication that these might be issues (but not guaranteeing that).

I.e., this is intended to be a simple check to find potential issues, but the user will have to look into them all further to decide if they are critical or not.

idaes/core/util/model_diagnostics.py
andrewlee94: I had forgotten to make that correction - thank you.

idaes/core/util/model_diagnostics.py
andrewlee94: I think I've fixed this - it would be good if you could double check my logic however.

idaes/core/util/model_diagnostics.py
```diff
@@ -1052,6 +1073,105 @@ def display_near_parallel_variables(self, stream=None):
...
+            Constraint, descend_into=descend_into
+        ):
+            _, mm, cc, k = walker.walk_expression(c.expr)
+
+            for i in mm:
+                mismatch.append(f"{c.name}: {i}")
+            for i in cc:
+                cancellation.append(f"{c.name}: {i}")
+            if k:
+                constant.append(c.name)
```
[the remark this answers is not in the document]
andrewlee94: This one would take some more work and require changing other `collect` methods as well. I think that might be better as a separate issue/PR.

idaes/core/util/model_diagnostics.py
andrewlee94: @jsiirola This walker was written specifically with Constraints in mind, so I had not considered that. The check for mismatched terms would make sense for an Objective however, so this could be extended to handle them as well. However, how would the expression walker know if it was dealing with an Objective or a Constraint (the input argument is the expression, not the component)?

dallan-keylogic (commented):
Something that occurred to me is that it might be useful for the user to call this not on the entire model, but on a subset of constraints (perhaps identified by the SVD toolbox, perhaps identified some other way). That probably doesn't need to be present during the initial rollout, but is something to keep in mind.

idaes/core/util/tests/test_model_diagnostics.py
```diff
@@ -999,6 +1014,111 @@ def test_display_near_parallel_variables(self):
...
+    @pytest.mark.component
+    def test_collect_constraint_mismatches(self):
+        m = ConcreteModel()
+        m.v1 = Var(initialize=2)
+        m.v2 = Var(initialize=3)
+
+        # Constraint with no free variables
+        m.c1 = Constraint(expr=m.v1 == m.v2)
+        m.v1.fix()
+        m.v2.fix()
```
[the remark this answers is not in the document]
andrewlee94: As it stands, Arc constraints will get lumped in here, however from a strict sense they should fall in here (as they can be trivially eliminated). However, from a structural stand point we probably want to ignore these.

What do we want to allow? One option would be the special case of `a=b` where `a` and `b` are leaf nodes (i.e. not expressions themselves).

idaes/core/util/model_diagnostics.py
```diff
@@ -4122,3 +4265,273 @@ def _collect_model_statistics(model):
...
+            return self._check_other_expression(node, data)
+
+        raise TypeError(
+            f"An unhandled expression node type: {str(nodetype)} was encountered while "
+            f"analyzing constraint terms {str(node)}"
+        )
+
+    def walk_expression(self, expr):
+        """
+        Main method ot call to walk an expression and return analysis.
```
andrewlee94, line 4518:
```suggestion
        Main method to call to walk an expression and return analysis.
```

idaes/core/util/model_diagnostics.py
```diff
@@ -4122,3 +4265,273 @@ def _collect_model_statistics(model):
...
+
+        Args:
+            expr - expression ot be analysed
```
andrewlee94, line 4521:
```suggestion
            expr - expression to be analyzed
```

idaes/core/util/tests/test_model_diagnostics.py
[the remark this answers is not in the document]
andrewlee94: @dallan-keylogic I believe I have fixed this - the cancellation checks are only run if at least one of the sides in an (in)equality is a sum expression. This means that constraints with the form `a == b` and `a == f(b)` will not be flagged as cancelling.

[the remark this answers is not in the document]
andrewlee94: Also, regarding you initial comment, this a tool to do that could be written fairly easily as it would just iterate over a `ComponentSet` containing the `Constraint` of interest and call this walker for each `Constraint`. This walker is not part of the `DiagnosticsToolbox` class, so it can easily be called from other places too (such as the `SVDToolbox`) if we can come up with a way to identify a `set of `Constraints` of interest.

idaes/core/util/model_diagnostics.py
```diff
@@ -4122,3 +4264,277 @@ def _collect_model_statistics(model):
...
+        self.canceling_terms = ComponentSet()
+        self.mismatched_terms = ComponentSet()
+
+    def _get_value_for_sum_subexpression(self, child_data):
+        return sum(i for i in child_data[0])
+
+    def _sum_combinations(self, values_list):
+        sums = []
+        for i in chain.from_iterable(
+            combinations(values_list, r) for r in range(2, len(values_list) + 1)
```
mrmundt, line 4310: 2 is a magic number! Why 2?

idaes/core/util/model_diagnostics.py
```diff
@@ -4122,3 +4264,277 @@ def _collect_model_statistics(model):
...
+            const = node.fixed
+        else:
+            const = True
+        return [value(node)], const, False
+
+    def _check_equality_expression(self, node, child_data):
+        # (In)equality expressions are a special case of sum expressions
+        # First, need to negate one side of the expression
+        mdata = []
+        for i in [0, 1]:
```
mrmundt, line 4348: It's possible I am being naive, but why bother with a for loop if there are only two cases and both cases are guaranteed to happen?

idaes/core/util/model_diagnostics.py
```diff
@@ -4122,3 +4264,277 @@ def _collect_model_statistics(model):
...
+        # We will also ignore any constraints where a == b and neither a nor b are sum expressions
+        if not child_data[0][2] and not child_data[1][2]:
+            # Linking constraint, skip checks
+            pass
+        elif node not in self.mismatched_terms and not any(d[1] for d in mdata):
+            # No constant terms, check for cancellation
+            # First, collect terms from both sides
+            t = []
+            for d in mdata:
+                t += d[0]
```
mrmundt, line 4372: This is fine but could be a list comprehension.

idaes/core/util/model_diagnostics.py
```diff
@@ -4122,3 +4264,277 @@ def _collect_model_statistics(model):
...
+            )
+        except ValueError:
+            raise ValueError(
+                f"Error in ConstraintTermAnalysisVisitor: error evaluating {str(node)}."
+            )
+        except ZeroDivisionError:
+            raise ZeroDivisionError(
+                f"Error in ConstraintTermAnalysisVisitor: found division with denominator of 0 "
+                f"({str(node)})."
+            )
```
mrmundt, line 4397: Do you want to add a raise in case there is any other error?

mrmundt (changes requested):
I did not review the tests cases yet because I do have some suggestions for changes in the main file that may require changes in test cases.

idaes/core/util/model_diagnostics.py
```diff
@@ -1052,6 +1073,105 @@ def display_near_parallel_variables(self, stream=None):
...
+        )
+
+        mismatch = []
+        cancellation = []
+        constant = []
+
+        for c in self._model.component_data_objects(
+            Constraint, descend_into=descend_into
+        ):
+            _, mm, cc, k = walker.walk_expression(c.expr)
```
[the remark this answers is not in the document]
andrewlee94: Mismatched terms, cancelling terms and constant - the names grew organically whilst I was testing the underlying code and made their way here.

idaes/core/util/model_diagnostics.py
```diff
@@ -1052,6 +1073,105 @@ def display_near_parallel_variables(self, stream=None):
...
+            stream: I/O object to write report to (default = stdout)
+
+        Returns:
+            None
+
+        """
+        if stream is None:
+            stream = sys.stdout
+
+        mismatch, _, _ = self._collect_constraint_mismatches()
```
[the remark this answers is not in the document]
andrewlee94: The issue with this is cache invalidation - we have no way of knowing if the cached values are still valid. We discussed this with some of the other methods, and decided that for now the simplest way was to just re-collect the information each time. Seeing as this is user-interaction anyway, the time hit is not that important.

idaes/core/util/model_diagnostics.py
andrewlee94: We are looking for any combination of terms which cancel, thus the minimum number of terms to consider is 2 (a single term cannot cancel with itself). I can add a comment.

idaes/core/util/model_diagnostics.py
andrewlee94: Is it necessary to have one? I thought that if a different type of exception was encountered it would be raised directly.

idaes/core/util/model_diagnostics.py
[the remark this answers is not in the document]
mrmundt: Ahhh, okay, fair. Then consider this resolved.

idaes/core/util/model_diagnostics.py
mrmundt: Necessary? No. Good practice? Yes. It will make the error messages look more standardized / not weirdly disappear into the stack trace.

idaes/core/util/model_diagnostics.py
dallan-keylogic: What this block of code does is go through combinations of terms in an expression first in groups of 2, then in groups of 3, all the way up to groups of `len(values_list)`. So if we have the sum expression `a+b+c+d`, we first iterate through `(a, b), (a, c), (a, d), (b, c), (b, d), (c, d)`, then iterate through `(a, b, c,), (a, b, d), (a, c, d), (b, c, d)`, then `(a,b,c,d)`.

However, the number of terms you're checking grows exponentially in expression size. In particular, if `len(values_list) == m`, you'll be checking `2 ** m - m -1` terms. I expect that this sort of check will take an extremely long time on any model with `Expressions` of any significant length, much less an extreme chonker like eNRTL.

idaes/core/util/model_diagnostics.py
andrewlee94: What is the best practices way to do this?

mrmundt:
```
try:
   (whatever)
except SpecificError:
   (blah)
except Exception as err:
    print(f"Unexpected {err=}, {type(err)=}")
    raise
```
^^ An example!

andrewlee94: Will pylint complain about the `except Exception`, or does it just complain about `except:`?

mrmundt: I think it complains specifically if you `except:` and then don't raise anything. I don't _think_ it yells if you catch general exceptions _and also_ raise them.

andrewlee94:
@mrmundt I think I have addressed all of your comment. If you have time to do another review, it would be appreciated.

idaes/core/util/tests/test_model_diagnostics.py
```diff
@@ -1358,7 +1483,7 @@ def test_report_numerical_issues_ok(self):
 
 ====================================================================================
 """
-
+        print(stream.getvalue())
```
mrmundt, line 1486: Leftover?

idaes/core/util/tests/test_model_diagnostics.py
```diff
@@ -4053,3 +4179,771 @@ def test_output(self, model):
...
+            11,  # 3-term sums starting with 2
+            12,  # 3-term sum starting with 3
+            10,
+            11,
+            12,
+            13,
+            14,  # 4-term sums
+            15,  # 5-term sum
+        ]
+        print(sums)
```
mrmundt, line 4221: Leftover?

idaes/core/util/tests/test_model_diagnostics.py
```diff
@@ -4053,3 +4179,771 @@ def test_output(self, model):
...
+        m.v1 = Var(["a", "b"], initialize=5e6)
+        m.v2 = Var(initialize=1e7)
+        m.v3 = Var(initialize=0)
+
+        expr = m.v3 == sum(v for v in m.v1.values()) - m.v2
+        vv, mm, cc, k = ConstraintTermAnalysisVisitor().walk_expression(expr=expr)
+
+        assert vv == [0, pytest.approx(0, abs=1e-12)]
+        assert len(mm) == 0
+        print(cc)
```
mrmundt, line 4870: Leftover?

andrewlee94:
@mrmundt Thank you - I fixed the left over print statements.

idaes/core/util/model_diagnostics.py
```diff
@@ -4122,3 +4266,297 @@ def _collect_model_statistics(model):
...
+        mdata.append(child_data[1])
+
+        # Next, call the method to check the sum expression
+        vals, const, _ = self._check_sum_expression(node, mdata)
+
+        # Next, we need to check for canceling terms.
+        # In this case, we can safely ignore expressions of the form constant = sum()
+        # We can also ignore any constraint that is already flagged as mismatched
+        # We will also ignore any constraints where a == b and neither a nor b are sum expressions
+        if not child_data[0][2] and not child_data[1][2]:
```
mrmundt, line 4384: I am... mildly irked about all of the magic numbers here. Perhaps it is more well-known than me what is inside `child_data`, but as someone who doesn't know, what is in `[0][0]` and why does it need to be negated?

mrmundt (changes requested):
I only have issues with one thing which is "magic numbers" - otherwise it looks good!

idaes/core/util/model_diagnostics.py
andrewlee94: Seeing as I had to remind myself what these meant, the point is well taken. I have added some comments that hopefully explain what is happening.

mrmundt (approved):
I am generally content with this now!

idaes/core/util/model_diagnostics.py
```diff
@@ -1052,6 +1080,168 @@ def display_near_parallel_variables(self, stream=None):
     # TODO: Block triangularization analysis
     # Number and size of blocks, polynomial degree of 1x1 blocks, simple pivot check of moderate sized sub-blocks?
 
+    def _collect_constraint_mismatches(self, descend_into=True):
+        walker = ConstraintTermAnalysisVisitor(
+            term_mismatch_tolerance=self.config.constraint_term_mismatch_tolerance,
```
dallan-keylogic, line 1096: Even though this function is private, shouldn't we have some sort of description about what it does?

idaes/core/util/model_diagnostics.py
```diff
@@ -1052,6 +1080,168 @@ def display_near_parallel_variables(self, stream=None):
...
+            title="The following constraints have mismatched terms:",
+            end_line="Call display_problematic_constraint_terms(constraint) for more information.",
+            header="=",
+            footer="=",
+        )
+
+    def display_constraints_with_canceling_terms(self, stream=None):
+        """
+        Display constraints in model which contain additive terms which potentially cancel each other.
+
```
dallan-keylogic, line 1158: Worth mentioning that, while we've caught most common forms, some perfectly benign equality constraints will run afoul of this test when the constraint is satisfied.

dallan-keylogic (commented):
```
    Cancellation in fs.reflux_mixer.rich_solvent_state[0.0].enth_mol_phase[Liq]. Terms 6 (-75.234960403619), 10 (203.81521375405413), 11 (-135.9668692510444)
    Cancellation in fs.reflux_mixer.mixed_state[0.0].enth_mol_phase[Liq]. Terms 6 (-74.70060388842397), 10 (201.96256194082713), 11 (-137.1777586826791)
```
It would be clearer if it said "Cancellation in expression `expr`".

idaes/core/util/model_diagnostics.py
```diff
@@ -4359,17 +4448,24 @@ def __init__(
         term_mismatch_tolerance: float = 1e6,
         term_cancellation_tolerance: float = 1e-4,
         term_zero_tolerance: float = 1e-10,
+        max_canceling_terms: int = 4,
+        max_cancellations_per_node: int = 5,
```
dallan-keylogic, line 4452: Why do these have default values in the object when we're also giving them default values in the `ConfigBlock`?

dallan-keylogic (approved):
Just one minor quibble. Otherwise looks good.

idaes/core/util/model_diagnostics.py
andrewlee94: Different objects - the is the visitor class which exists outside the `DiagnosticsToolbox` class and so can be used separately if desired. Thus, the visitor has the ability to take these as arguments and thus has default values for them.

PR #1466, opened 2024-08-08
add UI return, then visualize() can return now

CopyDemon:
## Fixes
Add return to UI.visualize() then user can use visualize() returned value and functions

-
-

### Legal Acknowledgement


idaes/core/base/flowsheet_model.py
```diff
@@ -275,9 +275,10 @@ def visualize(self, model_name, **kwargs):
             **kwargs: Additional keywords for :func:`idaes.core.ui.fv.visualize()`
 
         Returns:
-            None
+            VisualizeResult : The return from UI, namedtuple("VisualizeResult", ["store", "port", "server", "save_diagram"]) etc...
         """
-        UI().visualize(self, model_name, **kwargs)
+        VisualizeResult = UI().visualize(self, model_name, **kwargs)
+        return VisualizeResult
```
lbianchi-lbl, line 281: The convention in Python is to use `snake_case` for variables/instances, and `PascalCase` for classes. Since the value returned here is an instance, I think `snake_case` is the appropriate one:

```suggestion
        visualize_result = UI().visualize(self, model_name, **kwargs)
        return visualize_result
```

idaes/core/base/flowsheet_model.py
```diff
@@ -275,9 +275,10 @@ def visualize(self, model_name, **kwargs):
             **kwargs: Additional keywords for :func:`idaes.core.ui.fv.visualize()`
 
         Returns:
-            None
+            VisualizeResult : The return from UI, namedtuple("VisualizeResult", ["store", "port", "server", "save_diagram"]) etc...
```
lbianchi-lbl, line 278: Using the `class` and `meth` Sphinx roles for the Python domain (https://www.sphinx-doc.org/en/master/usage/domains/python.html#role-py-class) will generate clickable references in the Sphinx docs for this module (the references might not be enabled until the Intersphinx machinery is done, but it should not generate an error before that).

```suggestion
            visualize_result: The :class:`idaes_ui.fv.fsvis.VisualizeResult` instance returned by :meth:`UI.visualize`
```

lbianchi-lbl (changes requested):
Looks good apart from these minor changes.

lbianchi-lbl:
This is how this looks in the Sphinx docs:

image

idaes/core/base/flowsheet_model.py
```diff
@@ -275,9 +275,10 @@ def visualize(self, model_name, **kwargs):
             **kwargs: Additional keywords for :func:`idaes.core.ui.fv.visualize()`
 
         Returns:
-            None
+            visualize_result: The :class:`idaes_ui.fv.fsvis.VisualizeResult` instance returned by :meth:`UI.visualize`
```
lbianchi-lbl, line 278: After checking the docs build, I think the `visualize_result:` is being interpreted by Sphinx to be a return type rather than the name of the returned variable, which causes an extra "Return type" field to be added:

image

 To avoid this, we should remove it before merging:

```suggestion
            The :class:`idaes_ui.fv.fsvis.VisualizeResult` instance returned by :meth:`UI.visualize`
```

lbianchi-lbl (changes requested):
One more minor issue with the Sphinx rendering.

lbianchi-lbl (approved):

dangunter (approved):
Looks good, but a test would be nice. Will add another issue

PR #1467, opened 2024-08-09
Remove support for Python 3.8

### LINKED ISSUE #1462: Drop support for Python 3.8
Python 3.8 is related to a growing number of issues, mostly due to support being dropped by dependencies such as NumPy and matplotlib, including:

- #1401
- #1453

lbianchi-lbl:
- Remove Python 3.8 from supported versions in docs and `setup.py`
- Remove Python 3.8 from CI workflows

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -88,7 +88,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.9', '3.10', '3.11', '3.12']
         os:
           - linux
           - win64
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -132,7 +132,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.9', '3.10', '3.11', '3.12']
         os:
           - linux
           - win64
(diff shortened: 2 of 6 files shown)


bpaul4 (approved):
Just one non-critical comment.

There is one mention on Python 3.8 on the FAQ page: https://github.com/IDAES/idaes-pse/blob/main/docs/explanations/faq.rst. The `pywin32` issue may be relevant to other Python version, in which case we could leave the FAQ as-is.

lbianchi-lbl:
> Just one non-critical comment.
> 
> There is one mention on Python 3.8 on the FAQ page: https://github.com/IDAES/idaes-pse/blob/main/docs/explanations/faq.rst. The `pywin32` issue may be relevant to other Python version, in which case we could leave the FAQ as-is.

Good point, I did notice this as well and opted for leaving it in for the same reason you mentioned. I propose to leave that in unless someone objects.

andrewlee94:
@lbianchi-lbl @bpaul4 I suggest we keep the discussion as well, but add a note to indicate that Python 3.8 is no longer officially supported (i.e. make it clear we left this comment there for those who wish to try (or have similar issues), but that we don't guarantee that the rest of the code will work).

lbianchi-lbl:
> @lbianchi-lbl @bpaul4 I suggest we keep the discussion as well, but add a note to indicate that Python 3.8 is no longer officially supported (i.e. make it clear we left this comment there for those who wish to try (or have similar issues), but that we don't guarantee that the rest of the code will work).

@andrewlee94 I've added a note there (and the same to `README.md` for good measure):

image

ksbeattie (approved):
LGTM

PR #1403, opened 2024-05-03
Address many unexpected warnings in pytest

andrewlee94:
## Summary/Motivation:
This fixes a lot of unexpected warnings being emitted during pytest runs.

## Changes proposed in this PR:
- Fix instances of non-test classes with names `TestX`
- Fix warnings about regex escapes
- Remove a number of redundant tests for surrogate plotting. Also get plotting methods to return the plot objects so we can close them.
- Remove some `return` statements in test methods/functions.

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

Issue #1363, opened 2024-03-01
Ensure no unexpected warnings are emitted during test suite runs


lbianchi-lbl:
- Removing the `-W ignore` flag in #1362 caused warnings emitted during pytest run to be shown again (see below for a list)
- Ideally, this list would be empty:
  - Expected warnings should be inspected within the testing code using `pytest.warns()` (which as I understand is equivalent to `pytest.raises()` used for exceptions)
  - Unexpected warnings should be addressed in the code being tested

### `SyntaxWarning` invalid escape sequence

- Most often found in `pytest.raises(match=expected_str)`:
- If `expected_str` is supposed to be a plain string (i.e. containing no regexes), use `re.escape()`
- If `expected_str` contains both regex constructs and regex symbols that should not be interpreted as regexes, define `expected_str` as raw literals using `r"..."`, e.g. `r"[a-z]+.my_dict\[key\]"`

### `PytestCollectionWarning`: cannot collect test class 'TestXYZ' because it has a __init__ constructor

- If the `TestXYZ` class is supposed to be a pytest test class (i.e. its methods should be collected as pytest test functions), remove the `__init__()` definition
- If the `TestXYZ` class is not supposed to be a pytest class, change its name so that it doesn't collide with the `Test` prefix used by pytest, e.g.
  - `SampleXYZ`
  - `XYZForTesting`

### `PytestReturnNotNoneWarning`: Expected `None`, but `<test-function>` returned `<object>`, which will be an error in a future version of pytest.  Did you mean to use `assert` instead of `return`?

- Test functions in pytest are not supposed to be called directly
- Basically all the tests in `idaes/apps/matopt` have this

---

```
idaes/apps/grid_integration/tests/util.py:25
  /home/runner/work/idaes-pse/idaes-pse/idaes/apps/grid_integration/tests/util.py:25: PytestCollectionWarning: cannot collect test class 'TestingModel' because it has a __init__ constructor (from: idaes/apps/grid_integration/tests/test_bidder.py)
    class TestingModel:
idaes/apps/grid_integration/tests/util.py:267
  /home/runner/work/idaes-pse/idaes-pse/idaes/apps/grid_integration/tests/util.py:267: PytestCollectionWarning: cannot collect test class 'TestingForecaster' because it has a __init__ constructor (from: idaes/apps/grid_integration/tests/test_bidder.py)
    class TestingForecaster(AbstractPrescientPriceForecaster):
[...]

PR #1472, opened 2024-08-17
PySMO: fix warnings

OOAmusat:
## Fixes
#1403, #1363 

## Summary/Motivation:
Recent changes to the numpy backend have resulted in a large number of deprecation warnings (>7500) being raised in PySMO. This PR makes changes to address these warnings.

- Fix all numpy deprecation warnings, primarily arising from assigning scalar values to arrays, e.g., `y[i, 0] = x`. This is resolved in this PR by either extracting the specific elements of the scalar (`y[i, 0] = x[0]`) or using the item method (`y[i, 0] = x.item()`)
- Replacing `warnings.warn` with the idaes logger warnings, and fix tests to use `caplog` to reflect this change
- Adding  `@pytest.mark.filterwarnings` to specific tests to ignore runtime warnings around `np.log`

### Legal Acknowledgement


idaes/core/surrogate/pysmo/tests/test_kriging.py
```diff
@@ -97,6 +97,7 @@ def test__init__07(self, array_type):
 
     @pytest.mark.unit
     @pytest.mark.parametrize("array_type", [np.array, pd.DataFrame])
+    @pytest.mark.filterwarnings("ignore::RuntimeWarning")
```
lbianchi-lbl, line 100: Do you have a list of the `RuntimeWarning`s that are being filtered here?

OOAmusat: I am filtering two types of runtime warnings:

`RuntimeWarning: invalid value encountered in log
RuntimeWarning: divide by zero encountered in log`
  
However, I'm happy to remove the filters - the runtime warnings are few (about 10 of them). I'm on the fence about whether to filter them or not.

lbianchi-lbl: I see, thanks. If I understand how `pytest.mark.filterwarnings` works, I think you should be able to use a regular expression to catch both of those cases:

```suggestion
    @pytest.mark.filterwarnings("ignore:(?:invalid value|divide by zero) encountered in log:RuntimeWarning")
```

The goal that we're aiming to is to have no warnings at all that are not expected, so `pytest.mark.filterwarnings` is a good approach, as long as the filter is as specific is possible. In this case, for example, not adding a regex would let possible future `RuntimeWarning` being raised go undetected (which we don't want).

OOAmusat: Oh, I didn't know I could do that. 

I agree on the "specific" bit - that is why I was on the fence about adding the filters to start with. 

I'll make the change and see it it works. 

Thanks!

OOAmusat: Resolved

lbianchi-lbl (approved):
Thanks a lot @OOAmusat. From a quick look in the pytest job output log, I believe this resolves all of the unanticipated warnings from the PySMO submodules (i.e., I wasn't able to spot any).

The only one that looks like it might be related is the following:

image

However, if you feel this belongs outside of PySMO's "jurisdiction", then feel free to disregard and we can address it at a later time.

idaes/core/surrogate/pysmo/tests/test_kriging.py
```diff
@@ -97,6 +97,9 @@ def test__init__07(self, array_type):
 
     @pytest.mark.unit
     @pytest.mark.parametrize("array_type", [np.array, pd.DataFrame])
+    @pytest.mark.filterwarnings(
+        "ignore:invalid value encountered in log:RuntimeWarning"
```
andrewlee94, line 101: Just for my understanding, where are these warnings coming from and are we sure that they do not indicate an underlying issue somewhere?

andrewlee94 (approved):
This looks good to me. My only questions is just to confirm that the warnings about invalid values in logs is not a sign of underlying issues.

idaes/core/surrogate/pysmo/tests/test_kriging.py
OOAmusat: The warnings are coming from zeros and NaNs when computing logs. The specific warning is coming from this computation:

```
 RuntimeWarning: invalid value encountered in log:
 log_like = (0.5 * ns * np.log(ssd)) + (0.5 * lndetcov)
```
The warnings are expected: ssd and lndetcov involve matrix manipulations (e.g. multiplication, diagonalization, cholesky factorization), and in some cases that may lead to NaNs or zeros during computation.

PR #1474, opened 2024-08-19
[WIP] Add a command to install CyIpopt

Robbybp:
## Fixes
In conjuction with #1473, addresses part of https://github.com/IDAES/idaes-ext/issues/261. (Not sure if this works on Windows...)

## Summary/Motivation:
CyIpopt needs to be able to find a copy of the Ipopt shared library and header files, which it does (at least on Linux and Mac) using `pkg-config`. To install with Ipopt in a non-standard location (such as IDAES's Ipopt), we need to set the `PKG_CONFIG_PATH` environment variable (to include `.idaes/bin/lib/pkgconfig`, which will likely change in new idaes-ext releases). To make this a little bit easier for users, I propose to add a command to install CyIpopt in the current Python environment, e.g.:
```
idaes install-cyipopt
```
This basically just runs `pip install cyipopt` with `PKG_CONFIG_PATH` set appropriately.

I believe users will still need to set `LD_LIBRARY_PATH` (or `DYLD_LIBRARY_PATH`/`PATH` on Mac/Windows) to actually use CyIpopt. `import idaes` does set these variables, but I believe `import cyipopt` uses the value of the variables at the time the Python process started, not the current values in `os.environ`.

**This will only work with idaes-ext 3.4.2**

- Unpack `idaes-local*` when the `get-extensions` command is run
- Add an `install-cyipopt` command

### Legal Acknowledgement

### DIFF
--- a/idaes/commands/util/download_bin.py
+++ b/idaes/commands/util/download_bin.py
@@ -402,8 +402,21 @@ def download_binaries(
     _verify_checksums(checksum, pname, ptar, ftar)
 
     # Extract solvers
+    tars_to_unpack = list(ptar)
+    idaes_local_tarname = f"idaes-local-{platform}.tar.gz"
+    if "darwin" in idaes_local_tarname:
+        # We neglected to change the name of the idaes-local tar.gz file from *aarch64
+        # to *arm64 on ARM Mac, so we have to manually patch the name here. I believe
+        # this is fixed in the idaes-ext main branch, so this can be removed when
+        # we update to binaries compiled from this branch. -RBP
+        idaes_local_tarname = idaes_local_tarname.replace("aarch64", "arm64")
+    idaes_local_tarname = os.path.join(to_path, idaes_local_tarname)
+    # In addition to the tar.gz files we downloaded, we want to unpack the idaes-local
+    # tar.gz file. This is included in idaes-solvers, so we can unpack it in the same
+    # directory after we've unpacked idaes-solvers.
+    tars_to_unpack.append(idaes_local_tarname)
     links = {}
-    for n, p in zip(pname, ptar):
+    for p in tars_to_unpack:
         _log.debug(f"Extracting files in {p} to {to_path}")
         with tarfile.open(p, "r") as f:
             _verify_tar_member_targets(f, to_path, links)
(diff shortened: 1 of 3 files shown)


idaes/config.py
```diff
@@ -752,14 +752,33 @@ def setup_environment(bin_directory, use_idaes_solvers):
...
-            [oe.get("LD_LIBRARY_PATH", ""), bin_directory]
+            [
+                oe.get("LD_LIBRARY_PATH", ""),
+                bin_directory,
+                # We add .idaes/bin/lib to LD_LIBRARY_PATH to pick up any libraries
+                # we introduced by unpacking the idaes-local tar.gz file. This
+                # directory should be changed when/if the IDAES extensions directory 
+                # structure changes (e.g. to .idaes/lib). -RBP
+                os.path.join(bin_directory, "lib"),
+            ]
```
Robbybp, line 774: I'm not sure this is necessary (or at all beneficial), as the CyIpopt library appears to use the value of the appropriate variable at Python-start-up-time, rather than (somehow) using `os.environ`.

idaes/commands/util/download_bin.py
```diff
@@ -402,8 +402,21 @@ def download_binaries(
...
+    # In addition to the tar.gz files we downloaded, we want to unpack the idaes-local
+    # tar.gz file. This is included in idaes-solvers, so we can unpack it in the same
+    # directory after we've unpacked idaes-solvers.
+    tars_to_unpack.append(idaes_local_tarname)
     links = {}
-    for n, p in zip(pname, ptar):
+    for p in tars_to_unpack:
         _log.debug(f"Extracting files in {p} to {to_path}")
         with tarfile.open(p, "r") as f:
             _verify_tar_member_targets(f, to_path, links)
```
Robbybp, line 422: Here we unpack `idaes-local*` into `.idaes/bin`, but I think there is no real reason not to unpack it into `.idaes`, in which case we would end up with bin, lib, include, and share subdirectories of `.idaes`.

A downside of unpacking `idaes-local*` is that now we actually distribute two copies of `libipopt.so`: one in `.idaes/bin`, and one in `idaes/bin/lib`... This will be fixed in new versions of idaes-ext.

Robbybp:
Closing as I'm not actively working on this. Will re-open if it looks like this will be worth the effort.

PR #1482, opened 2024-09-04
Add basis to conc example on property naming conventions docs

### LINKED ISSUE #1481: Property naming conventions for vapor pressure and saturation vapor pressure
I have ```relative_humidity = vapor_pressure / saturated_vapor_pressure```. 

According to [the IDAES docs page](https://idaes-pse.readthedocs.io/en/latest/explanations/conventions.html#thermophysical-and-transport-properties) on IDAES property naming conventions, ```vapor_pressure``` should be ```pressure_sat```, which I am not sure I understand (since ```vapor_pressure``` would not be saturated unless relative humidity is 100%... or maybe I am missing something). But then does that mean ```saturated_vapor_pressure``` should be ```pressure_sat_equil```? In this case, ```vapor_pressure``` is the partial pressure of water under given conditions and ```saturated_vapor_pressure``` is the vapor pressure of water in equilibrium with an open liquid surface.

I suppose the issue here is:
- how to name these properties such that they would adhere to the current IDAES naming conventions
- if ```pressure_sat``` on the docs page should be ```pressure_vap``` (or something else)
- if there needs to be a new conditional qualifier ```_sat```... or really how the IDAES community feels about this in general? 
- separated but related issue: the IDAES docs pages has ```conc_phase_comp['liq', 'H2O']``` as a valid property name, but does concentration not require a basis? What would be the units of such a property?

Or maybe there is a different way to do it that would adhere to the conventions as they are already laid out? 

(This may be better suited as a discussion rather than an issue so I am happy to move it if so)

kurbansitterley:
## Fixes #1481

Add basis to ```conc``` example on property naming conventions docs page.
see #1481 

## Summary/Motivation:

Concentration needs a basis (mass, mol, etc) to be a valid property name but was not included in the docs example.

- Add basis to ```conc``` example on property naming conventions docs page.
-

### Legal Acknowledgement


dangunter:
@lbianchi-lbl Do you know why the spell checker would fail on a DMF file, which certainly wasn't altered by this PR?

lbianchi-lbl:
> @lbianchi-lbl Do you know why the spell checker would fail on a DMF file, which certainly wasn't altered by this PR?

The first reason that comes to mind is that a new version of the spellchecker was released, with new typos being added, which would cause typos that used to be ignored with previous versions to be flagged instead.

This is a consequence of the typos package version being unpinned. This was a conscious decision we took being aware that it would result in occasional failures in unrelated PRs, but I've been thinking about reconsidering this and make the version update (and therefore the possibility of new typos being flagged) be "opt-in" instead.

**TL;DR** New version of spellchecker flags new typos that were already present in the codebase.

bpaul4:
@lbianchi-lbl @dangunter we could pin between releases, and on each release cycle update the version and fix any flagged typos. Might be a "best of both worlds" approach so PRs don't fail when the `typos` pushes a new version.

docs/explanations/conventions.rst
```diff
@@ -223,7 +223,7 @@ Variable Name                  Meaning
 enth                           Specific enthalpy of the entire mixture (across all phases)
 flow_comp["H2O"]               Total flow of H2O (across all phases)
 entr_phase["liq"]              Specific entropy of the liquid phase mixture
-conc_phase_comp["liq", "H2O"]  Concentration of H2O in the liquid phase
+conc_mass_phase_comp["liq", "H2O"]  Concentration of H2O in the liquid phase on a mass basis
```
andrewlee94, line 226: As you noted in the associated issue, there are a few others here that need to be updated with a baisis. Would you be able to update the following

`enth` to `enth_mol`
`flow` to `flow_vol`
`entr_phase['Liq']` to `entr_mol_phase['Liq']`

(I chose a variety of bases to demonstrate their uses).

andrewlee94 (approved):
Thank you for noting this - this is important for understanding the naming convention and was definitively wrong (and thus misleading).

bpaul4 (approved):

PR #1484, opened 2024-09-05
Remove idaes.core.dmf and related code

lbianchi-lbl:
## Changes proposed in this PR:
- Remove most files in `idaes/core/dmf`
- Edit `idaes/core/dmf/__init__.py` to add deprecation to be shown when importing `idaes.core.dmf`
- Trim logic used to skip tests that need optional dependencies
- Remove code that was only used for DMF from `setup.py`
- Remove `dmf_data` directory

## To review

- [ ] Check if the (very minimal) deprecation infrastructure is OK (@andrewlee94?)
- [ ] Check if the pages that have been modified in the docs make sense:
  - https://idaes-pse--1484.org.readthedocs.build/en/1484/how_to_guides/opt_dependencies.html
  - https://idaes-pse--1484.org.readthedocs.build/en/1484/how_to_guides/workflow/data_rec_parmest.html
- [ ] See if you spot any mention of the DMF in the code without a comment/deprecation or something like that

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1488, opened 2024-09-11
Add minimal pytest plugin to verify that generated files are cleaned up by tests

lbianchi-lbl:
## Summary/Motivation:

- @andrewlee94 discovered that a conspicuous number of untracked files are generated when running the test suite
- I was able to confirm his intuition that this was most likely a regression introduced by #1484
- It's not entirely clear how exactly the presence of the DMF code was causing generated files not to be saved when running test modules in different parts of the codebase
- My best guess (mostly based on gut feeling, i.e. without any rigorous investigation) is that it might have something to do with the code formerly in `idaes.core.dmf.util` through manipulation of temporary directories and/or the working directory

### Legal Acknowledgement

### DIFF
--- a/idaes/conftest.py
+++ b/idaes/conftest.py
@@ -16,7 +16,9 @@
 
 import importlib.abc
 import importlib.machinery
+import subprocess
 import sys
+from pathlib import Path
 from typing import Dict
 from typing import Iterable
 from typing import List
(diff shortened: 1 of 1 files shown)


ksbeattie (approved):
LGTM

blnicho (approved):

PR #1493, opened 2024-10-02
Wrap 0D Fixed Bed TSA unit model in `extra_models/fixed_bed_tsa0d` for Flowsheet Processor UI

dangunter:
## Fixes

Documentation: https://idaes-pse.readthedocs.io/en/latest/reference_guides/model_libraries/models_extra/temperature_swing_adsorption/fixed_bed_tsa0d.html 
Example:[ https://idaes-examples.readthedocs.io/en/latest/docs/flowsheets/temperature_swing_adsorption/temperature_swing_adsorption_doc.html](https://idaes-examples.readthedocs.io/en/latest/docs/flowsheets/temperature_swing_adsorption/temperature_swing_adsorption_doc.html)

## Summary/Motivation:
Initial UI wrapper for IDAES

- Wrap the model so it can be displayed in the Flowsheet Processor UI

### Legal Acknowledgement


dangunter:
One issue I see immediately is that the "model options" for the TSA has a number of values and they will _probably_ be changed at the same time / along with the model's other parameters and variables. However currently the UI sets these options in different, initial step, as shown by this screenshot for a watertap model:

image

This is, imho not ideal. We should have these sorts of model options more included in the main configuration page, even though they are not Pyomo vars.

@Daison2102 Do I have this roughly right about the `FixedBedTSA0D` model class?
@sspoon thoughts on the design?

MichaelPesce:
@dangunter the original purpose of those options were to allow for pre-build configurations, so it had made sense to be able to view them before running the build. however, they should also be available from the main workflow (after build step), where the model can be re-built with different model configurations

dangunter:
@MichaelPesce ok, though I think it might be better to have them as normal parameters

idaes/models_extra/temperature_swing_adsorption/tests/test_fixed_bed_tsa0d_ui.py
```diff
@@ -0,0 +1,109 @@
+import pytest
+import logging
+import pandas as pd
+from idaes.models_extra.temperature_swing_adsorption.util import tsa_summary
```
lbianchi-lbl, line 4:
```suggestion
from idaes.models_extra.temperature_swing_adsorption.util import tsa_summary
pytest.importorskip("idaes_flowsheet_processor.api", reason="idaes-flowsheet-processor must be installed to run this test")
```

lbianchi-lbl (changes requested):
Before merging, we should (* denotes items that are not strictly required):

- [ ] Add a dev dependency on `idaes-flowsheet-processor @ git+https://github.com/watertap-org/idaes-flowsheet-processor@0.1.dev1`
- [x] (*) Add an entry point in `setup.py` pointing to the `idaes.model_extra.temperature_swing_adsorption.fixed_bed_tsa0d_ui` module under the `idaes.flowsheets` group
- [ ] (*) Enable testing using the `idaes-flowsheets` pytest plugin

MichaelPesce:
@lbianchi-lbl @dangunter in order to get this working in the `idaes-electron-build`, we have to change the import `from watertap.ui import fsapi as api` to `from idaes_flowsheet_processor import api`, as suggested above

lbianchi-lbl:
> @lbianchi-lbl @dangunter in order to get this working in the `idaes-electron-build`, we have to change the import `from watertap.ui import fsapi as api` to `from idaes_flowsheet_processor import api`, as suggested above

@MichaelPesce done in d0209e7

lbianchi-lbl (approved):

PR #1495, opened 2024-10-05
Fix nested flowsheet time domain construction

NishantGiridhar:
## Summary/Motivation:
In the current implementation of FlowsheetBlock, nested flowsheets with a time domain can only inherit time from the parent flowsheet. 

- Moved code to create a new time domain into a private method.
- Create a new time domain for a nested FlowsheetBlock if time_set is provided during construction.
- Added tests.

### Legal Acknowledgement


idaes/core/base/flowsheet_model.py
```diff
@@ -342,35 +367,24 @@ def _setup_dynamics(self):
                     "{} was set as a dynamic flowsheet, but time domain "
                     "provided was not a ContinuousSet.".format(self.name)
                 )
+            
+            if self.config.dynamic is False and not isinstance(
```
andrewlee94, line 371: 1. Is this check actually necessary? This would appear to preclude putting steady-state flowsheets inside dynamic flowsheets.
2. `if self.config.dynamic is False` can be written as `if not self.config.dynamic`

idaes/core/base/flowsheet_model.py
```diff
@@ -342,35 +367,24 @@ def _setup_dynamics(self):
...
-                    self._time = ContinuousSet(initialize=self.config.time_set)
-                else:
-                    # For steady-state, use an ordered Set
-                    self._time = pe.Set(initialize=self.config.time_set, ordered=True)
-                self._time_units = self.config.time_units
-
-                # Set time config argument as reference to time domain
-                self.config.time = self._time
+                _create_time_domain()
+            elif self.config.time_set is not None:
```
andrewlee94, line 385: I think this is enough, but just to check is there any additional validation we should do here? Also, an in-line comment to explain what this case is (sub-flowsheet with user-provided time domain).

idaes/core/base/tests/test_flowsheet_model.py
```diff
@@ -400,6 +400,39 @@ def test_dynamic_external_time(self):
...
+    @pytest.mark.unit
+    def test_dynamic_nested_time(self):
+        m = ConcreteModel()
+        m.s1 = ContinuousSet(initialize=[4, 5])
+        m.s2 = ContinuousSet(initialize=[1,2,3])
+        m.fs = FlowsheetBlock(dynamic=True, time_units=units.s)
+        m.fs.sub = FlowsheetBlock(dynamic=True, time=m.s2, time_units=units.s)
+
+        assert m.fs.sub.config.dynamic is True
+        assert isinstance(m.fs.sub.time, ContinuousSet)
```
andrewlee94, line 412: Here you should do `assert m.fs.time is m.s2`. - i.e. check that the time domain is the correct set, not just the correct type.

idaes/core/base/tests/test_flowsheet_model.py
```diff
@@ -400,6 +400,39 @@ def test_dynamic_external_time(self):
...
+
+    @pytest.mark.unit
+    def test_dynamic_nested_time_from_time_set(self):
+        m = ConcreteModel()
+        m.fs = FlowsheetBlock(dynamic=True, time_set = [1,2,3], time_units=units.s)
+        m.fs.sub = FlowsheetBlock(dynamic=True, time_set = [1, 2], time_units=units.s)
+
+        assert m.fs.sub.config.dynamic is True
+        assert isinstance(m.fs.sub.time, ContinuousSet)
+        assert m.fs.sub.time_units is units.s
```
andrewlee94, line 423: You need to check that `m.fs.sub.time` is correct - you should check that it has length 2 and the values are 1 and 2.

idaes/core/base/tests/test_flowsheet_model.py
```diff
@@ -400,6 +400,39 @@ def test_dynamic_external_time(self):
...
+
+    @pytest.mark.unit
+    def test_dynamic_parent_time_indexed_ss_child(self):
+        m = ConcreteModel()
+        m.fs = FlowsheetBlock(dynamic=True, time_set = [1,2,3], time_units=units.s)
+        m.fs.sub = FlowsheetBlock(dynamic=False, time_set = [0, 1], time_units=units.dimensionless)
+
+        assert m.fs.sub.config.dynamic is False
+        assert isinstance(m.fs.sub.time, Set)
+        assert m.fs.sub.time_units == units.dimensionless
```
andrewlee94, line 433: As above.

idaes/core/base/tests/test_flowsheet_model.py
```diff
@@ -400,6 +400,39 @@ def test_dynamic_external_time(self):
...
+
+        assert m.fs.sub.config.dynamic is True
+        assert isinstance(m.fs.sub.time, ContinuousSet)
+        assert m.fs.sub.time_units is units.s
+
+    @pytest.mark.unit
+    def test_dynamic_parent_time_indexed_ss_child(self):
+        m = ConcreteModel()
+        m.fs = FlowsheetBlock(dynamic=True, time_set = [1,2,3], time_units=units.s)
+        m.fs.sub = FlowsheetBlock(dynamic=False, time_set = [0, 1], time_units=units.dimensionless)
```
andrewlee94, line 429: Time should not be dimensionless - it does not matter for the test, but we should not accidentally encourage people to do think this is OK.

andrewlee94 (changes requested):
This looks like a good start - I only had a few minor comments.

WE should also check the documentation to make sure that what we say there is correct.

dangunter:
@NishantGiridhar any update on this?

idaes/core/base/flowsheet_model.py
NishantGiridhar: That's exactly what it was checking for. Since we don't want to allow adding a nested dynamic flowsheet inside a parent steady-state flowsheet, I will remove this check from this PR.

idaes/core/base/flowsheet_model.py
NishantGiridhar: Based on my testing I think this should be good enough. The `_create_time_domain`  method makes sure that the specified `time_set` list is of the right length. So the same checks that apply for time domain construction in a parent flowsheet should apply to the nested flowsheet.

@dallan-keylogic could Are there any edge cases I am overlooking?

idaes/core/base/tests/test_flowsheet_model.py
```diff
@@ -400,6 +400,40 @@ def test_dynamic_external_time(self):
         assert m.fs.sub.config.time is m.s
         assert m.fs.sub.time_units == units.minute
 
+    @pytest.mark.unit
+    def test_dynamic_nested_time(self):
+        m = ConcreteModel()
+        m.s1 = ContinuousSet(initialize=[4, 5])
```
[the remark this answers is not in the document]
NishantGiridhar: Updated this. I use it in the parent flowsheet now.

idaes/core/base/flowsheet_model.py
```diff
@@ -342,35 +367,18 @@ def _setup_dynamics(self):
...
-                else:
-                    # For steady-state, use an ordered Set
-                    self._time = pe.Set(initialize=self.config.time_set, ordered=True)
-                self._time_units = self.config.time_units
-
-                # Set time config argument as reference to time domain
-                self.config.time = self._time
+                _create_time_domain()
+            # Creates a new time domain for a sub-flowsheet with user provided time domain
+            elif self.config.time_set not in [None, [0]]:
```
andrewlee94, line 379: Why is the `[0]` here? I would have thought that if we are handed an argument we should use it. To put it another way, if the user provides `[0]` as an argument we will ignore it without saying why.

ksbeattie:
@dallan-keylogic will look at this and talk to @NishantGiridhar about it.

ksbeattie:
@dallan-keylogic ask @NishantGiridhar about this.

dallan-keylogic (commented):
This needs the most recent version of `main` merged into it, but after one minor change it should be good to go.

blnicho:
@NishantGiridhar could you explain the motivation/use-case for wanting to support different time domains in different parts of the flowsheet?

dallan-keylogic:
@blnicho, the original motivation was the creation of a model with both a fast and slow timescale, each with its own time set. However, I pointed out that such a model would be better implemented as two flowsheets living on the same Pyomo `ConcreteModel` instead of trying to nest one flowsheet inside another one.

In general, I think that subflowsheets should be required to have the same time domain as the parent flowsheet. If we're sticking a flowsheet on another flowsheet, the child flowsheet should have ports that can be connected by arc constraints to flowsheet elements on the parent flowsheet, which typically implies having the same time domain. The only possible exception might be some dynamic model where you want to take more timesteps or use a different discretization on the child flowsheet than on the parent flowsheet. In such a case, you would need to be careful about which variables are put on the subflowsheet's ports, but I think that's doable.

andrewlee94:
I vaguely recall there is a bit more to this PR than that. Whilst we decided the original intent was best avoided, I recall that in the process we found a bug in the current implementation as well, which I think was being fixed in this PR.

blnicho:
@NishantGiridhar @dallan-keylogic do either of you have an example of how this would be used in practice? The small examples in the tests check that multiple time domains can be stored but it doesn't test whether a model with nested time domains is compatible with other IDAES tools (like the PETSc simulator for example). I'm worried that this is introducing "allowable" dynamic IDAES models that won't be checked for or handled correctly by the rest of the dynamic modeling tools and infrastructure in IDAES.

dallan-keylogic:
@NishantGiridhar , did the outer flowsheet have a time domain that was a subset of that of the inner flowsheet? My impression is that the time domains weren't even related in that way, but rather one contained a time domain over a 24 hour representative day, while the other flowsheet had one time point for each month. I think the correct method to handle the latter situation *is* to have two different flowsheets linked via constraints on a Pyomo `IndexedBlock`, rather than attempting to use a `Flowsheet` block for that purpose

PR #1500, opened 2024-10-11
Scaler for equilibrium reactor and saponification properties

andrewlee94:
## Summary/Motivation:
This PR adds a new scaler for the Equilibrium Reactor unit model and the saponification property packages. It also fixes a couple of bugs found whilst writing these.

- New EquilibriumReactorScaler
- New SaponificationThermoScaler
- New SaponificationReactionScaler

### Legal Acknowledgement


idaes/models/properties/examples/saponification_thermo.py
```diff
@@ -135,12 +136,48 @@ def define_metadata(cls, obj):
...
+
+    DEFAULT_SCALING_FACTORS = {
+        "flow_vol": 1e2,
+        "conc_mol_comp": 1e-2,
+    }
+
+    def variable_scaling_routine(
+        self, model, overwrite: bool = False, submodel_scalers: dict = None
+    ):
+        self.scale_variable_by_default(model.flow_vol, overwrite=overwrite)
```
dallan-keylogic, line 153: Is there a way for the user to provide a default value, rather than just using whatever happens to be the default default value?

dallan-keylogic (approved):
Looks good for a first pass.

idaes/models/properties/examples/saponification_reactions.py
```diff
@@ -111,12 +113,77 @@ def define_metadata(cls, obj):
         )
 
 
+class SaponificationReactionScaler(CustomScalerBase):
+    DEFAULT_SCALING_FACTORS = {"reaction_rate": 1e2}
```
bpaul4, line 117: How does the scaling framework know to use this value?

bpaul4 (approved):
I have one question, which is mainly for my own understanding of the code. The changes and additions look good.

idaes/models/properties/examples/saponification_thermo.py
andrewlee94: Yes, although it might not be the best way to do it. The default values are in a class attribute (dict) which users can interact with. That said, that is probably not the most obvious API for it (we could have a method to update the defaults).

idaes/models/properties/examples/saponification_reactions.py
```diff
@@ -111,12 +113,77 @@ def define_metadata(cls, obj):
...
+            # First check to see if k_rxn is already scaled
+            sf = get_scaling_factor(model.k_rxn)
+
+            if sf is not None and not overwrite:
+                # k_rxn already has a scaling factor and we are not set to overwrite - move on
+                pass
+            else:
+                # Hopefully temperature has been scaled, so we can get the nominal value of k_rxn
+                # by walking the expression in the constraint.
+                nominals = self.get_expression_nominal_values(model.arrhenius_eqn)
```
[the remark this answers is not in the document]
andrewlee94: For this specific case probably, but this also stands as an example of how to do it for a more general case.

PR #1505, opened 2024-10-18
Fix issue with enthalpy based states trying to use EoS before construciton

### LINKED ISSUE #1483: Modular Cubic EoS implementation fails with FPhx state
### Discussed in https://github.com/IDAES/idaes-pse/discussions/1452

The issue appears to be a chicken-and-egg issues with when different variables are created. Using FPhx as the state variables currently requires partial enthalpies which in turn depend on some of the cubic properties which do not get created until later in the workflow.  It appears that the issue is the construction of the sum of partial enthalpies equation which can be deferred until after the cubic properties are constructed, but it appears there may be an issue with the presence (or not) of the phase fraction variable.

<div type='discussions-op-text'>

<sup>Originally posted by **bertkdowns** July 19, 2024</sup>
I'm trying to figure out how to modify a Generic Property Package to work with different state definitions. I've got a simple flowsheet just trying to solve a heater, using a benzene-toluene mixture with ChemSep data. 

The property package works fine with the `state_definition:FTPx`, but doesn't work with `FPhx` at all.

image

I'm unsure what these errors mean, what could I do to look into them further? I've looked at the `ceos.py` file in idaes and as far as I can tell, `PR_am` should be created by the `common()` method. This method is called in the `generic_property.py` file, as part of the GenericStateBlockData build() method:

image

However, the problem seems to arise because `PR_am` is used in the define_state() method called just above on line 2419. - before the property has even been added.
image

Another part of the error message seems to say it's a misconfigured property set,
image

Is this a bug, or is it due to some misconfiguration of my state definitions & property packages?  

The script i'm building is avaliable here:
https://github.com/bertkdowns/model-predictive-control/tree/8e38e1b3323b3baea08ae94b8c141daa4e7150b7/testing_peng_robinson

and can be run with the following commands
```
git clone https://github.com/bertkdowns/model-predictive-control.git
cd model-predictive-control/testing_peng_robinson
python solve.py
```
tested using idaes-pse version 2.5.0 on python 3.11. It runs fine until you change "FTPx" for "FPhx" in the configuration.py file.
</div>

andrewlee94:
## Summary/Motivation:
Issue #1483 identified that enthalpy-based state definitions in the modular property framework triggered calls to components constructed by the equation of state module before these had been instantiated. This PR defers the construction of the constraint relating total molar enthalpy to the phase molar enthalpies until after the EoS components have been instantiated.

- Defer construction of `enth_mol_eqn` in enthalpy based state definitions.
- Fix some issues of cross-contamination in tests resulting in unexpected behaviors with this change
- Added units to partial molar enthalpy in DummyEoS to address unit consistency issue in tests

### Legal Acknowledgement

### DIFF
--- a/idaes/models/properties/modular_properties/state_definitions/tests/test_FPhx.py
+++ b/idaes/models/properties/modular_properties/state_definitions/tests/test_FPhx.py
@@ -47,6 +47,7 @@ import idaes.logger as idaeslog
 pytestmark = pytest.mark.unit
 
 
+# Note: The state definition is set by importing functions for the relevant module above
 @declare_process_block_class("DummyParameterBlock")
 class DummyParameterData(GenericParameterData):
     pass
@@ -137,7 +138,7 @@ class TestInvalidBounds(object):
 
 class Test1PhaseDefinedStateFalseNoBounds(object):
     # Test define_state method with no bounds and defined_State = False
-    @pytest.fixture(scope="class")
+    @pytest.fixture()
     def frame(self):
         m = ConcreteModel()
 
@@ -263,7 +264,7 @@ class Test1PhaseDefinedStateFalseNoBounds(object):
 
 class Test1PhaseDefinedStateTrueWithBounds(object):
     # Test define_state method with no bounds and defined_State = False
-    @pytest.fixture(scope="class")
+    @pytest.fixture()
     def frame(self):
         m = ConcreteModel()
 
@@ -389,7 +390,7 @@ class Test1PhaseDefinedStateTrueWithBounds(object):
 
 class Test2PhaseDefinedStateFalseNoBounds(object):
     # Test define_state method with no bounds and defined_State = False
-    @pytest.fixture(scope="class")
+    @pytest.fixture()
     def frame(self):
         m = ConcreteModel()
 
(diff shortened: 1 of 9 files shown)


idaes/models/properties/modular_properties/state_definitions/tests/test_FPhx.py
```diff
@@ -137,7 +138,7 @@ def test_mole_frac(self, caplog):
 
 class Test1PhaseDefinedStateFalseNoBounds(object):
     # Test define_state method with no bounds and defined_State = False
-    @pytest.fixture(scope="class")
+    @pytest.fixture()
```
dallan-keylogic, line 141: Just out of curiosity, why are we removing these scopes?

dallan-keylogic (approved):
LGTM

idaes/models/properties/modular_properties/state_definitions/tests/test_FPhx.py
andrewlee94: They resulted in tests polluting other tests which resulted in unexpected failures (mainly the first check for defined state effect had flow on effects to other tests)

PR #1510, opened 2024-10-25
Handle poorly configured streams

### LINKED ISSUE #1509: Handle poorly configured streams when building stream table
When the stream table is being built, it is possible (though probably not desirable) that some streams will not have the attribute `define_display_vars` and/or `define_state_vars`. Also it is possible that stream states may not be able to be treated as indexed components. Models can build and even solve with these conditions, so it would be better if the stream table gracefully handled them by simply not displaying information that isn't there.

This is related to the idaes-ui issue https://github.com/IDAES/idaes-ui/issues/54

dangunter:
Fixes #1509 

# Changes proposed in this PR:
- for missing `display_{state,display}_vars`, set dict to empty and issue a warning
- for states that cannot be treated as indexed components, issue a warning and return state without the index

### Legal Acknowledgement


lbianchi-lbl:
Before looking at the code changes: do you have an idea of how much work it would be to add a test exercising this error handling, i.e. a model with missing and/or invalid data that without this PR would result in an error when creating the stream table?

idaes/core/util/tables.py
```diff
@@ -235,10 +235,18 @@ class VariableTypes:
...
-        if true_state:
-            disp_dict = sb.define_state_vars()
-        else:
-            disp_dict = sb.define_display_vars()
+        try:
+            if true_state:
+                disp_dict = sb.define_state_vars()
+            else:
+                disp_dict = sb.define_display_vars()
+        except AttributeError:  # if define_(display,state)_vars doesn't exist(!)
```
andrewlee94, line 243: Have we actually seen cases of this? In order for `define_state_vars` not to exist then someone must have done something wrong and maybe we should be catching this earlier.

dallan-keylogic, line 243: Perhaps refactor to
```
try:
    if true_state:
        disp_func = sb.define_state_vars
    else:
        disp_func = sb.define_display_vars
except AttributeError:
   disp_func = None
   ....
if disp_func is not None:
    disp_dict = disp_func()
else:
    disp_dict = {}
```

That way we catch only the attribute error for the function not existing, but not attribute errors internal to the function.

dangunter: Yes, in Mayo's model

dangunter: However if you can suggest where "earlier" might be, I can try to catch it there

andrewlee94: So, I think this might need a bigger design discussion. The issue here is the the stream table methods were written only to consider things that were in "material streams", i.e. those related to state blocks in which case the `define_state_vars` method must exist.

What Mayo was doing was trying to add additional, non-state, data to the stream table which is not supported at the moment. I think a better solution to this might be to design an interface for adding this additional information rather than trying to shoehorn it into the existing interface. There were a few other catches to what Mayo was trying to do as well due to how state data is indexed versus things at the unit level that need to be resolved as well.

dallan-keylogic: Actually it looks like you can just use `getattr` and provide a default to avoid the `try`-`except` block. I'm not sure whether `None` would work as a default, though.

lbianchi-lbl:
@dangunter is still discussing whether this should be handled in the core (i.e. in this repository), or on the UI side.

lbianchi-lbl:
Closing this as we decided that it probably makes more sense for this functionality to fail "noisily" than to try to handle all possible error conditions.

PR #1512, opened 2024-10-26
Add support for Greybox model DOF counitng in degrees_of_freedom function 

avdudchenko:
## Fixes
The current degrees_of_freedom function does not properly account for GreyBoxModel as a set of constraints, resulting in an incorrect number of DOFs being presented.

## Summary/Motivation:
With increased usage of GreyBoxModel (specifically [Reaktoro-pse](https://github.com/watertap-org/reaktoro-pse)) there is a need to support DOF counting when building IDAES model with [GreyBoxModels](https://pyomo.readthedocs.io/en/stable/contributed_packages/pynumero/pynumero.interfaces.external_grey_box_model.html)

-add the capability to add unfixed input and output vars on GreyBox model to ComponentSet produced by unfixed_variables_in_activated_equalities_set
-add a counter for the number of "equalities" in gray box model in number_activated_equalities by adding  number_grey_box_equalities method 

_Counting the number of equalities in GreyBox model:_ 
GrayBox models do not provide traditional Pyomo Constraints that bind inputs to outputs as such, we need to assume how many GrayBox model DOFs there are. 

In most cases (at least I can't think of any exceptions) GreyBoxes must be 0DOF in the eyes of Pyomo/IPOPT as each output must be some sort of function of provided inputs, regardless if the internal model in the GreyBox is 0DOF, an optimization, or it exposes internal constraints to solve the problem. 
As such, it should be safe to assume that the number of outputs equals the number of equality constraints. 

### Legal Acknowledgement


idaes/core/util/model_statistics.py
```diff
@@ -377,7 +378,31 @@ def number_activated_equalities(block):
     Returns:
         Number of activated equality Constraint components in block
     """
-    return sum(1 for _ in activated_equalities_generator(block))
+    return sum(
+        1 for _ in activated_equalities_generator(block)
+    ) + number_grey_box_equalities(block)
+
+
+def number_grey_box_equalities(block):
```
dangunter, line 386: `number_of_X` is less ambiguous in English.
Can add `-> int` to indicate return type and, if it's possible, the expected type of `block`.
In docs: "Method to" is superfluous and at any rate this is a function.

idaes/core/util/model_statistics.py
```diff
@@ -377,7 +378,31 @@ def number_activated_equalities(block):
...
+    """
+    Method to return the number of equality constraints in GreyBox
```
dangunter, line 388: I would say, "Compute total number of equality constraints for all GreyBox objects in this block."

idaes/core/util/model_statistics.py
avdudchenko: done

idaes/core/util/model_statistics.py
avdudchenko: done

idaes/core/util/model_statistics.py
```diff
@@ -529,7 +555,7 @@ def deactivated_inequalities_generator(block):
         block : model to be studied
 
     Returns:
-        A generator which returns all indeactivated equality Constraint
+        A generator which returns all in deactivated equality Constraint
```
andrewlee94, line 558: The "in" should be removed here.

idaes/core/util/model_statistics.py
```diff
@@ -1034,6 +1060,19 @@ def unfixed_variables_in_activated_equalities_set(block):
     for v in variables_in_activated_equalities_set(block):
         if not v.fixed:
             var_set.add(v)
+
+    # Checks for greyboxes, and if they exist will add
+    # input and output vars to var_set if they are free
+    # inputs and outputs are defined names for greybox class and should always exist
+    for grey_box in _iter_indexed_block_data_objects(
```
andrewlee94, line 1067: I think it would be good to have a `number_of_greybox_variables` function to do this so that users can check that separately and so that we can test the code in isolation.

idaes/core/util/tests/test_model_statistics.py
```diff
@@ -685,6 +689,42 @@ def test_degrees_of_freedom(m):
     assert degrees_of_freedom(m.b2) == -1
 
 
+@pytest.mark.unit
```
andrewlee94, line 692: We should also have test for the functions to collect the number of greybox variables and constraints.

andrewlee94 (changes requested):
Thanks for this, it is a case we have not considered and Greybox models are becoming more important. To start with, I had a few suggestions for more tests and modularization of the code.

However, a more significant request would be to add these tools and (and some test cases) into the diagnostics toolbox as well. We should at a minimum check that the DoF and model statistics output (number of variables and constraints) accurately reflect the model with a Greybox present. It might also be nice to add additional sub-rows in the model statistics for greybox components when they are present. Finally, we should look for any edge cases that might occur with the rest of the toolbox and if there are other things we need to consider when using greybox models.

Robbybp:
> In most cases (at least I can't think of any exceptions) GreyBoxes must be 0DOF in the eyes of Pyomo/IPOPT as each output must be some sort of function of provided inputs, regardless if the internal model in the GreyBox is 0DOF, an optimization, or it exposes internal constraints to solve the problem.
As such, it should be safe to assume that the number of outputs equals the number of equality constraints.

IIRC, outputs just get translated to equality constraints of the form `y = GB(x)`, but the greybox can define other equality constraints as well. So the number of equality constraints is at least the number of outputs, but can be greater.

avdudchenko:
@Robbybp : Right, so you can include equality constraints in a graybox and send them to the solver (ipopt) - but the variables in those constraints are not actually in pyomo space- and thus we can't count them (hence why here I am only looking at inputs and outputs and ignoring any internal equality constraints). (e.g. looking at [this example](https://github.com/Pyomo/pyomo/blob/main/pyomo/contrib/pynumero/examples/external_grey_box/react_example/reactor_model_residuals.py))

At this point, we just want to look at the graybox as a constraint that binds exposed pyomo inputs to outputs on it, and assume that the users properly setup the internal graybox model to have the correct DOFS... ideally we would have some metadata on the graybox that tells us its internal # of DOFs so we can include it in the total count. But I opt to avoid that for now... unless we want to add support for this now.

idaes/core/util/model_statistics.py
```diff
@@ -377,7 +378,32 @@ def number_activated_equalities(block):
...
+
+    Returns:
+        Number of equality constraints in all GreyBox objects on the provided block
+    """
+    equalities = 0
+    for grey_box in _iter_indexed_block_data_objects(
+        block, ctype=ExternalGreyBoxBlock, active=True, descend_into=True
+    ):
+        equalities += len(grey_box.outputs)
+    return equalities
```
Robbybp, line 406: As written, this will be incorrect for grey box models that have a nonzero number of "explicitly defined equality constraints" (as opposed to the equality constraints that are created to define the outputs). I think all that's needed to correct this is:
```python
equalities += grey_box.get_external_model().n_equality_constraints()
```

Robbybp (changes requested):

idaes/core/util/model_statistics.py
avdudchenko: done

idaes/core/util/model_statistics.py
avdudchenko: added one for variables, and unfixed variables.

idaes/core/util/tests/test_model_statistics.py
avdudchenko: added tests

idaes/core/util/model_statistics.py
avdudchenko: Right, did not think about this scenario! Updated to test graybox with suggestion, this now includes an example constraint that binds 2 inputs together. 

```    class BasicGrayBox(ExternalGreyBoxModel):
        def input_names(self):
            return ["a1", "a2", "a3"]

        def output_names(self):
            return ["o1", "o2"]

        def equality_constraint_names(self):
            return ["a12"]

        def evaluate_equality_constraints(self):
            a1 = self._input_values[0]
            a2 = self._input_values[1]
            return [a1 * 0.5 - a2]

**Comment by avdudchenko** [comment]
> Thanks for this, it is a case we have not considered and Greybox models are becoming more important. To start with, I had a few suggestions for more tests and modularization of the code.
> 
> However, a more significant request would be to add these tools and (and some test cases) into the diagnostics toolbox as well. We should at a minimum check that the DoF and model statistics output (number of variables and constraints) accurately reflect the model with a Greybox present. It might also be nice to add additional sub-rows in the model statistics for greybox components when they are present. Finally, we should look for any edge cases that might occur with the rest of the toolbox and if there are other things we need to consider when using greybox models.

So I started adding display to diagnostics tools 
- the key question is should we include gray box stats with all others (e.g when counting blocks - should gray blocks count to total block count?, same with variables etc?, or should they be separate category only?)

File: idaes/core/util/model_statistics.py [hunk 7]
```diff
@@ -1034,6 +1060,19 @@ def unfixed_variables_in_activated_equalities_set(block):
...
+    # inputs and outputs are defined names for greybox class and should always exist
+    for grey_box in _iter_indexed_block_data_objects(
+        block, ctype=ExternalGreyBoxBlock, active=True, descend_into=True
+    ):
+        for in_var in grey_box.inputs:
+            if not grey_box.inputs[in_var].fixed:
+                var_set.add(grey_box.inputs[in_var])
+        for out_var in grey_box.outputs:
+            if not grey_box.outputs[out_var].fixed:
+                var_set.add(grey_box.outputs[out_var])
```
Robbybp: Here, I think you need to make sure that these variables don't already appear in `var_set`. If the `ExternalGreyBoxBlock` is constructed with the `inputs` or `outputs` argument, then the block will contain references to the user-provided variables, which may have already been counted.

Robbybp (changes requested):

lbianchi-lbl:
This is just waiting for @avdudchenko to extend the reporting functionality for diagnostics. This should be in the November release.

avdudchenko:
Ready for another round of reviews I upded model statistics and diagnostics to try and treat greybox models as part of full model, 
so variables and equities from greyboxes will be added with other stats as well, in addition to providing additional greybox model outputs via diagnostic toolbox which will show this:

```
    Activated Blocks: 1 (Deactivated: 0)
    Free Variables in Activated Constraints: 4 (External: 0)
        Free Variables with only lower bounds: 0
        Free Variables with only upper bounds: 0
        Free Variables with upper and lower bounds: 0
    Fixed Variables in Activated Constraints: 3 (External: 0)
    Activated Equality Constraints: 5 (Deactivated: 3)
    Activated Inequality Constraints: 0 (Deactivated: 0)
    Activated Objectives: 0 (Deactivated: 0)
    GreyBox statistics
        Activated GreyBox models: 1 (Deactivated: 1)
        Activated GreyBox Equalities: 3 (Deactivated: 3)
        Free Variables in Activated GreyBox Equalities: 4 (Fixed: 1)
```

Robbybp:
I'm curious what happens in the structural analysis when you have a greybox model. I can't imagine the under/overconstrained subsystems are handled correctly. Do you know what happens in this case?

idaes/core/util/tests/test_model_statistics.py
```diff
@@ -685,6 +689,69 @@ def test_degrees_of_freedom(m):
...
+    m.o1 = Var(initialize=1)
+    m.o1_eq = Constraint(expr=m.o1 == m.gb.outputs["o1"])
+    m.o1.fix()
+    assert degrees_of_freedom(m) == -1
+    assert number_variables_in_activated_constraints(m) == 7
+    assert number_total_constraints(m) == 5
+    assert number_total_equalities(m) == 5
+    assert number_deactivated_equalities(m) == 3
+    assert number_deactivated_constraints(m) == 3
+    report_statistics(m)
```
dallan-keylogic, line 752: What does this line test?

dallan-keylogic (commented):
Looks mostly good, just a few issues with docstrings to resolve.

dallan-keylogic:
You also should add a warning to the structural analysis tool that it's not compatible with grey box models, then make sure that `report_structural_issues` in the diagnostics toolbox has sensible results.

avdudchenko:
> You also should add a warning to the structural analysis tool that it's not compatible with grey box models, then make sure that `report_structural_issues` in the diagnostics toolbox has sensible results.

oof, I have not worked with this tool set deeply enough to even know where to start.

 Can we leave this for another PR. At this stage, I doubt many users would try to use diagnostics with Greybox models in general... so it should be okay. I would rather make a new PR after testing and finding all issues with diagnostic tools and then add appropriate doc/warnings.

Robbybp:
> Can we leave this for another PR. At this stage, I doubt many users would try to use diagnostics with Greybox models in general... so it should be okay. I would rather make a new PR after testing and finding all issues with diagnostic tools and then add appropriate doc/warnings.

I agree, but I don't want to confidently give wrong analyses if we're supporting GreyBox in one part of the tool (model statistics). Can you raise a "beta warning" or something in `report_structural_issues` and `report_numerical_issues` if greybox blocks are detected?

idaes/core/util/model_statistics.py
```diff
@@ -568,7 +571,7 @@ def deactivated_equalities_set(block):
 def number_deactivated_equalities(block):
     """
     Method to return the number of deactivated equality Constraint components
-    in a model.
+    in a model. This include number of deactivated equality constraints in Greybox models.
```
dallan-keylogic, line 574: "This include" -> "This will include"

avdudchenko:
> @avdudchenko Can you also open an issue about handling GBBs in structural and numerical diagnostics? If you provide a MWE to test on, I can try to update incidence analysis to do something reasonable.

Will do.

avdudchenko:
> > Can we leave this for another PR. At this stage, I doubt many users would try to use diagnostics with Greybox models in general... so it should be okay. I would rather make a new PR after testing and finding all issues with diagnostic tools and then add appropriate doc/warnings.
> 
> I agree, but I don't want to confidently give wrong analyses if we're supporting GreyBox in one part of the tool (model statistics). Can you raise a "beta warning" or something in `report_structural_issues` and `report_numerical_issues` if greybox blocks are detected?

I added error raising for those functions for now.

avdudchenko:
> The `Exceptions` being raised when `report_structural_issues` and `report_numerical_issues` are being called are good, but the user could still call the structural analysis tool by itself and run into issues. Why not just have a single exception when creating the `DiagnosticsToolbox` (and `SVDToolbox` and `DegeneracyHunter`) if the model has any grey box models? You'd have to comment out the tests on model statistics, but leave them in so we can use them later. Also add tests making sure the exception is raised when creating any of those objects.
> 
> We can sort out which methods support grey box models and which ones don't in another issue/PR and fine tune our warnings/errors later.

Done, I also did not remove prior tests as they directly call on report function.

idaes/core/util/model_statistics.py
```diff
@@ -1233,8 +1233,8 @@ def greybox_variables(block):
         block : model to be studied
 
     Returns:
-        A ComponentSet including all unfixed Var components which appear within
-        activated equality Constraints in block
+        A ComponentSet of all Var components which appear within
```
dallan-keylogic, line 1236: Again, are these actual `Vars`?

dallan-keylogic (approved):
LGTM

idaes/core/util/tests/test_model_diagnostics.py
```diff
@@ -527,6 +596,30 @@ def model(self):
...
+            def output_names(self):
+                return ["o1", "o2"]
+
+            def equality_constraint_names(self):
+                return ["a_sum"]
+
+            def evaluate_equality_constraints(self):
+                a1 = self._input_values[0]
+                a2 = self._input_values[1]
+                return [a1 * 0.5 + a2]
```
adam-a-a, line 615: Just wondering if this needs to be replicated within tests like you do now or if you could just create once outside of tests.

idaes/core/util/model_statistics.py
```diff
@@ -115,6 +116,104 @@ def activated_blocks_set(block):
...
+
+    Returns:
+        number of deactivated greybox blocks
+    """
+    return len(deactivated_greybox_block_set(block))
+
+
+def number_greybox_blocks(block):
+    """
+    Function to return a Number of activated Greybox Blocks components in a
```
adam-a-a, line 191:
```suggestion
    Function to return a number of Greybox Blocks components in a
```

adam-a-a: @avdudchenko since this will show up in the documentation, I figured this docstring should be unique when comparing with the docstring for `number_activated_greybox_blocks`.

adam-a-a (approved):
LGTM - I had one minor comment that you can consider or deflect. One other suggested change to commit--minor, but affects documentation. Approving regardless so as not to hold things up.

ksbeattie:
@lbianchi-lbl, I think you were going to take a look at this (when you get back)?

PR #1533, opened 2024-11-12
Prune down unused dev dependencies

### LINKED ISSUE #1532: Revisit developer dependencies and prune down `requirements-dev.txt`
- Even though we're still working on reproducing the issue, #1529 seems to be caused by `jupyter_contrib_nbextensions` erroring out during installation, preventing the rest of the packages (including `idaes-pse` itself) from being installed
- `jupyter_contrib_nbextensions` is part of the developer dependencies in `requirements-dev.txt`, but it's not clear (1) why it was originally added, and (2) whether we still need it
- From a quick look, the same can be said of other dependencies in `requirements-dev.txt`

Because of this, I think we should look through `requirements-dev.txt` and get rid of all the dependencies we're not using.

lbianchi-lbl:
### Legal Acknowledgement

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,9 +1,6 @@
---index-url https://pypi.python.org/simple/
-
 # Developer extra packages
 
 ### docs
-alabaster>=0.7.7
 # Newer sphinx needed for proper type hint support in docstrings
 sphinx>=3.0.0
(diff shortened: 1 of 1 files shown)


ksbeattie (approved):
LGTM

dangunter (approved):
LGTM

PR #1546, opened 2024-12-02
List binary platforms and architectures

dangunter:
## Summary/Motivation:
Add a command to provide some hopefully helpful output about available binary releases for IDAES 'extensions'.
The goal is to help people understand why get-extensions doesn't work for their platform. It probably won't be used by people if get-extensions *does* work without a hitch.

- Add `idaes get-extensions --info` to print the information (and do nothing else)
- Mention this option if `idaes get-extensions` itself fails

### Legal Acknowledgement


bpaul4 (approved):
Tested and confirmed that this works as intended.

mrmundt (approved):
I love this, it's beautiful and helpful.

PR #1558, opened 2025-01-15
Add support for isothermal conditions in control volumes.

andrewlee94:
## Summary/Motivation:

WateTAP has a number of cases where they would like to use isothermal conditions in place of an energy balance. Currently, they use a mix-in to support this, but this adds an addition argument that is effectively redundant. Instead, it would be better to have explicit support for isothermal conditions as part of the core ControlVolume classes.

However, isothermal conditions do not make sense for all unit models, and there are a few cases where unit models have effectively implemented their own isothermal conditions (notable PressureChanger). Thus, to avoid potential edge cases in unit models until they can be updated, a new ExtendedControlVolume0D and ExtenededControlVolume1D class have been created as an interim solution. It is envisioned that unit models will gradually be updated to be more explicit about what balance types make sense, at which point the new and old control volumes can be merged.

- Added new `isothermal` option to `EnergyBalanceType` `Enum`.
- Added `BalanceTypeNotSupported` exceptions to `ControlVolume0D` and `ControlVolume1D` if `isothermal` option is selected, with pointer to the new control volume classes.
- New `ExtendedControlVolume0D` and `ExtendedControlVolume1D` classes with support for isothermal balances.
- Docs and tests.

### Legal Acknowledgement

### DIFF
--- a/docs/reference_guides/core/control_volume_0d.rst
+++ b/docs/reference_guides/core/control_volume_0d.rst
@@ -282,3 +282,31 @@ A single pressure balance is written for the entire mixture.
 .. math:: 0 = s_{pressure} \times P_{in, t} - s_{pressure} \times P_{out, t} + s_{pressure} \times \Delta P_t + s_{pressure} \times \Delta P_{custom, t}
 
 The :math:`\Delta P_{custom, t}` term allows the user to provide custom terms  which will be added into the pressure balance.
+
+
+Extended 0D Control Volume Class
+--------------------------------
+
+The ExtendedControlVolume0DBlock block builds upon ControlVolume0DBlock by adding some new balance options. It is envisioned that this will
+merge with ControlVolume0DBlock, however to ensure backward compatibility these additions have been kept separate until unit models can
+be updated to restrict (or allow) these new options if necessary. The core functionality is the same as for ControlVolume0DBlock, with the
+addition of one extra energy balance type; isothermal.
+
+.. module:: idaes.core.base.extended_control_volume0d
+
+.. autoclass:: ExtendedControlVolume0DBlock
+    :members:
+
+.. autoclass:: ExtendedControlVolume0DBlockData
+    :members:
+
+add_isothermal_constraint
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A constraint equating temperature at the inlet and outlet of the control volume is written.
+
+**Constraints**
+
+`enthalpy_balances(t)`:
+
+.. math:: P_{in, t} == P_{out, t}
--- a/docs/reference_guides/core/control_volume_0d.rst
+++ b/docs/reference_guides/core/control_volume_0d.rst
@@ -307,6 +307,6 @@ A constraint equating temperature at the inlet and outlet of the control volume 
 
 **Constraints**
 
-`enthalpy_balances(t)`:
+`isothermal_constraint(t)`:
 
 .. math:: T_{in, t} == T_{out, t}
--- a/docs/reference_guides/core/control_volume_1d.rst
+++ b/docs/reference_guides/core/control_volume_1d.rst
@@ -311,3 +311,33 @@ The :math:`\Delta P_{custom, t, x}` term allows the user to provide custom terms
 `pressure_linking_constraint(t, x)`:
 
 This constraint is an internal constraint used to link the pressure terms in the StateBlocks into a single indexed variable. This is required as Pyomo.DAE requires a single indexed variable to create the associated DerivativeVars and their numerical expansions.
+
+
+Extended 1D Control Volume Class
+--------------------------------
+
+The ExtendedControlVolume1DBlock block builds upon ControlVolume1DBlock by adding some new balance options. It is envisioned that this will
+merge with ControlVolume1DBlock, however to ensure backward compatibility these additions have been kept separate until unit models can
+be updated to restrict (or allow) these new options if necessary. The core functionality is the same as for ControlVolume1DBlock, with the
+addition of one extra energy balance type; isothermal.
+
+.. module:: idaes.core.base.extended_control_volume1d
+
+.. autoclass:: ExtendedControlVolume1DBlock
+    :members:
+
+.. autoclass:: ExtendedControlVolume1DBlockData
+    :members:
+
+add_isothermal_constraint
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A constraint equating temperature along the length domain of the control volume is written.
+
+**Constraints**
+
+`enthalpy_balances(t)`:
+
+.. math:: P_{t, x-1} == P_{t, x}
+
+This constraint is skipped at the inlet to the control volume.
--- a/docs/reference_guides/core/control_volume_1d.rst
+++ b/docs/reference_guides/core/control_volume_1d.rst
@@ -336,7 +336,7 @@ A constraint equating temperature along the length domain of the control volume 
 
 **Constraints**
 
-`enthalpy_balances(t)`:
+`isothermal_constraint(t, x)`:
 
 .. math:: T_{t, x-1} == T_{t, x}
 
--- a/idaes/core/base/extended_control_volume0d.py
+++ b/idaes/core/base/extended_control_volume0d.py
@@ -73,7 +73,7 @@ class ExtendedControlVolume0DBlockData(ControlVolume0DBlockData):
                 Method should accept time and phase list as arguments.
 
         Returns:
-            Constraint object representing enthalpy balances
+            Constraint object representing isothermal constraint
         """
         if has_heat_transfer:
             raise ConfigurationError(
--- a/idaes/core/base/extended_control_volume0d.py
+++ b/idaes/core/base/extended_control_volume0d.py
@@ -1,0 +1,109 @@
+#################################################################################
+# The Institute for the Design of Advanced Energy Systems Integrated Platform
+# Framework (IDAES IP) was produced under the DOE Institute for the
+# Design of Advanced Energy Systems (IDAES).
+#
+# Copyright (c) 2018-2024 by the software owners: The Regents of the
+# University of California, through Lawrence Berkeley National Laboratory,
+# National Technology & Engineering Solutions of Sandia, LLC, Carnegie Mellon
+# University, West Virginia University Research Corporation, et al.
+# All rights reserved.  Please see the files COPYRIGHT.md and LICENSE.md
+# for full copyright and license information.
+#################################################################################
+"""
+0D Control Volume class with support for isothermal energy balance.
+"""
+
+__author__ = "Andrew Lee"
+
+# Import Pyomo libraries
+from pyomo.environ import Constraint, Reals, units as pyunits, Var, value
+
+# Import IDAES cores
+from idaes.core.base.control_volume0d import ControlVolume0DBlockData
+from idaes.core import declare_process_block_class
+from idaes.core.util.exceptions import ConfigurationError
+
+import idaes.logger as idaeslog
+
+_log = idaeslog.getLogger(__name__)
+
+
+@declare_process_block_class(
+    "ExtendedControlVolume0DBlock",
+    doc="""
+    ExtendedControlVolume0DBlock is an extension of the ControlVolume0D
+    block with support for isothermal conditions in place of a formal
+    energy balance.""",
+)
+class ExtendedControlVolume0DBlockData(ControlVolume0DBlockData):
+    """
+    Extended 0-Dimensional (Non-Discretized) ControlVolume Class
+
+    This class extends the existing ControlVolume0DBlockData class
+    with support for isothermal energy balances.
+    """
+
+    def add_isothermal_constraint(
+        self,
+        has_heat_of_reaction=False,
+        has_heat_transfer=False,
+        has_work_transfer=False,
+        has_enthalpy_transfer=False,
+        custom_term=None,
+    ):
+        """
+        This method constructs an isothermal constraint for the control volume.
+
+        Arguments are supported for compatibility with other forms but must be False
+        or None otherwise an Exception is raised.
+
+        Args:
+            has_heat_of_reaction: whether terms for heat of reaction should
+                be included in enthalpy balance
+            has_heat_transfer: whether terms for heat transfer should be
+                included in enthalpy balances
+            has_work_transfer: whether terms for work transfer should be
+                included in enthalpy balances
+            has_enthalpy_transfer: whether terms for enthalpy transfer due to
+                mass transfer should be included in enthalpy balance. This
+                should generally be the same as the has_mass_transfer
+                argument in the material balance methods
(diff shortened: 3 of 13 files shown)


docs/reference_guides/core/control_volume_0d.rst
```diff
@@ -282,3 +282,31 @@ A single pressure balance is written for the entire mixture.
...
+add_isothermal_constraint
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A constraint equating temperature at the inlet and outlet of the control volume is written.
+
+**Constraints**
+
+`enthalpy_balances(t)`:
+
+.. math:: P_{in, t} == P_{out, t}
```
adam-a-a, line 312:
```suggestion
.. math:: T_{in, t} == T_{out, t}
```
Did you mean T rather than P?

docs/reference_guides/core/control_volume_1d.rst
```diff
@@ -311,3 +311,33 @@ The :math:`\Delta P_{custom, t, x}` term allows the user to provide custom terms
...
+add_isothermal_constraint
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A constraint equating temperature along the length domain of the control volume is written.
+
+**Constraints**
+
+`enthalpy_balances(t)`:
+
+.. math:: P_{t, x-1} == P_{t, x}
```
adam-a-a, line 341:
```suggestion
.. math:: T_{t, x-1} == T_{t, x}
```

idaes/core/base/extended_control_volume1d.py
```diff
@@ -0,0 +1,123 @@
...
+                and x == b.length_domain.first()
+            ) or (
+                b.config.transformation_scheme == "FORWARD"
+                and x == b.length_domain.last()
+            ):
+                return Constraint.Skip
+
+            return (
+                b.properties[t, b.length_domain.prev(x)].temperature
+                == b.properties[t, x].temperature
```
adam-a-a, line 122: It looks like the constraint would be skipped at the outlet when using forward scheme, but what does this mean for temperature at the outlet?

andrewlee94: Actually, most of this is not needed as there is no partial derivative. The way this is written, it has to skip the first point no matter what.

andrewlee94: Fixed.

idaes/core/base/tests/test_control_volume_0d.py
```diff
@@ -2280,6 +2280,24 @@ def test_add_total_energy_balances():
         m.fs.cv.add_total_energy_balances()
 
 
+@pytest.mark.unit
+def test_add_isothermal_energy_balances():
+    m = ConcreteModel()
+    m.fs = Flowsheet(dynamic=False)
+    m.fs.pp = PhysicalParameterTestBlock()
+    m.fs.rp = ReactionParameterTestBlock(property_package=m.fs.pp)
+    m.fs.pp.del_component(m.fs.pp.phase_equilibrium_idx)
```
adam-a-a, line 2289: Why is this necessary?
```suggestion
    m.fs.pp.del_component(m.fs.pp.phase_equilibrium_idx)
```

idaes/core/base/tests/test_extended_control_volume_0d.py
```diff
@@ -0,0 +1,167 @@
...
+
+    assert isinstance(m.fs.cv.enthalpy_balances, Constraint)
+    assert len(m.fs.cv.enthalpy_balances) == 4
+    for t in m.fs.time:
+        assert str(m.fs.cv.enthalpy_balances[t].expr) == str(
+            m.fs.cv.properties_in[t].temperature
+            == m.fs.cv.properties_out[t].temperature
+        )
+
+    assert_units_consistent(m.fs.cv)
```
adam-a-a, line 86: Interesting- I thought I remember the units check failing for dynamic cases, or is that only when has_holdup=True as well?

idaes/core/base/tests/test_extended_control_volume_0d.py
```diff
@@ -0,0 +1,167 @@
...
+    m.fs = Flowsheet(dynamic=False)
+    m.fs.pp = PhysicalParameterTestBlock()
+
+    m.fs.cv = ExtendedControlVolume0DBlock(property_package=m.fs.pp)
+
+    with pytest.raises(
+        ConfigurationError,
+        match="fs.cv: isothermal energy balance option requires that has_heat_transfer is False. "
+        "If you are trying to solve for heat duty to achieve isothermal operation, please use "
+        "a full energy balance as add a constraint to equate inlet and outlet temperatures.",
```
adam-a-a, line 101: This error message needs some revision: "a full energy balance **as add a constraint** to equate inlet and outlet temperatures."

idaes/core/base/tests/test_extended_control_volume_0d.py
andrewlee94: It will for anything with a partial derivative. Here we get aways with it because we don't have a derivative.

idaes/core/base/tests/test_control_volume_0d.py
andrewlee94: Cleaned up

idaes/core/base/tests/test_extended_control_volume_0d.py
```diff
@@ -0,0 +1,167 @@
...
+    m.fs = Flowsheet(dynamic=False)
+    m.fs.pp = PhysicalParameterTestBlock()
+
+    m.fs.cv = ExtendedControlVolume0DBlock(property_package=m.fs.pp)
+
+    with pytest.raises(
+        ConfigurationError,
+        match="fs.cv: isothermal energy balance option requires that has_work_transfer is False. "
+        "If you are trying to solve for work under isothermal operation, please use "
+        "a full energy balance as add a constraint to equate inlet and outlet temperatures.",
```
adam-a-a, line 118: Same issue- just need to revise this segment of the error message and propagate through.

idaes/core/base/tests/test_extended_control_volume_0d.py
andrewlee94: Fix: "and add"

idaes/core/base/tests/test_extended_control_volume_0d.py
andrewlee94: As above.

idaes/core/base/tests/test_extended_control_volume_1d.py
```diff
@@ -0,0 +1,203 @@
...
+        finite_elements=10,
+    )
+
+    m.fs.cv.add_geometry()
+    m.fs.cv.add_state_blocks(has_phase_equilibrium=False)
+
+    m.fs.cv.add_isothermal_constraint()
+
+    assert isinstance(m.fs.cv.enthalpy_balances, Constraint)
+    assert len(m.fs.cv.enthalpy_balances) == 1  # x==0 is skipped
```
adam-a-a, line 62: I thought the length of enthalpy_balances would be fine_elements - 1. Maybe I am missing something (or the test is failing).

andrewlee94: The model has not been discretised, so there are only 2 points.

idaes/core/base/tests/test_extended_control_volume_1d.py
```diff
@@ -0,0 +1,203 @@
...
+        finite_elements=10,
+    )
+
+    m.fs.cv.add_geometry()
+    m.fs.cv.add_state_blocks(has_phase_equilibrium=False)
+
+    m.fs.cv.add_isothermal_constraint()
+
+    assert isinstance(m.fs.cv.enthalpy_balances, Constraint)
+    assert len(m.fs.cv.enthalpy_balances) == 4  # x==0 is skipped
```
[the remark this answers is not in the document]
adam-a-a: finite elements=10 not 2 here

idaes/core/base/tests/test_extended_control_volume_1d.py
adam-a-a: Oh! I see now. Didn't catch that--didn't apply transformation.

idaes/core/base/tests/test_extended_control_volume_1d.py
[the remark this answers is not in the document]
adam-a-a: OK, just saw your previous response. Any reason not to apply transformation and check?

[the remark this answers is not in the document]
andrewlee94: Reduce test overhead.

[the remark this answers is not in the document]
adam-a-a: Could be good just to do a transformation with, say, 3 finite elements.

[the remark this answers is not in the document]
andrewlee94: I debated it, but decided that was as much a test of Pyomo functionality as of IDAES (and thus I was going to trust Pyomo).

adam-a-a (approved):
LGTM - I don't think it would hurt too much to add a test with transformation applied to 1d CV (with 2 - 3 finite elements), but I'll leave that to other reviewers and @andrewlee94

idaes/core/base/tests/test_extended_control_volume_1d.py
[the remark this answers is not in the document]
andrewlee94: I realised there is a reason to test this, but not for the obvious reason - when using the `domain.prev()` method it works on the domain as it appears at the time of the call, meaning that it will always create `T[0, 0] == T[0, 1]` and then fill in intermediate points. In this case, it should not matter, but it needs to be tested.

idaes/core/base/tests/test_extended_control_volume_0d.py
dallan-keylogic: Not necessarily. Generally for spatial domains we normalize the `ContinuousSet` to run from 0 to 1, adding in the flow length later. Those should pass. Any `ContinuousSet` with dimensions, however, will cause it to fail.

idaes/core/base/extended_control_volume0d.py
```diff
@@ -0,0 +1,106 @@
...
+    with support for isothermal energy balances.
+    """
+
+    def add_isothermal_constraint(
+        self,
+        has_heat_of_reaction=False,
+        has_heat_transfer=False,
+        has_work_transfer=False,
+        has_enthalpy_transfer=False,
+        custom_term=None,
```
dallan-keylogic, line 50: Should we add Type Hints here?

idaes/core/base/extended_control_volume0d.py
```diff
@@ -0,0 +1,106 @@
...
+            )
+        if custom_term is not None:
+            raise ConfigurationError(
+                f"{self.name}: isothermal energy balance option does not support custom terms."
+            )
+
+        # Add isothermal constraint
+        @self.Constraint(self.flowsheet().time, doc="Energy balances")
+        def enthalpy_balances(b, t):
+            return b.properties_in[t].temperature == b.properties_out[t].temperature
```
dallan-keylogic, line 106: Can we rename this constraint to something that better describe what it does? Otherwise you might, for example, get enthalpy scaling on a constraint that should have temperature scaling.

idaes/core/base/extended_control_volume1d.py
dallan-keylogic: Can we also rename `enthalpy_balances` to `isothermal_constraint` or something here?

dallan-keylogic (commented):
Some minor changes, otherwise it looks good.

idaes/core/base/extended_control_volume0d.py
andrewlee94: Might as well - if we don't do it now it probably won't get done.

idaes/core/base/extended_control_volume0d.py
andrewlee94: I debated whether the change the name - my reason for keeping it the same was for consistency of naming between different configurations (so that the "energy balance" would always have the same name no matter what option you chose).

However, I did not think about scaling, and having a descriptive name would be far more useful there for people doing custom scaling. I'll change it.

docs/reference_guides/core/control_volume_0d.rst
```diff
@@ -282,3 +282,31 @@ A single pressure balance is written for the entire mixture.
...
+add_isothermal_constraint
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A constraint equating temperature at the inlet and outlet of the control volume is written.
+
+**Constraints**
+
+`enthalpy_balances(t)`:
+
+.. math:: T_{in, t} == T_{out, t}
```
dallan-keylogic, line 312: This is now outdated

docs/reference_guides/core/control_volume_1d.rst
```diff
@@ -311,3 +311,33 @@ The :math:`\Delta P_{custom, t, x}` term allows the user to provide custom terms
...
+add_isothermal_constraint
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A constraint equating temperature along the length domain of the control volume is written.
+
+**Constraints**
+
+`enthalpy_balances(t)`:
+
+.. math:: T_{t, x-1} == T_{t, x}
```
dallan-keylogic, line 341: Outdated.

idaes/core/base/extended_control_volume0d.py
```diff
@@ -0,0 +1,112 @@
...
+            has_enthalpy_transfer: whether terms for enthalpy transfer due to
+                mass transfer should be included in enthalpy balance. This
+                should generally be the same as the has_mass_transfer
+                argument in the material balance methods
+            custom_term: a Python method which returns Pyomo expressions representing
+                custom terms to be included in enthalpy balances.
+                Method should accept time and phase list as arguments.
+
+        Returns:
+            Constraint object representing enthalpy balances
```
dallan-keylogic, line 76: Might be good to update this to reflect isothermal constraint.

idaes/core/base/extended_control_volume1d.py
```diff
@@ -0,0 +1,121 @@
...
+            has_enthalpy_transfer: whether terms for enthalpy transfer due to
+                mass transfer should be included in enthalpy balance. This
+                should generally be the same as the has_mass_transfer
+                argument in the material balance methods
+            custom_term: a Python method which returns Pyomo expressions representing
+                custom terms to be included in enthalpy balances.
+                Method should accept time and phase list as arguments.
+
+        Returns:
+            Constraint object representing enthalpy balances
```
dallan-keylogic, line 77: Might be good to update this to reflect isothermal constraint.

dallan-keylogic (commented):
A few more documentation quibbles.

dallan-keylogic (approved):
LGTM

Issue #1560, opened 2025-01-21
Small bug in CubicComplementarityVLE calculate_teq method


Mahakileach:
**Issue**
Within [`idaes\models\properties\modular_properties\phase_equil\smooth_VLE_2.py`](https://github.com/IDAES/idaes-pse/blob/d43c5dd43da766be7b309c6d058ec672aa66cf2a/idaes/models/properties/modular_properties/phase_equil/smooth_VLE_2.py#L51) the [`calculate_teq(blk, pp)`](https://github.com/IDAES/idaes-pse/blob/d43c5dd43da766be7b309c6d058ec672aa66cf2a/idaes/models/properties/modular_properties/phase_equil/smooth_VLE_2.py#L232) method checks if [`v_only_comps is None`](https://github.com/IDAES/idaes-pse/blob/d43c5dd43da766be7b309c6d058ec672aa66cf2a/idaes/models/properties/modular_properties/phase_equil/smooth_VLE_2.py#L250-L270). The implementation of [`identify_VL_component_list(blk, phase_pair)`](https://github.com/IDAES/idaes-pse/blob/d43c5dd43da766be7b309c6d058ec672aa66cf2a/idaes/models/properties/modular_properties/base/utility.py#L307) always returns a list and as such `v_only_comps is None` always evaluate to False.

**Proposed fix**
Replace `v_only_comps is None` with `v_only_comps == []` or `len(v_only_comps) == 0`

dallan-keylogic:
Fixed in #1563 .

PR #1563, opened 2025-01-30
Fix bugs in SmoothVLE2 initialization

dallan-keylogic:
## Fixes
- Able to initialize using `SmoothVLE2` when non-condensables are present.
- Correctly takes into account presence of non-volatiles in dew point calculations

## Summary/Motivation:
Addresses #1560 .

### Legal Acknowledgement


idaes/models/properties/modular_properties/phase_equil/tests/test_smooth_VLE_2.py
```diff
@@ -339,3 +343,81 @@ def test_calculate_ceos_derivative_slacks(frame):
...
+
+
+@pytest.mark.unit
+def test_calculate_teq_permanent_gas(H2O_H2_model):
+    m = H2O_H2_model
+    m.props[1].pressure.set_value(1e5)
+    m.props[1].temperature.set_value(300)
+    CubicComplementarityVLE.calculate_teq(m.props[1], ("Vap", "Liq"))
+    assert not m.props[1].is_property_constructed("temperature_dew")
+    assert not m.props[1].is_property_constructed("temperature_bubble")
```
bpaul4, line 423: It appears that the framework does not create these equilibrium properties because there is only one VLE component in the system, which makes sense. Are there tests for other scenarios, for example one or more non-condensable or non-volatile components in a system with two or more VLE components? In those cases, these properties should still be created.

dallan-keylogic: The `SmoothVLE2` method doesn't need bubble and dew points to operate, unlike the original `SmoothVLE`. The test coverage here is pretty poor, but I wanted to fix the outstanding issue quickly before moving onto other things.

idaes/models/properties/modular_properties/phase_equil/tests/test_smooth_VLE_2.py
```diff
@@ -339,3 +343,162 @@ def test_calculate_ceos_derivative_slacks(frame):
 
     assert value(gp["Vap"]) == pytest.approx(EPS_INIT, abs=1e-8)
     assert value(gn["Vap"]) == pytest.approx(EPS_INIT, abs=1e-8)
+
+
+class TestWithNonCondensable:
+    # TODO These tests could be fleshed out
```
agarciadiego, line 349: What would this TODO entail?

dallan-keylogic: Mostly adding in tests for the values of `gp`, `gn`, `_teq`, and other VLE variables both to make sure the model isn't changing and that the answer is reasonable. Right now, I'm testing that 1) the model initialization doesn't end up with an error and 2) bubble/dew points are not erroneously created.

bpaul4 (approved):

agarciadiego (approved):

PR #1566, opened 2025-02-11
Sparse SVD

dallan-keylogic:
## Summary/Motivation:
Presently, the `SVDToolbox` isn't widely utilized, in part because of the limited options for calculating the SVD. Scipy's dense SVD is  time-consuming on problems larger than 1000x1000, while its sparse SVD does not work well to find the singular values of the smallest magnitude. 

Therefore, I've implemented a custom version of the SVD that is based on Rayleigh-Ritz iteration.

I'd appreciate testers for this method before we merge it.

### Legal Acknowledgement


dallan-keylogic:
Pylint is complaining about "Possible unbalanced tuple unpacking". The source of the issue is that the function returns three variables, U, svals, and V, if m=n but it additionally returns a basis for the (left) null space if m!=n. There are multiple ways to make this warning go away, but I want some feedback about the return signature of the function. Should I always return the null space, or only if a specific option is included?

blnicho:
> Pylint is complaining about "Possible unbalanced tuple unpacking". The source of the issue is that the function returns three variables, U, svals, and V, if m=n but it additionally returns a basis for the (left) null space if m!=n. There are multiple ways to make this warning go away, but I want some feedback about the return signature of the function. Should I always return the null space, or only if a specific option is included?

My recommendation is to always return the same number of items from the function. We've been burned in the past with APIs that return different numbers of things based on arguments/options passed to the function.

ksbeattie:
Blocked on funding

Issue #1577, opened 2025-02-23
Issue within state initialisation of modular FTPx state definition


Mahaki-Leach:
**Issue**

Within the [modular properties FTPx state initialization](https://github.com/IDAES/idaes-pse/blob/ba3c5d4ae258d77faf2440360efd8337afef1ed6/idaes/models/properties/modular_properties/state_definitions/FTPx.py#L362) on [lines 412 to 423](https://github.com/IDAES/idaes-pse/blob/ba3c5d4ae258d77faf2440360efd8337afef1ed6/idaes/models/properties/modular_properties/state_definitions/FTPx.py#L412-L423) the following logic is written.
```
if len(vl_comps) > 0:
    # More than one VLE. Just use the default initialization for
    # now
    init_VLE = False
(
    l_phase,
    v_phase,
    vl_comps,
    henry_comps,
    l_only_comps,
    v_only_comps,
) = identify_VL_component_list(b, pp)
```

```vl_comps``` is only assigned after ```len(vl_comps) > 0``` is checked. As ```vl_comps``` is initialised to ```[]``` and left unmodified before this comparison, the VLE initialisation is never skipped for systems with more than one VLE.

**proposed fix**

Check ```len(vl_comps) > 0``` after ```identify_VL_component_list(b, pp)```

```
(
    l_phase,
    v_phase,
    vl_comps,
    henry_comps,
    l_only_comps,
    v_only_comps,
) = identify_VL_component_list(b, pp)
if len(vl_comps) > 0:
    # More than one VLE. Just use the default initialization for
    # now
    init_VLE = False
```

Mahaki-Leach:
Ah, i see. Thanks for the clarification 😃

PR #1578, opened 2025-02-24
VLE Init Edge Cases

dallan-keylogic:
## Fixes
-- Corrects VLE initialization behavior when only Henry components are present
-- Raises an exception when given a VLE with no components in both phases

## Summary/Motivation:
Addresses #1577

### Legal Acknowledgement


bpaul4 (approved):
This looks good to me.

idaes/models/properties/modular_properties/base/tests/test_utility.py
```diff
@@ -721,6 +720,102 @@ def test_all_types_components(self):
...
+            pressure_ref=100000.0,
+            temperature_ref=300,
+            base_units=self.base_units,
+        )
+
+        m.props = m.params.build_state_block([1], defined_state=False)
+        with pytest.raises(
+            PropertyPackageError,
+            match="Phase pair p1-p2 was identified as "
+            "being a VLE pair, however are no components present in both "
```
dallan-keylogic, line 814:
```suggestion
            "being a VLE pair, however there are no components present in both "
```

idaes/models/properties/modular_properties/state_definitions/FTPx.py
```diff
@@ -423,7 +418,7 @@ def state_initialization(b):
             ) = identify_VL_component_list(b, pp)
             pp_VLE = pp
 
-    if init_VLE:
+    if num_VLE == 1:  # Only support initialization when a single VLE is present
```
adam-a-a, line 421: Should there be an `else` clause if num_VLE somehow exceeds 1, where either an exception is raised or a warning is logged?

adam-a-a (commented):
I only have one question on whether we should handle the scenario where num_VLE > 1 by raising an exception or at least logging a warning. It may be the case that num_VLE can never be >1 because it is caught elsewhere, but I didn't do a deep enough to dive to check that.

dallan-keylogic:
> I only have one question on whether we should handle the scenario where num_VLE > 1 by raising an exception or at least logging a warning. It may be the case that num_VLE can never be >1 because it is caught elsewhere, but I didn't do a deep enough to dive to check that.

It's possible, in theory, for there to be multiple VLE if we have a vapor/liquid/liquid system. That's why the exception isn't raised. Such a system isn't explicitly supported at the moment, but a user who knows what they're doing could probably set one up.

adam-a-a:
> > I only have one question on whether we should handle the scenario where num_VLE > 1 by raising an exception or at least logging a warning. It may be the case that num_VLE can never be >1 because it is caught elsewhere, but I didn't do a deep enough to dive to check that.
> 
> It's possible, in theory, for there to be multiple VLE if we have a vapor/liquid/liquid system. That's why the exception isn't raised. Such a system isn't explicitly supported at the moment, but a user who knows what they're doing could probably set one up.

In that case, we should at least log a warning (or possibly raise a `NotImplementedError` that an advanced user could bypass with a try-except clause).

dallan-keylogic:
At this point, I'm inclined to merge this with or without Codecov's approval. I first created a warning to implement @adam-a-a 's suggestion, but I found I couldn't test it because initialization throws an exception if constraints have large residuals at the end. Before that, however, I had to correct an issue in which `results` was undefined in the initialization routine for the tortured model I had to use to test this issue.

PR #1603, opened 2025-04-01
Add Smooth Heaviside Function to Math

MarcusHolly:
## Summary/Motivation:
Addresses a comment made [here](https://github.com/watertap-org/watertap/pull/1563#discussion_r2019371484) to add the smooth_heaviside function into IDAES rather than WaterTAP

- Adds the smooth_heaviside function to math
- Updates the testing

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/math.py
+++ b/idaes/core/util/math.py
@@ -197,10 +197,10 @@ def smooth_heaviside(x, k):
     """
     Provides a smooth, continuous approximation of the Heaviside step function using the logistic function
     Args:
-        x : Independent variable; as x increases, the function decreases from 1 to 0
+        x : Independent variable; as x increases, the function increases from 0 to 1
         k : Smoothing parameter proportional to the slope of the discontinuity
     Returns:
-        function : Continuous approximation of a discontinuous function which approximately equals 0 when x > 0 and approximately equals 1 when x < 0
+        function : Continuous approximation of a discontinuous function which approximately equals 0 when x < 0 and approximately equals 1 when x > 0
     """
-    function = 1 / (1 + exp(k * x))
+    function = 1 / (1 + exp(-k * x))
     return function
--- a/idaes/core/util/math.py
+++ b/idaes/core/util/math.py
@@ -197,10 +197,10 @@ def smooth_heaviside(x, k):
     """
     Provides a smooth, continuous approximation of the Heaviside step function using the logistic function
     Args:
-        x : Independent variable; as x increases, the function increases from 0 to 1
+        x : Independent variable; as x increases, the function decreases from 1 to 0
         k : Smoothing parameter proportional to the slope of the discontinuity
     Returns:
-        function : Continuous approximation of a discontinuous function which approximately equals 1 when x > 0 and approximately equals 0 when x < 0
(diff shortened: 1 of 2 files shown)


idaes/core/util/math.py
```diff
@@ -191,3 +191,16 @@ def safe_log(a, eps=1e-4):
...
+
+
+def smooth_heaviside(x, k):
+    """
+    Provides a smooth, continuous approximation of a discontinuous function
+    Args:
+        x : Independent variable; as x increases, the function increases from 0 to 1
+        k : Smoothing parameter representing the slope of the discontinuity
+    Returns:
+        function : Continuous approximation of a discontinuous function
```
dallan-keylogic, line 203: Function which approximately equals 1 when x > 0 and approximately equals 0 when x < 0.

idaes/core/util/math.py
```diff
@@ -191,3 +191,16 @@ def safe_log(a, eps=1e-4):
...
+    """
+    function = 1 / (1 + exp(-2 * k * x))
```
dallan-keylogic, line 205: Out of curiosity, why `-2 * k * x` instead of just `-k * x`? I thought it might be to get the derivative to equal 1 at `x=0`, but that requires `-4*k*x` instead.

(The answer might be that you got it from me during our original discussions, in which case I have no idea why I chose it.)

idaes/core/util/math.py
```diff
@@ -191,3 +191,16 @@ def safe_log(a, eps=1e-4):
         approximately log(max(a, eps))
     """
     return log(smooth_max(a, eps, eps=eps))
+
+
+def smooth_heaviside(x, k):
+    """
+    Provides a smooth, continuous approximation of a discontinuous function
```
dallan-keylogic, line 198: "Provides a smooth, continuous approximation of the Heaviside step function using the logistic function"

dallan-keylogic (commented):
Could we augment this function with an additional function that is a drop-in replacement for `expr_if`? Maybe have two `expr_if` replacements, depending on whether the underlying expression is continuous (when we can use `smooth_max` or `smooth_min` or discontinuous, when we use `smooth_heaviside`.

idaes/core/util/math.py
MarcusHolly: Yea, this was pulled from our original discussions. That being said, I can (and probably should) still remove it so that we're not propagating numbers that we can't explain.

MarcusHolly:
> Could we augment this function with an additional function that is a drop-in replacement for `expr_if`? Maybe have two `expr_if` replacements, depending on whether the underlying expression is continuous (when we can use `smooth_max` or `smooth_min` or discontinuous, when we use `smooth_heaviside`.

After thinking about this a bit more (and trying to implement it), I'm finding it difficult to cover all the possible scenarios since there's not necessarily uniformity in how the conditional statement is set up (number of terms on the LHS and RHS). If we do assume a simple if x > y or x < y, I feel that it is fairly straightforward for the users to replace the Expr_if themselves. That way we don't need to keep Expr_ifs in the code at all (i.e. "the drop-in replacement" function would need to be passed the Expr_if)

dallan-keylogic (approved):
LGTM

PR #1606, opened 2025-04-10
Add Solver Config Option to IPOPT Convergence Analysis

MarcusHolly:
## Summary/Motivation:
Addresses an issue in https://github.com/watertap-org/watertap/pull/1563 where tests are passing locally but failing on GitHub. This is believed to be a result of this tool defaulting the solver to ipopt, whereas WaterTAP requires ipopt-watertap.

- Adds a config option to IPOPTConvergenceAnalysis

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/model_diagnostics.py
+++ b/idaes/core/util/model_diagnostics.py
@@ -3559,7 +3559,7 @@ class IpoptConvergenceAnalysis:
 
     CONFIG = CACONFIG()
 
-    def __init__(self, model, **kwargs):
+    def __init__(self, model, solver=SolverFactory("ipopt"), **kwargs):
         # TODO: In future may want to generalise this to accept indexed blocks
         # However, for now some of the tools do not support indexed blocks
         if not isinstance(model, BlockData):
@@ -3575,7 +3575,8 @@ class IpoptConvergenceAnalysis:
 
         self._model = model
 
-        solver = SolverFactory("ipopt")
+        self._solver = solver
+
         if self.config.solver_options is not None:
             solver.options = self.config.solver_options
 
--- a/idaes/core/util/model_diagnostics.py
+++ b/idaes/core/util/model_diagnostics.py
@@ -3559,7 +3559,7 @@ class IpoptConvergenceAnalysis:
 
     CONFIG = CACONFIG()
 
-    def __init__(self, model, solver=SolverFactory("ipopt"), **kwargs):
+    def __init__(self, model, solver=None, **kwargs):
         # TODO: In future may want to generalise this to accept indexed blocks
         # However, for now some of the tools do not support indexed blocks
         if not isinstance(model, BlockData):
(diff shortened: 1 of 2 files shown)


idaes/core/util/model_diagnostics.py
```diff
@@ -3559,7 +3559,7 @@ class IpoptConvergenceAnalysis:
 
     CONFIG = CACONFIG()
 
-    def __init__(self, model, **kwargs):
+    def __init__(self, model, solver=SolverFactory("ipopt"), **kwargs):
```
dallan-keylogic, line 3562: Having mutable Python objects as default arguments causes trouble, because it uses the _same_ object each time the function is called.

```
In [1]: def append_a(lst=[]):
   ...:     lst.append("a")
   ...:     return lst
   ...:

In [2]: lst = append_a()

In [3]: lst = append_a()

In [4]: lst
Out[4]: ['a', 'a']
```
We might expect the second instance of `lst` to contain a single element because it's using the default argument of `[]`, but because it uses the same instance of the list object each time the function is called, it contains two elements.

The way we work around this is by using `None` as the default argument, then creating an object if the argument is `None`:

```
def __init__(self, model, solver=None, **kwargs):
    if solver is None:
        solver = SolverFactory("ipopt")
```

Also, if you expect a solver object to be passed, you should make `solver_obj` the name of the function argument. It would be even better to add it as a config argument instead of an argument to `__init__`.

I assume that `ipopt_watertap` has some set of default options different than those in IDAES.

dallan-keylogic (changes requested):
Please avoid mutable objects as default arguments.

idaes/core/util/model_diagnostics.py
MarcusHolly: Thanks Doug! I didn't realize default arguments for each instance builds on top of the previous ones. I've set the default solver argument to None.

I'm not too familiar with the differences between ipopt and ipopt-watertap other than the fact that using ipopt has caused issues in the past.

idaes/core/util/model_diagnostics.py
```diff
@@ -3559,7 +3559,7 @@ class IpoptConvergenceAnalysis:
 
     CONFIG = CACONFIG()
 
-    def __init__(self, model, **kwargs):
+    def __init__(self, model, solver=None, **kwargs):
```
dallan-keylogic, line 3562: Could you please also change the name from `solver` to `solver_obj`? Typically if we had `solver` as the default argument, we'd pass the string `"ipopt"` instead of an IPOPT solver object.

dallan-keylogic (changes requested):

idaes/core/util/model_diagnostics.py
agarciadiego: You can find WaterTap solver options here https://github.com/watertap-org/watertap-solvers/blob/main/src/watertap_solvers/_base.py

agarciadiego (approved):

adam-a-a (approved):

PR #1579, opened 2025-02-25
Add Clustering Method to PriceTaker

MarcusHolly:
## Summary/Motivation:
Adds a literature-based method to the pricetaker model for determining the optimal number of clusters that a dataset should be broken into

## Changes proposed in this PR:
- Adds the clustering method
- Updates tests
- Addresses some of Adam's miscellaneous feedback in #1358 

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1633, opened 2025-06-10
Add Storage and Surrogate Support to the Price-taker class

radhakrishnatg:
## Summary/Motivation:
This PR adds new features to the price-taker class. This is WIP. This PR needs to be merged after merging #1579 

- Fix a bug: Capacity limit constraints, startup-shutdown constraints, etc. would not work for indexed operation blocks. This is fixed in this PR.
- Added methods for retrieving all operational and design variable values as `pd.DataFrame` and `dict`, respectively.
- Added a method for plotting operational profiles.
- Added a method for plotting LMP histograms
- Added a generalized storage model class. If an instance of `StorageModel` is found in the flowsheet instance, then linking constraints are now added automatically.
- Added support for automatic construction of polynomial-type surrogates.

### Legal Acknowledgement

### DIFF
--- a/idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
+++ b/idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
@@ -11,13 +11,78 @@
 # for full copyright and license information.
 #################################################################################
 
-from pyomo.environ import Binary, Param, Var
-from pyomo.common.config import Bool, ConfigDict, ConfigValue
+from pyomo.environ import Binary, Constraint, Expression, NonNegativeReals, Param, Var
+from pyomo.common.config import (
+    Bool,
+    ConfigDict,
+    ConfigValue,
+    IsInstance,
+    NonNegativeFloat,
+)
 from idaes.core.base.process_base import declare_process_block_class
 from idaes.core.base.process_base import ProcessBlockData
+from idaes.core.util.config import ConfigurationError, is_in_range
 import idaes.logger as idaeslog
 
 _logger = idaeslog.getLogger(__name__)
+
+
+def _format_data(coeffs):
+    """
+    Helper function to correctly format the surrogate model coefficients
+    """
+    if isinstance(coeffs, (int, float, Param, Var)):
+        # Here, correlation is assumed to be y = a_1 * design_var
+        return [0, coeffs]
+
+    if isinstance(coeffs, (list, tuple)):
+        # Here, correlation is assumbe to be
+        # y = a_0 + a_1 * design_var + a_2 * design_var**2 + ...
+        return coeffs
+
+    raise ConfigurationError(
+        f"Unrecognized data structure {coeffs} for auxiliary variable coefficients."
+    )
+
+
+def is_valid_variable_design_data(data: dict):
+    """Validates the arguments received for the variable design case"""
+    # Ensure that design_var and design_var_bounds are present
+    if "design_var" not in data:
+        raise ConfigurationError("design_var is not specified")
+
+    if "design_var_bounds" not in data:
+        raise ConfigurationError("design_var_bounds is not specified")
+
+    new_data = {
+        "design_var": data.pop("design_var"),
+        "design_var_bounds": data.pop("design_var_bounds"),
+        "auxiliary_vars": {},
+    }
+
+    # The rest of the variables are assumed to be correlations
+    for var_name, coeff in data.items():
+        new_data["auxiliary_vars"][var_name] = _format_data(coeff)
+
+    return new_data
+
+
+def is_valid_polynomial_surrogate_data(data: dict):
+    """Validates the arguments received for the variable design case"""
+    # Ensure that operation_var is present present
+    if "operation_var" not in data:
+        raise ConfigurationError("operation_var is not specified")
+
+    new_data = {
+        "operation_var": data.pop("operation_var"),
+        "auxiliary_vars": {},
+    }
+
+    # The rest of the variables are assumed to be correlations
+    for var_name, coeff in data.items():
+        new_data["auxiliary_vars"][var_name] = _format_data(coeff)
+
+    return new_data
 
 
 # pylint: disable = attribute-defined-outside-init, too-many-ancestors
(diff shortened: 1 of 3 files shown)


idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
```diff
@@ -80,6 +145,20 @@ def my_design_model(m, p_min, p_max, cost):
             doc="Dictionary containing arguments needed for model_func",
         ),
     )
+    CONFIG.declare(
+        "fixed_design_data",
+        ConfigValue(
+            domain=dict,
+            doc="Dictionary containing parameters associated with unit/process design",
```
adam-a-a, line 152: doc string is exactly same for "variable_design_data". Please update so the user can distinguish the purpose of having these as two separate config options

adam-a-a: Ah- there is a minor difference--- this mentions "parameters" while the latter mentions "variables"

idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
```diff
@@ -89,7 +168,20 @@ def build(self):
             doc="Binary: 1, if the unit is installed, 0 otherwise",
         )
 
-        if self.config.model_func is None:
+        if self.config.fixed_design_data is not None:
+            # The design is fixed, so all the desired quantities are defined as parameters
+            for param_name, param_value in self.config.fixed_design_data.items():
+                setattr(self, param_name, Param(initialize=param_value, mutable=True))
```
adam-a-a, line 174: Any reason why you chose mutable Params over fixing Vars ?

idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
```diff
@@ -213,14 +346,174 @@ def build(self):
...
+                    Constraint(expr=getattr(self, expr_name) == expr),
+                )
+
+            else:
+                # Declare the expression as a Pyomo Expression
+                setattr(self, expr_name, expr)
+
+
+@declare_process_block_class("StorageModel")
+class StorageModelData(ProcessBlockData):
```
adam-a-a, line 502: Is there any reason why we don't just have this in a separate file?

idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
```diff
@@ -213,14 +346,174 @@ def build(self):
...
+                    ),
+                )
+
+            else:
+                getattr(self, var_name).setub(bound)
+
+        _add_upper_bound("charge_rate", self.config.max_charge_rate)
+        _add_upper_bound("discharge_rate", self.config.max_discharge_rate)
+        _add_upper_bound("final_holdup", self.config.max_holdup)
+        _add_upper_bound("initial_holdup", self.config.max_holdup)
```
adam-a-a, line 595: If I understand correctly, this `_add_upper_bound` function essentially provides the upper bound for a variable by adding a constraint (named m.charge_rate_ub_con in the case of charge_rate, for example) when a Var or Expression is provided.

However, when a float, int, etc., is used, this constraint would not be constructed and the upper bound of the Var would just be set directly. 

While it seems like a trivial and subtle difference, it also seems less than ideal to allow variation in model structure like this (where one version includes upper bound constraints and another analogous version does not, simply based on input provided). 

Is there a reason why we don't just (1) check whether instance is a Var or Expression, and if so, just do `setub(pyo.value(bound))` ? The only potential issue that I can foresee with that is if an indexed Var or Expression is passed in, but haven't checked. Are there other reasons?

idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
radhakrishnatg: Not quite sure which option is preferable. If we define a `Var` and fix it, it would allow the user to `unfix` it. If they do that, then the design variable would be unconstrained and it may lead to an unexpected behavior.
I feel that declaring it as a `Param` would prevent unfixing and it ensures that the design is always fixed. I can change it to fixing Var if that is the convention we are following everywhere.

ksbeattie:
@radhakrishnatg should now be able to revisit this.

idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
radhakrishnatg: It is a short piece of code, so I just retained it in the same file. Anyways, users will be importing models from `idaes.apps.grid_integration`, so I thought it doesn't matter if we put it in this file or in a separate file.

idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
radhakrishnatg: The `Var` and `Expression` are needed for the case where the design is a decision variable. Since the value of the variable is unknown, `setub(pyo.value(bound))` will not work.
I agree that having different behavior for int/float vs. Var/Expression is not ideal. But, we cannot avoid it. To keep it consistent, we can implement all them as constraints regardless of the data type. But, that may impact the performance of some solvers. So, for now, I thought we can leave it as it is.

idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
```diff
@@ -117,6 +206,38 @@ def build(self):
             )
             self.fom = 0
 
+    def _build_variable_design_model(self):
+        data = self.config.variable_design_data
```
MarcusHolly, line 210: Missing a docstring

idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
```diff
@@ -173,15 +294,104 @@ def my_operation_model(m, design_blk):
...
+        "commodity",
+        ConfigValue(
+            domain=str,
+            doc="Variable on which capacity and ramping limits are needed",
+        ),
+    )
+    CONFIG.declare(
+        "capacity",
+        ConfigValue(doc="Maximum capacity of the commodity"),
+    )
```
MarcusHolly, line 339: Specify domain for this?

MarcusHolly (approved):
Minor comments - otherwise LGTM

idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
radhakrishnatg: The input can be anything: Var, Expression, int, or float. So, I did not specify the domain.

PR #1636, opened 2025-06-26
Butane and Isobutane in Helmholtz

bertkdowns:
## Fixes
Adds Butane and Isobutane in helmholtz, based on:

>Reference Equations of State for the Thermodynamic Properties of Fluid Phase n-Butane and Isobutane 
D. Bücker; W. Wagner

https://doi.org/10.1063/1.1901687 

## Summary/Motivation:

We wanted to be able to model butane in helmholtz for https://github.com/waikato-ahuora-smart-energy-systems/PropertyPackages, may as well contribute upstream too!

- Added and generated Butane and IsoButane helmholtz files
- Added property tests based on values in the paper, with some properties inferred to additional precision in the same way that John Eslick did orignally.

### Legal Acknowledgement


bertkdowns:
Good point! Here's some visualisations:
image
image
image
image


And the equivalent from https://tlk-energy.de/en/phase-diagrams/pressure-enthalpy
image
image

From https://www.researchgate.net/figure/T-s-diagram-of-butane-covering-the-whole-liquid-range-in-temperature-from-triple-point_fig2_318372826
image


I couldnt find a T-S curve for isobutane though. The scales are a bit different but seem to match.

Here's the code to generate the graphs if anyone in the future is trying to do the same thing:

```
import os
from pyomo.common.fileutils import this_file_dir
from idaes.models.properties.general_helmholtz.helmholtz_parameters import (
    WriteParameters,
)
from idaes.models.properties.general_helmholtz import (
    HelmholtzParameterBlock,
    PhaseType,
    StateVars,
)
import pyomo.environ as pyo
from idaes.core import FlowsheetBlock

m = pyo.ConcreteModel()

m.fs = FlowsheetBlock(dynamic=False)
m.fs.helmholtz = HelmholtzParameterBlock(
    pure_component="butane",
    phase_presentation=PhaseType.LG,
    state_vars=StateVars.PH,
)

fig, ax = m.fs.helmholtz.ts_diagram()

import matplotlib.pyplot as plt
fig.show()
plt.show()

```

idaes/models/properties/general_helmholtz/tests/test_parameter_util.py
```diff
@@ -51,12 +51,25 @@
...
 def _common_sat(sat_thermo_data, we):
     # Check the EoS expressions, the tolerance is a little loose
     # due to lack of sig. figs. in reported data
     for pnt in sat_thermo_data.values():
+
+        # Uncomment this if the tests are failing because the
+        # pressure calculated from rhol is not the same as rhov.
+        # This can infer a better value to use for rhol
+        # with more decimal places of precision
+        # calc_liq_value(pnt,we)
```
dallan-keylogic: Why isn't this uncommented by default?

idaes/models/properties/general_helmholtz/tests/test_parameter_util.py
```diff
@@ -402,3 +415,121 @@ def test_r32():
...
+            "sv": -0.22517272904982927,
+        },
+    }
+
+    we = isobutane_main(dry_run=True)
+    _common_sat(sat_thermo_data, we)
+
+
+def calc_liq_value(pnt, we):
+    import math
```
dallan-keylogic: Is there a reason to delay importing `math` after the header?

idaes/models/properties/general_helmholtz/tests/test_parameter_util.py
```diff
@@ -402,3 +415,121 @@ def test_r32():
...
+        # Converge on the correct density by replacing the upper/lower bound
+        # (basic bisection solving method )
+        if rhol_pressure > pressure:
+            rhol_max = rhol
+        elif rhol_pressure < pressure:
+            rhol_min = rhol
+        rhol = (rhol_max + rhol_min) / 2
+        rhol_pressure = we.calculate_pressure(rho=rhol, T=pnt["T"])
+
+    print(f"update the rhol to this value: {rhol}")
```
dallan-keylogic: When is this method intended to be used? Is `rhol` a parameter in some file?

idaes/models/properties/general_helmholtz/tests/test_parameter_util.py
bertkdowns: John Eslick notes in the top of the file:

> On some test points, I had to infer extra significant figures for density.  For example,
this density of a liquid near the triple point is insensitive to pressure, so to calculate
the pressure from density of a liquid at the triple point, a lot of significant figures 
are needed.

I also had this problem - the test data had a density of `"rhov": 1.4236` at one of my test cases, and a pressure of `"p": 48.482`. That pressure does give a density of approximately 1.4236, but as density is insensitive to pressure, the inverse is also true- calculating pressure from density is extremely sensitive. Using the rounded value from literature (`1.4236`) didn't give enough accuracy, and so the pressure was wildly off. However, when I used the helmholtz function to calculate rohv, it gave a density of `1.423641341935`, and so if I used this value with "inferred decimal places" it was able to calculate pressure accurately.

Presumably John Eslick wrote a method to infer these decimal places, but didn't include it in the codebase. that's what my calc_liq_value function does - I figured it'd be good to include it as it will be needed in the future when adding more tests. Though perhaps I should't include that directly in the test file, should I extract it to a seperate script of give it a bit more explanation?

idaes/models/properties/general_helmholtz/tests/test_parameter_util.py
bertkdowns: not really, other than it isn't needed in normal usage-  calc_liq_value will only be used if you're adding a new test.

idaes/models/properties/general_helmholtz/tests/test_parameter_util.py
bertkdowns: See above. Rhol (liquid density) is the parameter that needs extra precision correctly calculate pressures, because at the triple point pressure is extremely sensitive to liquid density.

idaes/models/properties/general_helmholtz/tests/test_parameter_util.py
bertkdowns: I've moved the import to the top of the file.

idaes/models/properties/general_helmholtz/tests/test_parameter_util.py
bertkdowns: I've updated the explanations and renamed the method, I think it should be a little clearer now.

Issue #1287, opened 2023-11-14
Bug in subcritical powerplant report methods


andrewlee94:
### Discussed in https://github.com/IDAES/idaes-pse/discussions/1286

<div type='discussions-op-text'>

<sup>Originally posted by **finAtoZ** November 13, 2023</sup>
i run the  subcritical_power_plant.py without modification， but produced several errors.
the first error is 
<img width="1138" alt="5adc7898b47ce3abbf38f3e8f2bd909" src="https://github.com/IDAES/idaes-pse/assets/135591017/11b17a28-62c3-4536-95ee-61b546f0e35e">

and the second is 
```
Traceback (most recent call last):
  File "D:\IDAES\miniconda3\envs\idaes-pse\Lib\site-packages\idaes\models_extra\power_generation\flowsheets\subcritical_power_plant\subcritical_power_plant.py", line 2284, in <module>
    m_ss = main_steady_state()
  File "D:\IDAES\miniconda3\envs\idaes-pse\Lib\site-packages\idaes\models_extra\power_generation\flowsheets\subcritical_power_plant\subcritical_power_plant.py", line 325, in main_steady_state
    m = get_model(dynamic=False)
  File "D:\IDAES\miniconda3\envs\idaes-pse\Lib\site-packages\idaes\models_extra\power_generation\flowsheets\subcritical_power_plant\subcritical_power_plant.py", line 869, in get_model
    blr.initialize(m)
  File "D:\IDAES\miniconda3\envs\idaes-pse\lib\site-packages\idaes\models_extra\power_generation\flowsheets\subcritical_power_plant\subcritical_boiler_flowsheet.py", line 792, in initialize
    fs.aECON.initialize(outlvl=outlvl)
  File "D:\IDAES\miniconda3\envs\idaes-pse\lib\site-packages\idaes\core\base\unit_model.py", line 540, in initialize
    flags = blk.initialize_build(*args, **kwargs)
  File "D:\IDAES\miniconda3\envs\idaes-pse\lib\site-packages\idaes\models_extra\power_generation\unit_models\boiler_heat_exchanger_2D.py", line 3219, in initialize_build
    res = opt.solve(blk, tee=slc.tee)
  File "D:\IDAES\miniconda3\envs\idaes-pse\lib\site-packages\pyomo\opt\base\solvers.py", line 627, in solve
    raise ApplicationError("Solver (%s) did not exit normally" % self.name)
[...]

PR #1639, opened 2025-07-15
Fix subcritical power plant model

aspitarl:
## Fixes

This fixes issue #1287. 
Possibly fixes the handling of calculate_initial_integral in idaes/models/control/controller.py. 

## Summary/Motivation:

This is a few minimal fixes to get the subcritical power plant model running. 

There appears to have been some changes to PID models that was not coordinated with the subcritical power plant model. See the commit made by @dallan-keylogic here: https://github.com/aspitarl/idaes-pse/blame/db028bdab41c92cabe1b5c05e63aeeb84ba8a82b/idaes/models_extra/power_generation/flowsheets/subcritical_power_plant/tests/test_subcritical_flowsheets.py#L151. The second commit of this PR does some fixes regarding the calculate_initial_integral keyword argument that seemed to work, but need to be checked.

- During tagging of the power plant svg in `print_pfd_results`, there is now a check that chemical species exist within a tagged stream. This was the underlying issue behind #1287 as there were multiple property packages and the flue gas species that the code was trying to check for did not exist in the other property package. 
- Some changes to PID models, in particular a potential fix to the logic of the calculate_initial_integral argument. 
    - https://github.com/aspitarl/idaes-pse/commit/db028bdab41c92cabe1b5c05e63aeeb84ba8a82b#diff-19b5677ae20b9cb50f50e22505a513e4a19bbfcf4f39a8b38ba266708651671bR472 
- Fixed the dynamic power plant test degrees of freedom, and added a new test that separates out flowsheet construction from initialization. 

### Legal Acknowledgement

### DIFF
--- a/idaes/models/control/controller.py
+++ b/idaes/models/control/controller.py
@@ -469,7 +469,9 @@ class PIDControllerData(UnitModelBlockData):
 
         # deactivate the time 0 mv_eqn instead of skip, should be fine since
         # first time step always exists.
-        if self.config.calculate_initial_integral:
+        #TODO: should this happen for True or False? Seems this change caused subcritical power plant to fail
+        # setting to not gets subcritical power plant to work, but steam cycle fails
+        if not self.config.calculate_initial_integral:
             self.mv_eqn[t0].deactivate()
 
         if self.config.controller_type in [ControllerType.PI, ControllerType.PID]:
--- a/idaes/models_extra/power_generation/flowsheets/subcritical_power_plant/subcritical_power_plant.py
+++ b/idaes/models_extra/power_generation/flowsheets/subcritical_power_plant/subcritical_power_plant.py
@@ -573,12 +573,15 @@ def main_dynamic():
     m_dyn.fs_main.fs_stc.spray_ctrl.mv_ref.value = (
         m_dyn.fs_main.fs_stc.spray_valve.valve_opening[t0].value
     )
+
+    #TODO: integral_component_ref was replaced with mv_integral_component_dot
+    # 'ref' does not exist anymore, not sure what this should be set to
     # For two bounded controllers, set the initial mv_integral_component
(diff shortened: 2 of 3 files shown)


idaes/models/control/controller.py
```diff
@@ -469,7 +469,9 @@ def mv_eqn(b, t):
 
         # deactivate the time 0 mv_eqn instead of skip, should be fine since
         # first time step always exists.
-        if self.config.calculate_initial_integral:
+        #TODO: should this happen for True or False? Seems this change caused subcritical power plant to fail
+        # setting to not gets subcritical power plant to work, but steam cycle fails
+        if not self.config.calculate_initial_integral:
             self.mv_eqn[t0].deactivate()
```
dallan-keylogic, line 475: The idea behind this code block is that if we're calculating the initial integral we have the associated manipulated variable fixed. However, if we're *not* calculating the initial integral, the manipulated variable should be free and calculated by `mv_eqn` at `t0`. It's possible that the supercritical coal plant operating in a way where it's fixing the manipulated variable at `t0` despite having `calculate_initial_integral == False`, but the recourse there should be to unfix the associated MV(s).

idaes/models_extra/power_generation/flowsheets/subcritical_power_plant/subcritical_power_plant.py
```diff
@@ -692,7 +695,7 @@ def main_dynamic():
     time_used = end_time - start_time
     _log.info("simulation time={}".format(time_used))
     write_data_to_txt_file(plot_data)
-    plot_results(plot_data)
+    # plot_results(plot_data)
```
dallan-keylogic, line 698: Since we're not plotting dynamic results here (and we shouldn't be), you should return `plot_data` so the user can plot it themselves.

PR #1642, opened 2025-07-24
Location factor

Lingyan90:
## Fixes
The location factor is an important metric for analyzing different market location. Here, move the location factor from PrOMMiS to IDAES for broader application.  (This originally started in PrOMMiS in PR https://github.com/prommis/prommis/pull/143)

## Summary/Motivation:
Move the location factor to IDAES so that other repository such as PrOMMiS, WaterTAP, etc. could also access these values.

- Add location factor data in `costing_base.py`
- Added location factor data bank into `location_factors.jason`
- Added testing code in `test_costing_base.py`
- Added location factor in `SSLW.py`

### Legal Acknowledgement


idaes/core/base/tests/test_costing_base.py
```diff
@@ -84,6 +86,70 @@ def test_register_idaes_currency_units():
...
+    no_city_match = test_location not in locations
+    no_country_fallback = (test_location[0], None) not in locations
+
+    if country_exists and no_city_match and no_country_fallback:
+        with pytest.raises(StopIteration):
+            _ = next(
+                entry
+                for entry in location_data
+                if (entry["country"], entry["city"]) == test_location
+            )
```
dallan-keylogic: The test name indicates that this should raise a `KeyError` but the test itself indicates it raises a `StopIteration` error. It seems like you should catch the `StopIteration` error and raise a `KeyError` yourself.

Also, why are we iterating through a list entry-by-entry to check whether the country-city pair exists? Doesn't it make more sense to implement it an object (like a `dict`) that uses a hash lookup?

idaes/core/base/tests/test_costing_base.py
```diff
@@ -84,6 +86,70 @@ def test_register_idaes_currency_units():
...
+    else:
+        pytest.skip("Test assumptions not satisfied for invalid city test.")
```
dallan-keylogic: Why are we skipping the test here? The only reason we should be skipping tests is if the test requires some dependency (like a solver) that not every environment has.

dallan-keylogic (commented):
I have some questions about the implementation here.

idaes/core/base/costing_base.py
```diff
@@ -98,6 +102,41 @@ def register_idaes_currency_units():
         )
 
 
+def load_location_factor():
+    """
+    Estimate the cost of constructing the same plant in different global locations using location factors.
```
bpaul4:
```suggestion
    Estimate the cost of constructing the same plant in different regions using location factors.
```

idaes/core/base/costing_base.py
```diff
@@ -98,6 +102,41 @@ def register_idaes_currency_units():
         )
 
 
+def load_location_factor():
```
bpaul4:
```suggestion
def load_location_factors():
```

idaes/core/base/costing_base.py
```diff
@@ -169,6 +208,9 @@ def build(self):
         self.base_currency = None
         self.base_period = pyo.units.year
 
+        # Set the location factor
```
bpaul4:
```suggestion
        # Set the location factor, default U.S. Washington DC, location_factor = 1.00
```

idaes/core/base/tests/test_costing_base.py
```diff
@@ -28,6 +28,8 @@
     register_idaes_currency_units,
 )
 
+from idaes.core.base.costing_base import load_location_factor
```
bpaul4:
```suggestion
from idaes.core.base.costing_base import load_location_factors
```

idaes/core/base/tests/test_costing_base.py
```diff
@@ -84,6 +86,70 @@ def test_register_idaes_currency_units():
         )
 
 
+@pytest.mark.unit
+def test_all_valid_locations_have_factors():
+    location_data = load_location_factor()
```
bpaul4: And everywhere else in the file where this appears.

```suggestion
    location_data = load_location_factors()
```

idaes/core/base/tests/test_costing_base.py
```diff
@@ -84,6 +86,70 @@ def test_register_idaes_currency_units():
...
+def test_invalid_country_raises_keyerror():
+    location_data = load_location_factor()
+    valid_locations = {(entry["country"], entry["city"]) for entry in location_data}
+    invalid_country = ("Brunei", None)
+    assert (
+        invalid_country not in valid_locations
+    ), f"Test assumption failed: {invalid_country} unexpectedly exists in data"
+
+
+@pytest.mark.unit
```
bpaul4: How would this work in practice? If a user tried to call a country which an invalid region/city, would it return a default?

idaes/models/costing/SSLW.py
```diff
@@ -55,6 +55,8 @@
     register_idaes_currency_units,
 )
 
+from idaes.core.base.costing_base import load_location_factor
```
bpaul4:
```suggestion
from idaes.core.base.costing_base import load_location_factors
```

idaes/models/costing/SSLW.py
```diff
@@ -208,6 +210,7 @@ class BlowerMaterial(StrEnum):
 class SSLWCostingData(FlowsheetCostingBlockData):
     # Register currency and conversion rates based on CE Index
     register_idaes_currency_units()
+    load_location_factor()
```
bpaul4: This should be assigned to an object, so it can be referenced later.

I think "location" should also be a configuration argument like "CE_index_year", so the `UnitModelCostingBlock` will save the attribute `self.location = ("United States", "Washington DC")` and then apply it to the capital cost equation.

Finally, please add tests to the SSLW test file applying location factors and check built objects, object values, and cost results.

idaes/models/costing/SSLW.py
```diff
@@ -222,6 +225,8 @@ def build_global_params(self):
         self.base_currency = pyo.units.USD_2018
         # Set a base period for all operating costs
         self.base_period = pyo.units.year
+        # Chose location and set location factor
+        self.location_factor = ("United States", "Washington DC")
```
bpaul4:
```suggestion
        self.location = ("United States", "Washington DC")
```
And then internal to the costing method, we can define `self.location_factor = location_data[self.location]`.

bpaul4 (commented):
Thanks @Lingyan90 for opening this, I have a few comments.

Lingyan90 (commented):
@bpaul4, The code looks very good to me. 

One issue that worth discussion is if the base cost in not in USD, it's in CAD for example, and user want use that base CAD cost to convert to Finland EUR, then the location factor application should be something like TPC_finland = TPC_Canada * LocationFactor_finland / LocationFactor_CAD.

For more general use, I would suggest modify the current location application equation to something like above, e.g. TPC_finland = TPC_baselocation * LocationFactor_finland / LocationFactor_baselocation.

And would also worth clarify that the current database are based on USD, if user use other currency, they should be consistent with their referenced currency.

idaes/models/costing/QGESS.py
```diff
@@ -297,6 +308,67 @@ def build_global_params(self):
...
+                    f"Country {country} not supported; please check the data file at "
+                    f"IDAES.idaes-pse.idaes.core.base.locations_factors.json for spelling and supported countries."
+                )
+
+            # check if city is valid
+            if city not in self.location_factor_dictionary[country]:
+                # city lists within countries are not too long, list them here for the user
+                raise KeyError(
+                    f"City {city} not supported for country {country}; valid cities include "
+                    f"{self.location_factor_dictionary[country].keys()}"
```
Akintomiwaojo: dict_keys([...]) may appear in the message output. This might be an alternative:

f"{list(self.location_factor_dictionary[country].keys())}"

idaes/models/costing/QGESS.py
```diff
@@ -297,6 +308,67 @@ def build_global_params(self):
...
+            # alias country and city to make code more readable
+            country = self.config.location[0]
+            city = self.config.location[1]
+
+            # check if country is valid
+            if country not in self.location_factor_dictionary:
+                # too many countries to list, point users to the data file
+                raise KeyError(
+                    f"Country {country} not supported; please check the data file at "
+                    f"IDAES.idaes-pse.idaes.core.base.locations_factors.json for spelling and supported countries."
```
Akintomiwaojo:
```suggestion
                    f"IDAES.idaes-pse.idaes.core.base.location_factors.json for spelling and supported countries."
```

idaes/core/base/costing_base.py
```diff
@@ -98,6 +102,62 @@ def register_idaes_currency_units():
...
+        C_{TPI, corrected} = F_{ISF} \times C_{TPI}
+
+    where:
+    - :math:`C` represents cost,
+    - :math:`F` represents a factor,
+    - :math:`TPI` is the total plant investment, and
+    - :math:`ISF` is the investment site factor (i.e., location factor).
+
+    Location factors for 139 countries are sourced from Compass International (2017):
+    https://www.compassinternational.net/wp-content/uploads/2017/01/Worldwide-Industrial.pdf
```
Akintomiwaojo, line 130: Could you please check this link again? It appears to be unavailable.

Akintomiwaojo: > Could you please check this link again? It appears to be unavailable.

Is it in-house data that you requested? I checked www.compassinternational.net, and it looks like I need to request it.

idaes/core/base/costing_base.py
```diff
@@ -98,6 +102,62 @@ def register_idaes_currency_units():
...
+
+    Location factors for U.S. regions are sourced from Table 16.13 in the Seider reference above. The values
+    are converted appropriately assuming the base factor of 1.0 corresponds to Washington D.C. and U.S. Northeast.
+
+    Note: For some countries, multiple city-specific location factors are provided.
+    The benchmark location is Washington, D.C., USA.
```
Akintomiwaojo, line 137: For clarification, does this imply that the benchmark for country factors, pulled from Compass International, is also Washington, D.C., USA? If  not, we might be erroneously mixing factors with different benchmarks

Akintomiwaojo (commented):
Some comments and clarifications in-line.

idaes/core/base/costing_base.py
bpaul4: @Lingyan90 could you please comment on this?

idaes/core/base/costing_base.py
bpaul4: The factors from Compass International have a benchmark of 1.0 for Washington, D.C., USA and everything else is in relation to that. The regional US factors from Seider were scaled so that Northeast (which is assumed to be the same region as DC) is 1.0.

idaes/core/base/costing_base.py
Lingyan90: They probably have removed it or replaced it with different file. Could change the citation to something like this: `Compass International, Inc. (2017). 2017 worldwide industrial [Sample edition]. https://www.compassinternational.net/wp-content/uploads/2017/01/Worldwide-Industrial.pdf (Accessed March 25, 2025).`

bpaul4:
> @bpaul4, The code looks very good to me.
> 
> One issue that worth discussion is if the base cost in not in USD, it's in CAD for example, and user want use that base CAD cost to convert to Finland EUR, then the location factor application should be something like TPC_finland = TPC_Canada * LocationFactor_finland / LocationFactor_CAD.
> 
> For more general use, I would suggest modify the current location application equation to something like above, e.g. TPC_finland = TPC_baselocation * LocationFactor_finland / LocationFactor_baselocation.
> 
> And would also worth clarify that the current database are based on USD, if user use other currency, they should be consistent with their referenced currency.

At the moment we only support USD for currencies. Users could define their own, however the current costing framework assumes that currencies are of the form "USD" + some descriptor about the index value or year/month so it would require an additional option. As it is we don't have any reference costs in other currencies so we can keep this in mind as a potential expansion for future work.

idaes/models/costing/tests/test_QGESS.py
```diff
@@ -168,6 +171,234 @@ def test_valid_years_for_phaseout(self, CEPCI_year):
...
+        # Create a Concrete Model as the top level object
+        m = pyo.ConcreteModel()
+
+        # Add a flowsheet object to the model
+        m.fs = FlowsheetBlock(dynamic=False)
+        with pytest.raises(
+            KeyError,
+            match=re.escape(
+                "City notacity not supported for country United States; valid cities include "
+                "['Washington DC / Northeast', 'Gulf Coast', 'Southwest', 'Midwest', 'West Coast']"
```
dallan-keylogic, line 377: It's weird that the `city` field accepts one city and five regions.

idaes/models/costing/tests/test_QGESS.py
```diff
@@ -168,6 +171,234 @@ def test_valid_years_for_phaseout(self, CEPCI_year):
...
+                "is not defined, the default value is `average`."
+            ),
+        ):
+            m.fs.costing = QGESSCosting(
+                tech=1,
+                location={
+                    "country": "United States",
+                    "city": "West Coast",
+                    "val": "min",
+                },
```
dallan-keylogic, line 265: All things being equal, I'd prefer to provide a dictionary rather than a list. That's not a reason to hold up this PR, though.

idaes/models/costing/QGESS.py
```diff
@@ -225,23 +226,33 @@ class QGESSCostingData(FlowsheetCostingBlockData):
...
+            domain=lambda v: (
+                ListOf(str)(v)
+                if (isinstance(v, list) and all(isinstance(i, str) for i in v))
+                else (_ for _ in ()).throw(
+                    ValueError(
+                        "Argument 'location' must be a list of strings in form ['country', 'city/region', 'val'] "
+                        "where 'val' can be 'min', 'max', or 'average'. Must define `country` and `city`. If `val` "
+                        "is not defined, the default value is `average`."
+                    )
+                )
```
dallan-keylogic: Defining a function using `def` is preferable to using a lambda function. Lambda functions have some undesirable properties, but I forget what those were.

dallan-keylogic (commented):
Just a few comments. Replace the `lambda` function with a full function defined using `def` and I'll approve it.

idaes/models/costing/tests/test_QGESS.py
bpaul4: The "city" field is intended for subsections of countries that have more than just an aggregate value, for the US we happen to have location factors for regions rather than cities.

idaes/models/costing/tests/test_QGESS.py
bpaul4: Noted, thanks, will keep in mind for a future update.

idaes/models/costing/QGESS.py
bpaul4: Done. Quick search says that lambda functions are great when it's a simple one line check, but functions are more readable, easier to update later, and separate the domain modifier from the actual config defintion.

dallan-keylogic: Here's the issue with the rationale for not including lambdas: https://github.com/IDAES/idaes-pse/issues/1240

I'm not sure whether it applies in this case, but it's easier to say "Don't use lambdas" and just use `def` instead.

dallan-keylogic (approved):
LGTM

Akintomiwaojo (approved):
My previous comments are addressed. It looks good to me.

PR #1645, opened 2025-08-02
Add options to DiagnosticsToolbox to use scaled Jacobian when reporting numerical issues

jasonmbray-p66:
## Summary/Motivation:
When the Jacobian is calculated in several of the DiagnosticsToolbox methods, the optional `scaled` argument in `idaes.core.util.scaling.get_jacobian` (which actually defaults to True) is set to False every time it is called. Because it is hard-coded into the class, there is no option for the user to change this choice. One of the purposes of the numerical diagnostics is to identify scaling issues, but the `scaled=False` argument whenever the Jacobian is calculated means that the numerical diagnostics always report the same results every time they run, no matter what scaling the user has added to the model to address the problems identified. Once any scaling is added to the model, they are flying blind as to whether the scaling factors are working or if other variables and constraints need additional scaling.

- Make a new config argument called `use_scaled_jacobian` that the user can set when creating the DiagnosticsToolbox object. This option is used throughout the class whenever the Jacobian needs to be calculated.
- Add an optional `use_scaled_jacobian` argument to several class methods that are often called directly by the user, giving the user flexibility about which way to calculate the Jacobian without having to re-create the DiagnosticsToolbox object.
- Store the Jacobian as a property in the DiagnosticsToolbox object and retrieve the stored value rather than re-calculating it every time it is needed. This saves a lot of time when running several diagnostic functions back to back, while currently the exact same Jacobian is recalculated over and over.
- Add an optional `force_jacobian_recalc` argument to several class methods where the stored Jacobian is used by default if present. There could be a situation where the model is re-run with new values, making the stored Jacobian outdated. The user needs to be able to force these methods to recalculate the Jacobian based on the updated model. This argument gives the user control over when that happens.

### Legal Acknowledgement


dallan-keylogic:
The original intention was for the user to perform a [Pyomo scaling transformation](https://pyomo.readthedocs.io/en/stable/api/pyomo.core.plugins.transform.scaling.ScaleModel.html#pyomo.core.plugins.transform.scaling.ScaleModel), then to create the diagnostic toolbox with the transformed model. That came with two downsides
1) The Pyomo scaling transformation is slow
2) Most initialization routines fail when called on the transformed model and will not use the scaling factors when called on the original model.

However, a solution is on the way. Pyomo's method to scale the model automatically when a .nl file is written [is finally going to be fixed.](https://github.com/Pyomo/pyomo/pull/3673). Once that is finished, 
```
solver = get_solver(
    solver="ipopt_v2",
    solver_options={"max_iter": 50},
    writer_config={
        "scale_model": True,
        "linear_presolve": False,
    }
)
```
will allow the scaling factors to be used at the solver-write time. (The linear presolve is turned off because, as of this moment, the choice of pivots to determine which variables to eliminate does not take into account scaling factors. That is going to be fixed soon, but it's not clear it will make the August Pyomo release). We can then work on updating the [IDAES default configuration](https://github.com/IDAES/idaes-pse/blob/c84a7fef8eb779d3efc0f4050acddb33a7e11bb7/idaes/config.py#L383) to use the writer-time scaling. (Theoretically, we should be able to just change the option and everything will solve.)

The plan for the Diagnostics Toolbox, then, will be to have it *only* report scaled versions of the variables and Jacobian, because those are what IPOPT actually sees.

jasonmbray-p66:
Thanks for the update! I like that your implementation addresses the need for including scaling when running the diagnostic tools.

I think it would be a valuable addition to give the user the flexibility of choosing whether to use scaling or not, which is something I included in this PR by both setting a config value and by providing an optional argument in several of the public functions. Your implementation captures the majority of use cases, but I can think of at least one situation where I would want to be able to run the diagnostics both ways (scaling on and scaling off) to see how they differ. Additionally, by just deleting `scaled=False` rather than changing it to `scaled=True`, you are susceptible to surprise behavior changes if the default in `get_jacobian()` ever changes.

Let me know if you'd like my help implementing this small enhancement, but I may not have time to work on it in the near future.

Issue #1648, opened 2025-08-08
Outstanding issues with the new scaling tools


dallan-keylogic:
1. Scaling factors for `Expressions` have been deprecated. I have a system of scaling hints for Expressions available on a branch
2. The team decided that the distinction between mutable class attributes `scaler.DEFAULT_SCALING_FACTORS` and `scaler.UNIT_SCALING_FACTORS` and their local copies `scaler.default_scaling_factors` and `scaler.unit_scaling_factors` should be handled via documentation. This documentation needs to be written.
3. The scaling API doesn't allow us to filter `scaler.default_scaling_factors` for unused inputs (either because a modeler forgot to perform scaling or because a user wanted to set some default value), nor does it allow us to apply all default scaling factors automatically.
4. The scaling API doesn't give us an easy way to communicate to the user which scaling factors they *need* to set on their own.
5. In order to do handle property packages of any complexity, we *need* scaling factors for more than just the state variables, yet the scaling API doesn't give us a method to solicit the user for such scaling factors, and the reliance on `CustomScalerBase.propagate_state_scaling` prevents us from rolling out such scaling factors in a general manner.
6. Order-of-magnitude scaling for variables is harmful, yet the entire API guides the user towards using it.

PR #1649, opened 2025-08-12
Scaling hints

dallan-keylogic:
## Summary/Motivation:
While the old scaling toolbox supported setting scaling factors for both `Vars` and `Expressions`, the Pyomo team indicated to us that we were abusing the scaling suffix by setting scaling factors for `Expressions` because they would not be exported to the NLP. However, using the `NominalValueExpressionWalker` for the sorts of giga-expressions that are regularly found in IDAES thermodynamic packages can be dangerous. As a result, I implemented a system of "scaling hints" that could be assigned to named `Expressions`, sequestering them in the `scaling_hint` suffix.

Additionally, I modified the `call_submodel_scaler_method` so that it passes the parent `Scaler`'s configuration and map of submodel scalers to the child `Scaler` objects. Note that some child scalers were not written expecting the `submodel_scalers` argument, so this does break backwards compatibility, but I doubt that there has been any major adoption of these new scaling tools.

This resolves point (1) in #1648.

- Add system of scaling hints for Expressions
- Pass config and `submodel_scalers` to child scalers in `call_submodel_scaler_method` of a parent scaler
- Add `scale_constraint_by_component` method to `CustomScalerBase` in order to take the scaling factor from a `Var` or `Expression` and assign it to a constraint.
- Modify `scale_variable_from_default` in order to avoid silent failures, as well as to make it clear to the user when their input into scaling is necessary.
- Raise an exception if people try to set scaling factors on indexed components (this is related to point (10) in #1648)

### Legal Acknowledgement


idaes/core/scaling/custom_scaler_base.py
```diff
@@ -370,16 +428,28 @@ def scale_constraint_by_default(self, constraint, overwrite: bool = False):
...
-        else:
-            _log.debug(
-                f"Could not set scaling factor for {constraint.name}, "
-                f"no default scaling factor set."
-            )
+        if not isinstance(constraint, ConstraintData):
+            raise TypeError(f"{constraint} is not a constraint (or is indexed).")
+        self._scale_component_by_default(component=constraint, overwrite=overwrite)
+
+    def get_expression_nominal_value(self, expression):
```
MarcusHolly: Is there not a better way to combine/differentiate the three nominal value functions?

idaes/core/scaling/custom_scaler_base.py
```diff
@@ -541,16 +632,36 @@ def propagate_state_scaling(
...
+        target_vars = target_state_data.define_state_vars()
+        source_vars = source_state_data.define_state_vars()
+
+        for state, var in target_vars.items():
+            for vidx, vardata in var.items():
+                self.scale_variable_by_component(
+                    target_variable=vardata,
+                    scaling_component=source_vars[state][vidx],
+                    overwrite=overwrite,
+                )
```
MarcusHolly: Similar point with this - is there a simple way to combine this with the existing function?

idaes/core/scaling/scaling_base.py
```diff
@@ -45,6 +46,7 @@
 CONFIG.declare(
     "max_variable_scaling_factor",
     ConfigValue(
+        # default=float("inf"),
```
MarcusHolly: Better to just delete these default values?

MarcusHolly (approved):
Minor comments, but nothing that should hold this back

idaes/core/scaling/custom_scaler_base.py
dallan-keylogic: Is the proposed behavior to have `propagate_state_scaling` check whether the inputs are indexed or unindexed, then handle those cases appropriately?

idaes/core/scaling/custom_scaler_base.py
dallan-keylogic: We have `get_expression_nominal_values` deprecated, and it can be removed in the next release. The difference between `get_sum_terms_nominal_values` and `get_expression_nominal_value` is that the former returns a list while the latter returns a float.

idaes/core/scaling/scaling_base.py
dallan-keylogic: I really want to get rid of these max and min values, because we're going to need to override them in pretty much every scaler in PrOMMiS, but they're required for the `Autoscalers` to function. I can certainly get rid of these comments, though.

idaes/core/scaling/custom_scaler_base.py
```diff
@@ -265,28 +292,59 @@ def scale_variable_by_bounds(self, variable, overwrite: bool = False):
...
+        sf = self.get_default_scaling_factor(component)
+        if sf is None or sf == DefaultScalingRecommendation.userInputRequired:
+            # Check to see if the user manually set a scaling factor
+            sf = get_scaling_factor(component)
+            if sf is None or overwrite:
+                # If the user told us to overwrite scaling factors, then
+                # accepting a preexisiting scaling factor is not good enough.
+                # They need to go manually alter the default entry to
+                # DefaultScalingRecommendation.userInputRecommended
+                raise KeyError(f"No default scaling factor set for {component}.")
```
strahlw: Is this really a KeyError? I would suggest a ValueError

idaes/core/scaling/custom_scaler_base.py
```diff
@@ -265,28 +292,59 @@ def scale_variable_by_bounds(self, variable, overwrite: bool = False):
...
+                component=component, scaling_factor=sf, overwrite=overwrite
             )
 
+    def scale_variable_by_default(self, variable, overwrite: bool = False):
+        """
+        Set scaling factor for variable or scaling hint for named expression
+        based on default scaling factor.
+
+        Args:
+            variable: variable/expression to set scaling factor for
```
strahlw, line 351: Personal preference to be more explicit with types: Var, Expression, any others expected? Also I am not sure what the convention is for IDAES, but including type hints in the function signature has been helpful on other projects.

idaes/core/scaling/custom_scaler_base.py
```diff
@@ -370,16 +428,28 @@ def scale_constraint_by_default(self, constraint, overwrite: bool = False):
...
-            self.set_constraint_scaling_factor(
-                constraint=constraint, scaling_factor=sf, overwrite=overwrite
-            )
-        else:
-            _log.debug(
-                f"Could not set scaling factor for {constraint.name}, "
-                f"no default scaling factor set."
-            )
+        if not isinstance(constraint, ConstraintData):
+            raise TypeError(f"{constraint} is not a constraint (or is indexed).")
```
strahlw: Is it difficult to add check that will give the user more specific information if it is indexed versus if it is not a constraint? If it's a lot of work, don't bother.

idaes/core/scaling/custom_scaler_base.py
```diff
@@ -265,28 +292,59 @@ def scale_variable_by_bounds(self, variable, overwrite: bool = False):
...
+
+        Args:
+            variable: variable/expression to set scaling factor for
+            overwrite: whether to overwrite existing scaling factors
+
+        Returns:
+            None
+        """
+        if not (isinstance(variable, VarData) or isinstance(variable, ExpressionData)):
+            raise TypeError(f"{variable} is not a variable/expression (or is indexed).")
```
strahlw: See comment on scale_variable_by_units. I think it would be beneficial to have clearer feedback to the user if they encounter these types of errors.

idaes/core/scaling/custom_scaler_base.py
MarcusHolly: Correct

strahlw: ^^^ I think this is right. The function signature and return value are the same, so might as well allow it to handle both types of inputs - that being said, I would make sure that the documentation clearly explains it.

idaes/core/scaling/scaling_base.py
```diff
@@ -162,6 +185,35 @@ def set_variable_scaling_factor(
...
+            variable: VarData component to set scaling factor for.
+            scaling_factor: nominal scaling factor to apply. May be limited by max and min values.
+            overwrite: whether to overwrite existing scaling factor (if present).
+              Defaults to Scaler config setting.
+
+        Returns:
+            None
+
+        Raises:
+            TypeError if variable is not an instance of VarData
```
strahlw, line 200: VarData -> ExpressionData

idaes/core/scaling/tests/test_custom_scaling_integration.py
```diff
@@ -269,7 +269,7 @@ def test_scale_constraint_by_nominal_derivative_2norm_imperfect_information():
         scaler.scale_constraint_by_nominal_derivative_norm(c)
 
     scaled = jacobian_cond(model, scaled=True)
-    assert scaled == pytest.approx(1.06128e11, rel=1e-5)
+    assert scaled == pytest.approx(3.5727e10, rel=1e-3)
```
strahlw, line 272: Is it understood why this changed?

idaes/core/scaling/tests/test_util.py
```diff
@@ -698,110 +965,264 @@ def test_scaling_factors_from_dict_suffix_verify_fail(self, model):
...
 
         # Ensure we have not mutated original dict
-        assert sdict == {
-            "v[1]": 50,
-            "v[2]": 100,
-            "c[1]": 50,
-            "c[2]": 250,
-            "block_name": "unknown",
-        }
+        assert sdict == sdict_copy
```
strahlw, line 1191: duplicate of line 1188

idaes/core/scaling/util.py
```diff
@@ -55,45 +58,123 @@
...
+        )
+
+    try:
+        sfx = blk.scaling_hint
+    except AttributeError:
+        # No existing suffix, create one
+        _log.debug(f"Created new scaling hint suffix for {_filter_unknown(blk)}")
+        sfx = blk.scaling_hint = Suffix(direction=Suffix.EXPORT)
+
+    return sfx
```
strahlw, line 140: There's duplicate code in 84-94 and 121-131. I suggest breaking it out into a separate function and having an argument that determines whether it uses "hints" or "factors" in the type error message.

idaes/core/scaling/util.py
```diff
@@ -304,25 +421,30 @@ def _suffix_to_dict(suffix):
...
     for k, v in json_dict.items():
-        # Safety catch in case this gets left in by other functions
-        if k == "parent_name":
-            continue
-
         comp = parent_block.find_component(k)
         if comp is not None:
+            if valid_types is not None and not any(
+                [isinstance(comp, cls) for cls in valid_types]
+            ):
```
strahlw, line 434: You can just use the generator here, no need to encapsulate in a list

idaes/models/unit_models/tests/test_equilibrium_reactor.py
```diff
@@ -766,5 +769,5 @@ def test_example_case(self):
         sm = TransformationFactory("core.scale_model").create_using(m, rename=False)
         jac, _ = get_jacobian(sm, scaled=False)
         assert (jacobian_cond(jac=jac, scaled=False)) == pytest.approx(
-            4.987e05, rel=1e-3
+            5.445e05, rel=1e-3
```
strahlw, line 772: I take it this changes because we now scale fixed variables for the "nominal" values.

idaes/core/scaling/tests/test_nominal_value_walker.py
```diff
@@ -153,7 +128,7 @@ def test_var_fixed_value(self):
         set_scaling_factor(m.var, 1 / 4)
 
         # Nominal value should be value
-        assert NominalValueExtractionVisitor().walk_expression(expr=m.var) == [-1]
+        assert NominalValueExtractionVisitor().walk_expression(expr=m.var) == [-4]
```
strahlw, line 131: Assuming this is an understood/intended change -> looking further down in the PR it looks like a "fundamental" change going on is no longer using fixed values as "nominal" values. If this is a correct understanding, please update the comment on line 130.

idaes/core/scaling/tests/test_custom_scaling_integration.py
strahlw: This is attributable to using scaled values for nominal values in the case the variables are fixed?

idaes/core/scaling/tests/test_scaling_profiler.py
```diff
@@ -544,15 +544,15 @@ def test_write_profile_report():
...
-Inverse Minimum          || 7.346E+18 | Failed 49  || 5.601E+12 | Solved 16 
-Nominal L1 Norm          || 1.189E+16 | Failed 61  || 2.060E+06 | Solved 4  
-Nominal L2 Norm          || 1.188E+16 | Failed 53  || 3.074E+06 | Solved 4  
-Actual L1 Norm           || 1.461E+09 | Failed 29  || 2.987E+03 | Solved 6  
-Actual L2 Norm           || 6.613E+08 | Failed 29  || 2.511E+03 | Solved 6  
+Harmonic                 || 3.788E+18 | Failed 135 || 2.839E+12 | Solved 17 
+Inverse Sum              || 9.077E+15 | Failed 200 || 1.502E+06 | Solved 6  
+Inverse Root Sum Squares || 1.083E+16 | Failed 200 || 9.600E+05 | Solved 6  
+Inverse Maximum          || 1.066E+16 | Failed 200 || 7.846E+05 | Solved 6  
+Inverse Minimum          || 7.543E+18 | Failed 200 || 5.599E+12 | Solved 16
```
strahlw, line 553: Is it clear why only the inverse minimum changed for the "perfect scaling" case. That seems a little odd to me (although percentage-wise it is a small change). Also, the user-scaling cases fail many times more - is this understood behavior, or just an observed response? (Does it need to be understood?)

idaes/core/scaling/custom_scaler_base.py
dallan-keylogic: The original error was simply if there was no corresponding entry in the default scaling dictionary. Since the workflow has become more complex, I'll change it to a `ValueError`.

idaes/core/scaling/custom_scaler_base.py
dallan-keylogic: The fact that we have separate functions for variables and constraints is legacy code that I don't care for. We need Variables and Expressions to be in the same function for things like the Modular Properties where whether something is an `Var` or an `Expression` depends on the state variables. I'll add some additional type hints.

idaes/core/scaling/custom_scaler_base.py
dallan-keylogic: No, I can do it. I'll probably extract the checking to a helper function.

PR #1653, opened 2025-08-25
Change to scale_model=True

dallan-keylogic:
## Summary/Motivation:
Changes `scale_model` to `True` in the .nl writer config to enable IDAES to take advantage of writer-time scaling. Additionally, `Exceptions` have been added when the user attempts to use `set_scaling_factor` on a constraint that has already been transformed using `constraint_scaling_transform` and, conversely, when the user attempts to use `constraint_scaling_transform` on a constraint which already has a scaling factor.

### Legal Acknowledgement

### DIFF
--- a/idaes/config.py
+++ b/idaes/config.py
@@ -385,7 +385,7 @@ def _new_idaes_config_block():
         "scale_model",
         pyomo.common.config.ConfigValue(
             domain=Bool,
-            default=False,  # TODO: Change to true once transition complete
+            default=True,
             description="Whether to apply model scaling in writer",
         ),
     )
--- a/idaes/core/scaling/tests/test_util.py
+++ b/idaes/core/scaling/tests/test_util.py
@@ -39,6 +39,7 @@ from idaes.core.scaling.util import (
     scaling_factors_from_json_file,
     report_scaling_factors,
 )
+from idaes.core.util.scaling import constraint_scaling_transform
 import idaes.logger as idaeslog
 
 currdir = this_file_dir()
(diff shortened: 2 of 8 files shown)


idaes/core/util/scaling.py
```diff
@@ -230,6 +233,26 @@ def set_scaling_factor(c, v, data_objects=True, overwrite=True):
...
+                "Please use only one of set_scaling_factor and constraint_scaling_transform "
+                "per constraint to avoid double scaling."
+            )
+    elif isinstance(c, Constraint):
+        for condata in c.values():
+            xform_factor = get_constraint_transform_applied_scaling_factor(condata)
+            if xform_factor is not None:
+                raise RuntimeError(
+                    "Attempted to set constraint scaling factor for indexed constraint "
+                    "with transformed ConstraintData children. Please use only one of "
```
adam-a-a:
```suggestion
                    f"Attempted to set constraint scaling factor for indexed constraint {c.name}"
                    f"with transformed ConstraintData children. Please use only one of "
```
I think it could be useful to directly point to the constraint where the issue arises, in the case that the user might not be applying scaling to constraints manually, one at a time.

idaes/core/util/scaling.py
```diff
@@ -460,6 +483,14 @@ def constraint_scaling_transform(c, s, overwrite=True):
     s = pyo.value(s)
     if not isinstance(c, ConstraintData):
         raise TypeError(f"{c} is not a constraint or is an indexed constraint")
+    scaling_factor = sfFinder.find(component_data=c)
+    if scaling_factor is not None:
+        raise RuntimeError(
+            "Attempted to transform constraint with existing scaling factor. "
```
adam-a-a:
```suggestion
            f"Attempted to transform Constraint {c.name} with existing scaling factor. "
```

adam-a-a: Again, it'd be nice to call out the exact constraint with the issue when the user is not carefully scaling constraints one at a time.

idaes/config.py
```diff
@@ -385,7 +385,7 @@ def _new_idaes_config_block():
         "scale_model",
         pyomo.common.config.ConfigValue(
             domain=Bool,
-            default=False,  # TODO: Change to true once transition complete
+            default=True,
```
adam-a-a, line 388: high-level: what happens when we do not set scale_model to True by default? At this point, I am a bit mixed up on what's happening with regard to scaling and how the solver will "see" things.

dallan-keylogic: When `scale_model == False`, then scaling factors are not applied to the model at the time an .nl file is written for the solver. Constraints transformed through `constraint_scaling_transform` will have their scaled form passed to the solver. If you pass the `nlp_scaling_method=user-scaling` option to IPOPT, IPOPT will use the user-provided scaling factors when computing and inverting the barrier hessian instead of its own `gradient-scaling` method. However, it will *not* use scaled values of the variables when computing the distance of a variable to its bounds or when computing the constraint residual (the latter fact being the reason that `constraint_scaling_transform` was created).

Now that `scale_model==True`, solvers that support the new .nl writer (definitely IPOPT, I'm not sure about other solvers) will only see the scaled model. The Jacobian will be computed using the scaled variables, and IPOPT will apply its gradient-based scaling on top of the user scaling factors (which is a good thing). Scaled values of variables will be used to compute the distance to bounds and scaled values will be used in computing the constraint residual (so we no longer need `constraint_scaling_transform`).

bpaul4: In light the benefits described in your second paragraph, why would one want to set `scale_model == False`? Is there a situation when that would be more useful than `True`?

idaes/core/scaling/tests/test_util.py
```diff
@@ -1136,6 +1137,52 @@ def test_set_scaling_factor_overwrite_false(self, caplog):
...
+        for idx in m.c4:
+            constraint_scaling_transform(m.c4[idx], 1e-3)
+
+        with pytest.raises(RuntimeError, match=match("c1")):
+            set_scaling_factor(m.c1, 1e-3)
+        with pytest.raises(RuntimeError, match=match("c2")):
+            set_scaling_factor(m.c2, 1e-3)
+        with pytest.raises(RuntimeError, match=match("c3")):
+            set_scaling_factor(m.c3, 1e-3)
+        with pytest.raises(
```
bpaul4, line 1778: Do you have a test of the correct way to apply scaling to `c4`?

idaes/config.py
dallan-keylogic: Prior to Pyomo 6.9.3, there was a bug in how scaling factors were applied in linear named `Expressions`---namely, they weren't applied in such expressions. Because those `Expressions` gave incorrect values, most models in the IDAES repo either failed to solve or gave erroneous results. It was fixed in this release, which is why I'm making this change now.

The only situation where you wouldn't want to scale the model is if you suspect a constraint has both a scaling factor set and has been transformed through `constraint_scaling_transform`. Then using `scale_model == True` results in double scaling, whereas with `scale_model == False` the double scaling would be less of a problem. That's why I went through and added a bunch of `Exceptions` if the user tried double scaling things.

idaes/core/scaling/tests/test_util.py
dallan-keylogic: I'm afraid I don't follow. I test applying a scaling factor both to the constraint as a whole as well as the individual indices. What other case do you want to test?

bpaul4: Okay, I see now that the allowed case is also being tested. I only saw the error check on `c4` before.

bpaul4 (approved):

adam-a-a (approved):
LGTM - just some pylint test failing

PR #1656, opened 2025-09-03
Modular properties fixes

dallan-keylogic:
## Fixes
- Modular properties initializer now correctly uses the passed log level
- Updated initialization to ensure we have 0 degrees of freedom for FpcTP initialization
- Updated initialization to ensure that the postcheck doesn't raise an error because the VLE constraints are not satisfied
- VLE in `Ideal` equation of state now uses equilibrium temperature rather than actual temperature
- Log fugacity in `Ideal` now uses log-form variables for mole fraction
- Antoine equation now returns log expression directly for `RPP5`
- `RPP5` now responds to configuration about whether or not to include enthalpy of formation in enthalpy calculations

### Legal Acknowledgement

### DIFF
--- a/idaes/core/initialization/initializer_base.py
+++ b/idaes/core/initialization/initializer_base.py
@@ -275,9 +275,9 @@ class InitializerBase:
         Returns:
             None
         """
-        try:
+        if hasattr(model, "fix_initialization_states"):
             model.fix_initialization_states()
-        except AttributeError:
+        else:
             _log.info_high(
                 f"Model {model.name} does not have a fix_initialization_states method - attempting to continue."
             )
--- a/idaes/models/properties/modular_properties/base/generic_property.py
+++ b/idaes/models/properties/modular_properties/base/generic_property.py
@@ -1290,7 +1290,7 @@ class ModularPropertiesInitializer(InitializerBase):
         scaler_block = [blk for blk in model.values()][0]
 
         if hasattr(scaler_block, "inherent_equilibrium_constraint") and (
-            not scaler_block.params._electrolyte
+            not scaler_block.params._electrolyte # TODO why do we need this check?
             or scaler_block.params.config.state_components == StateIndex.true
         ):
             init_log.debug(
--- a/idaes/models/properties/modular_properties/base/generic_property.py
+++ b/idaes/models/properties/modular_properties/base/generic_property.py
@@ -1274,10 +1274,10 @@ class ModularPropertiesInitializer(InitializerBase):
         """
         # Setup loggers
         init_log = idaeslog.getInitLogger(
-            model.name, self.config.output_level, tag="properties"
+            model.name, self.get_output_level(), tag="properties"
         )
         solve_log = idaeslog.getSolveLogger(
-            model.name, self.config.output_level, tag="properties"
+            model.name, self.get_output_level(), tag="properties"
         )
 
         # Create solver object
@@ -1286,6 +1286,32 @@ class ModularPropertiesInitializer(InitializerBase):
             solver_options=self.config.solver_options,
             writer_config=self.config.solver_writer_config,
         )
+
+        scaler_block = [blk for blk in model.values()][0]
+
+        if hasattr(scaler_block, "inherent_equilibrium_constraint") and (
+            not scaler_block.params._electrolyte
+            or scaler_block.params.config.state_components == StateIndex.true
+        ):
+            init_log.debug(
+                "Cannot converge inherent reaction constraints "
+                "at the state block level. They need to be solved "
+                "at the control volume level, when material and "
+                "energy balances are included. Ignoring "
+                "constraints with large residuals."
+            )
+            self.config.constraint_tolerance = float("inf")
+
+        if "flow_mol_phase_comp" or "mole_frac_phase_comp" in scaler_block.define_state_vars():
+            init_log.debug(
+                "Cannot converge phase equilibrium constraints "
+                "at the state block level due to using phase component "
+                "flows as state variables. These constraints need to be "
+                "solved at the control volume level, when material and "
+                "energy balances are included. Ignoring "
+                "constraints with large residuals."
+            )
+            self.config.constraint_tolerance = float("inf")
 
         init_log.info("Starting initialization routine")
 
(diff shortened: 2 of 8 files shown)


idaes/core/initialization/initializer_base.py
```diff
@@ -275,9 +275,9 @@ def fix_initialization_states(self, model: Block):
...
             None
         """
-        try:
+        if hasattr(model, "fix_initialization_states"):
             model.fix_initialization_states()
-        except AttributeError:
+        else:
             _log.info_high(
                 f"Model {model.name} does not have a fix_initialization_states method - attempting to continue."
             )
```
dallan-keylogic, line 283: This change revealed that there was an error in the legacy Cubic EoS initialization---the `fix_initialization_states` method was trying to deactivate `sum_mol_frac_out`, but that constraint doesn't appear to exist (at least, not with that name), which was raising an `AttributeError` that was swallowed in the try/except block. That illustrates why this change was necessary.

Since we said that we were going to remove the legacy CubicEoS in the May release, the straightforward solution is to just remove it now.

idaes/models/properties/modular_properties/base/utility.py
```diff
@@ -103,10 +103,13 @@ def get_method(self, config_arg, comp=None, phase=None):
...
-        mthd = c_arg.return_expression
-    except AttributeError:
-        mthd = c_arg
+    if not log:
+        try:
+            mthd = c_arg.return_expression
+        except AttributeError:
+            mthd = c_arg
+    else:
+        mthd = c_arg.return_log_expression
```
dallan-keylogic: Do we want to require the user to add a `return_log_expression` function to their implementation of Antoine's law?

tannerpolley (approved):

idaes/models/properties/modular_properties/base/generic_property.py
```diff
@@ -1287,6 +1287,35 @@ def initialization_routine(
             writer_config=self.config.solver_writer_config,
         )
 
+        scaler_block = [blk for blk in model.values()][0]
+
+        if hasattr(scaler_block, "inherent_equilibrium_constraint") and (
+            not scaler_block.params._electrolyte  # TODO why do we need this check?
```
agarciadiego: is TODO necessary? I beleive it creates confusion on who we are talking to

idaes/models/properties/modular_properties/base/generic_property.py
```diff
@@ -1287,6 +1287,35 @@ def initialization_routine(
...
+            or scaler_block.params.config.state_components == StateIndex.true
+        ):
+            init_log.debug(
+                "Cannot converge inherent reaction constraints "
```
agarciadiego, line 1297: This warning may be confusing. When we say "they need to be solved at the control volume level," the user may not know if that is something they must now do.

idaes/models/properties/modular_properties/base/generic_property.py
```diff
@@ -1651,12 +1683,19 @@ def fix_initialization_states(self):
...
                 k.sum_mole_frac_out.deactivate()
             except AttributeError:
                 pass
+            # Don't need equilibrium constraint for phase component flows
+            if (
+                "flow_mol_phase_comp" in k.define_state_vars()
+                or "mole_frac_phase_comp" in k.define_state_vars()
+            ):
+                k.equilibrium_constraint.deactivate()
+            # TODO Inherent reactions with a true component basis will fail here too
```
agarciadiego: Do not understant TODO message

idaes/models/properties/modular_properties/base/generic_property.py
```diff
@@ -2267,6 +2306,20 @@ def calculate_scaling_factors(self):
                 # for small molecules. For large molecules, this value will be inappropriate
                 iscale.set_scaling_factor(self.cp_mol_phase[p], 1 / 50, overwrite=False)
 
+        if self.is_property_constructed("fug_phase_comp"):
+            for idx in self.fug_phase_comp:
+                sf_x = iscale.get_scaling_factor(
+                    self.mole_frac_phase_comp[idx],
+                    default=1e3,  # I'd prefer 10, but this is consistent with existing scaling
```
agarciadiego: Remove "I prefer" comment or change wording

idaes/models/properties/modular_properties/eos/ideal.py
```diff
@@ -240,34 +240,49 @@ def entr_mol_phase_comp(b, p, j):
...
-                return log(b.get_mole_frac(p)[p, j]) + log(
-                    get_method(b, "pressure_sat_comp", j)(b, cobj(b, j), b.temperature)
-                )
-            else:
-                return Expression.Skip
+    @staticmethod
+    def log_fug_phase_comp_eq(b, p, j, pp):
+        if (p, j) not in b.phase_component_set:
+            raise KeyError(f"Component {j} is not present in phase {p}.")
+        # TODO does allowing a calculation with
```
agarciadiego: Change wording on TODO message

idaes/models/properties/modular_properties/eos/ideal.py
```diff
@@ -290,88 +305,96 @@ def fug_coeff_phase_comp_eq(b, p, j, pp):
     @staticmethod
     def log_fug_phase_comp_Tbub(b, p, j, pp):
         pobj = b.params.get_phase(p)
-        cobj = b.params.get_component(j)
+        if b.params.has_inherent_reactions:
+            raise PropertyNotSupportedError(
+                "Bubble/dew calculations for systems with inherent "
```
agarciadiego, line 309: This is only a question. What would be required to support this?

idaes/models/properties/modular_properties/base/generic_property.py
dallan-keylogic: Would you like me to remove the TODO or the entire comment?

idaes/models/properties/modular_properties/eos/ideal.py
dallan-keylogic: It depends. If all we want to do is have _any_ temperature at which we can solve the VLE equations for `SmoothVLE`, then we can probably just use the current system composition in the bubble/dew calculations. However, if we want `temperature_bubble` and `temperature_dew` to be the system's actual bubble and dew temperatures, we would need to write inherent reaction equilibrium equations not just at the system's temperature, but also at the bubble/dew temperature. That adds a lot of complexity and overhead, so I'd rather punt on the issue until we have a specific problem we're trying to solve.

We don't have this issue with the `CubicComplementarityVLE` because the equilibrium temperature there is not directly related to the bubble/dew temperatures.

dallan-keylogic: Of course, we're not going to be using `CubicComplementarityVLE` with the `Ideal` EoS anyway.

PR #1658, opened 2025-09-08
Scaling warning

dallan-keylogic:
## Summary/Motivation:
Add a warning when trying to get a scaling factor and it does not exist.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/scaling/tests/test_scaling_base.py
+++ b/idaes/core/scaling/tests/test_scaling_base.py
@@ -66,10 +66,32 @@ class TestScalerBase:
         assert not sb.config.overwrite
 
     @pytest.mark.unit
-    def test_get_scaling_factor(self, model):
+    def test_get_scaling_factor(self, model, caplog):
+        model.x = Var()
         sb = ScalerBase()
 
         assert sb.get_scaling_factor(model.v[1]) == 1
+
+        with caplog.at_level(idaeslog.WARNING):
+            sf = sb.get_scaling_factor(model.x)
+        assert len(caplog.text) == 0
+        assert sf is None
+
+        with caplog.at_level(idaeslog.WARNING):
+            sf = sb.get_scaling_factor(model.x, warning=True)
+        assert sf == None
+        assert "Missing scaling factor for x" in caplog.text
+        caplog.clear()
+
+        with caplog.at_level(idaeslog.WARNING):
+            sf = sb.get_scaling_factor(model.x, default=37, warning=False)
+        assert len(caplog.text) == 0
+        assert sf == 37
+
+        with caplog.at_level(idaeslog.WARNING):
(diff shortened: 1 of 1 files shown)


MarcusHolly (approved):
LGTM

idaes/core/scaling/util.py
```diff
@@ -461,12 +461,18 @@ def get_scaling_factor(component, default=None):
...
+            "Can only get scaling hints for named expressions, but component was an unnamed expression."
+        )
     sfx = get_component_scaling_suffix(component)
 
     try:
         return sfx[component]
     except (AttributeError, KeyError):
         # No scaling factor found, return the default value
+        if warning:
+            _log.warning(f"Missing scaling factor for {component.name}")
```
adam-a-a: Should we make the user aware that he or she can add a default, but preferably, should add an appropriate scaling factor?

bpaul4: I'm concerned this warning will result in a huge log output for many cases where users don't have scaling, or don't want/need to scale everything. Can we use this error check as a debug step, like a certain level of info reporting but otherwise don't display for every component? The current "missing scaling suffix" warnings can already get very verbose.

dallan-keylogic: Because of the way the scaling toolbox is set up now, values in the default dictionary will be ignored unless the scaler object specifically calls `scale_variable_by_default` (see #1648 item 3).

dallan-keylogic: We thought about that, but having the log output is a lesser evil than either 1) having an exception thrown or 2) having the scaling silently fail.

There were two contributing reasons to the log spam using the old tools. First, many times the scaling routines were looking for scaling factors in the wrong places, e.g., looking for scaling factors on an indexed `Var` when scaling factors had been set for the `VarData` children of that indexed `Var`. Now an exception is thrown if the user tries to call `get_scaling_factor` on an indexed `Var`. Second, the old scaling tools demanded the user set scaling factors for a ton of variables at the `ControlVolume` level, such as for heat, mass transfer, extent of reaction, etc. The `ControlVolume0DScaler` provides a reasonable initial guess for those scaling factors that should work for most use cases. Finally, we're now going to raise an exception if the user doesn't provide default scaling factors for the modular properties, so they'll be forced to provide a bare minimum of scaling.

Ideally, now the warning will be shown only if something has gone seriously wrong with scaling, either because a scaler was written incorrectly or there was no scaler written for the property package in question. We'll see how it works in practice.

One thing I'd like to do is enable the Pytest option that turns warnings into exceptions to catch any issues in the main repository. We still need to 1) write a log filter object to get rid of the Pyomo warnings about elements of a suffix not being exported (so the end user of IDAES doesn't even know Pyomo is logging those warnings) and 2) figure out a way to filter out warnings from the old scaling tools when tests are run.

PR #1659, opened 2025-09-10
Fix get default scaling factor

dallan-keylogic:
## Fixes
`get_default_scaling_factor` formerly looked for `component.local_name` in the `default_scaling_factors` dictionary. However, that caused problems for indexed variables, where the user could plausibly set `default_scaling_factors[mole_frac_phase_comp[Liq, H2O]` and have it be ignored because of the space after the comma. This PR fixes that using the component finder. The implementation, which iterates over all entries in the dictionary, isn't particularly efficient, but we don't expect these dictionaries to be that long---we can always optimize later.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/scaling/custom_scaler_base.py
+++ b/idaes/core/scaling/custom_scaler_base.py
@@ -229,18 +229,31 @@ class CustomScalerBase(ScalerBase):
         Returns:
             default scaling factor if it exists, else None
         """
-        try:
-            return self.default_scaling_factors[component.local_name]
-        except KeyError:
-            # Might be indexed, see if parent component has default scaling
-            parent = component.parent_component()
-            try:
-                return self.default_scaling_factors[parent.local_name]
-            except KeyError:
-                # No default scaling found, give up
-                pass
+        blk = component.parent_block()
+        parent_default = None
+        # We're not just returning self.default_scaling_factors[component.local_name]
+        # because the component finder has additional logic to handle, e.g., spaces
+        # separating indices for elements of an indexed component. We want the user
+        # to be able to provide either "var[a,b]" or "var[a, b]" (with a space following
+        # the comma) and still find the right component
 
-            # Log a message and return nothing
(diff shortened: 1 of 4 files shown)


blnicho:
@dallan-keylogic the `lock_attribute_creation_context` is new to me but this seems like its intended purpose.

idaes/core/scaling/custom_scaler_base.py
```diff
@@ -229,18 +229,47 @@ def get_default_scaling_factor(
...
+        # from getting triggered through this lookup.
+        if hasattr(blk, "_lock_attribute_creation"):  # pylint: disable=protected-access
+            lock_attribute_creation_orig = (
+                blk._lock_attribute_creation  # pylint: disable=protected-access
+            )
+            blk._lock_attribute_creation = True  # pylint: disable=protected-access
+        for key in self.default_scaling_factors:
+            comp2 = blk.find_component(key)
+            if comp2 is component:
+                comp_default = self.default_scaling_factors[key]
```
blnicho, line 260:
```suggestion
            if comp2 is component:
                comp_default = self.default_scaling_factors[key]
                break
```
In the large comment above you mention that if multiple matching string names appear for a component in `default_scaling_factors` then the first value encountered when iterating will be returned. I think you need a `break` here to get that behavior.

idaes/core/scaling/util.py
```diff
@@ -444,7 +444,7 @@ def _suffix_from_dict(
             )
 
 
-def get_scaling_factor(component, default: float = None, warning=True):
+def get_scaling_factor(component, default: float = None, warning: bool = False):
```
blnicho, line 447: `default` and `warning` are not described in the function docstring

blnicho (commented):
@dallan-keylogic I have a few minor change requests but otherwise this looks fine

agarciadiego (approved):

PR #1663, opened 2025-09-16
Convert setup py content to a toml file

sufikaur:
## Fixes

[Issue 1655](https://github.com/issues/assigned?issue=IDAES%7Cidaes-pse%7C1655)
## Summary/Motivation:
SInce setup is a less modern approach it could be causing warnings/ isssues like described in in [issue  1476](https://github.com/issues/assigned?issue=IDAES%7Cidaes-pse%7C1476)
-
-

### Legal Acknowledgement


blnicho:
Tagging @mrmundt to take a look at this since she recently did the same conversion for Pyomo

pyproject.toml
```diff
@@ -0,0 +1,138 @@
...
+    "Topic :: Scientific/Engineering :: Chemistry",
+    "Topic :: Software Development :: Libraries :: Python Modules",
+]
+keywords=["IDAES", "idaes-pse", "energy systems", "chemical engineering", "process modeling"]
+dynamic = ["version"]
+dependencies = [
+  "pyomo >= 6.9.3",
+  "pint >= 0.24.1", # required to use Pyomo units. Pint 0.24.1 needed for Python 3.9 support
+  "networkx", # required to use Pyomo network
+  "numpy>=1,<3", # pandas constraint added on 2023-08-30 b/c bug in v2.1  # see IDAES/idaes-pse#1253
```
mrmundt: This comment I suspect is on the wrong line

mrmundt (commented):
Some initial thoughts here:
1. I would recommend doing something similar to what we have in Pyomo. We made different categories for different types of dependencies, and you all could do the same with the dev dependencies. That way you don't need to have `requirements-dev.txt` anymore. (It looks like you started doing this with the optional-dependencies category, so go all the way!!)
2. If you do that ^^, then the docs will need to be updated: https://github.com/IDAES/idaes-pse/blob/main/docs/tutorials/advanced_install/index.rst
3. This is not your doing, but where is the Python version actually defined besides in the classifiers? Are they actually just polite suggestions?

pyproject.toml
sufikaur: I took this directly from the setup.py file. I think it may be okay?

A1/A2: As per the last dev call, I'll split this into two/ keep it in mind to discuss further. 

A3. I don't know if it is defined anywhere! @lbianchi-lbl

lbianchi-lbl: > This comment I suspect is on the wrong line

@mrmundt do you mean the comment on L43, i.e. where the `numpy` dependency is defined? If so, I think you might be right, as the comment refers to `pandas`. So that comment should be moved to the line where the `pandas` dependency is defined.

As for the Python version, AFAIK the classifiers are not enforced. From a cursory search (https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#python-requires), `requires-python` seems to be the way to define this in `pyproject.toml`

sufikaur: Thanks, Ludovico.
Is there a reason it was not enforced? Happy to add a line being stricter about enforcing python versions.

sufikaur: Fixed the comment line

pyproject.toml
```diff
@@ -0,0 +1,147 @@
...
+]
+
+# this will install all optional dependencies
+all = ["idaes-pse[ui,grid,omlt,coolprop]"]
+
+[tool.setuptools_scm]
+# https://setuptools-scm.readthedocs.io/en/v8.1.0/extending/
+local_scheme = "node-and-date"
+version_scheme = "only-version"
+
```
sufikaur, line 78: `ideas/ver.py` is still being used because it is being tested. 
As part of this PR, that file should be removed and remove the test associated with it. 
Must ensure, that in the top level `__init__.py` make sure that the attribute __version__ is set correctly using` importlib.metadata.version("idaes-pse")`.

PR #1673, opened 2025-09-26
Control Volume 1D Scaler

dallan-keylogic:
## Fixes
- Allows submodel scalers to be called on `BlockData` directly, not just `IndexedBlocks`

## Summary/Motivation:
Adds a scaler for the `ControlVolume1D`.

## Changes proposed in this PR:
- Adds scaler for `ControlVolume1D`
- Adds public `flow_direction` property to allow public access to the `_flow_direction` attribute.

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1676, opened 2025-10-01
Mixer Scaler

dallan-keylogic:
## Summary/Motivation:
Scaler object for the mixer unit model. Depends on #1673.

### Legal Acknowledgement

### DIFF
--- a/idaes/models/unit_models/tests/test_mixer.py
+++ b/idaes/models/unit_models/tests/test_mixer.py
@@ -1796,7 +1796,7 @@ class TestInitializersIAPWSEquality:
 
 # -----------------------------------------------------------------------------
 # Inherent reaction case
-# We do not want to make this property package an example because it has non-SI units.
+# We do not want to make this property package an example because it has non-SI base units.
 class LeachSolutionScaler(CustomScalerBase):
     """
     Scaler object for leach solution properties


idaes/models/unit_models/tests/test_mixer.py
```diff
@@ -1742,10 +1885,56 @@ def fix_initialization_states(self):
...
+        "conc_mass_comp[HSO4]": 1e-2,
+        "conc_mass_comp[SO4]": 1e-1,
+    }
+
+    def variable_scaling_routine(
+        self, model, overwrite: bool = False, submodel_scalers: ComponentMap = None
+    ):
+        self.scale_variable_by_default(model.flow_vol, overwrite=overwrite)
+        for var in model.conc_mass_comp.values():
+            self.scale_variable_by_default(var, overwrite=overwrite)
```
MarcusHolly: What happens if you do this, but one or more of the components doesn't have a default scaling factor. Would this still run successfully? If so, what scaling factor gets set for the unspecified component? If this fails, how does it fail? Warning/error message and, if so, is this tested?

idaes/models/unit_models/tests/test_mixer.py
```diff
@@ -1742,10 +1885,56 @@ def fix_initialization_states(self):
                 self[k].hso4_dissociation.deactivate()
 
 
+class LeachSolutionScaler(CustomScalerBase):
```
MarcusHolly: Why is this being built in this file?

idaes/models/unit_models/tests/test_mixer.py
```diff
@@ -1809,62 +1998,153 @@ def define_state_vars(self):
...
+        )
+
+        assert value(unit.inherent_reaction_extent[0, "Ka2"]) == pytest.approx(
+            0, abs=1e-6
+        )
+
+    @pytest.mark.component
+    @pytest.mark.solver
+    def test_no_numerical_warnings(self, model):
+        # TODO revisit once the diagnostics toolbox takes scaling into account
```
MarcusHolly, line 2140: Just out of curiosity, is this something that is being actively worked on? I think it'd be amazing if the toolbox could take scaling into account, but I recall Andrew mentioning that this was easier said than done.

idaes/core/base/control_volume1d.py
```diff
@@ -77,6 +81,258 @@ class DistributedVars(Enum):
...
+
+        if hasattr(model, "material_flow_dx"):
+            for idx, v in model._flow_terms.items():  # pylint: disable=protected-access
+                # As domain is normalized, derivative should have same
+                # scale as flow
+                self.scale_variable_by_component(
+                    model.material_flow_dx[idx], v, overwrite=overwrite
+                )
+
+        # TODO elemental flows
```
bpaul4, line 174: Will this be part of another PR?

idaes/core/base/control_volume1d.py
```diff
@@ -77,6 +81,258 @@ class DistributedVars(Enum):
...
+                    v, model.enthalpy_flow_linking_constraint[idx], overwrite=overwrite
+                )
+
+        if hasattr(model, "enthalpy_flow_dx"):
+            for (
+                idx,
+                v,
+            ) in model._enthalpy_flow.items():  # pylint: disable=protected-access
+                # Normalized domain, so scale should be the same as flow
+                # TODO is this correct?
```
bpaul4, line 217: Does this still need to be addressed?

idaes/core/base/control_volume_base.py
```diff
@@ -110,6 +111,18 @@ class ControlVolumeScalerBase(CustomScalerBase):
...
+    # String name of the reference state block to use when scaling
+    # balance equations and material/heat transfer terms. This should be
+    # set by the scaler object that inherits from this base
+    _state_block_ref = None
+
+    # Attribute name to use as a weight when scaling material and energy
+    # terms. Presently (9/25/25), this attribute exists to take into
+    # account the fact that all the material and energy terms in the
+    # ControlVolume1D are given on the basis of material or energy per
+    # unit length, so we want to weigh them accordingly.
```
bpaul4:
```suggestion
    # unit length, so we want to weight them accordingly.
```

idaes/core/base/tests/test_control_volume_0d_scaler_object.py
```diff
@@ -34,6 +34,10 @@
 from idaes.core.base.control_volume0d import ControlVolume0DScaler
 
 
+def approx(x):
+    return pytest.approx(x, rel=1e-15, abs=0)
```
bpaul4, line 38: This tolerance is extremely tight, could there be failures in the future from this?

idaes/models/unit_models/tests/test_mixer.py
dallan-keylogic: We needed a test case with inherent reactions. I considered pulling both this scaler and the parameter block it's designed to scale out of this test file, but I saw that it had non-SI base units and didn't want to spread that further in the IDAES code base.

idaes/models/unit_models/tests/test_mixer.py
dallan-keylogic: It raises an exception:
```
raise ValueError(
    "This scaler requires the user to provide a default "
    f"scaling factor for {component}, but no default scaling "
    "factor was set."
)
```

idaes/models/unit_models/tests/test_mixer.py
dallan-keylogic: It's actually straightforward to do now. We had issues when we had `scale_model=False` in the writer config, so it was unclear whether the `user_scaling` option was being used for IPOPT. Now we have `scale_model=True`, so we can always return the scaled version of the Jacobian, constraint residuals, and variable distance from bounds.

idaes/models/unit_models/tests/test_mixer.py
MarcusHolly: In that case, could you just leave an in-line comment/TODO note stating what you've said here.

idaes/models/unit_models/tests/test_mixer.py
```diff
@@ -1653,6 +1796,50 @@ def test_block_triangularization(self, model):
 
 # -----------------------------------------------------------------------------
 # Inherent reaction case
+class LeachSolutionScaler(CustomScalerBase):
```
dallan-keylogic, line 1800:
```suggestion
# We do not want to make this property package an example because it has non-SI units.
class LeachSolutionScaler(CustomScalerBase):
```

idaes/models/unit_models/tests/test_mixer.py
```diff
@@ -1653,6 +1796,51 @@ def test_block_triangularization(self, model):
 
 # -----------------------------------------------------------------------------
 # Inherent reaction case
+# We do not want to make this property package an example because it has non-SI units.
```
dallan-keylogic:
```suggestion
# We do not want to make this property package an example because it has non-SI base units.
```

idaes/core/base/control_volume1d.py
dallan-keylogic: Yes.

idaes/core/base/control_volume1d.py
dallan-keylogic: Yes, but now isn't the time to do it. I need to take some time to sit down with the advection equations and explore the conditioning of the solution, both in the case of a single `ControlVolume1D` and a counter-current unit model such as the `HeatExchanger1D`. The current scaling is adequate for many purposes.

idaes/core/base/tests/test_control_volume_0d_scaler_object.py
dallan-keylogic: Here we're trying to match scaling factors like `1/10 * 1/5 * 3/10`. Floating point arithmetic is not associative, so `(1/10 * 1/5)*(3/10) !=(1/10) * (1/5*3/10) `. However, it will be within a relative tolerance of machine epsilon, which is `2.22e-16` for double precision arithmetic in Python. So no, there shouldn't be failures in the future from this.

PR #1681, opened 2025-10-15
MSContactor Scaler

dallan-keylogic:
## Summary/Motivation:
Adds scaler object for MSContactor unit model.

### Legal Acknowledgement

### DIFF
--- a/idaes/models/unit_models/mscontactor.py
+++ b/idaes/models/unit_models/mscontactor.py
@@ -55,7 +55,6 @@ from idaes.core.initialization.initializer_base import StoreState
 from idaes.core.util.model_serializer import to_json, from_json
 import idaes.logger as idaeslog
 from idaes.core.util.units_of_measurement import report_quantity
-from idaes.core.util.exceptions import ConfigurationError
 from idaes.core.scaling import (
     ConstraintScalingScheme,
     CustomScalerBase,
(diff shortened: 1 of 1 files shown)


idaes/models/unit_models/mscontactor.py
```diff
@@ -54,11 +55,617 @@
...
+    Scaler for the MSContactor.
+    """
+
+    DEFAULT_SCALING_FACTORS = {
+        # We could scale volume by magnitude if it were being fixed
+        # by the user, but we often have the volume given by an
+        # equality constraint involving geometry in the parent
+        # unit model.
+        "volume": DefaultScalingRecommendation.userInputRequired,
+        "volume_frac_stream": 2,
```
MarcusHolly, line 78: Why is this 2 as opposed to 1 or 10? Just because the average volume fraction is 0.5?

idaes/models/unit_models/mscontactor.py
```diff
@@ -54,11 +55,617 @@
...
+        # Phase fraction may have already been created by the property package.
+        # Also, no "phase_fraction" Var exists on the MSContactor, only
+        # "stream_name_phase_fraction" Vars. Because we do not know the stream
+        # names a priori, this value is copied for those streams as part of
+        # the variable_scaling_routine (if the user hasn't already set a value)
+        "phase_fraction": 10,
```
MarcusHolly, line 84: Same comment as above... why not 1 or 2?

MarcusHolly (approved):
I agree with former comments that it'd be nice to flesh out the documentation we have on the current scaling tools, but I recognize that 's not as high priority as getting these things merged before the end of the month

idaes/models/unit_models/mscontactor.py
dallan-keylogic: Yeah. There will typically be two streams, and, without additional user information, we'd expect the volume to be evenly divided between streams.

idaes/models/unit_models/mscontactor.py
dallan-keylogic: I started using 10 for phase fraction because it's what I was using for mole fraction. 2 might have been a better choice, but I put 10 elsewhere so I put it here to be consistent.

PR #1683, opened 2025-10-17
Update pylint to 3.3.9

blnicho:
## Summary/Motivation:
The Pyomo 6.9.5 release broke pylint testing in IDAES which is pinned to a very old version of pylint. The easiest fix is to bump the version of pylint to something more recent. This PR bumps pylint to 3.3.9 and fixes new pylint failures stemming from new checks that have been introduced namely `used-before-assignment` and `possibly-used-before-assignment`. To fix the warnings I used a combination of code refactoring or skipping the checks.

- Update to pylint 3.3.9
- Fix or disable new pylint warnings

### Legal Acknowledgement

### DIFF
--- a/idaes/core/base/unit_model.py
+++ b/idaes/core/base/unit_model.py
@@ -244,8 +244,9 @@ Must be True if dynamic = True,
                         _idx = block.length_domain.last()
 
                     port, ref_name_list = sblock.build_port(
+                        doc,
                         # pylint: disable-next=possibly-used-before-assignment
-                        doc, slice_index=(slice(None), _idx)
+                        slice_index=(slice(None), _idx),
                     )
 
                 except AttributeError:
@@ -343,8 +344,9 @@ Must be True if dynamic = True,
                         _idx = block.length_domain.last()
 
                     port, ref_name_list = sblock.build_port(
+                        doc,
                         # pylint: disable-next=possibly-used-before-assignment
-                        doc, slice_index=(slice(None), _idx)
+                        slice_index=(slice(None), _idx),
                     )
 
                 except AttributeError:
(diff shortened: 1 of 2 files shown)


.github/workflows/core.yml
```diff
@@ -67,7 +67,7 @@ jobs:
           pip --no-cache-dir install --progress-bar off "$black_requirement"
       - name: Run Black to verify that the committed code is formatted
         run: |
-          black --check .
+          black --check --diff .
```
dallan-keylogic, line 70: Is the idea that we need to run `black` on only the files that have changes in them?

idaes/core/base/unit_model.py
```diff
@@ -342,7 +344,9 @@ def add_outlet_port(self, name=None, block=None, doc=None):
                         _idx = block.length_domain.last()
 
                     port, ref_name_list = sblock.build_port(
-                        doc, slice_index=(slice(None), _idx)
+                        doc,
+                        # pylint: disable-next=possibly-used-before-assignment
+                        slice_index=(slice(None), _idx),
```
dallan-keylogic, line 349: Ideally, in a situation like this, we would have an `else` clause that raises `BurntToast` (even if CodeCov doesn't like us adding code that is theoretically inaccessible). Skipping the Pylint checks is a good solution for now, though.

.github/workflows/core.yml
blnicho: The `--diff` command will display the changes that black would like to make. It can be convenient to fix 1 character issues reported by black directly in the GitHub interface without having to run black locally.

idaes/core/base/unit_model.py
blnicho: Exactly! There were several instances where I skipped the pylint check because I wasn't sure of a "safe" default value and I wanted to make sure an error was thrown if the enum options ever changed.

PR #1688, opened 2025-10-22
Scaler object for Helmholtz Properties

dallan-keylogic:
## Summary/Motivation:
Adds a Scaler object for the Helmholtz properties.

### Legal Acknowledgement

### DIFF
--- a/idaes/models/properties/general_helmholtz/tests/test_stateblock.py
+++ b/idaes/models/properties/general_helmholtz/tests/test_stateblock.py
@@ -25,7 +25,6 @@ from idaes.models.properties.general_helmholtz import (
     StateVars,
 )
 from idaes.core import FlowsheetBlock
-from idaes.core.util.scaling import get_jacobian, jacobian_cond
 
 
 @pytest.mark.unit


idaes/models/properties/general_helmholtz/tests/test_stateblock.py
```diff
@@ -37,6 +40,156 @@ def test_set_not_fixed():
...
+    assert sb.scaling_hint[sb.temperature] == 1e-1
+    assert sb.scaling_hint[sb.vapor_frac] == 10
+    assert sb.scaling_hint[sb.enth_mol_sat_phase["Liq"]] == 1e-2
+    assert sb.scaling_hint[sb.entr_mol_phase["Vap"]] == 1e-1
+    assert sb.scaling_hint[sb.cp_mol_phase["Vap"]] == 1e-2
+    assert sb.scaling_hint[sb.cv_mass_phase["Liq"]] == 1e-3
+    assert sb.scaling_hint[sb.dens_mol_phase["Vap"]] == 1
+    assert sb.scaling_hint[sb.energy_internal_mass] == 1e-3
+    assert sb.scaling_hint[sb.flow_vol] == pytest.approx(
+        1 / (137 * 18.015 / 1000 * pyo.sqrt(5 * 1000)), rel=1e-3
```
bpaul4, line 90: This is very specific, what is this from?

bpaul4 (approved):
One question for my own knowledge, but otherwise this looks good.

idaes/models/properties/general_helmholtz/tests/test_stateblock.py
dallan-keylogic: `1 / 137` is the scaling factor for the molar flow rate. `18.015/1000` is the molar mass of water in kg. The density of steam at ~ 1 atm and 100 C is ~ 5 kg/m^3, and the density of water is ~1000 kg/m^3. We take the geometric mean of those density values because we don't know which phase is going to be present.

idaes/models/properties/general_helmholtz/helmholtz_state.py
```diff
@@ -50,10 +50,119 @@
...
+        self.scale_variable_by_default(model.flow_vol, overwrite=overwrite)
+        if self.get_scaling_factor(model.flow_vol) is None:
+            nom_flow_mass = self.get_expression_nominal_value(model.flow_mass)
+            # Geometric mean of the mass density of steam at 100 C and 1 MPa and liquid water
+            nom_dens_mass = pyo.sqrt(1000 * 5)
+            self.set_component_scaling_factor(
+                model.flow_vol, 1 / (nom_flow_mass * nom_dens_mass), overwrite=overwrite
+            )
+
+        for varname in self.default_scaling_factors.keys():
```
MarcusHolly, line 144: So it seems like this for loop would only set the scaling factor for flow_vol if the code in Lines 140-141 hasn't already set it. Is that correct? And do we want to give priority to the method above (i.e.  if both methods are trying to set the scaling factor, is there a reason we should defer value given by the default scaling factor?)

idaes/models/properties/general_helmholtz/tests/test_stateblock.py
```diff
@@ -12,6 +12,8 @@
 #################################################################################
 """Test the parts of helmholtz_state that aren't hit by other tests in this."""
 
+import re
+
```
MarcusHolly, line 16: Seems like there's no comparison of the condition number before vs after scaling. Is this intentional?

idaes/models/properties/general_helmholtz/helmholtz_state.py
dallan-keylogic: You're right, I should skip `flow_vol` in this loop.

We want to defer to the value given by the default scaling factor because we don't know which phase the user expects (therefore we use the geometric mean of vapor and liquid densities) while the user, hopefully, knows which phase they expect and can choose an appropriate scaling factor. If the user has a unit operation that expects a vapor phase and encounters a liquid phase, or one that expects a liquid phase and encounters a vapor phase, they should expect their performance equations to go haywire anyway. That's partially why cavitation in pumps and droplets in turbines are such a big deal.

idaes/models/properties/general_helmholtz/tests/test_stateblock.py
dallan-keylogic: Yes. There are zero constraints most of the time, and with TPX state variables there is only a single active constraint, so the condition number is trivially 1.

idaes/models/properties/general_helmholtz/tests/test_stateblock.py
```diff
@@ -25,6 +25,7 @@
     StateVars,
 )
 from idaes.core import FlowsheetBlock
+from idaes.core.util.scaling import get_jacobian, jacobian_cond
```
MarcusHolly: These imports aren't needed. I'm assuming you accidentally left them while looking into my former comment

PR #1685, opened 2025-10-21
Separator and Flash Scaler

dallan-keylogic:
## Summary/Motivation:
Adds scaler objects for the `Separator` and `Flash` unit models.

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.txt file at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1689, opened 2025-10-22
Scaler for Feed, Product, and StateJunction unit models

dallan-keylogic:
## Summary/Motivation:
Adds a scaler for the `Feed`, `Product`, and `StateJunction` unit models. Because these unit models have no unit-level variables or constraints, the scaler simply calls the property package's scaler.

Depends on #1685 .

### Legal Acknowledgement

### DIFF
--- a/idaes/core/scaling/custom_scaler_base.py
+++ b/idaes/core/scaling/custom_scaler_base.py
@@ -23,20 +23,20 @@ from pyomo.core.base.units_container import UnitsError
 
 from pyomo.core.base.block import Block, BlockData
 from pyomo.core.base.constraint import ConstraintData
-from pyomo.core.base.var import VarData
-from pyomo.core.base.expression import ExpressionData
+from pyomo.core.base.var import VarData, ScalarVar
+from pyomo.core.base.expression import ExpressionData, ScalarExpression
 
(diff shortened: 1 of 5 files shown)


idaes/models/unit_models/product.py
```diff
@@ -26,7 +26,10 @@
 
 # Product blocks can reuse the Feed initializer
 # For consistency and future proofing, import with new name
-from idaes.models.unit_models.feed import FeedInitializer as ProductInitializer
+from idaes.models.unit_models.feed import (
+    FeedInitializer as ProductInitializer,
+    FeedScaler as ProductScaler,
+)
```
MarcusHolly, line 32: Is there a better name we can give the FeedInitializer and FeedScaler since they'll also be used for products and statejunctions?

MarcusHolly (approved):
One minor comment, but LGTM

idaes/models/unit_models/product.py
dallan-keylogic: Yes---we could call them `SingleStateBlockInitializer` and `SingleStateBlockScaler`. I don't want to manage a deprecation cycle for the old name right now, though.

PR #1690, opened 2025-10-24
Scale Variable from Definition Constraint

dallan-keylogic:
## Summary/Motivation:
Changes `scale_variable_from_definition_constraint` to take scaling hints into account (which are extremely useful when dealing with enthalpy and energy terms).

Also pins the Pyomo version to 6.9.4 until such time as #1683 is merged.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/scaling/custom_scaler_base.py
+++ b/idaes/core/scaling/custom_scaler_base.py
@@ -465,7 +465,7 @@ class CustomScalerBase(ScalerBase):
         """
         Helper function for scale_variable_by_definition_constraint.
 
-        Walks the constraint to build a map between Variables or Expressions
+        Iterates over the constraint to build a map between Variables or Expressions
         appearing in Constraint and the scaling factor or scaling hint assigned
         to them. (Because constraints can contain non-local variables and expressions,
         getting the scaling factor and scaling hint suffixes from the parent block of
@@ -616,7 +616,7 @@ class CustomScalerBase(ScalerBase):
             var_set = ComponentSet(var for var in identify_variables(replaced_expr))
             if not (len(var_set) == 1 and variable in var_set):
                 raise BurntToast(
-                    "Something has gone wrong in the variable replacement scheme. "
+                    "An unexpected error has occurred in the variable replacement scheme. "
                     "Please open an issue on the IDAES GitHub so that this problem "
                     "can be addressed."
                 )


idaes/core/scaling/custom_scaler_base.py
```diff
@@ -461,6 +461,69 @@ def scale_variable_by_units(self, variable, overwrite: bool = False):
...
+        the constraint is not good enough.)
+
+        Args:
+            variable: Variable we are scaling by the definition constraint "constraint".
+            constraint: Definition constraint that we are using to scale "variable" and
+                that is getting walked in this function.
+            safety_mode: Flag about whether or not to screen named expressions to see if
+                "variable" appears in them. Screening them takes longer (especially for
+                extremely deep expression trees), but if we do not screen them we run
+                the risk of removing "variable" from "constraint" entirely.
```
bpaul4: Why is this true? The screening is not actually deleting components, is it?

idaes/core/scaling/custom_scaler_base.py
```diff
@@ -482,6 +545,17 @@ def scale_variable_by_definition_constraint(
         Returns:
             None
         """
+        if variable.is_indexed():
+            raise TypeError(
+                f"Variable {variable} is indexed. Call with VarData "
+                "children instead."
```
bpaul4, line 554: Listing the index keys or "children" might be helpful if this is thrown.

idaes/core/scaling/custom_scaler_base.py
```diff
@@ -498,38 +572,64 @@ def scale_variable_by_definition_constraint(
...
+            replaced_expr = replace_expressions(
+                constraint.expr,
+                substitution_map=replacement_map,
+                descend_into_named_expressions=True,
+                remove_named_expressions=True,
+            )
+            var_set = ComponentSet(var for var in identify_variables(replaced_expr))
+            if not (len(var_set) == 1 and variable in var_set):
+                raise BurntToast(
+                    "Something has gone wrong in the variable replacement scheme. "
```
bpaul4: Maybe make this line a bit more formal sounding, since it's exposed to users.

setup.py
```diff
@@ -76,7 +76,7 @@ def __getitem__(self, key):
     # Put abstract (non-versioned) deps here.
     # Concrete dependencies go in requirements[-dev].txt
     install_requires=[
-        "pyomo >= 6.9.3",
+        "pyomo == 6.9.4",  # Temporary pin to avoid Pylint issues
```
bpaul4: Should we mark a TODO here, as well?

bpaul4 (commented):
Just a few comments.

idaes/core/scaling/custom_scaler_base.py
dallan-keylogic: You mean iterate over `variable` to list every single `VarData` child? I don't think being bombarded with a list of the form

```
fs.unit.control_volume.material_holdup[0,0,"H2O"]
fs.unit.control_volume.material_holdup[0,0,"NaOH"]
fs.unit.control_volume.material_holdup[0,0, "EthylAcetate"]
fs.unit.control_volume.material_holdup[0,0, "SodiumAcetate"]
fs.unit.control_volume.material_holdup[0,0, "Ethanol"]
fs.unit.control_volume.material_holdup[0,0.1,"H2O"]
fs.unit.control_volume.material_holdup[0,0.1,"NaOH"]
fs.unit.control_volume.material_holdup[0,0.1, "EthylAcetate"]
fs.unit.control_volume.material_holdup[0,0.1, "SodiumAcetate"]
fs.unit.control_volume.material_holdup[0,0.1, "Ethanol"]
etc...
```
is going to be particularly helpful (unless they don't know the difference between an `IndexedVar` and `VarData` in the first place).

idaes/core/scaling/custom_scaler_base.py
MarcusHolly: Something like "An unexpected failure has occurred in the variable replacement scheme. Pleas open an issue..."

idaes/core/scaling/custom_scaler_base.py
dallan-keylogic: The screening is not deleting components, but `replace_expressions` is building a new expression in which every variable and named expression in the keys of `replacement_map` is replaced by the expression in the values of `replacement_map`. Since we're going to be using `calculate_variable_from_constraint` in order to get the scaling factor, we still need `variable` to show up in the replaced expression.

idaes/core/scaling/custom_scaler_base.py
```diff
@@ -461,6 +461,69 @@ def scale_variable_by_units(self, variable, overwrite: bool = False):
                 "in self.unit_scaling_factors"
             )
 
+    def _build_replacement_map(self, variable, constraint, safety_mode: bool):
+        """
+        Helper function for scale_variable_by_definition_constraint.
+
+        Walks the constraint to build a map between Variables or Expressions
```
MarcusHolly: This doesn't _sound_ right to me. What is meant by "walks"? A synonym for "helps"?

idaes/core/scaling/custom_scaler_base.py
dallan-keylogic: Theoretically, the user should never see this line---we handle the case of `variable` not appearing in `constraint` using the `variable_in_constraint ` flag, while `safety_mode` *should* make sure that a named expression containing `variable` is not replaced. If we end up here anyway, it's because some unknown error has occurred, so we can't be more specific with the error message.

setup.py
dallan-keylogic: Done

PR #1697, opened 2025-11-04
Diagnostics Toolbox Scaling

dallan-keylogic:
## Summary/Motivation:
Changes the Diagnostics Toolbox to take model scaling into account.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/scaling/util.py
+++ b/idaes/core/scaling/util.py
@@ -1185,7 +1185,7 @@ def get_jacobian(
     vlist = nlp.vlist = nlp.get_pyomo_variables()
 
     if include_scaling_factors:
-        # Preallocaying arrays with NaNs to make it apparent if
+        # Preallocating arrays with NaNs to make it apparent if
         # some element doesn't get set for some reason
         var_scaling_factors = np.nan * np.ones(len(vlist))
         for i, var in enumerate(vlist):


idaes/core/scaling/util.py
```diff
@@ -1092,3 +1124,127 @@ def exitNode(self, node, data):
...
+    # save them on the NLP for later, since generating them seems to take a while
+    if equality_constraints_only:
+        clist = nlp.clist = nlp.get_pyomo_equality_constraints()
+    else:
+        clist = nlp.clist = nlp.get_pyomo_constraints()
+
+    vlist = nlp.vlist = nlp.get_pyomo_variables()
+
+    if include_scaling_factors:
+        # Preallocaying arrays with NaNs to make it apparent if
```
MarcusHolly:
```suggestion
        # Preallocating arrays with NaNs to make it apparent if
```

idaes/core/scaling/util.py
```diff
@@ -1092,3 +1124,127 @@ def exitNode(self, node, data):
             f"An unhandled expression node type: {str(nodetype)} was encountered while "
             f"retrieving the nominal value of expression {str(node)}"
         )
+
+
+# Based off of John Eslick's functions in the old scaling tools
+def get_jacobian(
+    m,
+    include_scaling_factors=True,
+    include_ipopt_autoscaling=False,
```
MarcusHolly, line 1133: Is it problematic if a user sets both of these options to True? If so, an error should be raised in these cases

dallan-keylogic: No, it's not. Now that we've set the `scale_model` writer setting to `True`, the IPOPT autoscaling is an additional layer of scaling on top of the user-provided scaling factors. This option reflects that.

PR #1704, opened 2025-11-14
Block Triangularization fixes

dallan-keylogic:
## Summary/Motivation:
Until Pyomo/pyomo#3785 is addressed, `solve_strongly_connected_components` does not take scaling into account. However, the initialization postcheck does, so constraints may have too large of a scaled residual to past the postcheck. Here we fix that issue by performing a block solve after `solve_strongly_connected_components` returns.

Additionally, I've switched block triangularization from using the Pyomo `SolverFactory` to `get_solver` so it uses IDAES's default solver settings. Using `get_solver` had caused issues in the past, but hopefully scaling has addressed them.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/util/model_diagnostics.py
+++ b/idaes/core/util/model_diagnostics.py
@@ -112,11 +112,6 @@ from idaes.core.util.model_statistics import (
     large_residuals_set,
     variables_near_bounds_set,
 )
-from idaes.core.util.scaling import (
-    extreme_jacobian_columns,
-    extreme_jacobian_rows,
-    extreme_jacobian_entries,
-)
 from idaes.core.scaling.util import (
     get_jacobian,
     get_scaling_factor,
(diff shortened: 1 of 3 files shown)


agarciadiego (approved):

idaes/core/initialization/block_triangularization.py
```diff
@@ -168,6 +167,12 @@ def _solve_block_data(self, block_data, solver):
         """
         Call solve_strongly_connected_components on a given BlockData.
         """
+        if number_activated_constraints(block_data) == 0:
```
bpaul4, line 170: Why is this here? Should there be some kind of warning or error thrown?

dallan-keylogic: This is here because Pyomo throws an error when `solve` is called on a block with no constraints. We need to support initialization for blocks with no constraints because block triangularization is the default initializer for state blocks. Certain state blocks end up having no constraints (typically when `defined_state == True`). It's easier to put this check in the block triangularization rather than building out infrastructure for every property package to handle this circumstance.

PR #1713, opened 2025-12-02
Add Various Startup Types Support to the Price-taker Class

Xinhe-Chen:
## Fixes
This PR adds new features to the price-taker class.

## Summary/Motivation:
The unit commitment model in the current price-taker class only supports a single type of startup. However, some generators have different startup types (e.g., hot and cold start) because their physical components heat up and cool down over time, so the time since last shutdown affects how much fuel it costs, and how much wear-and-tear is incurred to bring them back online.

The startup model is based on Knueven, Bernard, James Ostrowski, and Jean-Paul Watson. "On mixed-integer programming formulations for the unit commitment problem." INFORMS Journal on Computing 32, no. 4 (2020): 857-876.

- Add the method for supporting the various startup types for the idaes price-taker class.
- Add tests for the method. 

### Legal Acknowledgement


idaes/apps/grid_integration/pricetaker/price_taker_model.py
```diff
@@ -871,6 +882,9 @@ def add_startup_shutdown(
         setattr(self, start_shut_blk_name, Block(self.set_days))
         start_shut_blk = getattr(self, start_shut_blk_name)
 
+        if not startup_transition_time:
+            self._multiple_startup_types = True
```
ruonanli61: Just wondering why `_multiple_startup_types` is introduced, and why it is assigned as `True` if `startup_transition_time` is not provided. Could you please add a brief comment to the code?

ruonanli61 (changes requested):

idaes/apps/grid_integration/pricetaker/price_taker_model.py
Xinhe-Chen: Thanks, self._multiple_startup_types is a variable I forgot to remove during the development. I have removed these two lines.

ruonanli61 (approved):

idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
```diff
@@ -395,6 +395,14 @@ def my_operation_model(m, design_blk):
         ),
     )
 
+    CONFIG.declare(
+        "startup_types",
+        ConfigValue(
+            domain=dict,
```
radhakrishnatg: Suggest using a more rigorous domain validation method. 

```python
def is_valid_startup_types(data):
    if not isinstance(data, dict):
        raise TypeError("Data must be a dict")

    if len(data) == 0:
        raise ConfigurationError("Received an empty dictionary for startup types")

    for key, value in data.items():
        if not isinstance(key, str):
            raise TypeError("key must be a str")

        if not isinstance(value, int):
            raise TypeError("value must be an int")

    # Values must be unique, as they correspond to different startup types
    if len(data.values()) > len(set(data.values())):
        raise ConfigurationError("Startup time for two or more startup types is the same.")

    # Return a dictionary after sorting based on values
    return dict(sorted(data.items(), key=lambda item: item[1]))
```

You can use this domain validator as:
```python
ConfigValue(domain=is_valid_startup_types, doc="...")
```
This way, you don't have to check for empty dictionaries in your main code.

idaes/apps/grid_integration/pricetaker/price_taker_model.py
```diff
@@ -156,6 +156,14 @@
     ),
 )
 
+CONFIG.declare(
+    "startup_types",
+    ConfigValue(
+        domain=dict,
```
radhakrishnatg: You can use the above domain validator: `is_valid_startup_types`

idaes/apps/grid_integration/pricetaker/unit_commitment.py
```diff
@@ -176,6 +182,68 @@ def minimum_down_time_con(_, t):
             <= install_unit - op_blocks[t].op_mode
         )
 
+    if not startup_transition_time:
+        # if there is only one startup type, return
+        # startup_transition_time can be None or empty dict
```
radhakrishnatg: With the domain validator, empty dictionaries are no longer valid inputs. You just need to check that it is not `None`. For improved code readability, I suggest: 
```python
if startup_transition_time is None:
    return
```

idaes/apps/grid_integration/pricetaker/unit_commitment.py
```diff
@@ -176,6 +182,68 @@ def minimum_down_time_con(_, t):
...
+        return
+
+    # multiple startup types
+    if startup_transition_time:
```
radhakrishnatg: You can remove this `if` condition and de-indent the code.

idaes/apps/grid_integration/pricetaker/unit_commitment.py
```diff
@@ -176,6 +182,68 @@ def minimum_down_time_con(_, t):
...
+        # there will be at least two types of startup
+        startup_names = list(startup_transition_time.keys())
+
+        # assume the first should be max(min_down_time, startup_transition_time["hot"])
+        startup_transition_time[startup_names[0]] = max(
+            minimum_down_time, startup_transition_time[startup_names[0]]
+        )
+
+        # add a check to ensure the startup time is monotonically increasing.
```
radhakrishnatg: Domain validator returns a sorted dictionary. So, this check can now be removed.

radhakrishnatg (changes requested):
Added a few comments to improve code readability and maintainability. We can discuss the comments in our next meeting.

idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
Xinhe-Chen: I have updated the codes based on your comments.

idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
```diff
@@ -85,6 +85,46 @@ def is_valid_polynomial_surrogate_data(data: dict):
     return new_data
 
 
+def is_valid_startup_types(data):
+    """Validate if the startup_types received is valid"""
+
+    if data is None:
```
radhakrishnatg: This check is not needed. If the argument value is `None`, then Pyomo's `ConfigDict` would not call the domain validator.

idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
```diff
@@ -416,6 +460,14 @@ def build(self):
                 doc="Binary: 1 if the shutdown is initiated, 0 otherwise",
             )
 
+        if self.config.startup_types:
+            # self.config.startup_types can be None or an empty dict
```
radhakrishnatg: Minor change: For better code readability, please change the condition to `if self.config.startup_types is not None:`. Also, please remove the comment, as it is no longer applicable (may create confusion for future maintainers).

idaes/apps/grid_integration/pricetaker/unit_commitment.py
```diff
@@ -176,6 +187,62 @@ def minimum_down_time_con(_, t):
...
 
+    if startup_transition_time is None or len(startup_transition_time) == 0:
+        # if there is only one startup type, return
+        # startup_transition_time can be None.
+        # This is a double insurance check, that empty dict or None will skip the following.
+        return
+
+    # multiple startup types
+    if startup_transition_time:
+        # there will be at least two types of startup
```
radhakrishnatg: As mentioned earlier, the `if` condition check is now unnecessary. You can remove the check, and remove the additional indentation.

idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
```diff
@@ -85,6 +85,42 @@ def is_valid_polynomial_surrogate_data(data: dict):
     return new_data
 
 
+def is_valid_startup_types(data):
+    """Validate if the startup_types received is valid"""
+
+    if not isinstance(data, dict):
+        raise TypeError("Data must be a dictionary.")
+
+    if len(data) == 0:
```
radhakrishnatg, line 94: Based on your tests, I'm wondering if we should change this to `if len(data) <= 1:`. This method must be used only when there is more than one startup type. If you have only startup type, then this feels unnecessary (Also, we will have to check the minimum uptime and startup time values to make sure they are consistent). Thoughts?

radhakrishnatg (changes requested):
The code looks good, but I still have a few minor changes.

idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
Xinhe-Chen: Done

idaes/apps/grid_integration/pricetaker/unit_commitment.py
Xinhe-Chen: Done

idaes/apps/grid_integration/pricetaker/design_and_operation_models.py
Xinhe-Chen: Yes, I think the suggestion makes sense to me. Eq. 54 in Ben's paper once confused me until I asked for clarification from him. If {"hot": 4, "warm": 8, "cold": 12}, when the downtime is between 4 to 8, it is a hot start; 8 to 12, warm start and 12 to infinity, a cold start. This is important to understand eq54, if only one startup_transition_time  is given, eq54 will break down. 
I will add this example to the comments.

idaes/apps/grid_integration/pricetaker/tests/test_design_and_operation_models.py
```diff
@@ -397,6 +458,15 @@ def test_is_valid_data_type_for_storage_model():
     )
 
 
+@pytest.mark.unit
```
radhakrishnatg: Please move this after line 72 (keep all the tests related to a function in one place for easy maintenance).

idaes/apps/grid_integration/pricetaker/price_taker_model.py
```diff
@@ -156,6 +157,14 @@
...
 )
 
+CONFIG.declare(
+    "startup_types",
+    ConfigValue(
+        domain=is_valid_startup_types,
+        doc="Dictionary of startup types and their transition times for the unit/process",
+    ),
+)
+
```
radhakrishnatg, line 167: Please add this line to line 416: `startup_transition_time=blk.config.startup_types`. This method exists for adding the startup and shutdown constraints automatically based on the inputs provided to the operation model.

radhakrishnatg (approved):
Everything looks good to me. I have two minor changes. Please make those changes if possible. Thanks @Xinhe-Chen !

Issue #1684, opened 2025-10-21
Bad stream table formatting for Flash with Modular Properties


dallan-keylogic:
Right now, if we try to generate a stream table for the `Flash` unit model with the modular properties, we get something like this:

```
                                     Units     Inlet  Vapor Outlet  Liquid Outlet
Total Molar Flowrate         mole / second       1.0           NaN            NaN
Total Mole Fraction benzene  dimensionless       0.5           NaN            NaN
Total Mole Fraction toluene  dimensionless       0.5           NaN            NaN
Temperature                         kelvin     368.0           NaN            NaN
Pressure                            pascal  101325.0           NaN            NaN
flow_mol                     mole / second       NaN          50.0           50.0
mole_frac_comp benzene       dimensionless       NaN           0.5            0.5
mole_frac_comp toluene       dimensionless       NaN           0.5            0.5
temperature                         kelvin       NaN         300.0          300.0
pressure                            pascal       NaN      100000.0       100000.0
```

The root problem appears to be in [how the stream table is being generated](https://github.com/IDAES/idaes-pse/blob/ce7f3bd8eeb0883aad9d8d38e98f8a431c643e4a/idaes/models/unit_models/flash.py#L340):

```
        if self.config.ideal_separation:
            # If using ideal separation, get values from Ports and hope they map
            # to names in Inlet
            # TODO: Add a better way to map these if necessary
```
The values on the port are not mapped to names in `Inlet`, so they're getting values of `NaN` instead.

PR #1719, opened 2025-12-23
Fix issue 1684: bad stream table formatting for flash

Lingyan90:
## Fixes
Fixed a issue that has bad stream table formatting for Flash with Modular Properties (https://github.com/IDAES/idaes-pse/issues/1684)

- Updated the table formatting in `flash.py` under ideal separation scenario
- Uncommented the existing test for the related stream table formatting in `test_flash.py`

### Legal Acknowledgement


idaes/models/unit_models/flash.py
```diff
@@ -409,5 +385,27 @@ def _get_stream_table_contents(self, time_point=0):
                             quant = report_quantity(port_obj.vars[k][time_point, i[1:]])
                             stream_attributes[n][k + " " + kname] = quant.m
                             stream_attributes["Units"][k + " " + kname] = quant.u
+        else:
+            # If not using idea separation, we can get outlet state directly
```
bpaul4:
```suggestion
            # If not using ideal separation, we can get outlet state directly
```

idaes/models/unit_models/tests/test_flash.py
```diff
@@ -406,49 +405,48 @@ def test_get_performance_contents(self, bt_modular):
...
+                "mole_frac_comp toluene": getattr(
+                    pyunits.pint_registry, "dimensionless"
+                ),
+                "temperature": getattr(pyunits.pint_registry, "K"),
+                "pressure": getattr(pyunits.pint_registry, "Pa"),
+            },
+            "Inlet": {
+                "flow_mol": pytest.approx(
+                    1.00, rel=1e-4
+                ),  # This is different from it's property package's (BT_idea) properties_out value of 100
```
bpaul4:
```suggestion
                ),  # This is different from its property package's (BT_ideal) properties_out value of 100
```

bpaul4 (approved):
Just a couple of typos, otherwise looks good.

idaes/models/unit_models/tests/test_flash.py
Lingyan90: updated all the typos in new push

idaes/models/unit_models/tests/test_flash.py
```diff
@@ -406,49 +405,48 @@ def test_get_performance_contents(self, bt_modular):
...
+                "mole_frac_comp toluene": getattr(
+                    pyunits.pint_registry, "dimensionless"
+                ),
+                "temperature": getattr(pyunits.pint_registry, "K"),
+                "pressure": getattr(pyunits.pint_registry, "Pa"),
+            },
+            "Inlet": {
+                "flow_mol": pytest.approx(
+                    1.00, rel=1e-4
+                ),  # This is different from its property package's (BT_ideal) properties_out value of 100
```
dallan-keylogic, line 428: Why is this value different?

Lingyan90: This PR aims to fix the table formatting issue when using ideal property packages. As noted in the issue description (https://github.com/IDAES/idaes-pse/issues/1684), the input and output tables currently have different row names when an ideal property package is used. This PR allows `flash.py` to generate different table formats depending on the chosen property package and adds additional tests in `test_flash.py` to verify the expected table format under different property package configurations.

The lines you commented on (426–428) are part of a test that verifies the table format when using an ideal property package, where the input and output rows have different names. This test uses `bt_modular` within the same class (TestBTIdealModular). In `bt_modular`, the inlet `flow_mol` is fixed at `1`, which differs from the configuration used in the `BTIdeal` property package. I also noticed that the outputs report `flow_mol` = 50 for both the liquid and vapor streams, which sum to 100, exceeding the inlet value of `1`. For this reason, I added a comment noting that the inlet `flow_mol` differs from the configuration used in the property package.

Changing the fixed value of `flow_mol` in `bt_modular` from `1` to `100` causes failures in other parts of the existing test suite. Since the purpose of this PR is to address the table formatting issue—and this particular test checks formatting without running the full flash model—the mismatch between inlet and outlet `flow_mol` values falls outside the scope of this PR.

dallan-keylogic (approved):
LGTM

PR #1730, opened 2026-02-04
Pipe Unit Model

dallan-keylogic:
## Summary/Motivation:
This is a unit model for a pipe without reaction. Although piping can typically be ignored in steady-state modeling, it is required for dynamics when liquid holdup is important.

### Legal Acknowledgement

### DIFF
--- a/idaes/models/unit_models/pipe.py
+++ b/idaes/models/unit_models/pipe.py
@@ -11,8 +11,9 @@
 # for full copyright and license information.
 #################################################################################
 """
-Standard IDAES Pipe model.
+Model of an empty pipe, derived from the PFR model by removing reaction terms.
 """
+
 # Import Pyomo libraries
 from pyomo.environ import ComponentMap, Constraint, Reference, Var
 from pyomo.common.config import ConfigBlock, ConfigValue, In, ListOf, Bool
@@ -31,7 +32,7 @@ from idaes.core.util.config import is_physical_parameter_block
 from idaes.core.util.misc import add_object_reference
 from idaes.core.scaling import CustomScalerBase, DefaultScalingRecommendation
 
-__author__ = "Andrew Lee, John Eslick"
+__author__ = "Douglas Allan, Andrew Lee, John Eslick"
 
 
 class PipeScaler(CustomScalerBase):
--- a/idaes/models/unit_models/pipe.py
+++ b/idaes/models/unit_models/pipe.py
@@ -27,10 +27,7 @@ from idaes.core import (
     UnitModelBlockData,
     useDefault,
 )
-from idaes.core.util.config import (
-    is_physical_parameter_block,
-    is_reaction_parameter_block,
-)
+from idaes.core.util.config import is_physical_parameter_block
 from idaes.core.util.misc import add_object_reference
 from idaes.core.scaling import CustomScalerBase, DefaultScalingRecommendation
 
--- a/idaes/models/unit_models/pipe.py
+++ b/idaes/models/unit_models/pipe.py
@@ -1,0 +1,361 @@
+#################################################################################
+# The Institute for the Design of Advanced Energy Systems Integrated Platform
+# Framework (IDAES IP) was produced under the DOE Institute for the
+# Design of Advanced Energy Systems (IDAES).
+#
+# Copyright (c) 2018-2026 by the software owners: The Regents of the
+# University of California, through Lawrence Berkeley National Laboratory,
+# National Technology & Engineering Solutions of Sandia, LLC, Carnegie Mellon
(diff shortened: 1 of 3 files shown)


idaes/models/unit_models/tests/test_pipe.py
```diff
@@ -0,0 +1,358 @@
...
+
+    assert m.fs.unit.config.material_balance_type == MaterialBalanceType.useDefault
+    assert m.fs.unit.config.energy_balance_type == EnergyBalanceType.useDefault
+    assert m.fs.unit.config.momentum_balance_type == MomentumBalanceType.pressureTotal
+    assert not m.fs.unit.config.has_heat_transfer
+    assert not m.fs.unit.config.has_pressure_change
+    assert not m.fs.unit.config.has_phase_equilibrium
+    assert m.fs.unit.config.property_package is m.fs.properties
+
+    # Make sure that we got rid of all the PFR options
```
bpaul4, line 85: It doesn't look like the PFR model is imported anywhere, why are we mentioning it here?

idaes/models/unit_models/tests/test_pipe.py
```diff
@@ -0,0 +1,358 @@
...
+        m.fs.unit.inlet.conc_mol_comp[0, "Ethanol"].set_value(0.0)
+
+        m.fs.unit.inlet.temperature[0].set_value(303.15)
+        m.fs.unit.inlet.pressure[0].set_value(101325.0)
+
+        m.fs.unit.length.fix(0.5)
+        m.fs.unit.area.fix(0.1)
+
+        m.fs.unit.heat_duty.fix(0)
+        m.fs.unit.deltaP.fix(0)
```
bpaul4, line 300: Since we are specifying the pipe length and cross-sectional area, are we planning on calculating the pressure drop from the pipe flow equation?

idaes/models/unit_models/pipe.py
```diff
@@ -0,0 +1,358 @@
...
+    EnergyBalanceType,
+    MomentumBalanceType,
+    UnitModelBlockData,
+    useDefault,
+)
+from idaes.core.util.config import is_physical_parameter_block
+from idaes.core.util.misc import add_object_reference
+from idaes.core.scaling import CustomScalerBase, DefaultScalingRecommendation
+
+__author__ = "Andrew Lee, John Eslick"
```
MarcusHolly:
```suggestion
__author__ = "Douglas Allan"
```
Unless you've just resurrected some older code

idaes/models/unit_models/pipe.py
```diff
@@ -0,0 +1,358 @@
...
+        self.control_volume.add_momentum_balances(
+            balance_type=self.config.momentum_balance_type,
+            has_pressure_change=self.config.has_pressure_change,
+        )
+
+        self.control_volume.apply_transformation()
+
+        # Add Ports
+        self.add_inlet_port()
+        self.add_outlet_port()
```
MarcusHolly, line 329: Do we want to consider having multiple inlets/outlets where the user has to specify an inlet list and an outlet list (like the mixer)?

idaes/models/unit_models/tests/test_pipe.py
dallan-keylogic: This unit model was created by taking the PFR model and deleting all the chemical reactions.

idaes/models/unit_models/tests/test_pipe.py
dallan-keylogic: The user can supply their own correlation if they want to. The goal here is to represent the material holdup in the piping, which introduces a time delay between when a change occurs in one process and when that change is registered in a downstream process.

idaes/models/unit_models/pipe.py
dallan-keylogic: We're assuming that there is no interactions between streams.

idaes/models/unit_models/pipe.py
dallan-keylogic: It is largely copy-and-pasted code from the PFR. I'm not sure exactly how we should handle authorship in that case.

bpaul4: I'd say updating the author list per Marcus's suggestion is appropriate here, as it's a new unit model.

idaes/models/unit_models/tests/test_pipe.py
bpaul4: Maybe just clean up any references, since it's adapted from that model but doesn't actually leverage it.

MarcusHolly (approved):
LGTM - I'll let you decide how to handle the author list

bpaul4 (approved):

PR #1734, opened 2026-02-20
Utility Functions for PrOMMiS

dallan-keylogic:
## Summary/Motivation:
I ended up creating a handful of utility functions for work in PrOMMiS that fit better in IDAES. This PR brings them in, as well as adds testing.

- `scale_time_discretization_equations` is a function for scaling `DerivativeVars` and discretization equations en masse. This function existed for the old scaling tools, but this version is updated to not use `constraint_scaling_transform`.
- `copy_values_from_point` is a generalization of a function I was using to copy the steady-state values from t=0 forward to other time points.
- `assert_solution_equivalent` is a way to compare the solution of a model to a cached version. We would typically compare values variable-by-variable. However, that means we only see the values that changed one at a time. In order to see what else has changed, we have to run the test again or look forward in the tests checking value by value manually. This function compares all values, generates a report about which variables have changed more than their tolerances, then raises an `AssertionError` at the end if there are variables who changed outside their tolerances.

### Legal Acknowledgement


idaes/core/util/tests/test_testing.py
```diff
@@ -0,0 +1,587 @@
...
+# University of California, through Lawrence Berkeley National Laboratory,
+# National Technology & Engineering Solutions of Sandia, LLC, Carnegie Mellon
+# University, West Virginia University Research Corporation, et al.
+# All rights reserved.  Please see the files COPYRIGHT.md and LICENSE.md
+# for full copyright and license information.
+#################################################################################
+"""
+Who will test the testers?
+
+This file was created with the assistance of AI
```
MarcusHolly: Is this an acknowledgement that has to be made? If so, perhaps it would also make sense to specify which AI and/or specify which parts of the code were AI-assisted (unless most of the file was made by AI)

idaes/core/util/tests/test_testing.py
```diff
@@ -0,0 +1,587 @@
...
+
+class TestAssertSolutionEquivalent:
+    @pytest.fixture
+    def model(self):
+        """Creates a standard Pyomo model for testing."""
+        m = ConcreteModel()
+        m.i = Set(initialize=["A", "B", "C"])
+
+        # Indexed Variable
+        m.x = Var(m.i, initialize={"A": 1.0, "B": 2.5, "C": 3.0})
```
MarcusHolly, line 62: Imo, the most common use case for the new function is to check the values of a flowsheet, so we should have a test that reflects a conventional flowsheet setup (m.fs. = FlowsheetBlock()). I think the current tests are fine as they are, so perhaps just consider adding a test that creates a small flowsheet (one unit) and do some checks on it

idaes/core/util/tests/test_testing.py
dallan-keylogic: Until such time as we decide on an official policy, this lets us keep track of what code is and isn't generated using AI.

Most of the code here was generated by AI. I went through and did minor edits to extend coverage.

idaes/core/scaling/tests/test_util.py
```diff
@@ -2591,3 +2605,139 @@ def test_condition_number_none(self, model):
...
+            behave differently, this conditional might have to be updated
+
+    Returns:
+        None
+    """
+
+    def approx(expected):
+        return pytest.approx(expected, rel=1e-10)
+
+    m, _, _, _, _, _, _ = dae_with_non_time_indexed_constraint(
```
bpaul4, line 2628: Would the six expected values be used anywhere other than the test file? If not, we can simplify in the method `return` so we don't need all the underscores every time we invoke this method.

idaes/core/scaling/tests/test_util.py
```diff
@@ -2591,3 +2605,139 @@ def test_condition_number_none(self, model):
...
+            except KeyError:
+                assert get_scaling_factor(m.y_t_cont_eq[t, 1]) == approx(1)
+                assert get_scaling_factor(m.y_t_cont_eq[t, 2]) == approx(10)
+                assert get_scaling_factor(m.y_t_cont_eq[t, 3]) == approx(0.1)
+                assert get_scaling_factor(m.y_t_cont_eq[t, 4]) == approx(1000)
+                assert get_scaling_factor(m.y_t_cont_eq[t, 5]) == approx(13)
+                assert get_scaling_factor(m.y_t_cont_eq[t, 6]) == approx(1)
+
+
+@pytest.mark.unit
```
bpaul4, line 2685: Are these just smoke tests? Are there results or values we can check?

idaes/core/scaling/util.py
```diff
@@ -1247,3 +1251,102 @@ def jacobian_cond(m=None, scaled=True, jac=None):
...
+        time_set: Time set object. For an IDAES flowsheet object fs, this is fs.time.
+        time_scaling_factor: Scaling factor to use for time
+
+    Returns:
+        None
+    """
+
+    tname = time_set.local_name
+
+    # Copy and pasted from solvers.petsc.find_discretization_equations then modified
```
bpaul4: Do we need this comment? or maybe say "adapted from" instead of "copy and pasted, modified"

MarcusHolly (approved):

idaes/core/util/tests/test_testing.py
```diff
@@ -0,0 +1,725 @@
...
+        raises an AssertionError with a proper message.
+        """
+        expected_results = {
+            "x": {
+                "A": (1.0, 1e-6, 1e-8),
+                "B": (2.5, 1e-6, 1e-8),
+                "C": (3.1, 1e-6, 1e-8),  # Incorrect value
+            },
+        }
+        with pytest.raises(AssertionError, match=r"Found 1 mismatch\(es\)"):
```
bpaul4: Should this also indicate that "C" is the mismatch?

bpaul4 (commented):
I have a few comments.

idaes/core/scaling/tests/test_util.py
dallan-keylogic: The value testing occurs in `discretization_tester`

idaes/core/scaling/tests/test_util.py
dallan-keylogic: The returned values are used in the PETSc test from which this function is being imported.

bpaul4 (approved):

PR #1739, opened 2026-02-27
Add structfs action model classes

### LINKED ISSUE #1738: Add type to model_variables action for structfs flowsheet runner
This is now being done in #1742 (@dangunter, please correct if that's not right)

dangunter:
## Fixes
 
Fixes #1738 

-
-

### Legal Acknowledgement


CopyDemon (approved):

idaes/core/util/structfs/model_vars_schema.py
```diff
@@ -0,0 +1,98 @@
...
+    # [index, value]
+    v: list[tuple[IndexValue, JsonScalar]]
+
+
+class VariableLeaf(BaseModel):
+    """Leaf node for variable values."""
+
+    t: Literal["V"]
+    # [index, value, fixed, stale, lower_bound, upper_bound]
+    v: list[tuple[IndexValue, JsonScalar, bool, bool, JsonScalar, JsonScalar]]
```
dallan-keylogic, line 74: You probably want to track the Pyomo variable domain as well. That can have unexpected interactions when trying to change variable bounds. For example, if a variable domain is `NonnegativeReals`, if you try to set a lower bound below zero, it will instead get set to zero.

Maybe we could track variable scaling factor? That's a bit tricky, though, because the variable scaling factor doesn't live on the `Var` object, but on a suffix attached to any block in the variable's parent blocks.

idaes/core/util/structfs/runner_actions.py
```diff
@@ -433,27 +447,25 @@ def _extract_vars(self, m):
                 subtype = self.PARAM_TYPE
             else:
                 continue  # ignore other components
+            if self._dbg:
+                self.log.debug(f"_extract_vars: component {c} ({subtype})")
```
dallan-keylogic, line 451: I assume the logging level for the flowsheet runner can be set independent of the logging level for, e.g., unit initialization.

dallan-keylogic (commented):
One minor suggestion and one question.

dangunter:
incorporated into #1742

PR #1742, opened 2026-03-03
Enhanced reporting for structured flowsheet wrapper

### LINKED ISSUE #1741: Add diagnostics 'action' to structfs flowsheet wrapper
The 'structured flowsheet' wrapper includes 'actions' that can automatically get information from flowsheets during and after their runs, and return them as Pydantic/JSON objects for other tools to easily consume.

One action that would be useful is to fetch some set of information provided by the diagnostic toolkit.

dangunter:
## Fixes
 
Fixes #1741 

## Summary/Motivation:
The Prommis UI needs information from the model.
The general strategy is to generate this information with FlowsheetRunner 'actions' and report them back as Pydantic objects.

- Add diagnostics (interface with Andrew's refactor) FlowsheetRunner action
- Add model variables FlowsheetRunner action
- Add Mermaid diagram FlowsheetRunner action
- Improved interface to the FlowsheetRunner in general

### Legal Acknowledgement


idaes/core/util/structfs/actions/diagnostics.py
```diff
@@ -0,0 +1,263 @@
...
+        elif "unused variable" in msg_lower or "unused variables" in msg_lower:
+            category = "unused_variables"
+
+        return StructuralIssueData(
+            kind=kind,
+            category=category,
+            count=count,
+            message=body,
+            details=lines[1:],
+        )
```
dallan-keylogic: It might be better, in the long run, for the Diagnostics Toolbox to return these data objects directly and then generate a report from that rather than writing a report to a stream and parsing it after the fact.

idaes/core/util/structfs/actions/diagnostics.py
```diff
@@ -0,0 +1,263 @@
...
+                continue
+            if line == "Suggested next steps:":
+                section = "next_steps"
+                continue
+            if line in (
+                "No warnings found!",
+                "No cautions found!",
+                "If you still have issues converging your model consider:",
+            ):
+                continue
```
dallan-keylogic: So does this section not get shown to the user at all?

dallan-keylogic (commented):
We should think about how the Diagnostics Toolbox returns data. In this PR, the text stream from the Diagnostics Toolbox is parsed in order to make the data available through the GUI, but it might make sense for the DiagnosticsToolbox to return `StructuralData` and `NumericalData` classes, and then we can either write the report or make the data available in the GUI from that.

@andrewlee94 told me that he was going to open a couple of PRs breaking the `diagnostics_toolbox.py` file into smaller subfiles and making them compatible with the Pyomo GreyBox interface, so if we elect to change what the `DiagnosticsToolbox` outputs, then we should wait into those PRs are merged in.

idaes/core/util/structfs/actions/diagnostics.py
dangunter: I agree. In that case it would make more sense to have a separate PR to make those changes, then this PR becomes much simpler -- essentially selecting which tests to run and returning the objects. @dallan-keylogic do you have any time to help with this? I need a partner in crime to at least review changes

idaes/core/util/structfs/actions/diagnostics.py
dangunter: yeah. the idea was it wasn't really that informative

ksbeattie:
@dangunter is this, now that #1747 has been merged, still blocked?

idaes/core/util/structfs/actions/diagnostics.py
dangunter: this file is removed now

dangunter:
@dallan-keylogic The parsing of diagnostics is gone, and now things call what used to be private but are now public methods in the refactored diagnostics code -- to create Pydantic objects for the UI

ksbeattie:
@dangunter is this still blocked?

dangunter:
This got merged into #1772

PR #1745, opened 2026-03-07
Diagnostics output objects

### LINKED ISSUE #1744: Natively produce data structures from diagnostics toolkit
As noted in discussion for #1742 , it would make sense for the diagnostics toolkit to produce output first as objects, then as formatted text -- as opposed to only the latter. This would make it much easier for downstream consumers that are not consoles to use the output.

dangunter:
Fixes #1744 

- While preserving existing output..
- Provide Pydantic model objects for all operations
- Add a 'history' option that can record past input/result pairs so the iterative process can be better tracked

### Legal Acknowledgement


dangunter:
putting this on hold

PR #1751, opened 2026-03-25
Update power plant O&M costing calculation

shermanjasonaf:
## Summary/Motivation:

Update a TPC calculation in the power plant operation and maintenance (O&M) costing library, based on discussions with @aostace01 and @dallan-keylogic 

- Update power plant O&M costing calculation

### Legal Acknowledgement


bpaul4:
Hi @shermanjasonaf, @aostace01, @dallan-keylogic, could you please share a bit more detail on why this change was implemented? Fixing the total TPC variable and then invoking it in the costing equations should be sufficient for LCOE optimization.

aostace01:
Hi @bpaul4 , happy to share more. Fixing the TPC removes it from the active variable set and that causes problems in an optimization context. We ran into this in practice when minimizing LCOE with absorber/stripper sizes, solvent flowrate, and HX area as free variables: the optimization converged without error, but the results didn't make sense until we traced it back to `total_TPC` being frozen at its pre-optimization value. The optimizer was effectively blind to how equipment sizing decisions affected cost. Replacing `.fix()` with an equality constraint keeps `total_TPC` active, so the cost objective correctly responds to design decisions. Happy to discuss further!

bpaul4:
If `fixed_TPC` is not changing, wouldn't we expect `total_TPC` to remain constant even while optimizing? Also, if we're _not_ using `fixed_TPC` and are using the TPC calculations based on `UnitModelCostingBlock` constraints, then the `else` stuff should never happen.

dallan-keylogic:
After further discussion with @aostace01 , we decided this PR is unnecessary right now.

Issue #1755, opened 2026-03-31
Add parameter-sweep interface to flowsheet wrapper


dangunter:
Add the ability to perform a parameter sweep to the capabilities of the flowsheet wrapper.
API tbd, but something like:
```
from wrapped_flowsheet_module import FS  # FlowsheetWrapper instance

def do_sweep(sweep_parameters):
    FS.configure_sweep(sweep_parameters)
    FS.run_sweep()
```
Of course, this is sweeping (cough) the complexities under the rug of `sweep_parameters`.
But the basic idea is to invoke the existing machinery to actually run the sweep, using this interface to simplify the configuration and normalize the results to something the UI would want.

dangunter:
Not going to happen, but if it did, it would happen in https://github.com/prommis/flowsheet-inspector-lib

PR #1760, opened 2026-04-04
Json list db

dangunter:
## Fixes

Related to #1755 

## Summary/Motivation:

List of JSON documents, with accompanying index with some metadata and offset of each document in the file.

This is motivated by the need to keep a history of report objects, or the results of a parameter sweep, or both, from the `idaes.core.util.structfs.FlowsheetRunner`. These may be rather large, so the implementation keeps the offsets of each JSON document in a separate file, allowing O(1) access to a given document without parsing or reading any other document's data.

- new jsonlist module in idaes.core.util
- `JsonList` class

### Legal Acknowledgement


dangunter:
@CopyDemon if you get a chance could you look at this and approve? It's an added utility not used inside core, at the moment, so should "do no harm" as long as it builds (which you can see it does).

CopyDemon (approved):
I think this is good to merge—it doesn’t affect the core and all tests are passing.

dangunter:
Not useful right now

Issue #1686, opened 2025-10-22
Getting default scaler and default initializer broken for scaler state blocks


dallan-keylogic:
When trying to test the scaler object for the Helmholtz properties, I encountered the following error:

```
    @pytest.mark.unit
    def test_scaler_object_mole_basis():
        m = pyo.ConcreteModel()
        m.fs = FlowsheetBlock()

        m.fs.prop_water = HelmholtzParameterBlock(
            pure_component="h2o", amount_basis=AmountBasis.MOLE
        )

        m.fs.sb = m.fs.prop_water.build_state_block()

>       assert m.fs.sb.default_scaler is hstate.HelmholtzEoSScaler

tests\test_stateblock.py:52:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
..\..\..\core\base\property_base.py:602: in default_scaler
    return self.parent_component().default_scaler
..\..\..\core\base\property_base.py:602: in default_scaler
    return self.parent_component().default_scaler
E   RecursionError: maximum recursion depth exceeded while calling a Python object
!!! Recursion detected (same locals & position)
```
The problem here is that `m.fs.sb.parent_component() is m.fs.sb` because it is a scalar block. We have been attaching `default_scaler` and `default_initializer` attributes to the `StateBlock` instead of the `StateBlockData` children:
```
class _StateBlock(StateBlock):
    """
    This class contains methods which should be applied to Property Blocks as a
    whole, rather than individual elements of indexed Property Blocks.
    """

    # Set default initializer
    default_initializer = HelmholtzEoSInitializer
    default_scaler = HelmholtzEoSScaler
```
but evidentially that isn't sufficient.

PR #1766, opened 2026-04-13
Default scaler objects for properties and reactions

dallan-keylogic:
## Fixes
- #1686
- Item 5 of #1648

## Summary/Motivation:
When using the old scaling tools, we could set default scaling factors to use for all state blocks using a property package. With the new scaling tools, heretofore we have been forced to use changing the class attribute `DEFAULT_SCALING_FACTORS`. Global mutation is dangerous and ought be avoided when possible.

In this PR, I allow the user to specify a default scaler object to use for all instances of a property or reaction package in a flowsheet by setting the `default_state_scaler_object` or `default_reaction_scaler_object` on their respective parameter blocks.

Additionally, I fixed the issue with the default scaler object for scaler state and reaction blocks.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/base/property_base.py
+++ b/idaes/core/base/property_base.py
@@ -110,6 +110,7 @@ class PhysicalParameterBlock(ProcessBlockData, property_meta.HasPropertyClassMet
 
         # By default, property packages do not include inherent reactions
         self._has_inherent_reactions = False
+        self._default_state_scaler = None
 
     @property
     def state_block_class(self):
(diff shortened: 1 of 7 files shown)


MarcusHolly (approved):
LGTM

idaes/core/base/property_base.py
```diff
@@ -110,12 +110,16 @@ def build(self):
...
     def state_block_class(self):
         if self._state_block_class is not None:
             return self._state_block_class
         else:
+            # Python catches the error and raises its own AttributeError
+            # with its own message. Leave the error, though, in case
+            # the user looks through the code and ends up here.
             raise AttributeError(
                 "{} has not assigned a StateBlock class to be associated "
                 "with this property package. Please contact the developer of "
```
bpaul4: Does that mean we know that this will never happen? In that case we should preempt the default error with our own, or remove the error since it is never happening anyways. Same with other ones in this PR, and I think that will help you on code coverage as well.

idaes/core/scaling/tests/test_custom_scaler_base.py
```diff
@@ -150,6 +214,51 @@ def _does_not_create_component(self):
...
+        self.a = Var(initialize=1)
+
+    def _recursion1(self):
+        self.recursive_cons1 = Constraint(expr=self.recursion2 == 1)
+
+    def _recursion2(self):
+        self.recursive_cons2 = Constraint(expr=self.recursion1 == 1)
+
+    def _raise_exception(self):
+        # PYLINT-TODO
```
bpaul4: What needs to be done here?

idaes/core/base/property_base.py
dallan-keylogic: No. What happens is that, if this error is triggered, either Python or the create-on-demand property framework will intercept this error in a `try` block, then raise its own `AttributeError`. I thought about removing the error message, but thought it still might be useful for users tracking down why they're getting an error.

idaes/core/scaling/tests/test_custom_scaler_base.py
dallan-keylogic: I'm not sure---I copied that class from the tests for `ReactionBase`, which also has that TODO. I think I can strip out a significant amount of those methods for the test.

PR #1768, opened 2026-04-15
Update get-extensions and setup-environment logic for new major release of extensions

### LINKED ISSUE #1248: Update IPOPT version to 3.14
We should try to update the version of IPOPT we distribute and use as one of the first things after the August release. This will give us three months to work out any kinks that come up during the change over.

mrmundt:
## Summary/Motivation:

Who wants new extension versions?

### WE DO!

This PR implements the changes necessary to support both old and new versions of IDAES extensions in a lovely, compatible way. There are probably a plethora of opportunities for debate on exact implementations, so go forth and argue.

That being said, this does currently work for both <4 and >=4 releases. **NOTE** that I know there are some failing tests, and I am going to need some guidance/help on how to fix them. I wanted this open for the visibility of those and to enable discussion.

- Update supported OS' map
- Make failure messages a bit stronger
- No longer assume that all of the executables and libraries are in `bin_directory`; be more layout/release aware
- Add in functionality to parse old and new license/version info files
- Add in functionality to setup the environment correctly on all OS' for both old and new layout structures
- Warn if someone is trying to install >= 4 release stuff but they still have < 4 items in their IDAES path
- Make petsc-finding logic compatible for both old and new layouts
- Tests tests tests

### Legal Acknowledgement


ksbeattie:
@mrmundt is this (and the idaes-ext release) still looking possible for a May (end of the month) release?

ksbeattie:
Waiting on funding to arrive.

PR #1774, opened 2026-04-20
Solvent reboiler initializer object

dallan-keylogic:
## Summary/Motivation:
I want to use the `SolventReboiler` as an example in a scaling tutorial. I also want to use the new initialization API in this tutorial to help acclimate users to it, so I created a new initializer based on the old method.

- Initializer object for `SolventReboiler`
- Fixed bug in which constraint residual tolerance was getting set to infinity for state variables that could solve for VLE within the state block. (The tolerance remains at infinity for `FpcTP` and `FpTPxpc`, in which VLE cannot be solved within the state block alone---it needs to be part of a control volume).

### Legal Acknowledgement

### DIFF
--- a/idaes/models/properties/modular_properties/base/generic_property.py
+++ b/idaes/models/properties/modular_properties/base/generic_property.py
@@ -2041,7 +2041,7 @@ class ModularPropertiesInitializer(InitializerBase):
             self.config.constraint_tolerance = float("inf")
 
         if (
-            "flow_mol_phase_comp"
+            "flow_mol_phase_comp" in scaler_block.define_state_vars()
             or "mole_frac_phase_comp" in scaler_block.define_state_vars()
         ):
             init_log.debug(
--- a/idaes/models/properties/modular_properties/base/generic_property.py
+++ b/idaes/models/properties/modular_properties/base/generic_property.py
@@ -2038,6 +2038,8 @@ class ModularPropertiesInitializer(InitializerBase):
                 "initialization method may be needed."
                 "Ignoring constraints with large residuals."
             )
+            # TODO can we figure out how skip these constraints without
+            # mutating the initializer object?
             self.config.constraint_tolerance = float("inf")
 
         if (
@@ -2055,6 +2057,8 @@ class ModularPropertiesInitializer(InitializerBase):
                 "initialization method may be needed. "
                 "Ignoring constraints with large residuals."
             )
+            # TODO can we figure out how skip these constraints without
+            # mutating the initializer object?
             self.config.constraint_tolerance = float("inf")
 
         init_log.info("Starting initialization routine")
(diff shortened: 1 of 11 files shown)


dallan-keylogic:
I'm going to ignore the CodeCov issue. The untested part of the new initializer wasn't tested in the old initializer either.

idaes/models_extra/column_models/tests/test_solvent_reboiler.py
```diff
@@ -45,223 +45,139 @@
...
+        prefix + "temperature": 393.8,
+        prefix + "flow_mol": 74.33,
+    }
+    mole_frac_guesses = {"CO2": 0.0285, "H2O": 0.8491, "MEA": 0.1224}
+    for j, val in mole_frac_guesses.items():
+        liquid_guess[prefix + f"mole_frac_comp[{j}]"] = val
+
+    import idaes.logger as idaeslog
+
+    initializer_object = model.fs.unit.default_initializer(output_level=idaeslog.DEBUG)
```
dallan-keylogic, line 80:
```suggestion
    initializer_object = model.fs.unit.default_initializer()
```

idaes/models/properties/modular_properties/base/generic_property.py
```diff
@@ -2038,10 +2038,12 @@ def initialization_routine(
                 "initialization method may be needed."
                 "Ignoring constraints with large residuals."
             )
+            # TODO can we figure out how skip these constraints without
```
bpaul4, line 2041: Maybe temporarily deactivate them?

idaes/models_extra/column_models/tests/test_solvent_reboiler.py
```diff
@@ -45,223 +45,139 @@
...
+    liquid_guess = {
+        prefix + "pressure": 183700,
+        prefix + "temperature": 393.8,
+        prefix + "flow_mol": 74.33,
+    }
+    mole_frac_guesses = {"CO2": 0.0285, "H2O": 0.8491, "MEA": 0.1224}
+    for j, val in mole_frac_guesses.items():
+        liquid_guess[prefix + f"mole_frac_comp[{j}]"] = val
+
+    import idaes.logger as idaeslog
```
bpaul4: This is not used anywhere, since you're removing the argument below.

MarcusHolly (approved):
LGTM -  I think the tutorial will be very helpful

idaes/models/properties/modular_properties/base/generic_property.py
dallan-keylogic: `InitializerBase.postcheck` is called after variables and constraints are reverted, so that won't work. The answer is probably to add an attribute to `InitializerBase` containing a set of constraints whose large residuals should be ignored, but that's well beyond the scope of this PR.

bpaul4 (approved):

PR #1777, opened 2026-04-22
Fix negative scaling factor bug

dallan-keylogic:
## Summary/Motivation:
While creating the scaling tutorial notebook I discovered two things
  1. `CustomScalerBase.get_expression_nominal_value` could return negative numbers and
  2. `ScalerBase._set_scaling_factor` did not flag negative numbers as erroneous. Instead, it replaced them with the minimum scaling factor(`1e-10` by default).

- Add input validation to `ScalerBase._set_scaling_factor`
- Repair places where `CustomScalerBase.get_expression_nominal_value` is used to account for the fact it may be negative.

### Legal Acknowledgement

3. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

### DIFF
--- a/idaes/core/base/control_volume_base.py
+++ b/idaes/core/base/control_volume_base.py
@@ -208,8 +208,10 @@ class ControlVolumeScalerBase(CustomScalerBase):
                 prop_idx = idx[:-2]
                 p = idx[-2]
                 j = idx[-1]
-                nom = self.get_expression_nominal_value(
-                    props[prop_idx].get_material_flow_terms(p, j)
+                nom = abs(
+                    self.get_expression_nominal_value(
+                        props[prop_idx].get_material_flow_terms(p, j)
+                    )
                 )
                 self.set_component_scaling_factor(
                     rate_rxn_gen[idx], weight / nom, overwrite=overwrite
@@ -257,8 +259,10 @@ class ControlVolumeScalerBase(CustomScalerBase):
                 prop_idx = idx[:-2]
                 p = idx[-2]
                 j = idx[-1]
-                nom = self.get_expression_nominal_value(
-                    props[prop_idx].get_material_flow_terms(p, j)
+                nom = abs(
+                    self.get_expression_nominal_value(
+                        props[prop_idx].get_material_flow_terms(p, j)
+                    )
                 )
                 self.set_component_scaling_factor(
                     equil_rxn_gen[idx], weight / nom, overwrite=overwrite
@@ -306,8 +310,10 @@ class ControlVolumeScalerBase(CustomScalerBase):
                 prop_idx = idx[:-2]
                 p = idx[-2]
                 j = idx[-1]
-                nom = self.get_expression_nominal_value(
-                    props[prop_idx].get_material_flow_terms(p, j)
+                nom = abs(
+                    self.get_expression_nominal_value(
+                        props[prop_idx].get_material_flow_terms(p, j)
+                    )
                 )
                 self.set_component_scaling_factor(
                     inh_rxn_gen[idx], weight / nom, overwrite=overwrite
@@ -347,8 +353,10 @@ class ControlVolumeScalerBase(CustomScalerBase):
         if hasattr(model, "mass_transfer_term"):
             for prop_idx in props:
                 for p, j in phase_component_set:
-                    nom = self.get_expression_nominal_value(
-                        props[prop_idx].get_material_flow_terms(p, j)
+                    nom = abs(
+                        self.get_expression_nominal_value(
+                            props[prop_idx].get_material_flow_terms(p, j)
+                        )
                     )
                     self.set_component_scaling_factor(
                         model.mass_transfer_term[prop_idx, p, j],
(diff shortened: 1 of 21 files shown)


idaes/core/base/control_volume_base.py
```diff
@@ -459,8 +471,10 @@ def variable_scaling_routine(
...
                     nom_list.append(
-                        self.get_expression_nominal_value(
-                            props[prop_idx].get_enthalpy_flow_terms(p)
+                        abs(
+                            self.get_expression_nominal_value(
+                                props[prop_idx].get_enthalpy_flow_terms(p)
+                            )
                         )
                     )
                 # TODO we need to do some validation so that nom isn't zero or near-zero
```
bpaul4, line 480: Can we address this here?

idaes/core/scaling/tests/test_scaling_base.py
```diff
@@ -39,6 +39,11 @@ def c_rule(b, i):
 
     m.c = Constraint(m.s, rule=c_rule)
 
+    def e_rule(b, i):
+        return -b.v[i]
```
bpaul4, line 43: Why minus?

idaes/core/scaling/tests/test_scaling_base.py
```diff
@@ -245,10 +254,75 @@ def test_set_scaling_factor_min_limit(self, model, caplog):
...
+            (model.c[2], "constraint", 21),
+            (model.e[4], "expression", 1 / 5),
+        ]:
+            assert scaler_obj.get_scaling_factor(obj) == sf
+            # NaN scaling factor
+            with pytest.raises(
+                ValueError,
+                match=re.escape(
+                    "Scaling factors must be strictly positive and finite. Received "
+                    "value of nan instead."
```
bpaul4, line 275:
```suggestion
                    "value of NaN instead."
```

idaes/core/base/control_volume_base.py
dallan-keylogic: I don't see a method to handle this in the control volume alone. The main thing we're trying to do is avoid getting a nominal value of 0 for the molar enthalpy. However, since we do not specify any single way for the user to implement `get_enthalpy_flow_terms`, we are mostly at the mercy of the expression walker.

We could try to get a scaling factor or scaling hint from `get_enthalpy_flow_terms`, but as part of IDAES, we don't mandate that the enthalpy flow terms have any specific form. For example, the user could have
```
def get_enthalpy_flow_terms(self, p):
    return self.flow_mol * self.cp_mol[p] * (self.temperature - self.params.temperature_ref)
```
and would still expect that to work. (In fact, this is how `get_enthalpy_flow_terms` *are* implemented in the saponification properties.)

This has an obvious problem if the scaling factor of `temperature` is set to `1/temperature_ref`.

idaes/core/scaling/tests/test_scaling_base.py
dallan-keylogic: Why not?

idaes/core/scaling/custom_scaler_base.py
```diff
@@ -43,21 +43,6 @@
 # Set up logger
 _log = idaeslog.getLogger(__name__)
 
-
-def _filter_scaling_factor(sf):
-    # Cast sf to float to catch obvious garbage
```
MarcusHolly, line 48: What's the significance of moving this from `custom_scaler_base `to `scaling_base`?

idaes/models/unit_models/tests/test_pressure_changer.py
```diff
@@ -1221,11 +1212,7 @@ def test_example_case_isothermal(self, isothermal_model):
         assert check_optimal_termination(results)
 
         # Check condition number to confirm scaling
-        sm = TransformationFactory("core.scale_model").create_using(
-            isothermal_model, rename=False
-        )
-        jac, _ = get_jacobian(sm, scaled=False)
-        assert (jacobian_cond(jac=jac, scaled=False)) == pytest.approx(170.82, rel=1e-3)
+        assert jacobian_cond(isothermal_model) == pytest.approx(193.26, rel=1e-3)
```
MarcusHolly, line 1215: Are we expecting a noticeable change here?

idaes/models/unit_models/tests/test_pressure_changer.py
```diff
@@ -1338,11 +1325,7 @@ def test_example_case_isentropic(self, isentropic_model):
         assert check_optimal_termination(results)
 
         # Check condition number to confirm scaling
-        sm = TransformationFactory("core.scale_model").create_using(
-            isentropic_model, rename=False
-        )
-        jac, _ = get_jacobian(sm, scaled=False)
-        assert (jacobian_cond(jac=jac, scaled=False)) == pytest.approx(9.71e3, rel=1e-3)
+        assert jacobian_cond(isentropic_model) == pytest.approx(1186.00, rel=1e-3)
```
MarcusHolly, line 1328: Same with this case and the following one. Are these changes in condition number to be expected?

idaes/core/scaling/custom_scaler_base.py
dallan-keylogic: Previously, it was only used as a part of `CustomScalerBase`. However, since I need to filter the input to `_set_scaling_factor` in `ScalerBase`, I made it available to both functions.

idaes/models/unit_models/tests/test_pressure_changer.py
dallan-keylogic: I didn't know what to expect, but I'm not surprised it changed. Previously, an expression with a negative value would have its scaling factor replaced by `1e-10`, but now we correctly take the absolute value to make that expression positive.

I'm actually surprised that we had changes in only the pressure changer.

idaes/models/unit_models/tests/test_pressure_changer.py
dallan-keylogic: Yes, they are to be expected.

idaes/core/scaling/tests/test_scaling_base.py
dallan-keylogic: I don't care much either way, but Python has its NaNs in lower case:
```
In [1]: float("NaN")
Out[1]: nan
```
Edit:
Actually, we *do* need it to be lower case nan, because Python is printing the output as a generic float. I don't think creating an extra case so we can capitalize NaN is worthwhile.

MarcusHolly (approved):
LGTM (assuming all the checks pass)

bpaul4 (approved):
All my comments have been addressed.

PR #1785, opened 2026-05-04
New port classes

dallan-keylogic:
## Summary/Motivation:
A persistent problem in flowsheet visualization has been trying to determine whether a port corresponds to an inlet or outlet. The present "solution" has been to search for "inlet" or "outlet" in the port name. Here I added two subclasses of `Port`, so we can verify if something is an inlet, outlet, or neither using `isinstance`.

- `InletPort` and `OutletPort` classes have been implemented
- Unit models have been updated to use them

### Legal Acknowledgement

### DIFF
--- a/idaes/core/base/tests/test_ports.py
+++ b/idaes/core/base/tests/test_ports.py
@@ -1,0 +1,622 @@
+#################################################################################
+# The Institute for the Design of Advanced Energy Systems Integrated Platform
+# Framework (IDAES IP) was produced under the DOE Institute for the
+# Design of Advanced Energy Systems (IDAES).
+#
+# Copyright (c) 2018-2026 by the software owners: The Regents of the
+# University of California, through Lawrence Berkeley National Laboratory,
+# National Technology & Engineering Solutions of Sandia, LLC, Carnegie Mellon
+# University, West Virginia University Research Corporation, et al.
+# All rights reserved.  Please see the files COPYRIGHT.md and LICENSE.md
+# for full copyright and license information.
+#################################################################################
+"""
+Tests for flowsheet_model.
+
+Author: Douglas Lee
+"""
+
+import pytest
+import pyomo.common.unittest as unittest
+from io import StringIO
+
+from pyomo.environ import (
+    ConcreteModel,
+    AbstractModel,
+    Var,
+    Set,
+    NonNegativeReals,
+    Binary,
+    Reals,
+    Integers,
+    RangeSet,
+)
+from pyomo.network import Arc
+from idaes.core.base.ports import (
+    InletPortData,
+    InletPort,
+    ScalarInletPort,
+    IndexedInletPort,
+    OutletPortData,
+    OutletPort,
+    ScalarOutletPort,
+    IndexedOutletPort,
+)
+
+# These port tests were copied from Pyomo
+# ____________________________________________________________________________________
+#
+# Pyomo: Python Optimization Modeling Objects
+# Copyright (c) 2008-2026 National Technology and Engineering Solutions of Sandia, LLC
+# Under the terms of Contract DE-NA0003525 with National Technology and Engineering
+# Solutions of Sandia, LLC, the U.S. Government retains certain rights in this
+# software.  This software is distributed under the 3-clause BSD License.
+# ____________________________________________________________________________________
+
+
+# It's less than ideal that we end up creating inlet ports named OUT and outlet ports
+# named IN, but these port classes are intended to simply be renamed versions of
+# their Pyomo counterparts, so that we can use isinstance to check topology.
+class TestInletPort(unittest.TestCase):
+    _port_data_class = InletPortData
+    _port_class = InletPort
+    _scalar_port = ScalarInletPort
+    _indexed_port = IndexedInletPort
+
+    @pytest.mark.unit
+    def test_default_scalar_constructor(self):
+        model = ConcreteModel()
+        model.c = self._port_class()
+        self.assertEqual(len(model.c), 1)
+        self.assertEqual(len(model.c.vars), 0)
+
+        model = AbstractModel()
+        model.c = self._port_class()
+        self.assertEqual(len(model.c), 0)
+        # FIXME: Not sure I like this behavior: but since this is
+        # (currently) an attribute, there is no way to check for
+        # construction without converting it to a property.
+        #
+        # TODO: if we move away from multiple inheritance for
+        # simplevars, then this can trigger an exception (cleanly)
+        self.assertEqual(len(model.c.vars), 0)
+
+        inst = model.create_instance()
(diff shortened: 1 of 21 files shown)


dallan-keylogic:
The CodeCov failure is due to the same issue getting coverage for functions, classes and import statements and consequently will be ignored.

@andrewlee94 , let me know if you have any opinions about this change.

andrewlee94:
An alternative to consider that might be a bit more robust would be to give ports the following properties: `is_inlet_port` and `is_outlet_port` - these could point to a single underlying boolean attribute which can be used to decide which they are. Then you could use the properties instead of `isinstance` for classification, and would allow for possible  duck-typing in future.

dallan-keylogic:
> An alternative to consider that might be a bit more robust would be to give ports the following properties: `is_inlet_port` and `is_outlet_port` - these could point to a single underlying boolean attribute which can be used to decide which they are. Then you could use the properties instead of `isinstance` for classification, and would allow for possible duck-typing in future.

I thought about doing something like this, but 1) that would still require subtyping the full diamond structure of `Port` and 2) we would have to preface `port_obj.is_inlet_port()` with `hasattr(port_obj, "is_inlet_port"), because users need to switch over their models to using the new subclass. Nevertheless, those hoops might be worthwhile for full duck typing. 

The other method I thought about using was to look for ports in an arc as the "destination" field and treating those as inlets and ports in "source" field of an arc to treat that as an outlet. However, that method is also vulnerable to user error, and users also like to fix boundary conditions directly on inlet ports without using a `Feed` block. Also, the subclass of `InletPort` and `OutletPort` could be modified to raise an exception if the user tries to add them to an arc in an unexpected way (like making an `InletPort` a source or an `OutletPort` a destination).

blnicho:
@dallan-keylogic if inlet/outlet information is mostly captured in the `Arc` then I would push for using that rather than making custom port classes with duplicate information. Could you elaborate on the edge cases or issues that you mentioned in your previous comment that lead you to thinking this is not a viable path?

dallan-keylogic:
> @dallan-keylogic if inlet/outlet information is mostly captured in the `Arc` then I would push for using that rather than making custom port classes with duplicate information. Could you elaborate on the edge cases or issues that you mentioned in your previous comment that lead you to thinking this is not a viable path?

The problem is that there are multiple places where inlet/outlet information are added, but they do not necessarily agree with each other. The first is in the `source` and `destination` fields of `Arc` objects. However, if a port is an inlet not just for a block but for the entire flowsheet, it may not be attached to an `Arc`. The `Feed` block was supposed to fix this, but that requires user discipline. @dangunter can attest to the fact that users often can't be relied on to add `Feed` and `Product` blocks where appropriate.

The second place that information is located is in the unit model's initialization method. Almost every initialization method ends up fixing variables on inlet ports. However, there's no class features that force the user to do this.

The last resort for determining what's an inlet and what's an outlet is the port's name. In multiple places in the IDAES codebase, iterating over `Port` objects and looking for `inlet` in the name is used to try to determine what directionality a port has.

Most of these issues can be handled by experienced users, but new users can find them tricky. One of the motivations for this PR came from watertap-org/watertap#1757. It would be nice to tell the user which ports are inlets and which ports are outlets as part of a similar method in IDAES.

idaes/core/base/tests/test_ports.py
```diff
@@ -0,0 +1,622 @@
...
+# University of California, through Lawrence Berkeley National Laboratory,
+# National Technology & Engineering Solutions of Sandia, LLC, Carnegie Mellon
+# University, West Virginia University Research Corporation, et al.
+# All rights reserved.  Please see the files COPYRIGHT.md and LICENSE.md
+# for full copyright and license information.
+#################################################################################
+"""
+Tests for flowsheet_model.
+
+Author: Douglas Lee
```
bpaul4, line 16: Is this correct?

idaes/core/base/tests/test_ports.py
```diff
@@ -0,0 +1,622 @@
...
+        model.c = self._port_class()
+        self.assertEqual(len(model.c), 1)
+        self.assertEqual(len(model.c.vars), 0)
+
+        model = AbstractModel()
+        model.c = self._port_class()
+        self.assertEqual(len(model.c), 0)
+        # FIXME: Not sure I like this behavior: but since this is
+        # (currently) an attribute, there is no way to check for
+        # construction without converting it to a property.
```
bpaul4, line 78: We can't just say assert len(model.c,) == 0? Is that because `c` hasn't been constructed yet at this point?

idaes/core/base/tests/test_ports.py
```diff
@@ -0,0 +1,622 @@
...
+        self.assertEqual(pipe.OUT.polynomial_degree(), 1)
+
+        pipe.pIn.unfix()
+        self.assertEqual(pipe.OUT.polynomial_degree(), 2)
+
+        pipe.OUT.add(pipe.flow / pipe.pIn, "nonLin")
+        self.assertEqual(pipe.OUT.polynomial_degree(), None)
+
+    @pytest.mark.unit
+    def test_potentially_variable(self):
```
bpaul4, line 233: What does "potentially" variable mean?

idaes/core/base/tests/test_ports.py
```diff
@@ -0,0 +1,622 @@
...
+        : pressure :  3.14
+      b :     flow :    10
+        : pressure :  3.14
+      c :     flow :    20
+        : pressure :  3.14
+""",
+        )
+
+
+class TestOutletPort(TestInletPort):
```
bpaul4, line 618: Should this contain tests as well, or does this inherit and duplicate tests from the `TestInletPort` class?

idaes/core/base/tests/test_property_base.py
```diff
@@ -414,11 +416,27 @@ def define_port_members(blk):
...
+        assert not isinstance(port, InletPort)
+        assert not isinstance(port, OutletPort)
+    elif port_class is InletPort:
+        assert isinstance(port, InletPort)
+        assert not isinstance(port, OutletPort)
+    elif port_class is OutletPort:
+        assert not isinstance(port, InletPort)
+        assert isinstance(port, OutletPort)
+    else:
+        raise AssertionError()
```
bpaul4, line 439: Might be worth checking here that `InletPort` and `OutletPort` are also instances of `Port`.

idaes/core/base/tests/test_property_base.py
```diff
@@ -471,11 +490,27 @@ def define_port_members(blk):
...
```
bpaul4, line 513: same here

idaes/core/base/tests/test_property_base.py
```diff
@@ -530,11 +566,27 @@ def define_port_members(blk):
...
```
bpaul4, line 589: and here

PR #1788, opened 2026-05-13
Euclidean condition number

dallan-keylogic:
## Summary/Motivation:
Presently, the diagnostic toolbox uses the Frobenius condition number for `report_numerical_issues`. However, calculating that condition number requires forming the (pseudo) inverse of the Jacobian, which is inevitably a dense matrix. Consequently, it is extremely slow for larger models. This PR adds the option to calculate the 2-norm condition number, making use of the SVD method added by #1566.

### Legal Acknowledgement


ksbeattie:
Needs #1566

PR #1789, opened 2026-05-13
Add support for pandas 3.0 to enable conda-forge releases

### LINKED ISSUE #1253: Pysmo tests np.testing.assert_array_equal() fail because of ordering

sufikaur:
## Fixes

#1253 
#1254 

## Summary/Motivation:
In 2023, bounds to pandas versions were added as a specific change in pandas 2.1 and and versions about pandas 3 did not work, however, this is now a blocker for `conda-forge` packages getting created that use pandas versions above 3 (or do not have the same bounds). 

This PR is to determine if we need these restrictions still. I will continue to block pandas 2.1 since there seems to be a genuine issue with that version despite the fact it pip installs should not install that version. 

**Pandas 3.0 compatibility and dependency updates:**

* Updated to new release of egret through `gridx-prescient` compatible with pandas 3.
* Removed the upper version constraint on pandas in `pyproject.toml`, allowing installation of pandas 3.x and above.

**Code and test refactoring for pandas best practices:**

* Updated test files (`test_PEM_Parameterized_bidder.py`, `test_perfectforecaster.py`) to use lowercase `"1h"` for the frequency argument in `pd.date_range`, matching pandas recommendations and avoiding warnings. 

* Refactored example code in `docs/reference_guides/core/util/tags.rst` to build DataFrames by accumulating rows and to use `.iloc` instead of `.loc` for integer indexing, aligning with pandas best practices and future-proofing the code.

**Bug fix in data splitting utility:**

* Fixed the `split_dataframe` function in `idaes/core/surrogate/sampling/data_utils.py` to ensure correct splitting and index resetting, addressing potential issues with the previous implementation.

### Legal Acknowledgement

### DIFF
--- a/docs/reference_guides/core/util/tags.rst
+++ b/docs/reference_guides/core/util/tags.rst
@@ -130,7 +130,7 @@ for details.  The code below provides a simple example.
 
 .. testcode::
 
-  import pyomo.environ as pyo
+   import pyomo.environ as pyo
   import pandas as pd
   from idaes.core.util import ModelTag, ModelTagGroup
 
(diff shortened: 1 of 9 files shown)


ksbeattie:
@jsiirola will look into getting upstream egret fix for this.

PR #1790, opened 2026-05-14
QGESS Costing Reorganization

bpaul4:
## Summary/Motivation:
Integrate costing methods from the IDAES Power Plant Costing Framework and the PrOMMiS UKy REE Costing Framework into a single, unified QGESS module.

- New QGESS module with common methods
- Power plant costing module and dictionaries contain power plant relevant data and methods only
- REE specific data and methods will live in PrOMMiS, including handling of WaterTAP models as custom models

### Legal Acknowledgement


bpaul4:
I will continue to address spelling, formatting, etc. and add tests, but reviewers please take a look at the new QGESS module and let me know any feedback or suggestions.

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+        "USD_2008_Nov" in pyunits._pint_registry  # pylint: disable=protected-access
+        and "USD_2019_Sep" in pyunits._pint_registry  # pylint: disable=protected-access
+        and "USD_2018_Dec" in pyunits._pint_registry  # pylint: disable=protected-access
+        and "USD_2025" in pyunits._pint_registry  # pylint: disable=protected-access
+        and "USD_UKy_2019" in pyunits._pint_registry  # pylint: disable=protected-access
+    ):
+        # Assume that custom plant units have already been registered
+        # Log a message and end
+        _log.info(
+            "Custom plant currency units (USD_2022, USD_2025, USD_UKy_2019) "
```
MarcusHolly: USD_2022 is referenced here, but nowhere else in this function. Is this intentional?

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+    )
+
+    CONFIG.declare(
+        "variable_OM",
+        ConfigValue(
+            default=False,
+            domain=bool,
+            description="Boolean to enable variable operating & maintenance cost calculations.",
+        ),
+    )
```
adam-a-a, line 116: Why do we need these options? Should fixed and variable O&M costs always be 
accounted for, or are there edge cases that you are trying to account for here?

Also, if we do keep these config options, I might rename them so that it is clearer that they are Boolean options (e.g., `has_fixed_OM`).

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+
+    CONFIG.declare(
+        "fixed_OM",
+        ConfigValue(
+            default=True,
+            domain=bool,
+            description="Boolean to enable fixed operating & maintenance cost calculations.",
+        ),
+    )
```
MarcusHolly, line 107: Do we want this to be the only config option that defaults to True?

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+
+    CONFIG.declare(
+        "tech",
+        ConfigValue(
+            default=None,
+            domain=int,
+            description="Integer corresponding to supported technology libraries.",
+        ),
+    )
```
MarcusHolly, line 209: I'm a bit confused on what this means. Perhaps provide an in-line comment to provide some more context

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+
+        # check that the currency units are built-in or user-defined
+        try:
+            self.base_currency = getattr(pyunits, "USD_" + self.config.CE_index_year)
+        except AttributeError:
+            raise AttributeError(
+                f"CE_index_year {self.config.CE_index_year} is not a valid currency base option. "
+                f"Valid CE index options include CE500, CE394 and years from "
+                f"1990 to 2023."
+            )
```
MarcusHolly, line 273: Include the alternative syntaxes in this error: 2019_Sep and UKy_2019

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+                units=pyunits.dimensionless,
+            )
+        
+        self.CE_index_units = getattr(pyunits, "MUSD_" + self.config.CE_index_year)
+
+        # Set a base period for all operating costs
+        self.base_period = pyunits.year
+
+        # Set capacity factor for overnight and levelized costs
+        if self.config.tech == 10:  # UKy REE, assume operating 336 days per year
```
MarcusHolly, line 301: Is it possible/probable that people could add tech options in the future such that 10 no longer corresponds to UKy? If so, this might silently fail. Is there a way we can double check or replace the conditional with something that more directly checks for the UKy tech option?

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+                initialize=0.85,
+                mutable=True,
+                doc="capacity factor of the plant",
+                units=pyunits.dimensionless,
+            )            
+        
+        # if Lang_factor not None, add as param
+        # this will be used to estimate installation costs of cost accounts
+        # that do not already calculate construction fees and other costs, e.g.
+        # many power plant accounts have their own contruction cost factors
```
MarcusHolly:
```suggestion
        # many power plant accounts have their own construction cost factors
```

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+
+            self.Lang_factor = Param(
+                initialize=self.config.Lang_factor,
+                mutable=True,
+                doc="Lang factor",
+                units=pyunits.dimensionless,
+            )
+
+        # else, if tech == 10, use % factors, Lang = sum of % factors
+        elif self.config.tech == 10:
```
MarcusHolly, line 330: Same comment as above (also occurs at other places throughout the code, but I won't comment on any of the other ones)

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+            # this could be True even is variable_OM is False, as labor is required for production but not necessarily chemicals, electricity, etc.
+            if self.config.consider_taxes:
+
+                self.income_tax_percentage = Param(
+                    initialize=26,
+                    mutable=True,
+                    doc="Combined federal and state income tax percentage"
+                    "usually between 26 - 40%",
+                    units=pyunits.percent,
+                )
```
MarcusHolly, line 451: Is it convention for our costing estimations to provide a minimum cost of operation or an average cost of operation? If the former, 26 is fine. If the latte, perhaps we should use 33 as the tax percentage

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+                    self.phaseout_fractions = Param(
+                        initialize=[75, 50, 25],
+                        mutable=True,
+                        doc="list of fractions of the production incentive tax credit that are applied during the " \
+                        "phaseout years. Each value MUST be between 0 and 1, and the list MUST have the same length " \
+                        "as phaseout_years. Defaults to [75, 50, 25] based on OBBBA, which specifies a 25% " \
+                        "reduction in the first year of phaseout, 50% reduction in the second year, and 75% reduction "
+                        "in the third year before the tax credit is fully phased out.",
+                        units=pyunits.percent,
+                    )
```
MarcusHolly: Should we also check that the numbers are in descending order?

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+                        raise ValueError(
+                            f"CE_index_year does not start with a valid 4-digit year."
+                        )
+                    self.current_year = Param(
+                        initialize=int(current_year),
+                        mutable=True,
+                        doc="Current year for calculating production incentive charge phaseout",
+                        units=pyunits.dimensionless,
+                    )
+
```
MarcusHolly: Wasn't this defined previously?

idaes/models/costing/QGESS.py
bpaul4: There could be cases where there are only fixed operating costs, or cases like https://github.com/IDAES/examples/blob/ed708c8a155ec7efdbe339583fd233f4be7ac55d/idaes_examples/mod/power_gen/ngcc_soec_costing.py#L1052 where we are aggregating equipment costs only.

I like your suggestion to change to `has_X` for any Boolean arguments.

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+                the built-prices do not require input to this argument to be
+                used as long as an entry in the `resources` argument matches
+                a built-in price name.
+            land_cost: Expression, Var or Param to calculate land costs. If this
+                argument has units of cost per time, an annual land cost will be
+                included in the total variable OM cost, and plant overhead for
+                tech=10 (REE) plants. If this argument has units of cost, a
+                one-time land cost will be included in the overnight cost. If this
+                argument has any other units or no units, an error will be returned.
+            fuel: string to define which resource requires fuel storage and
```
MarcusHolly:
```suggestion
            fuel: string to define which resources require fuel storage and
```

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+                handling.
+            feedstock: string to define which resource requires feedstock storage and
```
MarcusHolly:
```suggestion
            feedstock: string to define which resources require feedstock storage and
```

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+                            pyunits.convert(
+                                o.variable_operating_cost,
+                                to_units=b.CE_index_units / pyunits.year,
+                            )
+                        )
+
+        b.total_BEC = Var(
+            initialize=100,
+            bounds=(0, None),
+            doc="Total TPC",
```
MarcusHolly:
```suggestion
            doc="Total BEC",
```
Probably worth spelling out the acronym in the docstring anyway

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+            # assume that total_plant_cost is in millions of
+            # USD_year, where year is the CE_index_year users set
+            units=b.CE_index_units,
+        )
+
+        b.total_TPC = Var(
+            initialize=100,
+            bounds=(0, None),
+            doc="Total TPC",
```
MarcusHolly: Same here

MarcusHolly (commented):
I'll admit that I have not done a thorough line-by-line review, but I've provided a few comments here. Will tests be added in this PR or a separate one? Does these changes also warrant a new tutorial/How-To on costing

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+            "call of custom_power_plant_currency_units."
+        )
+    else:
+        pyunits.load_definitions_from_strings(
+            [
+                # power plant cost account units
+                # from https://toweringskills.com/financial-analysis/cost-indices/
+                "USD_2008_Nov = 500/566.2 * USD_CE500",
+                "USD_2019_Sep = 500/599.3 * USD_CE500",
+                "USD_2018_Dec = 615.7/500 * USD_CE500",
```
Lingyan90: This line of code is not in same patter as other neaby lines of code. probably you mean `"USD_2018_Dec = 500/615.7 * USD_CE500"` instead

idaes/models/costing/QGESS.py
bpaul4: In answer to your question above, I'll be adding tests here but wanted to start the review cycle on the main code first. And yes we will need to update the IDAES and PrOMMiS QGESS costing examples too.

Some of these currencies are needed for the Power Plant and REE costing methods. We could enforce that users define these on the flowsheet level if needed, or register them in those files instead.

idaes/models/costing/QGESS.py
bpaul4: We could default this to False, I think this was originally done because most plants have fixed costs that scale with CAPEX but it's definitely not required so defaulting to False might be better.

idaes/models/costing/QGESS.py
bpaul4: The argument "tech" corresponds to the type of plant being costed: https://idaes-pse.readthedocs.io/en/latest/reference_guides/model_libraries/power_generation/costing/power_plant_costing_netl.html where 1-7 are various power plant types, 8-9 are specific case studies, and 10 is UKy REE. This determines many things such as maintenance factors, some equation forms, and parameter values.

idaes/models/costing/QGESS.py
bpaul4: No, we will enforce that 10 is always UKy REE and future tech's would have new numbers, e.g. 11, 12, and so on. We will update the dictionary on the PrOMMiS side to ensure the source # is 10 for the current accounts.

idaes/models/costing/QGESS.py
bpaul4: 26 was used to calculate the annualization factor, so I think this makes the most sense as a default value.

idaes/models/costing/QGESS.py
bpaul4: That could be useful. If we want to ensure it is always descending on construction and when new values are set, we could put this into the CONFIG options and write a validator method for the domain.

idaes/models/costing/QGESS.py
bpaul4: Oh you're right, I'll check and make sure things are not being defined or checked twice unnecessarily.

idaes/models/costing/QGESS.py
bpaul4: I think we currently only allow one resource to be a fuel resource, e.g. "natural gas", but now I think we should allow more i.e. like we do for "chemicals" and "waste".

idaes/models/costing/QGESS.py
bpaul4: Same here.

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+            description="Boolean to enable economy of numbers calculations.",
+        ),
+    )
+
+    CONFIG.declare(
+        "CE_index_year",
+        ConfigValue(
+            default="2021",
+            domain=str,
+            description="Basis year for costing. Must be a supported value; see the IDAES 'costing_base.py' module",
```
Lingyan90: could make it more specific, e.g. currently the `costing_base.py` have CEPCI index from 1990 to 2023. can change it to `description="Basis year for costing. Must be a supported value from 1990 to 2023. For details, see the IDAES 'costing_base.py' module",`

Also, we should probably rename the `CE_index_year` to `CEPCI_year`, since `CEPCI` is the name of the index, not `CE`.

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+    """
+    Define conversion rates for US Dollars based on CE Index.
+    """
+    register_idaes_currency_units()
+    if (
+        "USD_2008_Nov" in pyunits._pint_registry  # pylint: disable=protected-access
+        and "USD_2019_Sep" in pyunits._pint_registry  # pylint: disable=protected-access
+        and "USD_2018_Dec" in pyunits._pint_registry  # pylint: disable=protected-access
+        and "USD_2025" in pyunits._pint_registry  # pylint: disable=protected-access
+        and "USD_UKy_2019" in pyunits._pint_registry  # pylint: disable=protected-access
```
Lingyan90: Something to consider: does “currency” here refer to different countries’ currencies, or to different years of the same currency? My understanding is that these units are all USD-based, but the values are referenced to different cost years, similar to the CEPCI basis.

A name like `register_cost_year_basis` would be clearer to me. Since we may incorporate location factors later, users may naturally interpret “currency” as referring to local currencies from different countries rather than year-indexed USD. Therefore, I would suggest renaming the function accordingly and probably in another PR, updating the associated names in IDAES `costing_base.py` as well.

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+    #         description="Basis location for costing. Must be a supported value passed as 'country; city';"
+    #         "see the IDAES 'location_factors.json' dictionary. For entries with a country and no city, must "
+    #         "be of the form 'country; None'.",
+    #     ),
+    # )
+
+    def build_global_params(self):
+        """
+        This is where we can declare any global parameters such as default year,
+        currency, Lang factor, location factor, and so on.
```
Lingyan90: here mentioned default year and currency as well, which I belive they both means in USD currency, but potentially in different years. This eco my previous point of rename the function `custom_currency_units` to avoid confusing.

idaes/models/costing/QGESS.py
Lingyan90: The `USD_UKy_2019` unit appears to be specific to a particular case study. Since `QGESS.py` is intended for more general applications, I think we should clean up and standardize costing units across IDAES, PrOMMiS, and potentially WaterTAP and other platforms we are developing.

For example, instead of using a case-specific unit such as `USD_UKy_2019`, we should consider using a more general unit such as `USD_2019`, unless there is a strong reason to preserve the case-specific basis.

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+        if self.config.Lang_factor is not None:
+
+            self.Lang_factor = Param(
+                initialize=self.config.Lang_factor,
+                mutable=True,
+                doc="Lang factor",
+                units=pyunits.dimensionless,
+            )
+
+        # else, if tech == 10, use % factors, Lang = sum of % factors
```
Lingyan90, line 329: We should probably mention where users can check the tech number. It is unclear to me what these tech numbers represent and where they are stored or documented.

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+                
+                maintenance_percentages = {
+                    1: [0.4, 0.016, 0.016],
+                    2: [0.4, 0.016, 0.016],
+                    3: [0.35, 0.03, 0.03],
+                    4: [0.35, 0.03, 0.03],
+                    5: [0.35, 0.03, 0.03],
+                    6: [0.4, 0.019, 0.019],
+                    7: [0.4, 0.016, 0.016],
+                    8: [0.4, 0.019, 0.016],
```
Lingyan90: the tech 8 seems not in same trend as other tech. the labor percent and material percent of TPC is different. which makes the maintenance labor split (0.4) not exactly consistent with 0.019 / (0.019 + 0.016) ≈ 54.3%.

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+                    9: [0.4, 0.019, 0.019],
+                }
```
Lingyan90: I feel there is too much detail in this script. Currently, `QGESS.py` appears to serve as both the costing-estimation framework and a data bank. I would lean toward keeping the framework logic in this file and moving data tables, such as `maintenance_percentages`, to a separate data file.

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+        def income_tax_eq(c):
+            return c.income_tax == pyunits.convert(
+                b.income_tax_percentage, to_units=pyunits.dimensionless
+            ) * (
+                b.total_sales_revenue
+                - (
+                    b.total_variable_OM_cost[0]
+                    + b.total_fixed_OM_cost
+                    + b.annualized_cost / pyunits.year
+                )
```
Lingyan90, line 3262: Taxable income does not seem to account for depreciation costs. Typically, the IRS allows plant depreciation to be included as part of total production costs when calculating taxable income. Seider's book has mentioned 5 year, 7 year etc. MACRS plant depreciation schedule.

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+
+        else:
+            if self.config.consider_taxes:
+                # this is not allowed, there's no income or production to tax
+                raise ConfigurationError(
+                    "Cannot set consider_taxes to True if fixed_OM is False. Calculations of income tax, royalties, "
+                    "and mineral depletion require sales revenue, and production incentive is only applicable to plants "
+                    "producing saleable product streams. Sale revenue and associated administrative costs are included in "
+                    "the fixed O&M calculations."
+                )
```
Lingyan90, line 540: We should also account for cases where taxable income is negative. In those cases, taxes should not be applied.

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+            fixed_charge_doc = "Fixed charge rate from E. Lewis, S. McNaul, et al., Comparison"
+            " of Commercial, State-of-the-Art, Fossil-based Hydrogen Production"
+            " Technologies, National Energy Technology Laboratory, Pittsburgh,"
+            " September 30, 2022. based on CRF values, real for three/five years"
+            " X, nominal for three/five years Y"
+
+        elif self.config.tech == 8:
+
+            # 15% TPC for other owner's costs, 2.7% TPC for financing, 0.5% TPC for spare parts, 2% for preproduction
+            self.pct_TPC = Param(
```
Lingyan90, line 704: We should probably move these technology-specific parameters to a separate file to make the structure cleaner and the main script shorter

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+                initialize=int(current_year),
+                mutable=True,
+                doc="Current year for calculating production incentive charge phaseout",
+                units=pyunits.dimensionless,
+            )
+        
+        self.CE_index_units = getattr(pyunits, "MUSD_" + self.config.CE_index_year)
+
+        # Set a base period for all operating costs
+        self.base_period = pyunits.year
```
Lingyan90, line 298: This seems to repeat line 108. Also, I wonder why this is needed if a base year has already been specified for all costs.

Lingyan90 (commented):
Will continue to review this next week

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+                "10 for a 10% inflation rate. Set to 0 to indicate no inflation."
+            )
+
+            self.revenue_inflation_percentage = Param(
+                initialize=3,
+                units=pyunits.percent,
+                description="Inflation rate for revenue during the operating "
+                "period. The value should be a percentage, for example 10 for a "
+                "10% inflation rate. Set to 0 to indicate no inflation."
+            )
```
Lingyan90, line 633: Should the `operating_inflation_percentage` and `revenue_inflation_percentage` should be the same? If so, is there a reason to define them separately?

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+            annual_revenue: user-defined value (MUSD/year in base year),
+                Var, Param, or Expression for the total annual revenue from
+                saleable products.
+            debt_expression: user-defined value (MUSD in base year), Var,
+                Param, or Expression for the total debt assumed to finance
+                capital costs. This expression will override any value set for
+                debt_percentage_of_capex if both are passed.
+            feedstock_rate: Var, Param, or Expression for rate of feedstock
+                input to the process. If passed, report() method will include
+                a cost breakdown per unit of feedstock.
```
Lingyan90: It is not clear to me whether `feedstock_rate` represents an aggregate feedstock rate used for costing, or whether the model allows multiple feedstocks, each with its own flow rate and price.

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+            land_cost=None,
+            fuel=None,
+            feedstock=None,
+            waste=None,
+            additional_waste_cost=None,
+            chemicals=None,
+            additional_chemicals_cost=None,
+            chemicals_inventory=None,
+            transport_cost=None,
+            ):
```
Lingyan90: Chemicals could be treated as a subset of feedstocks, while fuels could be treated as a subset of utilities; together, feedstocks and utilities may fall under the broader category of resources. It may be helpful to clearly define how `feedstock`, `utilities`, `chemicals`, and `resources` are used in this model to avoid double counting and keep the terminology consistent with broader chemical engineering conventions.

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+                Param, or Expression for the total debt assumed to finance
+                capital costs. This expression will override any value set for
+                debt_percentage_of_capex if both are passed.
+            feedstock_rate: Var, Param, or Expression for rate of feedstock
+                input to the process. If passed, report() method will include
+                a cost breakdown per unit of feedstock.
+            annual_production_rate: Var, Param, or Expression for annual
+                production rate of power or saleable product generated by the
+                plant. If passed, cost of production will be calculated and
+                reported.
```
Lingyan90: Can it account for a combination of multiple products? For example, if the process produces both material and energy products, can the existing model handle that? In this case, would `annul_production_rate` return a list or dictionary of products, such as `{material, energy}`?

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+            SC is the scaled cost
+            RC is the reference cost
+            SP1 is the scaled operational/design parameter 1
+            RP1 is the reference operational/design parameter 1
+            w1 is the fraction of the cost scaled using
+            operational/design parameter 1
+            SP2 is the scaled operational/design parameter 2
+            RP2 is the reference operational/design parameter 2
+            w2 is the fraction of the cost scaled using
+            operational/design parameter 2
```
Lingyan90, line 1723: Dose `w1` + `w2` = 1?

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+        CO2 capture, Illinois No. 6 coal
+        4. Single-stage, slurry-feed, oxygen-blown gasifier with and without
+        CO2 capture, Illinois No. 6 coal
+        5. Single-stage, dry-feed, oxygen-blown, up-flow gasifier with
+        and without CO2 capture, Illinois No. 6 coal
+        6. Natural gas, air-fired, with and without CO2 capture
+        7. Advanced Ultrasupercritical PC
+        8. Polymer Layers accounts
+        9. Sensors & Controls accounts
+        10. University of Kentucky Fire Clay Seam (Hazard No. 4) Rejects
```
Lingyan90, line 1743: line 1321-1335, do you mean some equipment cost could be referenced from multiple resources? Or, is the number 1-10 here means different technologies and the associated equipment cost are only referenced from their corresponding technology?

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+            Exp is the scaling exponent
+
+        This equation is implemented when one operational/design parameter is
+        used to scale the cost
+
+        When two operational/design parameters are used to scale the cost,
+
+        the following equation is implemented:
+
+            SC = RC*w1*(SP1/RP1)^Exp + RC*w2*(SP2/RP2)^Exp
```
Lingyan90, line 1711: I’m curious which equipment uses two operational/design parameters for cost scaling.

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+
+            # append them
+            costing_params.append(BB_costing_params)
+
+
+        # add additional costing params - enforce that this is a list of dictionaries for backwards compatibility
+        if additional_costing_params is not None:
+
+            if isinstance(additional_costing_params, dict):
+                raise TypeError("additional_costing_params must be a list of dicts, not a single dict")
```
Lingyan90: What should be included in the list of dictionaries? It might be helpful to include an example in the message so it is easier to debug.

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+
+            if not (
+                isinstance(additional_costing_params, list)
+                and all(isinstance(item, dict) for item in additional_costing_params)
+            ):
+                raise TypeError("additional_costing_params must be a list of dicts")
```
Lingyan90: same as comment in line 1458

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+                            )
+                            if len(mixed_product_output_rates) > 0
+                            else 1e-12 * c.CE_index_units / pyunits.h
+                        )
+                    )
+                    * c.capacity_factor,
+                    c.CE_index_units / pyunits.year,
+                )
+
+        if annual_fixed_operating_cost is not None:  # we will calculate the fixed operating cost
```
Lingyan90: seems the above code is for tech = =10, and the below code if for tech other than 10. Is this line of code 2371 duplicate to line 2151?

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+                        c.debt_percentage_of_capex, to_units=pyunits.dimensionless
+                    )
+                    * c.capex,
+                    to_units=c.CE_index_units,
+                )
+
+        else:
+
+            b.loan_debt = Reference(b.config.debt_expression)
+
```
Lingyan90, line 3872: Could you help confirm whether the `debt_expression=None` argument is still needed in the model function `def calculate_NPV(b, debt_expression=None):`. Lines 3260–3276 appear to use `b.config.debt_expression` directly. To check whether the `debt_expression` argument is still being used, you could test the function by passing a value different from `b.config.debt_expression`, for example: `calculate_NPV(b, debt_expression=some_other_value)`. If the results do not change compared with using the default argument, then `debt_expression` is likely unused and could potentially be removed from the function signature.

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+        # define series present worth factor as an method so it can be called
+
+        def series_present_worth_factor(r, g, N):
+            """
+            Returns expression for series present worth factor where r is the discount rate
+            expressed as a decimal, N is the project lifetime, and g is the escalation rate
+            (e.g. inflation) expressed as a decimal.
+            """
+            return (1 - ((1 + g) ** (N)) * ((1 + r) ** (-N))) / (r - g)
```
Lingyan90: this could fail when `r==g`

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+                    + b.total_installation_cost
+                    + b.other_plant_costs
+                    + (b.land_cost if b.land_cost_reoccurence == "one_time" else 0 * b.CE_index_units)
+                )
+            )
+            b.opex = Expression(
+                expr=(
+                    b.total_fixed_OM_cost * pyunits.year
+                    + b.total_variable_OM_cost[0] * pyunits.year
+                    + (b.land_cost if b.land_cost_reoccurence == "annual" else 0 * b.CE_index_units/pyunits.year)
```
Lingyan90: the unit in line 3044 (USD/year) seems different from line 3042 & line 3043 (USD)?

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+
+        # input verification
+
+        try:
+            b.capex = Expression(
+                expr=(
+                    b.total_BEC
+                    + b.total_installation_cost
+                    + b.other_plant_costs
+                    + (b.land_cost if b.land_cost_reoccurence == "one_time" else 0 * b.CE_index_units)
```
Lingyan90: typo, `land_cost_reoccurrence`

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+                        len(c.config.capital_expenditure_percentages),
+                    )
+                ),
+                to_units=c.CE_index_units,
+            )
+
+        @b.Constraint()
+        def pv_revenue_constraint(c):
+            # Revenue starts after the capital expenditure period, so we need to account for a delay
+            # PV_Revenue = - REVENUE * [ P/A(r, g, Revenue_end_year) - P/A(r, g, CAPEX_end_year) ]
```
Lingyan90: the comment has negtive signe in right hand equation but the following code didn't

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+        Equations for cash flow expressions derived from the textbook
+        Engineering Economy: Applying Theory to Practice, 3rd Ed. by Ted. G. Eschenbach.
+
+        The net present value (NPV) is a representative measure of the "current
+        day" value of a chemical plant over the total lifetime, including all
+        cash flows.
+
+        This method supports capital expenditure, loan repayment, inflation,
+        and taxes if previously calculated. The NPV formulation assumes that
+        negative cash flows consists capital and operating costs scaled to a
```
Lingyan90:
```suggestion
        negative cash flows consist capital and operating costs scaled to a
```

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+                        b.production_incentive_charge_percent_list.append(0.0)
+
+        b.npv = Var(
+            initialize=(-b.capex + (b.total_sales_revenue - b.opex) * b.config.plant_lifetime),
+            bounds=(None, None),
+            doc="Present value of plant over entire capital and operation lifetime",
+            units=b.CE_index_units,
+        )
+
+        # define series present worth factor as an method so it can be called
```
Lingyan90:
```suggestion
        # define series present worth factor as a method so it can be called
```

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,3580 @@
...
+                # %year1 * CAPEX * (P/A_year1 - P/A_year0)     change from year 1 only
+                # + %year2 * CAPEX * (P/A_year2 - P/A_year1)   change from year 2 only
+                # + %year3 * CAPEX * (P/A_year3 - P/A_year2)   change from year 2 only
+                # + ...)
+                # P/A_year0 = 0, which places each CAPEX expenditure at the end of each period
+
+                return c.pv_capital_cost == -pyunits.convert(
+                    sum(
+                        pyunits.convert(
+                            c.config.capital_expenditure_percentages[idx]
```
Lingyan90: probably should use `c.capital_expenditure_percentages[idx]` instead, otherwise later user updates (e.g. line 3063)  to the paramter will not affect the NPV calculation.

Lingyan90 (commented):
Overall, this is very well-drafted code. I have a few additional comments below:

idaes/models/costing/QGESS.py
bpaul4: The operating costs escalate with labor costs, resource costs, etc. while the revenue escalates with sales prices and those are not necessarily going to be the same rate.

idaes/models/costing/QGESS.py
bpaul4: We only support one feedstock stream at the moment.

idaes/models/costing/QGESS.py
bpaul4: I'll add more in-line comments to clarify. Often we do want to double count resources if they contribute multiple OPEX and overnight cost components, for example resource requirements, plant overhead, and inventory duplicate some resources because extra supply is needed.

idaes/models/costing/QGESS.py
bpaul4: We only support one product as far as the levelized cost of production is concerned, the flowsheet could always be producing more than one product e.g. the byproduct recovery module.

idaes/models/costing/QGESS.py
bpaul4: Yes, it would be like a lever rule.

idaes/models/costing/QGESS.py
bpaul4: These are the potential technology options, the possible types of `tech` that could be used for unit model or plant costing. Each unit model or flowsheet costing block will only use one value of `tech`.

idaes/models/costing/QGESS.py
bpaul4: There are not any examples of this in the repository, I believe it was required for a case study at some point.

bpaul4:
Still working through Pylint and CodeCov, but otherwise this is ready for a final review. The QGESS module has some TODOs for things to update later, and there is a basic test for the power plant costing along with the new PowerPlantCosting dictionary and class modules. More power plant tests, updating examples/flowsheets, and the TODOs will be done in future PRs.

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,4027 @@
...
+        "tech",
+        ConfigValue(
+            default=None,
+            domain=int,
+            description="Integer corresponding to supported technology libraries, where 1-7 are various power plant types, 8-9 are specific case studies, and 10 is UKy REE.",
+        ),
+    )
+
+    # Categories:
+    #     1. Supercritical PC, air-fired, with and without CO2 capture,
```
MarcusHolly: Spell out acronym? (pulverized coal I'm assuming?)

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,4027 @@
...
+        currency (currently only USD is supported), Lang factor (TIC = Lang * BEC, TPC = BEC + TIC),
+        location factor (Washington, D.C. = 1), and so on.
+        """
+
+        # check that the user selected a technology
+        if self.config.tech is None:
+            raise ValueError(
+                "Must set a technology type. Valid options include: \n"
+                "1. Supercritical PC, air-fired, with and without CO2 capture, Illinois No. 6 coal \n"
+                "2. Subcritical PC, air-fired, with and without CO2 capture, Illinois No. 6 coal \n"
```
MarcusHolly:
```suggestion
                "1. Supercritical pulverized coal, air-fired, with and without CO2 capture, Illinois No. 6 coal \n"
                "2. Subcritical pulverized coal, air-fired, with and without CO2 capture, Illinois No. 6 coal \n"
```

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,4027 @@
...
+
+            expr_units = pyunits.get_units(additional_chemicals_cost)
+
+            if expr_units is None:
+                raise ValueError(
+                    "Expression additional_chemicals_cost has no units defined."
+                )
+
+            try:
+                # flag to include land cost in overnight cost
```
MarcusHolly:
```suggestion
                # flag to include additional chemicals cost in overnight cost
```

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,4027 @@
...
+                self.additional_chemicals_cost = Expression(
+                    expr=pyunits.convert(
+                        additional_chemicals_cost, to_units=self.CE_index_units
+                    )
+                )
+                self.additional_chemicals_cost_reoccurrence = "one_time"
+
+            except UnitsError:
+                try:
+                    # flag to include land cost in total variable OM cost
```
MarcusHolly:
```suggestion
                    # flag to include additional chemicals cost in total variable OM cost
```

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,4027 @@
...
+
+            expr_units = pyunits.get_units(additional_waste_cost)
+
+            if expr_units is None:
+                raise ValueError(
+                    "Expression additional_waste_cost has no units defined."
+                )
+
+            try:
+                # flag to include land cost in overnight cost
```
MarcusHolly:
```suggestion
                # flag to include waste cost in overnight cost
```

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,4027 @@
...
+                self.additional_waste_cost = Expression(
+                    expr=pyunits.convert(
+                        additional_waste_cost, to_units=self.CE_index_units
+                    )
+                )
+                self.additional_waste_cost_reoccurrence = "one_time"
+
+            except UnitsError:
+                try:
+                    # flag to include land cost in total variable OM cost
```
MarcusHolly:
```suggestion
                    # flag to include waste cost in total variable OM cost
```

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,4027 @@
...
+            SP2 is the scaled operational/design parameter 2
+            RP2 is the reference operational/design parameter 2
+            w2 is the fraction of the cost scaled using
+            operational/design parameter 2
+            Exp is the scaling exponent
+
+        The scaled cost is computed using ref values for different
+        technologies.
+        Categories:
+        1. Supercritical PC, air-fired, with and without CO2 capture,
```
MarcusHolly: Spell out acronym?

idaes/models/costing/QGESS.py
```diff
@@ -0,0 +1,4027 @@
...
+        Illinois No. 6 coal
+        2. Subcritical PC, air-fired, with and without CO2 capture,
+        Illinois No. 6 coal
```
MarcusHolly: Spell out acryonym?

MarcusHolly (commented):
Minor comments, but LGTM after failing tests are addressed

MarcusHolly (approved):

idaes/models/costing/QGESS.py
bpaul4: This will be addressed in a future PR, it has been marked with a TODO in the code.

idaes/models/costing/QGESS.py
bpaul4: This was actually changed, we now support one or multiple feedstock and fuel, just like we do for waste and chemicals, but this variable is a single aggregate so we can calculate a levelized cost per unit feedstock.

idaes/models/costing/QGESS.py
bpaul4: The production rate is a single total value, e.g. total REE recovered, to calculate the levelized cost of production, however the pure and mixed product rates are separately defined so it is broken down by component there. CO2 is also considered separately, so a process can be producing both power and captured CO2 and levelized costs are calculated for each.

Lingyan90 (approved):
LGTM

PR #1815, opened 2026-07-01
Update to support CoolPropv8

StephenBurroughs:
## Fixes
Updates coolprop wrapper tests to expect new assertion error for missing components.

This doesn't fix the compatability test, as it's in a separate repository. I have opened a pull request in that repository to update the tolerance of the haber bosch test which should fix it: [https://github.com/IDAES/idaes-compatibility/pull/13](url)
 

## Summary/Motivation:
Coolprop has updated to v8, the only breaking change is that it now returns a different error.

-Fix test
-

### Legal Acknowledgement


sufikaur:
Thank you for doing this!
I have temporarily restricted the version from pyproject.toml to be coolprop <8. You may want to override that in this PR, and when you merge into main can also merge in the req without the temporary "<8"

StephenBurroughs:
@ksbeattie I've made two changes here. One is to add a check for the version of CoolProp within the CoolPropWrapper test, then check that the thrown error is appropriate based on the version. This should be fine to go in as is given that it has no other change, although the compatibility repository will still need to have that PR merged to make that pass.
There's a second commit here that makes some additional changes as well. It seems that CoolPropV8 returns slightly different results for the BT PR tests in the CoolPropWrapper test. I'm not sure why this wasn't flagged previously during the GitHub test run, but it seems to be a consistent point of failure on my machine. I'm not exactly sure why this is happening (it's possible that the cubic EOS within CoolProp has been slightly modified, but I haven't had the time to do a deep dive yet and am not sure when that might happen) but it doesn't push things out by a large margin, so I've updated some of the tolerances within the relevant tests to account for this.

I'm happy to rollback that change and just go with the error message one if that's preferable for now, but it does raise a bit of conundrum...
If it's preferable to keep the tolerances as is, then it probably means that the change to the cubic EOS needs to be tracked down and updated. Given that the Benzene/Toluene formulations use the CoolPropWrapper to load data, it seems likely that the divergence is then in the IDAES/CoolProp PR formulations themselves. The challenge here is that if the IDAES PR formulation is updated, then it will pass with CoolPropV8 but fail with CoolPropV7 - basically the same issue but inversed.

Given that it seems impractical to have a case-based formulation and the results are still actually quite accurate (tolerance of 1e-6 vs 1e-4), it seems like updating the tolerances is probably the best course of action...

dallan-keylogic:
@StephenBurroughs , I loosened the tolerance of the failing verification test. If you merge in main, the tests should pass.

StephenBurroughs:
> @StephenBurroughs , I loosened the tolerance of the failing verification test. If you merge in main, the tests should pass.

The tolerance is still a little tight, if it shifts to 1e-3 then it will be resolved, this is actually in line with other tolerances within the testsuite as well.
There is a PR for this on the compatability repo: [IDAES/idaes-compatibility#13](https://github.com/IDAES/idaes-compatibility/pull/13)

StephenBurroughs:
I think if the compatibility test is rerun, it should pass now. I don't think I can manually trigger the workflow though.

dallan-keylogic:
The problem appears to be that the compatibility tests workflow uses the release version of compatibility tests instead of pulling from `main`. Therefore, the looser tolerances we've pushed to `main` aren't being see. @sufikaur , could you please change the workflow to pull from `main` instead?

sufikaur:
> The problem appears to be that the compatibility tests workflow uses the release version of compatibility tests instead of pulling from `main`. Therefore, the looser tolerances we've pushed to `main` aren't being see. @sufikaur , could you please change the workflow to pull from `main` instead?

If possible please mimic the one line change in this PR. https://github.com/IDAES/idaes-pse/pull/1837/changes. If I change the main, it seems like tests on main will break as indicated by the failing tests on the branch. 

Let me know if that works.

sufikaur:
@dallan-keylogic @StephenBurroughs maybe some more changes in compatibility tests are needed?

StephenBurroughs:
> @dallan-keylogic @StephenBurroughs maybe some more changes in compatibility tests are needed?

Yes, I'm just exploring this now and will open a PR in the compatibility repo to resolve things

StephenBurroughs:
I've opened a new PR within the compatability repo that addresses these issues: [Robustness fixes](https://github.com/IDAES/idaes-compatibility/pull/15).
I've also added a couple of lines into the diagnostics utils file that accounts for an edge case relating to one of the failing tests that tries to check an empty Jacobian column

sufikaur (approved):

idaes/models/properties/modular_properties/coolprop/tests/test_coolprop_wrapper.py
```diff
@@ -71,11 +72,17 @@ def test_load_component(self):
 
     @pytest.mark.unit
     def test_load_component_invalid(self):
-        with pytest.raises(
-            RuntimeError,
-            match="Failed to find component foo in CoolProp " "JSON database.",
-        ):
-            CoolPropWrapper._load_component_data("foo")
+        if float(version("CoolProp")[0]) < 8:
```
dallan-keylogic: @sufikaur , do we want to change the coolprop requirement to version 8?

I'm not sure whether casting the version to a float is robust.

StephenBurroughs: I'm happy to update to enforce this. At the time I was thinking that it might be worth keeping backwards compatibility, but I can't really recall if I had a good reason for this... If @sufikaur agrees then I'll update to pin the version instead.

sufikaur: Good point @dallan-keylogic. Lets pin coolprop to `>=8.0`. I don't think there is a need for backward compatibility and it would make the tests cleaner. Thanks @StephenBurroughs

dallan-keylogic (approved):
LGTM

PR #1816, opened 2026-07-02
Propagate scaling suffix to temporary block

dallan-keylogic:
## Summary/Motivation:
As part of the PETSc rewrite, I added the function to propagate submodel scaling to a temporary block to the overall scaling tools. As per Pyomo/pyomo#3800, not including scaling suffixes from the original model when a temporary block is solved is the intended behavior. Therefore, if we want to include scaling suffixes, we need to copy them over manually.

-Function to copy scaling suffixes to a temporary block
-Moved the `NominalValueExtractionVisitor` out of `scaling/util.py` into its own file.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/scaling/nominal_value_tools.py
+++ b/idaes/core/scaling/nominal_value_tools.py
@@ -1,5 +1,5 @@
 """
-The NominalValueExpressionWalker and supporting functions.
+The NominalValueExtractionVisitor and supporting functions.
 
 Authors: Andrew Lee, Douglas Allan
 """
--- a/idaes/core/scaling/nominal_value_tools.py
+++ b/idaes/core/scaling/nominal_value_tools.py
@@ -45,8 +45,6 @@ import idaes.logger as idaeslog
 
 _log = idaeslog.getLogger(__name__)
 
-TAB = " " * 4
-
 
 def get_nominal_value(component):
     """
--- a/idaes/core/scaling/nominal_value_tools.py
+++ b/idaes/core/scaling/nominal_value_tools.py
@@ -1,3 +1,15 @@
+#################################################################################
+# The Institute for the Design of Advanced Energy Systems Integrated Platform
+# Framework (IDAES IP) was produced under the DOE Institute for the
+# Design of Advanced Energy Systems (IDAES).
+#
+# Copyright (c) 2018-2026 by the software owners: The Regents of the
+# University of California, through Lawrence Berkeley National Laboratory,
+# National Technology & Engineering Solutions of Sandia, LLC, Carnegie Mellon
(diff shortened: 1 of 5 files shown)


dallan-keylogic:
CodeCov is complaining because of the outstanding issue with it missing import statements/function definitions when calculating coverage. The actual coverage is fine.

idaes/core/scaling/tests/test_util.py
```diff
@@ -266,6 +268,25 @@ def e2_rule(b, k):
     return m
 
 
+def _scale_model(m):
```
bpaul4, line 271: Why is this a private method?

idaes/core/scaling/tests/test_util.py
bpaul4, line 245: Same question as the new method.

idaes/core/scaling/nominal_value_tools.py
```diff
@@ -0,0 +1,355 @@
...
+
+from pyomo.common.numeric_types import native_types
+
+from idaes.core.scaling.util import get_scaling_factor
+from idaes.core.util.exceptions import BurntToast
+import idaes.logger as idaeslog
+
+_log = idaeslog.getLogger(__name__)
+
+TAB = " " * 4
```
bpaul4: This isn't used anywhere.

idaes/core/scaling/tests/test_util.py
dallan-keylogic: Generally, to discourage users from using methods from test files.

sufikaur:
This may temporarily be part of idaes-pse but will eventually be moved to pyomo

bpaul4 (approved):

idaes/core/scaling/tests/test_util.py
```diff
@@ -2741,3 +2747,39 @@ def diff_eqn(b, z, t):
...
+    propagate_scaling_factors_to_temporary_block(blk)
+    assert hasattr(blk, "scaling_factor")
+    assert len(blk.scaling_factor) == 5
+    assert blk.scaling_factor[m.v[1]] == 5
+    assert blk.scaling_factor[m.v[3]] == 15
+    assert blk.scaling_factor[m.b[2].v2] == 10
+    assert blk.scaling_factor[m.c[1]] == 5
+    assert blk.scaling_factor[m.c[3]] == 125
+
+    # Now test what happens if a scaling factor suffix already exits
```
MarcusHolly:
```suggestion
    # Now test what happens if a scaling factor suffix already exists
```

MarcusHolly (approved):
LGTM - just one typo

Issue #1818, opened 2026-07-15
Deadlock between Pytest and Pyomo output capture on Windows


dallan-keylogic:
While trying to test out a Pyomo PR in IDAES, I encountered an issue with certain tests hanging. After some testing, I found that the tests would pass if I stepped through them with pdb or suppressed Pytest output capturing (the `-o` option). The root cause appears to be related to Pytest's output capturing and Pyomo's output capturing fighting on Windows. Suppressing IDAES/Pyomo output  capture when running Pytest seems to be sufficient to fix the issue.

I was able to trace the issue (partially) to IDAES dropping `pywin32` as a dependency and bumping the version numbers of `pywinpty` and `tornado`.

ksbeattie:
@blnicho, @jsiirola, when you are able, could you look into this from the Pyomo side?  I seem to recall this, or something very similar, came up a few years ago.

PR #1819, opened 2026-07-15
Stop logging capture while running Pytest

dallan-keylogic:
## Fixes
Band-aid fix for #1818 .

## Summary/Motivation:
Fighting between Pytest and Pyomo's output capturing can cause tests to hang on Windows. This is a simple fix suggested by Gemini 3.5 to disable logger capture on Windows when Pytest is imported. We should discuss whether that is the way we should fix this or whether a different fix is desirable.

The test for solver logging capture is going to fail by design.

### Legal Acknowledgement

### DIFF
--- a/idaes/logger.py
+++ b/idaes/logger.py
@@ -287,8 +287,8 @@ class SolverLogInfo(object):
 def solver_log(logger, level=logging.ERROR):
     """Context manager to send solver output to a logger."""
     tee = logger.isEnabledFor(level)
-    # Deadlock between Pyomo and Pytest stdout capture observed on 7/23/26.
-    # See IDAES #1818 for more details.
+    # Temporary fix for deadlock between Pyomo and Pytest stdout
+    # capture observed on 7/23/26. See IDAES #1818 for more details.
     on_windows_pytest = sys.platform.startswith("win") and "pytest" in sys.modules
     if not solver_capture() or on_windows_pytest:
         yield SolverLogInfo(tee=tee)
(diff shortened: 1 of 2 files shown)


idaes/logger.py
```diff
@@ -285,7 +287,9 @@ def __init__(self, tee=True):
 def solver_log(logger, level=logging.ERROR):
     """Context manager to send solver output to a logger."""
     tee = logger.isEnabledFor(level)
-    if not solver_capture():
+    # Bypass capturing on Windows under Pytest to prevent crashes and deadlocks
```
dallan-keylogic:
```suggestion
    # Deadlock between Pyomo and Pytest stdout capture observed on 7/23/26.
    # See IDAES #1818 for more details.
```

idaes/logger.py
```diff
@@ -285,7 +287,10 @@ def __init__(self, tee=True):
 def solver_log(logger, level=logging.ERROR):
     """Context manager to send solver output to a logger."""
     tee = logger.isEnabledFor(level)
-    if not solver_capture():
+    # Deadlock between Pyomo and Pytest stdout capture observed on 7/23/26.
+    # See IDAES #1818 for more details.
```
dallan-keylogic:
```suggestion
    # Temporary fix for deadlock between Pyomo and Pytest stdout
    # capture observed on 7/23/26. See IDAES #1818 for more details.
```

sufikaur (approved):

ksbeattie (approved):
LGTM

PR #1824, opened 2026-07-27
PETSc Overhaul for DAE Index Reduction

### LINKED ISSUE #1399: Function to allow initial condition problem in PETSc to be debugged

dallan-keylogic:
## Fixes
#1399
#1347

## Summary/Motivation:
After some experience with the original PETSc integrator, I've found some shortcomings:
- If the user needs to call PETSc multiple times, they must flatten the model each time, which is slow
- It is hard to troubleshoot model issues for either the initial condition or timestep problem
- It is impossible to convert a time derivative variable into an algebraic variable---if the user wishes to convert a differential variable to an algebraic variable, they must fix the derivative to zero

Point 2 made handling index reduction, in which we regularly need to convert derivative variables into algebraic variables, rather difficult. This problem could be bypassed by adding dummy variables for the derivatives, then setting only a subset of them to equal actual `DerivativeVars`, but that would require large modifications to existing unit models to add these dummy variables to result in a situation in which the user gets confused about which variable is actually a `DerivativeVar` and which is just a dummy.

To address these issues, I decided to rework the PETSc interface into a class.
- The flattened model is cached on the integrator object. If the model structure doesn't change between calls to PETSc, the user can skip re-flattening the model.
- User-facing functions to construct the initial condition and timestamp problem have been created. These functions are called by the PETSc integrator directly when setting up subproblems as part of integration, so the problems the integrator solves are the same ones available to the user for troubleshooting.
- The model looks for the activation status of a `DerivativeVar` object's discretization equation. If that equation is deactivated, it treats it as an algebraic variable. This makes the process of index reduction easier.

- Create new PETSc object class

## Notes for reviewers
- Is checking for the activation status of the discretization equation the way we want to handle index reduction, or do we want to do something else, like use the expression replacement walker?
- Right now, I'm expecting the user to call `refresh_model` if the model structure changes. Do we want to try to save the user from themselves by calling `refresh_model` every time the user calls `dae_by_time_element` unless the user specifies otherwise using a function argument or config option?

### Legal Acknowledgement


idaes/core/solvers/petsc_object.py
```diff
@@ -0,0 +1,871 @@
...
+    active discretization or continuity equation.
+
+    This function should be re-run if the model structurally changes (i.e., a
+    variable is fixed or unfixed, or a constraint is activated/deactivated).
+
+    Args:
+        blk (BlockData): Block which will be searched for DerivativeVars
+        cont_set (ContinuousSet): This function searches for DerivativeVars
+            which have been differentiated by this set.
+        raise_higher_derivative_exception(Exception): Raise a ValueError if
```
dallan-keylogic:
```suggestion
        raise_higher_derivative_exception(bool): Raise a ValueError if
```

blnicho:
I haven't looked through the code yet but here are my initial thoughts/comments on your questions:

>     Is checking for the activation status of the discretization equation the way we want to handle index reduction, or do we want to do something else, like use the expression replacement walker?

Is the intent of the index reduction just to get a clean PETSc solve or is the intent to make permanent changes to the model (like a transformation)? Or a third possible intent would be as a diagnostic tool to alert users to index issues and suggest a fix that the user would have to manually apply. For the first and third intent I think checking the activation status is fine (although there is an implementation detail in how you handle the case where the indexed constraint is only partially deactivated). If we want to approach this like a transformation then the expression replacement walker might be the better option.

`    Right now, I'm expecting the user to call refresh_model if the model structure changes. Do we want to try to save the user from themselves by calling refresh_model every time the user calls dae_by_time_element unless the user specifies otherwise using a function argument or config option?`

Yes, I think "less efficient but safe" should be the default with an option to skip the refresh_model call for users who understand the implication. We could also take a look at code in the new Pyomo solver interfaces that @michaelbynum wrote for automatically detecting when the model structure changed and see if that could be reused here to know if calling refresh_model is needed.

dallan-keylogic:
> I haven't looked through the code yet but here are my initial thoughts/comments on your questions:
> 
> > ```
> > Is checking for the activation status of the discretization equation the way we want to handle index reduction, or do we want to do something else, like use the expression replacement walker?
> > ```
> 
> Is the intent of the index reduction just to get a clean PETSc solve or is the intent to make permanent changes to the model (like a transformation)? Or a third possible intent would be as a diagnostic tool to alert users to index issues and suggest a fix that the user would have to manually apply. For the first and third intent I think checking the activation status is fine (although there is an implementation detail in how you handle the case where the indexed constraint is only partially deactivated). If we want to approach this like a transformation then the expression replacement walker might be the better option.

The intent of the index reduction procedure is to make a permanent change to the model. One or more differential variables must be transformed into algebraic variables in order to result in a nonsingular timestep problem. Presently, I'm manually deactivating discretization equations for variables which are being so transformed and using the activation status of those discretization equations in order to tell PETSc which variables are "differential" and which are "algebraic". 

Typically, we want to turn a subset of chemical components into algebraic variables, so this is the case in which we're deactivating equations only for a subset of indicies.

The expression walker might cause more problems than it solves.  Right now, the `DerivativeVar` is the parent component of a bunch of `VarData` corresponding to derivatives. We'd have to recreate it with a restricted set of indices and substitute it in every equation in the model. Then we'd have to create a dummy derivative var object with the algebraic indices and substitute *those* into the model, because we typically don't want to totally remove the derivative variables, but rather solve for them algebraically from other equations. The benefit to that approach is that the structural change forces the user to grapple with the fact that only some are still differential variables with a defined initial condition, while the others are algebraic variables. However, I think PyomoDAE supports only `DerivativeVar` object pointing only at `IndexedVar` or `ScalarVar` objects, not a relationship between their `VarData` children, so major changes would probably have to be made there.

>> ` Right now, I'm expecting the user to call refresh_model if the model structure changes. Do we want to try to save the user from themselves by calling refresh_model every time the user calls dae_by_time_element unless the user specifies otherwise using a function argument or config option?`
> 
> Yes, I think "less efficient but safe" should be the default with an option to skip the refresh_model call for users who understand the implication. We could also take a look at code in the new Pyomo solver interfaces that @michaelbynum wrote for automatically detecting when the model structure changed and see if that could be reused here to know if calling refresh_model is needed.

Done---I made it a kwarg for the appropriate functions.

idaes/core/solvers/petsc_object.py
```diff
@@ -0,0 +1,923 @@
...
+                diffvar.local_name + "_" + cont_set_name + DAE_CONT_SUFFIX
+            )
+
+            for idx in var:
+                out_map = {}
+                if (
+                    disc_eq is not None
+                    and disc_eq.active
+                    and idx in disc_eq
+                    and disc_eq[idx].active
```
blnicho: Do you really need to check `active` on both the indexed constraint and the constraint data? I would have expected that checking `disc_eq[idx].active` is enough.

idaes/core/solvers/petsc_object.py
```diff
@@ -0,0 +1,923 @@
...
+                ):
+                    out_map["disc_eq"] = disc_eq[idx]
+                if (
+                    cont_eq is not None
+                    and cont_eq.active
+                    and idx in cont_eq
+                    and cont_eq[idx].active
```
blnicho: Same question about checking active twice

idaes/core/solvers/petsc_object.py
```diff
@@ -0,0 +1,923 @@
...
+            if time_point in con and con[time_point] not in self._disc_condata:
+                constraints.append(con[time_point])
+
+        variable_fixedness = to_json(
+            self._model, wts=StoreSpec.isfixed(), return_dict=True
+        )
+        t_block = create_subsystem_block(constraints, list(variables))
+        # Fix input variables (which shouldn't be indexed by time)
+        for vardata in t_block.input_vars.values():
+            vardata.fix()
```
Daison2102, line 693: Each time step saves the fixed/unfixed state of the whole model with `to_json`, then restores it with `from_json` after the solve. I see two potential problems with this.

- Speed: The save and restore cover every variable at every time point, and they run on every step. Their cost grows with the model size and the number of steps, so on a large flowsheet most of the step time can go to this bookkeeping. The `not in self._disc_condata` check adds to it, because it searches a list that holds every discretization equation.

- Errors: The input variables get fixed before the checks that can raise, and before the PETSc solve. If something fails in between, nothing unfixes them, and the user's model is left changed. That's easy to miss when debugging a model, which is what these new methods are for.

My Suggestion is to fix only the input variables that aren't already fixed, and unfix just those after the step. That drops the whole-model `to_json`/`from_json` and handles the speed issue. Making sure the unfix runs even when the step raises also covers the error case, like the old function did with `TemporarySubsystemManager(to_fix=...)`. Also, making `_disc_condata` a `ComponentSet` would make that lookup fast.

idaes/core/solvers/petsc_object.py
```diff
@@ -0,0 +1,923 @@
...
+            None
+        """
+        for v in self.time_variables:
+            if not v[t_to].fixed:
+                v[t_to].value = v[t_from].value
+
+    def dae_by_time_element(
+        self,
+        between=None,
+        refresh_model: bool = True,
```
Daison2102, line 741: On your second question about `refresh_model`: the PR description says users are expected to call `refresh_model()` when the model structure changes, but `dae_by_time_element`, `get_initial_condition_problem` and `get_timestep_problem` all default to `refresh_model=True`. So by default the model gets flattened again on every call, and the cache only helps users who know to pass `False`.

Both choices seem reasonable to me. Keeping `True` is safer, because a user who fixes a variable or deactivates a constraint doesn't have to remember to refresh. In that case, the docstrings should say the cache is opt-in. Switching to `False` gives the speedup by default, but the docstrings would need to tell users clearly to call `refresh_model()` after fixing, unfixing, activating or deactivating anything.

idaes/core/solvers/petsc_object.py
```diff
@@ -0,0 +1,923 @@
...
+                    timestep_block=timestep_block,
+                    time_point=t,
+                    previous_trajectory=trajectory,
+                    previous_variables=variables_prev,
+                )
+
+            # Revert variable fixedness after solving the timestep
+            from_json(self._model, sd=variable_fixedness, wts=StoreSpec.isfixed())
+            tprev = t
+            res_list.append(res)
```
Daison2102, line 848: Each step's solver result gets appended to `res_list` twice: once inside the `solver_log` block and again here at the end of the loop. So `results.results` has two entries for every time step instead of one.

Loops that only check each result's termination condition still work. But any code that counts results, or matches a result to its time step by position, gets the wrong answer. The old `petsc_dae_by_time_element` appends each result once.

I'd suggest to remove this second `res_list.append(res)`. A quick check on the length of `results.results` in one of the existing solve tests would catch it if it comes back.

idaes/core/solvers/petsc_object.py
```diff
@@ -0,0 +1,923 @@
...
+        if self.config.time_var is not None:
+            for t in itime:
+                self.config.time_var[t] = t
+            no_repeat.add(id(self.config.time_var[tlast]))
+        for var in self.time_variables:
+            if id(var[tlast]) in no_repeat:
+                continue
+            no_repeat.add(id(var[tlast]))
+            if isinstance(var[t0].parent_component(), DerivativeVar):
+                continue  # skip derivative vars
```
Daison2102, line 914: With index reduction, a derivative whose `_disc_eq` is deactivated becomes an algebraic variable, and PETSc solves it that way. But `_interpolate_results` skips every `DerivativeVar`, so these derivatives get skipped too. At the time points that aren't in `between`, they keep whatever value they had before the run.

`calculate_derivatives=True` doesn't cover them either, because `calculate_time_derivatives` also skips derivatives with an inactive discretization equation. So regular derivatives can be filled in, but index-reduced ones can't, and the model ends up inconsistent at those points without any warning.

I'd suggest skipping only the derivatives in `self._derivative_differential_vardata_map`, not every `DerivativeVar`. `_save_trajectory` needs the same change. Otherwise, when `between` has more than two points, the trajectory vectors for these derivatives aren't joined across steps, and interpolation fails on mismatched lengths. A test that combines index reduction with a `between` subset would catch this.

PR #1825, opened 2026-07-28
Cleanup Deprecations

dallan-keylogic:
## Summary/Motivation:
There is no prospect of a 3.0.0 version of IDAES in the foreseeable future. Therefore, we should change deprecations that were supposed to happen in 3.0.0 to instead happen in 2.14.0.

- Change deprecations from 3.0.0 to 2.14.0
- Add hard cast to float for old scaling tools to catch garbage input.

### Legal Acknowledgement


dallan-keylogic:
We can ignore CodeCov here---it's flagging the deprecated property as uncovered, but that's because it's deprecated and we're not using it anymore.

idaes/core/util/scaling.py
```diff
@@ -230,6 +230,10 @@ def set_scaling_factor(c, v, data_objects=True, overwrite=True):
         # doesn't exist.  This handles the case where you get a constant 0 and
         # need its scale factor to scale the mass balance.
         return 1
+
+    # Cast scaling factor to float to catch garbage input
+    v = float(v)
```
bpaul4, line 235: What is `v` supposed to be?

dallan-keylogic: The scaling factor. This is in the old tools, I tried to use more descriptive variable names in the new tools.

bpaul4: But why do we need to cast it to a float? I'd expect it to either already be one, or fail at some point because it's not in which case we should catch it and have a custom error message.

dallan-keylogic: Pyomo `Suffixes` allow pretty much anything as values. I've ended up assigning a Pyomo `Var` or `Expression` as a scaling factor before due to errors in my code. What happens is that `set_scaling_factor` allows it to be put on the `Suffix`, but then an error gets raised when it comes time to write the `.nl` file. This later error is hard to trace back to the original point where the mistake was made.

bpaul4 (approved):

ksbeattie (approved):
LGTM

sufikaur (approved):

PR #1829, opened 2026-07-30
Remove NonNegativeReals from main repo

### LINKED ISSUE #1795: NonnegativeReals considered harmful
Setting a variable's Pyomo domain to `NonnegativeReals` causes any attempt to set a lower bound lower than 0 to silently fail (as documented in pyomo/pyomo#3961). I am not aware of any advantage of using a domain instead of using a lower bound, with the possible exception of trying to save a user from themselves. Therefore, we should eliminate this bound wherever it appears.

dallan-keylogic:
## Fixes
#1795

## Summary/Motivation:
The domain `NonNegativeReals` is redundant when bounds are defined and silently truncates user bounds that are lower than zero.  In reviewing the repo, I found some modeling mistakes in the enthalpy state variables for the modular properties. The domain was improperly set to `NonNegativeReals`, and the user would be none the wiser if they used a negative lower bound.

### Legal Acknowledgement


idaes/models/properties/modular_properties/state_definitions/tests/test_FcPh_electrolyte.py
```diff
@@ -44,7 +44,7 @@
 from idaes.core.util.model_statistics import degrees_of_freedom
 from idaes.core.solvers import get_solver
 
-solver = get_solver("ipopt_v2")
+solver = get_solver("ipopt_v2", options={"constr_viol_tol": 1e-8})
```
bpaul4, line 47: Why is this needed?

idaes/models/properties/modular_properties/state_definitions/tests/test_FcPh_electrolyte.py
```diff
@@ -722,7 +723,4 @@ def test_solve_for_true_species(self, frame):
         ) == pytest.approx(0.8, rel=1e-5)
         assert value(
             m.fs.state[1].mole_frac_phase_comp_apparent["Liq", "KHCO3"]
-        ) == pytest.approx(1.6e-8, abs=1e-9)
-        assert value(
-            m.fs.state[1].mole_frac_phase_comp_apparent["Liq", "KHCO3"]
-        ) == pytest.approx(1.6e-8, abs=1e-9)
+        ) == pytest.approx(1.6e-8, abs=1e-7)
```
bpaul4, line 726: same question as above, and for similar changes in the other files

idaes/models/properties/modular_properties/state_definitions/tests/test_FcPh_electrolyte.py
dallan-keylogic: It makes no sense to ask for agreement to nine decimal places in the tests if we're only solving the system of equations to six decimal places.

idaes/models/properties/modular_properties/state_definitions/tests/test_FcPh_electrolyte.py
dallan-keylogic: For a problem solved to an approximate tolerance, the error in variable values is bounded by the error in solving the equations times the condition number. That's a worst-case bound, and we frequently do better. Nevertheless, it makes no sense to ask for six decimal places in the solution and test to nine decimal places. I'm having it solve to eight decimal places now, and testing to seven decimal places to take into account ill-conditioning.

bpaul4 (approved):
Looks good.

idaes/models/costing/SSLW.py
```diff
@@ -1287,7 +1285,7 @@ def cost_turbine(blk, integer=True):
         if integer is True:
             domain = pyo.Integers
         else:
-            domain = pyo.NonNegativeReals
+            domain = pyo.Reals
```
mrmundt, line 1365: Why is this one `Reals`?

idaes/models/costing/SSLW.py
```diff
@@ -1378,7 +1375,7 @@ def cost_pump(
...
```
mrmundt, line 1455: Why is this one `Reals`?

idaes/models/costing/SSLW.py
```diff
@@ -1687,7 +1683,7 @@ def _make_common_vars(blk, integer=True):
     if integer is True:
         domain = pyo.Integers
     else:
-        domain = pyo.NonNegativeReals
+        domain = pyo.Reals
```
mrmundt, line 1766: Why is this one `Reals`?

idaes/models/costing/SSLW.py
dallan-keylogic: It's reals here because the code structurally requires me to define a variable domain, and I don't want to use `NonNegativeReals`. Because the other branch defines an integer domain, I don't want to dispense with it altogether.

The overarching reason this PR exists is because I view `NonNegativeReals` as redundant with variable bounds and frequently get tripped up when I want to relax variable bounds (due to IPOPT not liking almost-active variable bounds in square problems). If I try to set the variable bound to less than zero, the domain silently truncates the lower bound to zero. From conversations with other developers, there's a good reason this behavior exists when dealing with integer domains, but I've found no good reason to use it for real-valued variables. Hence, I'm eliminating it.

idaes/models/costing/SSLW.py
dallan-keylogic: Ditto the first comment

idaes/models/costing/SSLW.py
dallan-keylogic: Ditto the first comment.

idaes/models/costing/SSLW.py
mrmundt: Gotcha. Okay, so, from a code quality view, everything here is fine. Someone else would have to weigh in on whether they agree with this logic. But that someone... _is not me_!

mrmundt (approved):

PR #1832, opened 2026-08-13
Dimensionless radius for Drum1D

dallan-keylogic:
## Summary/Motivation:
This stops a warning being emitted about constructing a range set from mutable data:
```
WARNING: Constructing RangeSet '[0.9..1.03]' from non-constant data (e.g., Var
or mutable Param).  The linkage between this RangeSet and the original source
data will be broken, so updating the data value in the future will not be
reflected in this RangeSet.  To suppress this warning, explicitly convert the
source data to a constant type (e.g., float, int, or immutable Param)
```
by converting the `ContinuousSet` into a dimensionless form and modifying the other equations appropriately.

### Legal Acknowledgement

### DIFF
--- a/idaes/__init__.py
+++ b/idaes/__init__.py
@@ -33,6 +33,7 @@ from . import config
 __version__ = version("idaes-pse")
 
 
+# TODO Is this obsolete?
 def _handle_optional_compat_activation(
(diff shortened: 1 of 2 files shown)


dallan-keylogic:
Moved to #1834 which isn't dependent on the Pyomo log filtering PR.

PR #1833, opened 2026-08-14
Mock structured flowsheet

dangunter:
## Fixes

.. not having a mock structured flowsheet?

To work in "non-mock" mode, requires https://github.com/prommis/flowsheet-inspector-lib/pull/28

- Structured flowsheet wrapper "mock" that can be used without actually depending on the library to do anything.
- Remove any existing (unused) wrapper support, which is now obsolete

### Legal Acknowledgement


dangunter:
Note that this tests whether the mock library works, not the real one.

docs/examples/structfs/flash_flowsheet.py
```diff
@@ -24,62 +24,73 @@
...
+    # Updated: 2023-06-01
+
+    @FS.step(Steps.build)
+    def build_model(ctx):
+        """Build the model."""
+        m = ConcreteModel()
+        m.fs = FlowsheetBlock(dynamic=False)
+        m.fs.properties = BTXParameterBlock(
+            valid_phase=("Liq", "Vap"), activity_coeff_model="Ideal", state_vars="FTPz"
+        )
```
dallan-keylogic, line 46: We really need to retire the BTX property package (but not as part of this PR).

idaes/core/surrogate/plotting/tests/test_alamo_plotting.py
```diff
@@ -31,6 +32,9 @@
     surrogate_residual,
 )
 
+# use non-interactive plotting backend for tests
+matplotlib.use("Agg")
```
dallan-keylogic, line 36: Thank you---fixing this issue was long overdue.

idaes/core/util/structfs/tests/test_mock.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+@pytest.mark.unit
+@pytest.mark.parametrize("mock", [True, False])
+def test_flowsheet_runner(mock):
+    FlowsheetRunner = load_flowsheet_runner(force_mock=mock)
+    if not mock and hasattr(FlowsheetRunner, "mock"):
+        return  # do nothing; real lib not available
+
+    print(f"\n** test {('real', 'mock')[mock]} flowsheet runner **")
```
dallan-keylogic, line 26: Do we want to leave a print statement in a test?

idaes/core/util/structfs/__init__.py
```diff
@@ -10,229 +10,170 @@
...
+        set_solver,
+        initialize,
+        set_operating_conditions,
+        set_scaling,
+        solve_initial,
+        set_autoscaling,
+        add_costing,
+        initialize_costing,
+        setup_optimization,
+        solve_optimization,
```
dallan-keylogic, line 174: I like the idea of this class, but I don't want to be wedded to these step names forever because they sounded good at the time. I'm not sure whether there's a way to make these "experimental", however, if we begin adding them to actual flowsheets.

The obvious choice is to skip the 2.13 release with these, but then we can't try using these for PrOMMiS until after the 2.14 release.

idaes/core/util/structfs/tests/test_mock.py
dangunter: yes - pytest swallows it unless there's an error. unless it's expensive or has side-effects, it is not a problem

idaes/core/util/structfs/__init__.py
dangunter: this is the same class used in the real library. we have to pick something for step names. it doesn't mean we can't change them later. the idea of the class is just to avoid having undetected typos or creative variations causing hard-to-understand behavior.

dangunter:
@dallan-keylogic , are you ok with this now? incidentally I think some local changes from this PR explain the test deletions that snuck into #1850

PR #1834, opened 2026-08-14
Dimensionless radial domain for Drum1D

dallan-keylogic:
## Summary/Motivation:
This stops a warning being emitted about constructing a range set from mutable data:
```
WARNING: Constructing RangeSet '[0.9..1.03]' from non-constant data (e.g., Var
or mutable Param).  The linkage between this RangeSet and the original source
data will be broken, so updating the data value in the future will not be
reflected in this RangeSet.  To suppress this warning, explicitly convert the
source data to a constant type (e.g., float, int, or immutable Param)
```
by converting the `ContinuousSet` into a dimensionless form and modifying the other equations appropriately.

Additionally, it stops a different warning from being emitted by the Helmholtz properties by explicitly asking to avoid creating phase equilibrium terms:

```
WARNING  idaes.models.properties.general_helmholtz.helmholtz_state:helmholtz_state.py:419 Helmholtz EoS packages using Mixed phase representation ignore the 'has_phase_equilibrium' configuration argument. However, setting this to True can result in errors when constructing material balances due to only having a single phase (thus phase transfer terms cannot be constructed).
```

As part of this PR, I had to tweak `dyn_utils.py` in order to avoid an exception caused by a time derivative term being a constant `Expression` equaling zero for the steady state model. `dyn_utils.py` was trying to copy over the steady state value to the dynamic model by accessing the `value` attribute, which `Expressions` do not have. Explicitly using the `value` *function* fixed that problem, but caused another problem when copying the value from a variable that wasn't initialized. That was dealt with a targeted try-except block.

### Legal Acknowledgement


idaes/core/util/dyn_utils.py
```diff
@@ -742,10 +742,22 @@ def copy_values_at_time(
...
+        def _safe_set_value(src, tgt):
+            try:
+                val = value(src)
+            except ValueError as err:
+                # Uninitialized variable
+                if "No value for uninitialized" in str(err):
+                    val = None
+                else:
+                    raise
+            tgt.set_value(val)
```
bpaul4, line 754: Is this supposed to set the value as None if the expected error is thrown? Would that cause errors later on?

idaes/models_extra/power_generation/flowsheets/subcritical_power_plant/subcritical_boiler_flowsheet.py
```diff
@@ -106,6 +106,7 @@ def add_unit_models(m):
     fs.blowdown_split = HelmSplitter(
         dynamic=False,
         property_package=prop_water,
+        property_package_args={"has_phase_equilibrium": False},
```
bpaul4, line 107: What happens if this is not explicitly set to False here?

bpaul4 (commented):
Just a couple comments, otherwise looks good.

idaes/core/util/dyn_utils.py
dallan-keylogic: Yes, it's supposed to set the value as `None`. It may cause errors later on, but it replicates the current behavior of `copy_values_at_time` when encountering a `None` value.

I needed to make this change because I wanted to have `dT/dt` as a constant `Expression` when the system is at steady state; the previous implementation checked `config.dynamic` to determine whether `dT/dt` should be added to constraints/expressions. Turning it into an `Expression` at steady state caused a problem because `copy_values_at_time` tried to look for the `value` attribute---previously, `copy_values_at_time` would note that the `dT/dt` variable did not exist on the steady state model and log a warning, which was then ignored. Changing to using the `value` function fixed the problem with `dT/dt`, but ran into problems for variables whose value was `None`. Previously, the `None` value would be propagated from one flowsheet to another, but using the `value` function raised the exception I'm catching. Catching the exception and assigning a `None` value replicates the previous behavior for `Vars` with a value of `None`.

idaes/models_extra/power_generation/flowsheets/subcritical_power_plant/subcritical_boiler_flowsheet.py
dallan-keylogic: The default for all `StateBlocks` is `has_phase_equilibrium=True`. However, that's inappropriate for Helmholtz property packages with a Mixed state representation. The Helmholtz state block logs this warning:
> WARNING  idaes.models.properties.general_helmholtz.helmholtz_state:helmholtz_state.py:419 Helmholtz EoS packages using Mixed phase representation ignore the 'has_phase_equilibrium' configuration argument. However, setting this to True can result in errors when constructing material balances due to only having a single phase (thus phase transfer terms cannot be constructed).

bpaul4 (approved):

jinliangma:
I reviewed the Drum1D code and did not find anything wrong. By using the dimensionless radial domain from 0 to 1, we no longer need to set the inside diameter and drum metal wall thickness as the model configuration parameters. The original version was trying to fix them as two configuration parameters such that the ContinuousSet is fixed. For the proposed revised code, I feel we should remove the "drum_inner_diameter" and "drum_thickness" configuration parameters and make them as the model variables (Var's). This will allow them as the design variables which can be optimized in a flowsheet. Certainly, we need to make corresponding changes in the files in the repository that use the Drum1D model. I also propose some minor revisions as listed below even though they will not change the model
1. Since we define drum_thickness as a variable (previous as a Param), we should replace "b.drum_ro - b.drum_ri" by "drum_thickness" (e.g. Lines 525, 559, 563). Note 
2. In some Constraint or Expression definitions, both "self.xxx" and "b.xxx" are used. It will be better if we always use "b.xxx" (e.g. Line 684).

jinliangma (approved):
I reviewed the revision. I approve this PR.

PR #1836, opened 2026-08-18
Arc constraint scaler

dallan-keylogic:
## Summary/Motivation:
The old scaling tools automatically scaled all arc constraints on the flowsheet, but there was no equivalent function in the new scaling tools to do so. This PR adds the `ArcConstraintScaler` class, which automatically applies scaling to all arc constraints.

### Legal Acknowledgement

### DIFF
--- a/idaes/core/scaling/arc_constraint_scaler.py
+++ b/idaes/core/scaling/arc_constraint_scaler.py
@@ -1,0 +1,109 @@
+#################################################################################
+# The Institute for the Design of Advanced Energy Systems Integrated Platform
+# Framework (IDAES IP) was produced under the DOE Institute for the
+# Design of Advanced Energy Systems (IDAES).
+#
+# Copyright (c) 2018-2026 by the software owners: The Regents of the
+# University of California, through Lawrence Berkeley National Laboratory,
+# National Technology & Engineering Solutions of Sandia, LLC, Carnegie Mellon
+# University, West Virginia University Research Corporation, et al.
+# All rights reserved.  Please see the files COPYRIGHT.md and LICENSE.md
+# for full copyright and license information.
+#################################################################################
+"""
+Tool for automatically scaling arc constraints on a Pyomo block.
+
+Author: Douglas Allan
+"""
+
+from pyomo.environ import Constraint
+from pyomo.network import Arc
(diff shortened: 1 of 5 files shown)


idaes/core/scaling/arc_constraint_scaler.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def scale_model(self, **kwargs):
+        """
+        Do not use this method, use one of the scale_arc_constraints
+        methods instead.
+        """
+        raise NotImplementedError(
+            "The scale_model method is not implemented for the ArcConstraintScaler. "
+            "Use one of the scale_arc_constraints methods instead."
+        )
```
MarcusHolly, line 49: Why include this method at all then? Is this something that will be implemented in the future?

MarcusHolly (commented):
One small comment, but I have some generic implementation questions as well. So it seems to me like these functions (`scale_arc_constraints_by_nominal_value `& `scale_arc_constraints_by_nominal_derivative_norm`) are meant to be called at the flowsheet level. So using BSM2-P as an example, I assume I would just replace the `elif "_expanded"` loop with `scale_arc_constraints_by_nominal_value `? And how would a user determine which of the two functions to use?

Image

idaes/core/scaling/arc_constraint_scaler.py
dallan-keylogic: I'm including this method because it lives on the `CustomScalerBase`. It doesn't make sense for this sort of scaler, so I want to intercept users with a helpful error message.

dallan-keylogic:
> One small comment, but I have some generic implementation questions as well. So it seems to me like these functions (`scale_arc_constraints_by_nominal_value `& `scale_arc_constraints_by_nominal_derivative_norm`) are meant to be called at the flowsheet level. So using BSM2-P as an example, I assume I would just replace the `elif "_expanded"` loop with `scale_arc_constraints_by_nominal_value `?

Correct.

> And how would a user determine which of the two functions to use?

They could try both and see which one results in a reasonable condition number.

MarcusHolly (approved):

bpaul4 (approved):

